PR #1, opened 2022-08-28
Create nitro-bypass.js

ArjixWasTaken:
Tricks the client into thinking you have nitro (so you can stream at 1080p and 60fps) and replaces nitro emojis with links


ArjixWasTaken:
If there come new patches that override the sendMessage function...then uhh, there should be a util that registers callbacks for this.

ArjixWasTaken:
Actually, should I do that?

Vendicated:
this should use in place patches instead of monkey patching, look at the experiments and message click patches for reference. I would find which method has the stream quality logic and simply remove the premiumType checks, that way you don't even need to modify the current user

Vendicated:
if you want to do this and need further help/pointers, just message me on discord. it should be trivial once you find the right function, just a regex replace

ArjixWasTaken:
That would be nice and all, if the goal was just to stream at 1080p w/ 60fps, I also want discord to show all the emojis as available.
Although what you said still applies, it would just be double the effort. (cause I'll also have to find the discord code that checks if you can use all the emojis)

ArjixWasTaken:
Although I am up to learn if you are willing to teach me

PR #5, opened 2022-09-02
make better

LandenStephenss:


ItzOnlyAnimal (changes requested):
not enough badges

Sepruko (changes requested):
ummm... make even better...

Sepruko:
> ummm... make even better...

true

Vendicated:
i hope you explode

PR #7, opened 2022-09-02
Add "Ify"

Cynosphere:
Disabes Spotify auto-pausing and premium checks


Ryuo1:
Doesn't work anymore.

PR #10, opened 2022-09-16
feat(installer) Windows powershell installer and uninstaller

MeguminSama:
Installer and Uninstaller for windows (powershell)

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -15,7 +15,48 @@ $PACKAGE_JSON = @"
 }
 "@
 
-$discord_root = "$env:LOCALAPPDATA\Discord"
+$branch_paths = Get-ChildItem -Directory -Path $env:LOCALAPPDATA |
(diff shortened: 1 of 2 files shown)


ArjixWasTaken:
I can't really verify this, but your scripts only target the stable discord client
What if I only have discord PTB?

MeguminSama:
> I can't really verify this, but your scripts only target the stable discord client What if I only have discord PTB?

The normal install script only does stable, so I copied that into this script

ArjixWasTaken:
no that is wrong, you can rename the discord-ptb command to be just discord
so any client version is supported, as long as the executable on PATH is named discord

MeguminSama:
Working on adding a branch selector

MeguminSama:
:)
image

Vendicated:
LGTM probably idk I never did pwsh

PR #11, opened 2022-09-16
fix(experiments): Patch canUseOverrides variable

MeguminSama:
This patches the canUseOverrides variable to re-enable experiments on newer builds.


Vendicated:
<a:yesyes:860233446110134303>

PR #17, opened 2022-09-30
feat(plugin): add muteNewGuild Plugin

GlitchChan:
add a plugin that will auto-magically mute newly joined guilds.

### DIFF
--- a/src/plugins/muteNewGuild.ts
+++ b/src/plugins/muteNewGuild.ts
@@ -1,23 +1,16 @@
 import definePlugin from "../utils/types";
-import { findByProps } from "../webpack";
 
 export default definePlugin({
-    name: "muteNewGuild",
+    name: "MuteNewGuild",
     description: "Mutes newly joined guilds",
     author: "Glitchy",
     patches: [
         {
             find: "acceptInvite:function",
             replacement: {
-                match: /(return [a-z]\|\|null==[a-z])/,
-                replace: "Vencord.Plugins.plugins.muteNewGuild.muteGuild(a);$1"
+                match: /(\w=null!==[^;]+)/,
+                replace: "$1;Vencord.Webpack.findByProps('updateGuildNotificationSettings').updateGuildNotificationSettings($1,{'muted': true, 'suppress_everyone': true, 'suppress_roles': true})"
             }
         }
     ],
-
-    muteGuild(a) {
-        const guildSettings = findByProps("updateGuildNotificationSettings")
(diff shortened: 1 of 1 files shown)


src/plugins/muteNewGuild.ts
```diff
@@ -0,0 +1,23 @@
...
+export default definePlugin({
+    name: "muteNewGuild",
+    description: "Mutes newly joined guilds",
+    author: "Glitchy",
+    patches: [
+        {
+            find: "acceptInvite:function",
+            replacement: {
+                match: /(return [a-z]\|\|null==[a-z])/,
+                replace: "Vencord.Plugins.plugins.muteNewGuild.muteGuild(a);$1"
```
Vendicated, line 13: don't hardcode variable names, match them with the regex, also u can use \w instead of [a-z]

the muteGuild call could just be inlined (Vencord.Webpack.findByProps......)

GlitchChan: Working on this right now

src/plugins/muteNewGuild.ts
```diff
@@ -0,0 +1,19 @@
...
+export default definePlugin({
+    name: "MuteNewGuild",
+    description: "Mutes newly joined guilds",
+    authors: [{
+        name:"Glitchy",
+        id: 269567451199569920n
+    }],
+    patches: [
+        {
+            find: "acceptInvite:function",
```
Vendicated, line 12: there are two modules containing this string
```suggestion
            find: ",acceptInvite:function",
```

src/plugins/muteNewGuild.ts
```diff
@@ -0,0 +1,19 @@
...
+            replacement: {
+                match: /(\w=null!==[^;]+)/,
+                replace: "$1;Vencord.Webpack.findByProps('updateGuildNotificationSettings').updateGuildNotificationSettings($1,{'muted': true, 'suppress_everyone': true, 'suppress_roles': true})"
```
Vendicated, line 15:
```suggestion
                replace: "$1;Vencord.Webpack.findByProps('updateGuildNotificationSettings').updateGuildNotificationSettings($1,{'muted':true,'suppress_everyone':true,'suppress_roles':true})"
```

Vendicated (changes requested):

GlitchChan:
I have done both now 😎

Vendicated (approved):
Thank you!

GlitchChan:
This was definitely a fun learning experience

br0gio:
This plugin is broken

PR #15, opened 2022-09-30
likely spammer gon

ItzOnlyAnimal:
i hope the regex is okay lol it works for me

PR #19, opened 2022-10-01
feat(plugin): nop spammer

botatooo:
continuation of #15 but i can't be bothered to do pr inception

### DIFF
--- a/src/plugins/nopSpammer.ts
+++ b/src/plugins/nopSpammer.ts
@@ -1,13 +1,11 @@
+import { Devs } from "../utils/constants";
 import definePlugin from "../utils/types";
 
 export default definePlugin({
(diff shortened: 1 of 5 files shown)


src/plugins/nopSpammer.ts
```diff
@@ -0,0 +1,23 @@
...
+        {
+            name: "Iryis",
+            id: 118437263754395652n,
+        },
+    ],
+    patches: [
+        {
+            find: "),{hasFlag:",
+            replacement: {
+                match: /(if\(e<=1<<30\)return)/,
```
Vendicated, line 18:
```suggestion
                match: /(if\(.{1,2}<=1<<30/,
```
then use that group in replace `if($1===`

PR #25, opened 2022-10-02
feat(plugins): Add isStaff plugin

MeguminSama:
:catnod:

image


botatooo:
isn't this covered by the exp snippet?

MeguminSama:
No, it is not

botatooo:
i see i seee

PR #26, opened 2022-10-02
Make NitroBypass only add spaces when there are none

Ruthenic:
IE, `test:emoji:` turns into `test https://discord[blahblahbla]`, but `test :emoji:` becomes   
`test https://discord[blahblah]` and *not* `test  https://discord[blah]` (which was the prior behaviour)


Vendicated:
I don't like this approach since it'll insert unnecessary spaces. Instead i would do the following:
Check character before and after. If
- undefined: no space
- any whitespace (\s): no space
- anything else: space

Ruthenic:
Should cover all those cases, now

Ruthenic:
(commit also fixes not checking for `emoji.require_colons` in the edit handler)

Vendicated (approved):

PR #34, opened 2022-10-04
feat: no blocked messages plugin

DiamondMiner88:
Hide all blocked messages from chat

### DIFF
--- a/src/plugins/noBlockedMessages.ts
+++ b/src/plugins/noBlockedMessages.ts
@@ -10,7 +10,7 @@ export default definePlugin({
             find: "safety_prompt:\"DMSpamExperiment\",response:\"show_redacted_messages\"",
             replacement: [
                 {
-                    match: /collapsedReason;return \w{1,2}.createElement/,
+                    match: /collapsedReason;return (?=\w{1,2}.createElement)/,
                     replace: "collapsedReason; return null; null.createElement"
                 }
             ]
--- a/src/plugins/noBlockedMessages.ts
+++ b/src/plugins/noBlockedMessages.ts
@@ -11,7 +11,7 @@ export default definePlugin({
             replacement: [
                 {
                     match: /collapsedReason;return (?=\w{1,2}.createElement)/,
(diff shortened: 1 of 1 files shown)


src/plugins/noBlockedMessages.ts
```diff
@@ -0,0 +1,19 @@
...
+    name: "NoBlockedMessages",
+    description: "Hides all blocked messages from chat completely.",
+    authors: [Devs.rushii],
+    patches: [
+        {
+            find: "safety_prompt:\"DMSpamExperiment\",response:\"show_redacted_messages\"",
+            replacement: [
+                {
+                    match: /collapsedReason;return \w{1,2}.createElement/,
+                    replace: "collapsedReason; return null; null.createElement"
```
Vendicated, line 14: null.createElement?

Vendicated: i mean it will work because you return before that but kinda weird xd

DiamondMiner88: You would still have to add some identifier so it parses correctly no?

src/plugins/noBlockedMessages.ts
```diff
@@ -0,0 +1,19 @@
...
+export default definePlugin({
+    name: "NoBlockedMessages",
+    description: "Hides all blocked messages from chat completely.",
+    authors: [Devs.rushii],
+    patches: [
+        {
+            find: "safety_prompt:\"DMSpamExperiment\",response:\"show_redacted_messages\"",
+            replacement: [
+                {
+                    match: /collapsedReason;return \w{1,2}.createElement/,
```
Vendicated, line 13:
```suggestion
                    match: /collapsedReason;return (?=\w{1,2}.createElement)/,
```

src/plugins/noBlockedMessages.ts
Vendicated, line 14:
```suggestion
                    replace: "collapsedReason; return null;"
```

Vendicated (approved):

PR #46, opened 2022-10-04
feat(plugin): add webhookTags

Cynosphere:


src/plugins/webhookTags.ts
```diff
@@ -0,0 +1,34 @@
...
+            },
+        },
+        {
+            find: '.BOT=0]="BOT"',
+            replacement: {
+                match: /case (.)\.BOT:default:(.)=/,
+                replace: (orig, types, text) =>
+                    `case ${types}.WEBHOOK:${text}="WEBHOOK";break;${orig}`,
+            },
+        },
```
Ruthenic, line 24: you can merge these into one patch object (`replacement` can be an array)

Ruthenic (changes requested):

src/plugins/webhookTags.ts
```diff
@@ -0,0 +1,34 @@
...
+        {
+            find: ".Types.ORIGINAL_POSTER",
+            replacement: {
+                match: /return null==(.)\?null:.\.createElement\((.)\.Z/,
```
Vendicated, line 28: is hardcoding Z safe?

adryd325: .default always turns to Z
.default when there are multiple exports turns to ZP

botatooo: you never know when that might change though

adryd325: It's consistent across every module we've seen when moved to swc

Cynosphere: There are modules dating before the main client switched to swc that have `Z` as `default`, namely RPC.

adryd325 (approved):

Vendicated (approved):

PR #47, opened 2022-10-05
feat(plugin): add clearURLs

adryd325:
At the moment this.onSend() doesn't get called for some reason; @Cynosphere and I couldn't figure out why.

I'd like to wait for whatever is causing this to get fixed, whether in my extension or in Vencord

Update: addPreSendListener is just broken, i'll fix that in this PR i guess


adryd325:
Forgot to include dependency

Vendicated:
> At the moment this.onSend() doesn't get called for some reason; @ Cynosphere and I couldn't figure out why.

Was the problem just not having the message api enabled? Or is it actually broken

adryd325:
The problem was not having the API enabled
The simplemarkdown patch was making me think unindent was working fine and it had cyn and i confused

Vendicated:
Cool! I think that plugin was made before dependencies were a thing and I forgot to update it

If I understood the rules correctly, the ones without @ apply to any URLs and the ones with an @ are paramToRemove@host? It would be helpful to add a little comment explaining how they work for people who want to modify them in the future

It would also be nice to add Spotify to the default rules, the parameters to remove would be si and utm_* (there's a few, I only know utm_source off the top of my head)

Another useful feature would be to have URLs automatically cleaned before opening them

Also, Plugin Settings will be a thing soon™

adryd325:
The rules format was inspired by https://github.com/Smile4ever/Neat-URL

I just added wildcard matching to params and added documentation on how to write rules

botatooo:
what about https://docs.clearurls.xyz/1.23.0/specs/rules/

adryd325:
> what about https://docs.clearurls.xyz/1.23.0/specs/rules/

I didn't feel like implementing the entirety of clearurls

adryd325:
this should be ready for merge

src/plugins/clearURLs/index.ts
```diff
@@ -0,0 +1,129 @@
...
+            const splitRule = rule.split("@");
+            paramRule = splitRule[0];
+
+            paramRule = new RegExp("^" + paramRule.replace(/[#-}]/g, '\\$&').replace(/\\\*/, ".+?") + "$");
+
+            if (splitRule[1]) {
+                hostRule = new RegExp(
+                    "^(www\\.)?" +
+                        splitRule[1]
+                            .replace(/[#-}]/g, '\\$&')
```
Vendicated, line 43: did you mean to escape this `-`? otherwise this is the charrange # to } which is a lot
image

adryd325: this is what happens when I grab an escape regex from stackoverflow instead of somewhere sensible like lodash lmao

shoud be fixed now

Vendicated:
Seems to work fine now! I would push a few changes (mainly formatting & typescript types stuff) if that's okay, then merge

Vendicated (approved):

Vendicated:
Thank you!

PR #52, opened 2022-10-05
feat: View raw button

ArjixWasTaken:


ArjixWasTaken:
Adds a view raw button to the shift+hover popup on messages
image


for now it just logs the message in the console, when the modals are fixed this will change, hence why it is a draft

ArjixWasTaken:
I intended to make this a util instead (so plugins can register their own buttons) but...
I couldn't figure out how to insert a button at runtime, so yeah...

Other plugins can just copy paste and modify this one instead.

ArjixWasTaken:
Ok, so this is ready, *but*, should I add an API to register action buttons (on messages) in this same PR.
*Or* make another PR that also updates this plugin?

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,85 @@
...
+        message: Message,
+        clickEvent: MouseEvent
+    ) {
+        openModal(() => {
+            return (
+                <div style={{ overflow: "scroll" }}>
+                    {
+                        highlightJsHelperFuncs.codeBlock.react(
+                            { lang: "json", content: JSON.stringify(message, null, 2) },
+                            highlightJs,
```
Vendicated, line 41: you can just pass null here

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,85 @@
+import type { Channel, Message } from "discord-types/general";
+import { ModalSize, openModal } from "../utils/modal";
+import definePlugin from "../utils/types";
+import { Devs } from "../utils/constants";
+import { lazyWebpack } from "../utils";
+import { filters } from "../webpack";
+
+const highlightJs = lazyWebpack(filters.byProps(["highlight"]));
+const highlightJsHelperFuncs = lazyWebpack(filters.byProps(["codeBlock"]));
```
Vendicated, line 9:
```suggestion
const parser = lazyWebpack(filters.byProps(["codeBlock"]));
```

src/plugins/viewRaw.tsx
ArjixWasTaken: where?

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    match: /(\w{1,2})\((\{key:"copy-link",.+?})\)/,
+                    replace: (m, Zn, menuItem) => {
+                        const items = menuItem.matchAll(
+                            /(\w+):((?:["'].+?["'])|(?:[^,}]+))/g
+                        );
+
+                        let duplicate = "{";
+
+                        for (const item of items) {
+                            if (item[1] === "key") item[2] = "\"view-raw\"";
```
Vendicated, line 66: switch block would be nicer here

Vendicated (changes requested):

src/plugins/viewRaw.tsx
Vendicated: highlightJs

ArjixWasTaken: are you sure about that? I just did what the discord code does

src/plugins/viewRaw.tsx
ArjixWasTaken: well...yes, but have you read my previous comments?
There is no point refactoring this if it will be removed entirely in case I add an API for adding icons.

src/plugins/viewRaw.tsx
Vendicated: image

ArjixWasTaken:
> Ok, so this is ready, _but_, should I add an API to register action buttons (on messages) in this same PR. _Or_ make another PR for that API that also updates this plugin (to use it instead of manually patching the buttons)?

@Vendicated

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ),
+    viewRaw: function (
+        channel: Channel,
+        message: Message,
+        clickEvent: MouseEvent
+    ) {
+        openModal(
+            () => {
+                return (
+                    <div style={{ overflow: "scroll" }}>
```
Vendicated, line 39: overflow scroll is very ugly, try text wrap

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,95 @@
...
+                );
+            },
+            { size: ModalSize.LARGE }
+        );
+
+        // i don't fucking know why discord does this, but oh well, it's easy to fix.
+        document.body.style.userSelect = "auto";
+    },
+    patches: [
+        // prettier-ignore
```
Vendicated, line 58: ?

src/plugins/viewRaw.tsx
Vendicated: Also you must now wrap your modal into ModalRoot. You should also consider using different Modal sections to add more info and perhaps add buttons to copy the raw content or raw json to your clipboard in this modal

Vendicated (changes requested):

src/plugins/viewRaw.tsx
ArjixWasTaken: in case you don't use prettier, that comment disables automatic formatting for the next block/statement

src/plugins/viewRaw.tsx
```diff
@@ -0,0 +1,95 @@
...
+        {
+            find: 'key:"configure",',
+            replacement: [{
+                match: /(\w{1,2})\((\{key:"copy-link",.+?})\)/,
+                replace: (m, Zn, menuItem) => {
+                    const items = menuItem.matchAll(
```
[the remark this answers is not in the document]
ArjixWasTaken: well, you never really answered my question, will this be made into an API or not?
there really is no reason to document this if i am going to erase it later either way...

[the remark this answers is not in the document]
Vendicated: probably api would make sense

src/plugins/viewRaw.tsx
somerandomcloud: The repo uses eslint??

ArjixWasTaken: > The repo uses eslint??

What does it matter that the repo uses eslint?

I have prettier as a vscode extension, the repo doesn't need to have it configured for it to work.

And also, most of the famous linters/formatters can detect the prettier-ignore comments, in the same way that prettier can detect eslint comments.

As long as the formatter you use is popular enough, it will support other formatters' comments.


Finally, it is just a fucking comment, what's your problem?

Vendicated:
superseded by #275

PR #62, opened 2022-10-08
feat: plugin css & clean up build scripts

DiamondMiner88:
Adds a convenience property `css` to `PluginDef` which is added as a style tag to the head when starting.

Adding it as a property allows for people to either embed it within the plugin file or put it externally, like at `plugins/messageLogger/index.css` which then can be imported as `import css from './index.css'`


MeguminSama:
Slightly confused here - if people wanted to mutate document.head in start() and stop(), we aren't stopping them. CSS Modules are practical for loading CSS in TSX files, and can easily be applied to document.head, SoI'm unsure of the use-case for the `PluginDef.css` property.

DiamondMiner88:
> So I'm unsure of the use-case for the PluginDef.css property.
Convenience and less duplication.

The practical issue is I have no clue how on earth to selectively import all `**/*.css` files in a way that esbuild's css plugin detects the import

PR #64, opened 2022-10-08

ActuallyTheSun:
plugin to show message links as embeds in chat
image

### DIFF
--- a/src/plugins/messageLinkEmbeds.ts
+++ b/src/plugins/messageLinkEmbeds.ts
@@ -9,23 +9,23 @@ export default definePlugin({
         id: 406028027768733696n
     }],
     patches: [{
-        find: '["disableReactionReads"',
+        find: "[\"disableReactionReads\"",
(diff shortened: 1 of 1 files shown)


Vendicated:
would be way cooler if you made it render as actual Message Components instead

ActuallyTheSun:
i did not realize that could be done
great idea

Vendicated:
it definitely can be done, just have to fetch the right components and figure out how to render them properly based on the fetched message

Cyn had a cumcord plugin that does this, it is outdated now but can still be used as reference
https://gitlab.com/Cynosphere/cumcord-plugins/-/tree/master/MessageLinkPreview

ActuallyTheSun:
yeah, i have no idea where to even start with that (mainly because i barely know what i am doing) and im fairly sure i need to completely rewrite this to show as a message
i'll just close the pr and return when i can figure out how to do that

PR #68, opened 2022-10-08

ActuallyTheSun:
plugin to show message links as embeds in chat

not gonna do them as message components because i am way too smooth brained for that, just embeds for now

also at the moment it only gets the messages from cache and will not fetch them, which i'm assuming is an issue. i'll fix soon after i figure out how to actually fetch stuff

plus the embed looks good now and has a decent amount of info (definitely didn't just copy the aliucord plugin embed :trollface:)


ArjixWasTaken:
All `findByProps` calls should only be called once, there is no need to call them multiple times.

ActuallyTheSun:
> All `findByProps` calls should only be called once, there is no need to call them multiple times.

I'm... not sure what you mean by that. It looks like they're all called only once to me. Or do you mean move them to someone where they're only declared on plugin start?

ArjixWasTaken:
> > All `findByProps` calls should only be called once, there is no need to call them multiple times.
> 
> I'm... not sure what you mean by that. It looks like they're all called only once to me. Or do you mean move them to someone where they're only declared on plugin start?

they are called whenever you create an embed, that doesn't seem only once to me

ActuallyTheSun:
yeah then good idea

Vendicated:
yeah you should use waitFor / lazyWebpack on the top level and reuse the same module to not unnecessarily fetch the module multiple times

Your patch also has a lot of inline logic, it'd be cleaner and nicer to maintain if the patch would only call a method on the plugin and have no logic in the actual patch

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,54 @@
...
+    }],
+    patches: [{
+        find: "_messageAttachmentToEmbedMedia",
+        replacement: {
+            match: /var (.{1,2})=(.{1,2})\.channel,(.{1,2})=.{1,2}\.message,(.{1,2})=.{1,2}\.renderSuppressEmbeds/,
+            replace: "const msgLink=$2.message.content?.match(Vencord.Plugins.plugins.MessageLinkEmbeds.messageLinkRegex) &&\
+            $2.message.content.match(Vencord.Plugins.plugins.MessageLinkEmbeds.messageLinkRegex)[1];\
+            if(msgLink)$2.message.embeds=[...$2.message.embeds,\
+            Vencord.Plugins.plugins.MessageLinkEmbeds.generateEmbed(msgLink, $2.message.embeds)].filter(item=>item);\
+            var $1=$2.channel,$3=$2.message,$4=$2.renderSuppressEmbeds"
```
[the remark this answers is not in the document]
ArjixWasTaken: you can do smth like
```javascript

const replacement = `
const msgLink = $2.message.content
    ?.match(Vencord.Plugins.plugins.MessageLinkEmbeds.messageLinkRegex)?.[1];

if (msgLink) {
    $2.message.embeds = [
        ...$2.message.embeds,
        Vencord.Plugins.plugins.MessageLinkEmbeds
          .generateEmbed(msgLink, $2.message.embeds)
    ].filter(item => item);
};

var 
  $1 = $2.channel,
  $3 = $2.message,
  $4 = $2.renderSuppressEmbeds;
`.trim().replace(/(?<=[^(?:const)|(?:var)])\s+/gm, "")

// ... and use the variable as is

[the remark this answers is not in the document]
Comment (ArjixWasTaken) [reply src/plugins/messageLinkEmbeds.ts:19]: this would be much easier to maintain as well

**Review by ArjixWasTaken — CHANGES_REQUESTED** [review]
Oops, forgot to submit my review...

File: src/plugins/messageLinkEmbeds.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (ActuallyTheSun) [reply src/plugins/messageLinkEmbeds.ts:19]: i tried to instead move most of the logic outside of the replacement string to a new function and i'm failing horribly at that so that's a better idea

File: src/plugins/messageLinkEmbeds.ts [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }],
+
+    messageLinkRegex: /https?:\/\/(?:\w+\.)?discord(?:app)?\.com\/channels\/((?:\d{17,19}|@me)\/\d{17,19}\/\d{17,19})/,
+
+    generateEmbed(messageURL: string, existingEmbeds: any[]) {
+        const [guildID, channelID, messageID] = messageURL.split("/");
+        if (existingEmbeds.find(i => i.id === `messageLinkEmbeds-${messageID}`)) return null;
+        const message = findByProps("getMessage").getMessage(channelID, messageID);
```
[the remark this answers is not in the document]
ArjixWasTaken: this review was written before ven commented, use `waitFor` / `lazyWebpack` instead

ActuallyTheSun:
okay so its really late for me and i wanna sleep so i'll just leave this as a dumpster fire and deal with it tomorrow

ArjixWasTaken:
Ok so, in `webpack/common.tsx`, add

```typescript
export const UserStore: Stores.UserStore = lazyWebpack(filters.byProps(["getCurrentUser", "initialize"]));
export const GuildStore: Stores.GuildStore = lazyWebpack(filters.byProps(["getGuildCount", "initialize"]));
export const ChannelStore: Stores.ChannelStore = lazyWebpack(filters.byProps(["getChannel", "initialize"]));
export const GuildMemberStore: Stores.GuildMemberStore = lazyWebpack(filters.byProps(["getMember", "initialize"]));
export const MessageStore: Stores.MessageStore = lazyWebpack(filters.byProps(["getMessage", "initialize"]));
```

and remove the conflicting `UserStore` lines

<br/>
<br/>
and in ur plugin change the code to

```typescript
import definePlugin from "../utils/types";
import { GuildStore, ChannelStore, MessageStore, GuildMemberStore } from "../webpack/common";

const replacement = `
const msgLink = $2.message.content?.match(Vencord.Plugins.plugins.MessageLinkEmbeds.messageLinkRegex)?.[1];

if (msgLink) {
    $2.message.embeds = [
        ...$2.message.embeds,
        Vencord.Plugins.plugins.MessageLinkEmbeds
          .generateEmbed(msgLink, $2.message.embeds)
    ].filter(item => item);
};

var
  $1 = $2.channel,
  $3 = $2.message,
  $4 = $2.renderSuppressEmbeds /* Don't add a semicolon here, there are more definitions after this. */
`.trim().replace(/(?<=[^(?:const)|(?:var)])\s+/gm, "");


export default definePlugin({
    name: "MessageLinkEmbeds",
    description: "Embeds message links",
    authors: [{
        name: "ActuallyTheSun",
        id: 406028027768733696n
    }],
    patches: [{
        find: "_messageAttachmentToEmbedMedia",
        replacement: {
            match: /var (.{1,2})=(.{1,2})\.channel,(.{1,2})=.{1,2}\.message,(.{1,2})=.{1,2}\.renderSuppressEmbeds/,
            replace: replacement
        }
    }],

    messageLinkRegex: /https?:\/\/(?:\w+\.)?discord(?:app)?\.com\/channels\/((?:\d{17,19}|@me)\/\d{17,19}\/\d{17,19})/,

    generateEmbed(messageURL: string, existingEmbeds: any[]) {
        const [guildID, channelID, messageID] = messageURL.split("/");
        if (existingEmbeds.find(i => i.id === `messageLinkEmbeds-${messageID}`)) return null;
        const message = MessageStore.getMessage(channelID, messageID);
        if (!message) return null;
        return {
            author: {
                iconProxyURL: `https://${window.GLOBAL_ENV.CDN_HOST}/avatars/${message.author.id}/${message.author.avatar}`,
                iconURL: `https://${window.GLOBAL_ENV.CDN_HOST}/avatars/${message.author.id}/${message.author.avatar}`,
                name: `${message.author.username}#${message.author.discriminator}`
            },
            color: GuildMemberStore.getMember(guildID, message.author.id)?.colorString,
            image: message.attachments.length != 0 ? {
                height: message.attachments[0].height,
                width: message.attachments[0].width,
                url: message.attachments[0].url,
                proxyURL: message.attachments[0].proxy_url
            } : (message.embeds[0]?.type === "image" ? { ...message.embeds[0].image } : null),
            rawDescription: message.content,
            footer: {
                text: "#" + ChannelStore.getChannel(channelID).name +
                    ` (${GuildStore.getGuild(guildID).name})`
            },
            timestamp: message.timestamp,
            id: `messageLinkEmbeds-${messageID}`,
            fields: [],
            type: "rich"
        };
    }
});
```

ActuallyTheSun:
the reason tenor gif's aren't embedded is because there's no convenient way of getting the gif itself, and the discord proxy only has the gif as a video which cannot be embedded. any non-tenor embeds should work though.

ArjixWasTaken:
Nooo

ArjixWasTaken:
Add the store definitions in `webpack/common.tsx`

ActuallyTheSun:
is it a good idea to add definitions to `webpack/common.tsx` in a pr that is most likely gonna get squash-merged

ArjixWasTaken:
Why would it not be?

ArjixWasTaken:
you still haven't changed it?

ActuallyTheSun:
i missed it 🗿

ArjixWasTaken:
So far so good, what about fetching messages that are not cached?

ActuallyTheSun:
currently no progress, but i think i know where to start
tomorrow i'll have a decent amount of time so i expect support to be added then

ArjixWasTaken:
worst case scenario you fetch the message using discord's REST API

but if you do that, make sure all the headers and stuff are present, so it is just like a normal request

ActuallyTheSun:
this is definitely not ready for a merge
- no support for actual embeds
- no support for videos or tenor gifs
- some other really bad code
will try to fix all of these today, worst case tomorrow

BanTheNons:
Maybe we can reuse the automod embed for this? Would look a lot better if that's possible

ActuallyTheSun:
> Maybe we can reuse the automod embed for this? Would look a lot better if that's possible

I've never seen an automod embed (i've never seen anything related to discord's automod) so idk. good idea though

somerandomcloud:
The automod embed is beautiful

ActuallyTheSun:
don't think it's possible, as the author of the automod message is the user that sent the message which got blocked, and there's no way to get it to show another user than the author. plus the message only has content and no image embeds/attachments/etc.

Cynosphere:
The component for it (pre-swc name: ChatMessageEmbedCard) isn't exported anymore due to module flattening.

ArjixWasTaken:
> The component for it (pre-swc name: ChatMessageEmbedCard) isn't exported anymore due to module flattening.

That ain't an issue since we can export it via a patch.

That is if we can find it via regex that is....

Cynosphere:
> If we can find it via regex that is....
`find: "().embedCard,"`, then figure out which function it is an export it.

ActuallyTheSun:
after probably close to 2 hours of trying, i can say i am not adding the automod embed right now. main reason is that adding a component to the message requires basic React knowledge which i do not have, and even then i'd have to look for ways for how to render everything that discord doesn't do automatically

also tried to simply add an `auto_moderator_message` to the embed but discord refuses to render it no matter what

ActuallyTheSun:
...good question

ActuallyTheSun:
yep, i give up on trying to add the automod message embed
the same error ("Invalid hook call. Hooks can only be called inside of the body of a function component.") has been stopping me from doing anything for 3 days, and i cant fix it (both because i have no idea what causes it and because i lack basic react knowledge)
[i'll still leave what i did so far here](https://cdn.discordapp.com/attachments/764071870336598016/1031225946704969758/messageLinkEmbeds.ts) (it crashes as soon as it sees a message link) so it doesn't go to waste, but i'm not fixing it

ActuallyTheSun:
i misclicked close with comment 🗿

Cynosphere:
MessageAccessoriesAPI is a thing now if you want to try again with the message card stuff without having to deal with embed limitations.

ActuallyTheSun:
yeah saw that, gonna mess with it later today or tomorrow

ActuallyTheSun:
added the automod embed 👍
does not embed video but should otherwise work
is there anything to put in settings other than the option to always use rich embeds?

src/plugins/messageLinkEmbeds.ts
```diff
@@ -106,6 +107,32 @@ interface Attachment {
     proxyURL?: string;
 }
 
+function getImages(message: Message): Attachment[] {
+    const attachments: Attachment[] = [];
+    if (message.attachments) message.attachments.forEach(a => {
```
ArjixWasTaken, line 112:
```suggestion
    message.attachments?.forEach(a => {
```

src/plugins/messageLinkEmbeds.ts
```diff
@@ -106,6 +107,32 @@ interface Attachment {
...
+        if (a.content_type!.startsWith("image/")) attachments.push({
+            height: a.height!,
+            width: a.width!,
+            url: a.url,
+            proxyURL: a.proxy_url!
+        });
+    });
+    if (!message.embeds) return attachments;
+    message.embeds.forEach(e => {
```
ArjixWasTaken, line 121:
```suggestion
    message.embeds?.forEach(e => {
```

ArjixWasTaken (commented):
some code style suggestions, nothing much

D3SOX:
I've been using this for a while now and it LGTM from a user perspective ^^

ArjixWasTaken:
:skull:

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    stop() {
+        // requires a restart to remove the rich embeds because that's not an accessory
+        // and even if they were removed all messages keep their embeds until rerendered
+        // is there a point in having a stop function?
+        removeAccessory("messageLinkEmbed");
+    },
+
+    // the > is kept to be checked for later; cause i have no idea how to make the whole regex fail if it's there
+    messageLinkRegex: /https?:\/\/(?:\w+\.)?discord(?:app)?\.com\/channels\/((?:\d{17,19}|@me)\/\d{17,19}\/\d{17,19}(?:>)?)/,
```
Vendicated, line 183:
```suggestion
    messageLinkRegex: /https?:\/\/(?:\w+\.)?discord(?:app)?\.com\/channels\/((?:\d{17,19}|@me)\/\d{17,19}\/\d{17,19}(?!>))/,
```
use a negative lookahead

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    messageEmbedAccessory(props: Record<string, any>): JSX.Element {
+        const { message } = props;
+        const Nothing = React.createElement("Fragment");
```
Vendicated, line 187: u can just return null

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+
+        const msgLink = message.content?.match(this.messageLinkRegex)?.[1];
+        if (!msgLink) return Nothing;
+        const [guildID, channelID, messageID] = msgLink.split("/");
```
Vendicated, line 191: why not use regex groups for this?

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+const messageCache: { [id: string]: { message?: Message, fetched: boolean; }; } = {};
+const elementCache: { [id: string]: { element: JSX.Element, shouldRenderRichEmbed: boolean; }; } = {};
+
+let get: (...query) => Promise<any>,
+    MessageEmbed: (...props) => JSX.Element,
+    parse: (content: string) => any[] /* (JSX.Element | string)[] i think */,
+    Endpoints: Record<string, any>;
+waitFor(["get", "getAPIBaseURL"], _ => ({ get } = _));
+waitFor(["MessageEmbed"], _ => ({ MessageEmbed } = _));
+waitFor(["parse", "parseTopic"], _ => ({ parse } = _));
```
Vendicated, line 43: use Parser from commons

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+        messageCache[message.id] = {
+            message: actualMessage,
+            fetched: true
+        };
+        if (originalMessage) dispatchBlankUpdate(originalMessage.channelID, originalMessage.messageID);
+    }
+    if (messageID in messageCache && !messageCache[messageID].fetched) return null;
+    if (messageCache[messageID]?.fetched) return callback(messageCache[messageID].message);
+    messageCache[messageID] = { fetched: false };
+    return get({
```
Vendicated, line 59: please use the Queue util to make sure this doesn't spam the api

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+            width: a.width!,
+            url: a.url,
+            proxyURL: a.proxy_url!
+        });
+    });
+    message.embeds?.forEach(e => {
+        if (e.type === "image") attachments.push(
+            e.image ? { ...e.image } : { ...e.thumbnail! }
+        );
+        if (e.type === "gifv" && !e.url!.match(/https:\/\/(?:www.)?tenor\.com/)) {
```
Vendicated, line 112: please use regex.test

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    start() {
+        addAccessory("messageLinkEmbed", props => this.messageEmbedAccessory(props), 4 /* just above rich embeds*/);
+    },
+
+    stop() {
+        // requires a restart to remove the rich embeds because that's not an accessory
+        // and even if they were removed all messages keep their embeds until rerendered
+        // is there a point in having a stop function?
+        removeAccessory("messageLinkEmbed");
```
Vendicated, line 179: you can keep track of the message ids and dispatch an update for each or perhaps attach a custom class to your embed somehow and use query selector to remove them. But imo stop cleanup is not that important, people should just restart

src/plugins/messageLinkEmbeds.ts
```diff
@@ -0,0 +1,344 @@
...
+
+        if (hasActualEmbed && !linkedMessage.content) {
+            elementCache[linkedMessage.id] = {
+                element: Nothing,
+                shouldRenderRichEmbed: true
+            };
+            return Nothing;
+        }
+
+        const MessageEmbedElement = React.createElement(MessageEmbed, {
```
Vendicated, line 222: please use jsx

Vendicated (changes requested):

Vendicated:
sorry for taking so long btw, thanks a lot for your effort!

src/plugins/messageLinkEmbeds.ts
Cynosphere: Alternatively there's a function with the pre-swc name of `tryParseChannelPath`

ActuallyTheSun:
image

PR #72, opened 2022-10-09
Adds an AV1 embed plugin

nate-moo:
enables sharing of AV1 encoded videos through a bodged link because discord sucks


Vendicated:
what's the use case?

Cynosphere:
I personally believe that plugins shouldn't depend on external services unless the service is deemed reliable by consensus (ex: GitHub, self-hosted alternatives to social media (as long as the base URL is configurable))

PR #69, opened 2022-10-08
feat: better extension dev + firefox support

DiamondMiner88:
Depends on #62 to prevent merge conflicts

- Adds web installing documentation
- Adds support for watching web build
- Support for launching chrome/firefox with test extension

PR #60, opened 2022-10-07
Plugin that removes canary subdomain from message

somerandomcloud:
Small plugin that removes `canary.` from message. Plinko, I guess

PR #74, opened 2022-10-09
Firefox based browsers: The PR

emir4169:
Makes it supported for firefox based stuff
and makes the description better


Vendicated:
This just combines #69 and #60 and for some reason includes moyai, what's the point

emir4169:
> This just combines #69 and #60 and for some reason includes moyai, what's the point

removing

PR #78, opened 2022-10-10
Add nix flake

pupbrained:
Still very much in progress, this is just currently some code that'll make it remotely work.


MeguminSama:
Why did you remove dist from git ignore 💀

pupbrained:
I was thinking of just making the folder pre-build for nix but pnpm would need internet to install and that's not allowed w/ the way this all works

MeguminSama:
why not use a separate repo for nix configs? I'm not a fan of the idea of adding scripts for every distro to the root of this project 💀

pupbrained:
Nix is a special case w/ flakes and it's much easier to work with it in-repo than making a completely new one specifically for it

That plus I don't think there's any other distros out there that would need this treatment

pupbrained:
The flake.nix and flake.lock are required to be in the root, but everything else can be in that scripts/nix folder

MeguminSama:
the idea of the scripts/nix folder is that the flake goes in there too, and the prepare:nix script (in package.json) will copy the flake files to root, where the user can then do stuff with it.

pupbrained:
> the idea of the scripts/nix folder is that the flake goes in there too, and the prepare:nix script (in package.json) will copy the flake files to root, where the user can then do stuff with it.

I'm not sure that's possible, since the first thing nix looks for on building is the flake and if it's not there it'll just fail

MeguminSama:
I'm on my phone but from what I gather through docs and guides...

git clone vencord
go into vencord folder
pnpm install
pnpm prepare:nix
nix-build

pupbrained:
> I'm on my phone but from what I gather through docs and guides...
> 
> git clone vencord go into vencord folder pnpm install pnpm prepare:nix nix-build

did the best i could do from this comment

pupbrained:
which is really hard w/ nix

pupbrained:
ok cool, dist no longer needs to be in the repo for nix to work

pupbrained:
...maybe trying to get nix working w/ pnpm instead of yarn is a useless endeavour

Cynosphere:
https://github.com/winterqt/pnpm2nix/tree/add-5.3-support

pupbrained:
> https://github.com/winterqt/pnpm2nix/tree/add-5.3-support

I didn't even see this, let me try editing my fork to reflect this

pupbrained:
Should be all good, finally got everything working w/ pnpm

PR #82, opened 2022-10-10
Fix Bug in isStaff module showing staff for all users

nmsturcke:
This fixes the bug (not sure how good the code is tho lol)


nmsturcke:
#83 just better

PR #83, opened 2022-10-10
fix(plugin): fix isStaff returning true for all users

BanTheNons:


src/plugins/isStaff.ts
```diff
@@ -5,18 +5,24 @@ export default definePlugin({
...
+            id: 460478012794863637n
+        }
+    ],
     patches: [
         {
             find: ".isStaff=function(){",
             replacement: [
                 {
-                    match: /(\w+)\.isStaff=function\(\){return\s*!1};/,
-                    replace: "$1.isStaff=function(){return true};",
```
Vendicated, line 15: why did you remove this patch?

BanTheNons: cuz its useless

Vendicated: then the last patch should also be removed because its only purpose is fixing isStaff() now returning true

BanTheNons: This patch does not matter because the function it's patching gets defined again a few lines later. The last patch is for that. You could keep this patch and set it to a random string and it would still not make a difference. You can try the fork to see it yourself.

Vendicated (approved):

PR #89, opened 2022-10-11
Implemented an owoify plugin

BanTheNons:
As you can see from its description, this plugin is different than uwuifier because it owoifies channels, guilds, and even menus. I've renamed the uwuifier plugin to "UwUify command" (with permission from @exhq) to help distinguish them. This was originally a console script written by @Benricheson101 a few months ago, I just turned it into a Vencord plugin with a few tweaks.


BanTheNons:
forgot to remove console.log lol

BanTheNons:
image
:(

PR #90, opened 2022-10-11
feat(plugin): UnminifyErrors

Cynosphere:


Vendicated:
since the json is fairly large, how about turning it into a github gist and fetching it on demand? Or if that's too tedious to implement, on plugin start

PR #94, opened 2022-10-12
added choices to CommandsAPI Options

MaiKokain:
current issue long image loading becuz bad website also dirty regex :husk:.


MaiKokain:
Added choices option, don't review.

MaiKokain:
removed libart.so plugin due to ads and images time limit (forgot about it)

PR #101, opened 2022-10-14
feat(plugin): Show Hidden Channels

sneksnake:
i've re-written it so it doesn't mess with permissions.
Hidden Channel


zachjmurphy:
It would be nice if the channel icon turned to a lock rather than adding [🔒] to the channel name.

ArjixWasTaken:
There is no stop method...disabling the plugin won't have an effect.

That's not nice.

sneksnake:
> There is no stop method...disabling the plugin won't have an effect.
> 
> That's not nice.

the plugin patches things so if you disable it you will have to restart discord.



> It would be nice if the channel icon turned to a lock rather than adding [🔒] to the channel name.

yeah i'm trying right now but the patcher doesn't want me to

ArjixWasTaken:
> the plugin patches things so if you disable it you will have to restart discord.

That's just an excuse, you can make it so that your callback only shows hidden channels when the plugin is enabled.

Even if the patch is still there until the user restarts, it doesn't mean that the user has to restart for it to take effect.

sneksnake:
> It would be nice if the channel icon turned to a lock rather than adding [🔒] to the channel name.

Channel with lock icon

BanTheNons:
> > the plugin patches things so if you disable it you will have to restart discord.
> 
> ~That's just an excuse, you can make it so that your callback only shows hidden channels when the plugin is enabled.~
> 
> ~Even if the patch is still there until the user restarts, it doesn't mean that the user has to restart for it to take effect.~
> 
> Edit: I reread the patches, yeah it ain't as simple as I described it, but it is not that hard either.
> 
> You can prob use
> 
> ```js
> Vencord.Plugins.plugins["Show Hidden Channels"].enabled
> ```
> 
> to access if the plugin is enabled directly from within the patches.
> 
> (I am on my phone and I can't test it)

Since it uses patches, it will prompt the user to restart the app anyway. Even when clicking "Later", the channels stay because the channel view is not updated unless it receives either CHANNEL_CREATE, CHANNEL_DELETE or CHANNEL_UPDATE. Switching guilds and coming back also removes the lock icon but doesn't hide the channel, so I think you should revert this. (or call handleViewUpdate somehow)

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,76 @@
...
+import { findByProps } from "../webpack";
+
+var can = (a, b) => true; // defaults to true while findByProps isn't loaded
+
+const VIEW_CHANNEL = 1024n;
+
+const VOICE_CHANNEL = 2;
+
+// https://github.com/Vendicated/Vencord/pull/101#issuecomment-1279601609
+var enabled = false;
```
x6r, line 12:
```suggestion
import { findByProps } from "../webpack";

const VIEW_CHANNEL = 1024n;
const VOICE_CHANNEL = 2;
```

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+    ],
+    patches: [
+        {
+            // RenderLevel defines if a channel is hidden, collapsed in category, visible, etc
+            // Here we just show every hidden channels
+            find: ".CannotShow",
+            replacement: {
+                match: /renderLevel:(\w+)\.CannotShow/g,
+                replace: "renderLevel:Vencord.Plugins.plugins[\"ShowHiddenChannels\"].isEnabled()?$1.Show:$1.CannotShow"
```
x6r, line 30:
```suggestion
                replace: "renderLevel:$1.Show"
```

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,76 @@
...
+            }
+        },
+        {
+            // Begining of the function that creates the elements.
+            // Changing the name here has a smaller footerprint than changing the permissions and real name
+            find: "l=e.hasActiveThreads",
+            replacement:
+            {
+                match: /hasActiveThreads,(.)=function\((\w+),(\w+),(\w+)\){switch\((\w+.type)\)/g,
+                replace: "hasActiveThreads,$1=function($2,$3,$4){switch(Vencord.Plugins.plugins[\"ShowHiddenChannels\"].getChannelTypeIcon($2))"
```
x6r, line 40:
```suggestion
                replace: "hasActiveThreads,$1=function($2,$3,$4){switch(Vencord.Plugins.plugins.ShowHiddenChannels.getChannelTypeIcon($2))"
```

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,76 @@
...
+        ({ can } = findByProps("can", "initialize"));
+        enabled = true;
+    },
+    stop() {
+        enabled = false;
+    },
+    isEnabled() {
+        return enabled;
+    }
+});
```
x6r, line 76:
```suggestion
    getChannelTypeIcon: (channel) => {
        const { can } = findByProps("can", "initialize");
        channel._isHiddenChannel = !can(VIEW_CHANNEL, channel);
        return channel._isHiddenChannel ? VOICE_CHANNEL : channel.type;
    },
});
```

x6r: i dont really understand what getChannelTypeIcon **actually** does but this is the way to do it probably:)

x6r (changes requested):

sneksnake (commented):
fixed in lastest commit + removed minified variables names from replacement + renamed "getChannelTypeIcon" to "isHiddenChannel"

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,39 @@
...
+                match: /renderLevel:(\w+)\.CannotShow/g,
+                replace: "renderLevel:$1.Show"
+            }
+        },
+        {
+            // This is where the logic that chooses the icon is, we overide it to be a locked voice channel if it's hidden
+            find: ".rulesChannelId))",
+            replacement: {
+                match: /(\w+)\.locked(.*?)switch\((\w+)\.type\)({case \w+\.\w+\.GUILD_ANNOUNCEMENT)/g,
+                replace: "Vencord.Plugins.plugins.ShowHiddenChannels.isHiddenChannel($3)?true:$1.locked$2switch($3._isHiddenChannel?2:$3.type)$4"
```
Vendicated, line 29:
```suggestion
                replace: "Vencord.Plugins.plugins.ShowHiddenChannels.isHiddenChannel($3)||$1.locked$2switch($3._isHiddenChannel?2:$3.type)$4"
```

ArjixWasTaken:
Ven is dum dum

sneksnake:
is there anything left to fix ?

src/plugins/showHiddenChannels.ts
```diff
@@ -0,0 +1,39 @@
...
+            find: ".rulesChannelId))",
+            replacement: {
+                match: /(\w+)\.locked(.*?)switch\((\w+)\.type\)({case \w+\.\w+\.GUILD_ANNOUNCEMENT)/g,
+                replace: "Vencord.Plugins.plugins.ShowHiddenChannels.isHiddenChannel($3)||$1.locked$2switch($3._isHiddenChannel?2:$3.type)$4"
+            }
+        }
+    ],
+    isHiddenChannel(channel) {
+        if (!channel) return false;
+        const { can } = findByProps("can", "initialize");
```
Vendicated, line 35: this should be moved to the top level with waitFor

Vendicated (approved):
looks good now. Please fix lint  then it can be merged

sneksnake:
👍

Vendicated:
Could you make it so that clicking a Hidden channel doesn't do anything? Or it could open a modal with more info on the channel like its topic

Currently it tries to load it which sends bad api requests

sneksnake:
Alright, i added a small model when you try to select an hidden channel.
The topic section doesn't appear if there is no topic.

NSFW channel with topic
Channel with no topic

Vendicated (approved):

Vendicated:
thank you!

PR #104, opened 2022-10-16
Add pronoundb plugin

TymanWasTaken:
Uses bulk requests to request all queued users every 300ms (can be changed), and caches all fetched pronouns (even if the request fails, it will not try to fetch it again)
This uses capitalized pronouns (`They/Them` as opposed to `they/them`) right now, but that can be changed

Also updated pnpm because it was being annoying about it

### DIFF
--- a/src/plugins/pronoundb/PronounComponent.tsx
+++ b/src/plugins/pronoundb/PronounComponent.tsx
@@ -1,0 +1,23 @@
+import { fetchPronouns } from "./utils";
+import { useAwaiter } from "../../utils/misc";
+import { MessageHeaderProps, PronounMapping } from "./types";
+import { findByProps } from "../../webpack";
+import { proxyLazy } from "../../utils";
+
+const styles: Record<string, string> = proxyLazy(() => findByProps("timestampInline"));
+
+export default function PronounComponent({ message }: MessageHeaderProps) {
+    // Don't bother fetching bot or system users
+    if (message.author.bot && message.author.system) return null;
+
+    const [result, error, isPending] = useAwaiter(() => fetchPronouns(message.author.id));
+
+    // If the fetching completed successfully and the result was not "unspecified", then return a span with the pronouns
+    if (!isPending && result && result !== "unspecified") return (
+        <span className={`${styles.timestampInline} ${styles.timestamp}`}>• {PronounMapping[result]}</span>
+    );
+    // If there was an error, log it
+    if (error) console.error("Error fetching pronouns: ", error);
+    // Return null so nothing else is rendered
+    return null;
+}
--- a/src/plugins/pronoundb/PronounComponent.tsx
+++ b/src/plugins/pronoundb/PronounComponent.tsx
@@ -1,23 +1,27 @@
 import { fetchPronouns } from "./utils";
-import { useAwaiter } from "../../utils/misc";
-import { MessageHeaderProps, PronounMapping } from "./types";
-import { findByProps } from "../../webpack";
-import { proxyLazy } from "../../utils";
+import { classes, lazyWebpack, useAwaiter } from "../../utils/misc";
+import { PronounMapping } from "./types";
+import { filters } from "../../webpack";
+import { Message } from "discord-types/general";
 
-const styles: Record<string, string> = proxyLazy(() => findByProps("timestampInline"));
+const styles: Record<string, string> = lazyWebpack(filters.byProps(["timestampInline"]));
 
-export default function PronounComponent({ message }: MessageHeaderProps) {
+export default function PronounComponent({ message }: { message: Message; }) {
     // Don't bother fetching bot or system users
     if (message.author.bot && message.author.system) return null;
 
-    const [result, error, isPending] = useAwaiter(() => fetchPronouns(message.author.id));
-
-    // If the fetching completed successfully and the result was not "unspecified", then return a span with the pronouns
-    if (!isPending && result && result !== "unspecified") return (
-        <span className={`${styles.timestampInline} ${styles.timestamp}`}>• {PronounMapping[result]}</span>
+    const [result, , isPending] = useAwaiter(
+        () => fetchPronouns(message.author.id),
+        null,
+        e => console.error("Fetching pronouns failed: ", e)
     );
-    // If there was an error, log it
-    if (error) console.error("Error fetching pronouns: ", error);
-    // Return null so nothing else is rendered
-    return null;
+
+    // If the promise completed, the result was not "unspecified", and there is a mapping for the code, then return a span with the pronouns
+    if (!isPending && result && result !== "unspecified" && PronounMapping[result]) return (
+        <span
+            className={classes(styles.timestampInline, styles.timestamp)}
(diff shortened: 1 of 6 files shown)


src/plugins/pronoundb/PronounComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default function PronounComponent({ message }: MessageHeaderProps) {
+    // Don't bother fetching bot or system users
+    if (message.author.bot && message.author.system) return null;
+
+    const [result, error, isPending] = useAwaiter(() => fetchPronouns(message.author.id));
+
+    // If the fetching completed successfully and the result was not "unspecified", then return a span with the pronouns
+    if (!isPending && result && result !== "unspecified") return (
+        <span className={`${styles.timestampInline} ${styles.timestamp}`}>• {PronounMapping[result]}</span>
```
Vendicated, line 17:
```suggestion
        <span className={`${styles.timestampInline} ${styles.timestamp}`}>• {PronounMapping[result] ?? result}</span>
```

shouldn't happen but just in case, noone wants to see undefined

Vendicated, line 17:
```suggestion
        <span className={classes(styles.timestampInline, styles.timestamp)}>• {PronounMapping[result]}</span>
```
and add import ofc

src/plugins/pronoundb/PronounComponent.tsx
```diff
@@ -0,0 +1,23 @@
+import { fetchPronouns } from "./utils";
+import { useAwaiter } from "../../utils/misc";
+import { MessageHeaderProps, PronounMapping } from "./types";
+import { findByProps } from "../../webpack";
+import { proxyLazy } from "../../utils";
+
+const styles: Record<string, string> = proxyLazy(() => findByProps("timestampInline"));
```
Vendicated, line 7:
```suggestion
const styles: Record<string, string> = lazyWebpack(filters.byProps(["timestampInline"]));
```

src/plugins/pronoundb/PronounComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+    if (message.author.bot && message.author.system) return null;
+
+    const [result, error, isPending] = useAwaiter(() => fetchPronouns(message.author.id));
+
+    // If the fetching completed successfully and the result was not "unspecified", then return a span with the pronouns
+    if (!isPending && result && result !== "unspecified") return (
+        <span className={`${styles.timestampInline} ${styles.timestamp}`}>• {PronounMapping[result]}</span>
+    );
+    // If there was an error, log it
+    if (error) console.error("Error fetching pronouns: ", error);
```
Vendicated, line 20: you should wrap this in an useEffect with the error as dependency so it's logged only once and not every time

Vendicated: alternatively, perhaps add some form of `onError` callback to useAwaiter?

src/plugins/pronoundb/types.ts
```diff
@@ -0,0 +1,40 @@
+// Not the full props, only what is necessary for pronoundb plugin to work
+export interface MessageHeaderProps {
+    message: {
```
Vendicated, line 3: you can import type "Message" from discord-types package

src/plugins/pronoundb/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+    });
+    return await req.json()
+        .then((res: PronounsResponse) => {
+            for (const [id, pronouns] of Object.entries(res)) {
+                cache[id] = pronouns;
+            }
+            return res;
+        })
+        .catch(e => {
```
Vendicated, line 51: this catch will only catch errors for req.json(), not errors for the fetch() itself. Is that intended?

TymanWasTaken: Nope, I didn't realize that

src/plugins/pronoundb/PronounComponent.tsx
TymanWasTaken: Ah, I tried lazyWebpack but it didn't work so I just used a lazy proxy
Does `lazyWebpack("timestampInline")` not work? I thought it should

src/plugins/pronoundb/PronounComponent.tsx
TymanWasTaken: no one also wants to see two letters
wouldn't just not showing anything if there is no mapping work better?

Vendicated: i guess so

src/plugins/pronoundb/PronounComponent.tsx
Vendicated: it expects a filter function. I'm not sure why i didn't also allow passing props shortcut, probably just forgot

src/plugins/pronoundb/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+// A map of cached pronouns so the same request isn't sent twice
+const cache: Record<string, PronounCode> = {};
+// A map of ids and callbacks that should be triggered on fetch
+const requestQueue: Record<string, ((pronouns: PronounCode) => void)[]> = {};
+
+// Executes all queued requests and calls their callbacks
+const bulkFetch = debounce(async () => {
+    const ids = Object.keys(requestQueue);
+    const pronouns = await bulkFetchPronouns(ids);
+    for (const [id, callbacks] of Object.entries(requestQueue)) {
```
Vendicated, line 13: you should loop over the "ids" local here.
Otherwise I'm pretty sure there could be a race condition where between the bulkFetch call and this, more ids are added to the queue, since the event loop processes new events while waiting for the fetch to complete

src/utils/misc.tsx
```diff
@@ -30,18 +30,19 @@ export function lazyWebpack<T = any>(filter: FilterFn): T {
...
+        error: null,
         pending: true
     });
 
     React.useEffect(() => {
         let isAlive = true;
         factory()
             .then(value => isAlive && setState({ value, error: null, pending: false }))
-            .catch(error => isAlive && setState({ value: null, error, pending: false }));
+            .catch(error => isAlive && setState({ value: null, error, pending: false }) || onError?.(error));
```
Vendicated, line 45: why is this an ||?

src/plugins/pronoundb/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    try {
+        const req = await fetch("https://pronoundb.org/api/v1/lookup-bulk?" + params, {
+            method: "GET",
+            headers: {
+                "Accept": "application/json"
+            }
+        });
+        return await req.json()
+            .then((res: PronounsResponse) => {
+                for (const [id, pronouns] of Object.entries(res)) {
```
Vendicated, line 48: Object.assign(cache, res)

src/utils/misc.tsx
TymanWasTaken: setState returns void so && wouldn't work

Vendicated (approved):

PR #105, opened 2022-10-16
add LinkChannels plugin

E-boi:
adds a link button to the channel item that copies the channel link

### DIFF
--- a/src/plugins/linkChannels.tsx
+++ b/src/plugins/linkChannels.tsx
@@ -1,0 +1,47 @@
+import { popNotice, showNotice } from "../api/Notices";
+import { classes, lazyWebpack } from "../utils";
+import definePlugin from "../utils/types";
+import { filters } from "../webpack";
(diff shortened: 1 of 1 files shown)


src/plugins/linkChannels.tsx
```diff
@@ -0,0 +1,47 @@
+import { popNotice, showNotice } from "../api/Notices";
+import { classes, lazyWebpack } from "../utils";
+import definePlugin from "../utils/types";
+import { filters } from "../webpack";
+import { Tooltip } from "../webpack/common";
+const icon = lazyWebpack(filters.byProps(["iconItem"]));
+
+function onClick(channel) {
+    if (!IS_WEB) {
+        window.window.DiscordNative.clipboard.copy(`<#${channel.id}>`);
```
Cynosphere, line 10: Replace with #127 if/when it gets merged.

PR #106, opened 2022-10-16
Add the VENCORD_USER_DATA_DIR environment variable

Kyiro:


src/ipcMain/index.ts
```diff
@@ -7,7 +7,11 @@ import IpcEvents from "../utils/IpcEvents";
 
 import "./updater";
 
-const DATA_DIR = join(app.getPath("userData"), "..", "Vencord");
+const DATA_DIR = process.env.VENCORD_USER_DATA_DIR ?? (
+    process.env.DISCORD_USER_DATA_DIR ?
+        join(process.env.DISCORD_USER_DATA_DIR, "..", "Vencord") :
```
Vendicated, line 12:
```suggestion
        join(process.env.DISCORD_USER_DATA_DIR, "..", "VencordData") :
```
I think this would be better cause otherwise it might conflict with Vencord source dir / be confusing

PR #111, opened 2022-10-17
show only the required plugins that are enabled

stijnvdkolk:
At the moment, all plugins that need the required plugins are shown, instead of only the enabled plugins.


src/components/PluginSettings/index.tsx
```diff
@@ -230,5 +230,6 @@ function makeDependencyList(deps: string[]) {
 }
 
 function dependencyCheck(pluginName: string, depMap: Record<string, string[]>): string[] {
-    return depMap[pluginName] || [];
+    const settings = useSettings();
```
Vendicated, line 233: just access Settings directly, this method also creates a state under the hood and does a bunch of unnecessary logic

you can import it from the same file

PR #119, opened 2022-10-19
feat(plugin): Ify improvements

Nuckyz:
Improve Spotify premium bypass, turn disable auto pause into an option and add option to keep the activity playing when idling

### DIFF
--- a/src/plugins/ify.ts
+++ b/src/plugins/ify.ts
@@ -27,7 +27,7 @@ export default definePlugin({
         predicate: () => Settings.plugins.Ify.keepSpotifyActivityOnIdle,
         replacement: {
             match: /(shouldShowActivity=function\(\){.{1,50})&&!.{1,6}\.isIdle\(\)(.{0,}?})/,
-            replace: "$1$2"
+            replace: (_, functionDeclarationAndExpression, restOfFunction) => `${functionDeclarationAndExpression}${restOfFunction}`
         }
     }],
     options: {
--- a/src/plugins/ify.ts
+++ b/src/plugins/ify.ts
@@ -1,18 +1,47 @@
 import { Devs } from "../utils/constants";
-import definePlugin from "../utils/types";
+import definePlugin, { OptionType } from "../utils/types";
+import { Settings } from "../Vencord";
 
 export default definePlugin({
     name: "Ify",
-    description: "Disables Spotify auto-pausing and Premium checks, allowing you to listen along with others.",
-    authors: [Devs.Cyn],
+    description: "Disables Spotify auto-pausing, allows activity to continue playing when idling and bypasses premium checks, allowing you to listen along with others.",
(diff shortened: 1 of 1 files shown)


Ruthenic:
Why in the hell would you want to re-enable autopausing

Nuckyz:
> Why in the hell would you want to re-enable autopausing

It doesn't hurt to have it as an option

src/plugins/ify.ts
```diff
@@ -1,18 +1,47 @@
...
+    ],
     patches: [{
-        find: '.displayName="SpotifyStore"',
+        find: 'dispatch({type:"SPOTIFY_PROFILE_UPDATE"',
         replacement: [{
-            match: /\.isPremium=.;/,
-            replace: ".isPremium=true;"
-        }, {
+            match: /(function\((.{1,2})\){)(.{1,6}dispatch\({type:"SPOTIFY_PROFILE_UPDATE")/,
+            replace: '$1$2.body.product=\"premium\";$3'
```
Vendicated, line 16: what's the point of changing this patch?

Nuckyz: I haven't found any difference but that patches the very source of where isPremium comes from, seems better to avoid issues

Vendicated: I disagree, it's much more complex than the old one so I would prefer the old one

src/plugins/ify.ts
```diff
@@ -1,18 +1,47 @@
...
+        predicate: () => Settings.plugins["Ify"].noSpotifyAutoPause === true,
+        replacement: {
             match: /function (.{1,2})\(\).{0,200}SPOTIFY_AUTO_PAUSED\);.{0,}}}}/,
             replace: "function $1(){}"
-        }]
-    }]
+        }
+    }, {
+        find: '.displayName="SpotifyStore"',
+        predicate: () => Settings.plugins["Ify"].keepSpotifyActivityOnIdle === true,
```
Vendicated, line 27: please prefer direct property access (Settings.plugins.Ify) and remove the comparison to true, it is obsolete

Nuckyz: I took inspiration from the experiments plugin, that's why it's like, gonna change it

Vendicated: Then that should be changed as well (not in this PR). Maybe there's an eslint rule which can enforce this in the future? 🤔

Vendicated:
could you please change the patches from using $1$2 to using a function?
`replace: (fullMatch, firstGroup, secondGroup) => ""`
this way its more readable since thanks to named groups you don't have to decipher what each group stands for

src/plugins/ify.ts
Nuckyz: Can this be considered resolved?

Vendicated (approved):

PR #121, opened 2022-10-19
add(BANger): Added settings to BANger

GlitchChan:
Adds a simple setting to change the video to whatever

### DIFF
--- a/src/plugins/banger.tsx
+++ b/src/plugins/banger.tsx
@@ -1,4 +1,4 @@
-import definePlugin from "../utils/types";
+import definePlugin, { OptionType } from "../utils/types";
 import { Devs } from "../utils/constants";
 
 export default definePlugin({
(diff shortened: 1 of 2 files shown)


Vendicated:
why the rename to tsx? it doesn't matter I'm just curious

src/plugins/banger.tsx
```diff
@@ -16,8 +16,16 @@ export default definePlugin({
             find: "BAN_CONFIRM_TITLE.",
             replacement: {
                 match: /src:\w\(\d+\)/g,
-                replace: 'src: "https://i.imgur.com/wp5q52C.mp4"'
+                replace: "src: Settings.plugins.BANger.source"
```
Vendicated, line 19: shouldn't this be 
```suggestion
                replace: "src: Vencord.Settings.plugins.BANger.source"
```
remember that not everyone has console shortcuts enabled

GlitchChan:
> why the rename to tsx? it doesn't matter I'm just curious

Because I was messing with things and forgot to change it back lol

src/plugins/banger.tsx
GlitchChan: Ah I did not know that was apart of console shortcuts

Vendicated: for future ref the only guaranteed window props are Vencord and VencordNative, anything else is console shortcuts

Vendicated (approved):

PR #127, opened 2022-10-20
feat(api): add Clipboard entry in webpack common

Cynosphere:
Gives access to copying on web if browser supports it.


Vendicated:
Does this just call navigator.clipboard.writeText?

Cynosphere:
No, it checks if its embedded (desktop app) and uses DiscordNative.copy, otherwise if it can use copy it'll offload to another module that does things with a textarea element and ranges.

I just think it's smarter to use existing function in the client if they exist than relying on DiscordNative.

Vendicated (approved):

Vendicated (changes requested):
there are multiple modules matching this filter, I matched some random react component

Vendicated (approved):

PR #29, opened 2022-10-02
feat(plugin): add KyzaMode

Ruthenic:

PR #139, opened 2022-10-22
New plugin: capitalize

luximus-hunter:
Please let me know if i need to improve stuff 😓

### DIFF
--- a/src/plugins/capitalize.ts
+++ b/src/plugins/capitalize.ts
@@ -16,9 +16,9 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import definePlugin, { OptionType } from "../../utils/types";
-import { addPreEditListener, addPreSendListener, MessageObject, removePreEditListener, removePreSendListener } from "../../api/MessageEvents";
-import { Settings } from "../../Vencord";
+import definePlugin, { OptionType } from "../utils/types";
+import { addPreEditListener, addPreSendListener, MessageObject, removePreEditListener, removePreSendListener } from "../api/MessageEvents";
+import { Settings } from "../Vencord";
 
 const blacklistSeperator = ";";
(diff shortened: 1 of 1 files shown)


src/plugins/capitalize.ts
```diff
@@ -0,0 +1,67 @@
...
+export default definePlugin({
+    name: "Remove Chat Buttons",
+    description: "Make every first letter of a sentence a capital letter.",
+    authors: [
+        {
+            id: 306692821358739456n,
+            name: "Luximus",
+        },
+    ],
+    patches: [],
```
MaiKokain, line 35: remove if not being used

luximus-hunter: My bad!, I removed it

src/plugins/capitalize.ts
```diff
@@ -0,0 +1,66 @@
...
+import definePlugin, { OptionType } from "../utils/types";
+import { addPreEditListener, addPreSendListener, MessageObject, removePreEditListener, removePreSendListener } from "../api/MessageEvents";
+import { Settings } from "../Vencord";
+
+const blacklistSeperator = ";";
+const urlRegex = new RegExp("^((http[s]?|ftp):\/)?\/?([^:\/\s]+)((\/\w+)*\/)([\w\-\.]+[^#?\s]+)(.*)?(#[\w\-]+)?$");
+
+export default definePlugin({
+    name: "Remove Chat Buttons",
+    description: "Make every first letter of a sentence a capital letter.",
```
taskylizard, line 28: i don't think this is intentional

Vendicated:
Thanks! But to be honest, this is pretty useless. Why Would Anyone Want To Talk Like This?

You can post it in the third party plugins channel if you want though

luximus-hunter:
> Why Would Anyone Want To Talk Like This?

It doesn't capitalize each word in a sentence. It just does the first word of it.

Vendicated:
Oh l see, my bad. In that case however it should be part of the KyzaMode plugin that already has an open PR

Vendicated:
#29

luximus-hunter:
Alright! I didn't see it as it has a name that doesn't suggest it 😓 

Thanks either way tho!

PR #150, opened 2022-10-23
fix: plugin dependencies not enabling

MeguminSama:


src/components/PluginSettings/index.tsx
```diff
@@ -48,37 +52,90 @@ function showErrorToast(message: string) {
...
+interface ReloadRequiredCardProps extends React.HTMLProps<HTMLDivElement> {
+    plugins: string[];
+}
+
+function ReloadRequiredCard({ plugins, ...props }: ReloadRequiredCardProps) {
+    if (plugins.length === 0) return null;
+
+    const pluginText = plugins.length === 1
+        ? `The plugin ${plugins[0]} requires a reload to apply changes.`
+        : `The following plugins require a reload to apply changes: ${plugins.join(", ")}.`;
```
Vendicated, line 64: why not wrap the plugins in ` like before?

src/components/PluginSettings/index.tsx
```diff
@@ -48,37 +52,90 @@ function showErrorToast(message: string) {
...
         Modals.openModalLazy(async () => {
             return modalProps => {
-                return <PluginModal {...modalProps} plugin={plugin} onRestartNeeded={onRestartNeeded} />;
+                return <PluginModal {...modalProps} plugin={plugin} onRestartNeeded={() => onRestartNeeded(plugin.name)} />;
             };
         });
     }
 
     function toggleEnabled() {
         const enabled = isEnabled();
```
Vendicated, line 99: since this is the old state it's kinda confusing to read, so please rename to wasEnabled or something along those lines

src/components/PluginSettings/index.tsx
MeguminSama: Not sure I understand what you mean. They are wrapped in `?

somerandomcloud: image
 maybe something like this? I have no clue

MeguminSama: Oh, <code> tags? That's doable.

PR #153, opened 2022-10-24
Add RemoveNitroAds plugin

ayes-web:
Adds a plugin that removes gift button and nitro button from dm list

### DIFF
--- a/src/plugins/removeNitroAds.ts
+++ b/src/plugins/removeNitroAds.ts
@@ -1,0 +1,60 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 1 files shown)


MeguminSama (changes requested):

Vendicated:
Thank you! However I don't really see the point of this because this is easier done with css and that won't break in the future. It also doesn't even cover all cases, there is for example still an obnoxious header on top of people's banners

ayes-web:
> Thank you! However I don't really see the point of this because this is easier done with css and that won't break in the future. It also doesn't even cover all cases, there is for example still an obnoxious header on top of people's banners

I feel like being able to easily configure which parts i want turned off is an improvement over CSS, you can't really do that with a CSS theme (without modifying the theme itself manually)

Vendicated:
Someone could make a css toggler that lets you add and toggle css snippets

ayes-web:
> Someone could make a css toggler that lets you add and toggle css snippets

Pretty good idea, might make it myself

Vendicated:
that'd be awesome, I can help you with that if you want

ayes-web:
Yeah sure :)
I guess i will close this PR and make a new one later

PR #157, opened 2022-10-25
feat: add new plugin ForceOwnerCrown

D3SOX:
~~See bfd20424971556edd3e628e9de1e31e081c453ed, let me know which replacement you like more.
The previous one might be more robust because if they change something it would most likely just no longer work, whereas with the current one it could possibly break.~~
Edit: Restored the old patch

### DIFF
--- a/src/plugins/forceOwnerCrown.ts
+++ b/src/plugins/forceOwnerCrown.ts
@@ -32,7 +32,7 @@ export default definePlugin({
     patches: [
         {
             // This is the logic where it decides whether to render the owner crown or not
-            find: ".renderOwner",
+            find: ".renderOwner=",
             replacement: {
                 match: /isOwner;return null!=(\w+)?&&/g,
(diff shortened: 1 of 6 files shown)


src/plugins/forceOwnerCrown.ts
```diff
@@ -0,0 +1,56 @@
...
+export default definePlugin({
+    name: "ForceOwnerCrown",
+    description: "Force the owner crown next to usernames even if the server is large.",
+    authors: [Devs.D3SOX],
+    patches: [
+        {
+            // This is the logic where it decides whether to render the owner crown or not
+            find: ".renderOwner=",
+            replacement: {
+                match: /(\w+).isOwner;/g,
```
Vendicated, line 35: this looks really sketchy

D3SOX: wdym?

Vendicated (approved):

PR #159, opened 2022-10-25
fix(vcDoubleClick): exclude text channel mentions

D3SOX:


Vendicated:
thanks! I noticed this as well but was too lazy to fix it

PR #161, opened 2022-10-25
feat(settings): add beforeSave check

MeguminSama:
This PR adds support for a `beforeSave` method that accepts synchronous and asynchronous functions. If `beforeSave` returns true, then the plugin will save & close the modal. If it returns a string, it will display this string to the user and not save.

### DIFF
--- a/src/components/PluginSettings/PluginModal.tsx
+++ b/src/components/PluginSettings/PluginModal.tsx
@@ -67,6 +67,7 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
     const [tempSettings, setTempSettings] = React.useState<Record<string, any>>({});
 
     const [errors, setErrors] = React.useState<Record<string, boolean>>({});
+    const [saveError, setSaveError] = React.useState<string | null>(null);
 
     const canSubmit = () => Object.values(errors).every(e => !e);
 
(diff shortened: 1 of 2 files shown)


src/components/PluginSettings/PluginModal.tsx
```diff
@@ -79,11 +80,20 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
...
 
-    function saveAndClose() {
+    async function saveAndClose() {
         if (!plugin.options) {
             onClose();
             return;
         }
+
+        if (plugin.beforeSave) {
+            const result = await Promise.resolve(plugin.beforeSave(tempSettings));
```
Vendicated, line 90: what's this Promise.resolve() for?

src/utils/types.ts
```diff
@@ -75,6 +75,11 @@ interface PluginDef {
      * Optionally provide settings that the user can configure in the Plugins tab of settings.
      */
     options?: Record<string, PluginOptionsItem>;
+    /**
+     * Check that this returns true before allowing a save to complete.
+     * If a string is returned, show the error to the user.
+     */
+    beforeSave?(options: Record<string, any>): (true | string) | Promise<true | string>;
```
Vendicated, line 82: You can use Promisable<true | string> from type-fest

PR #165, opened 2022-10-26
feat: custom components in settings

MeguminSama:


src/components/PluginSettings/components/SettingCustomComponent.tsx
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { PluginOptionComponent } from "../../../utils/types";
+import { ISettingElementProps } from ".";
+
+export function SettingCustomComponent({ option, onChange, onError }: ISettingElementProps<PluginOptionComponent>) {
+    return option.component({ onChange, setError: onError });
```
Vendicated, line 23: why not pass the option again? more comfy than using this imo

MeguminSama: Passing the option is useless though since the only external thing the user would ever need is onChange and setError. I don't believe there is anything in the `option` object that the developer would need.

The idea was to make the API as simple as possible to use e.g.

```jsx
{
	component: ({ onChange, setError }) => { /* do stuff with onChange and setError */ }
}
```

Vendicated: yeah but they don't have to destructure the option if they don't need it, so it doesn't make a difference. The reasoning behind passing the option is that the person may want to reuse the same component for multiple options

Vendicated (approved):

PR #167, opened 2022-10-27
fix(ShowHiddenChannels): better patch & visual bug fix

sneksnake:


D3SOX (approved):
LGTM, thx!

D3SOX:
This broke right-clicking on a hidden channel which could previously be used to mute them

ArjixWasTaken:
> This broke right-clicking on a hidden channel which could previously be used to mute them

Uhh...muting a hidden channel is useless, you should instead mention copying the ID of a hidden channel

D3SOX:
> muting a hidden channel is useless

I don't think so, I use it to clean up channels I never use, using the "Hide Muted Channels" option

ArjixWasTaken:
> > muting a hidden channel is useless
> 
> Not really a major issue but I don't think so, I use it to clean up channels I never use, using the "Hide Muted Channels" option

Hmm, then shouldn't we just mark the hidden channels as muted?

Like, it makes sense cause you won't get notifications from them...

D3SOX:
Already fixed by #170

PR #169, opened 2022-10-27
style(showHiddenChannels): change modal text

xafn:
* removes 'you don't have permission' text
* changes button text to 'close'

### DIFF
--- a/src/plugins/showHiddenChannels.tsx
+++ b/src/plugins/showHiddenChannels.tsx
@@ -140,8 +140,7 @@ export default definePlugin({
                             ))}
                         </Flex>
                     </ModalHeader>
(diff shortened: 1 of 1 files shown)


D3SOX:
I agree with the close button, but why remove that text?

somerandomcloud:
Yeah, why remove the text? Makes it easier to notice if you have access or not. Honestly, some aren't the sharpest tools in the shed and need the extra drop of info

D3SOX:
Can't see the issue because a modal like this is never used to indicate that you **have** access.
I think the text doesn't hurt, only benefits the peeps you mentioned, and it even looks better with it because it looks very empty without.

D3SOX:
I think it's good with the text, maybe add a lock icon to the modal title on the left

sneksnake:
i don't think we should remove the text too
but i renamed the button in my latest pr
https://github.com/Vendicated/Vencord/pull/170

PR #172, opened 2022-10-28
fix(vcDoubleClick): don't require double-click on active voice channel, fix stage channels patch

D3SOX:


src/plugins/vcDoubleClick.ts
```diff
@@ -63,6 +64,13 @@ export default definePlugin({
     schedule(cb: () => void, e: any) {
         // support from stage and voice channels patch
         const id = e?.id ?? e.props.channel.id as string;
+        const activeVoiceChannel = SelectedChannelStore.getVoiceChannelId();
+        if (activeVoiceChannel) {
```
Vendicated, line 68:
```suggestion
        if (SelectedChannelStore.getVoiceChannelId() === id) {
```

D3SOX: Why? The suggestion doesn't make sense, what did you want to achieve?

Vendicated: the outer if is unnecessary, you can just do the comparison right away. now you can also eliminate the variable

D3SOX: I just wanted to be sure that it's not undefined and `id` is also `undefined` and then it's equal but yeah

Vendicated: Id shouldn't ever be undefined, should it? in any case, I would just add a separate check that returns early if id is undefined, that's way prettier

D3SOX:
For the stage channels patch: I'm not sure if this is different in Canary and Stable right now. I can't test it with Canary rn but maybe it should cover both ways

Vendicated:
special canary treatment is kinda pointless and will be very cumbersome, I would just make sure it works on stable

D3SOX:
As of this PR, the patch works in stable

src/plugins/vcDoubleClick.ts
```diff
@@ -46,8 +47,8 @@ export default definePlugin({
                 },
                 // stage channels
                 {
-                    match: /onClick:(\w+)\?void 0:this\.handleClick,/g,
-                    replace: "onClick:$1?void 0:(...args)=>Vencord.Plugins.plugins.vcDoubleClick.schedule(()=>{this.handleClick(...args);}, args[0]),",
+                    match: /onClick:(.*)this\.handleClick,/g,
```
Vendicated, line 50: This greedy operator is dangerous, do something else or limit it to as litte chars as possible, such as `.{0,20}`

Vendicated (changes requested):

src/plugins/vcDoubleClick.ts
D3SOX: Updated

src/plugins/vcDoubleClick.ts
```diff
@@ -63,6 +64,10 @@ export default definePlugin({
     schedule(cb: () => void, e: any) {
         // support from stage and voice channels patch
         const id = e?.id ?? e.props.channel.id as string;
+        if (SelectedChannelStore.getVoiceChannelId() === id) {
+            cb();
```
Vendicated, line 68: nitpick but u can just `return void cb()`

D3SOX: Didn't know about that syntax and got this warning by SonarLint lol

image

PR #173, opened 2022-10-29
feat(plugin): ClipboardImageFix

Cynosphere:
Fixes not being able to paste images from Firefox into Discord on Linux. May also fix double pasting from Firefox on Windows, but that is unconfirmed currently.


Vendicated:
Updated the description because it doesn't only happen on Firefox. It fixes the problem nicely though, thanks I actually hated this bug!

PR #174, opened 2022-10-29
Implement Subcommands; fix errors due to Settings <-> Plugins circular imports

Vendicated:
Discord implements subcommands very strangely, in that each sub command is registered as a top level command with special properties subCommandPath and rootCommand. This pr makes Vencord's CommandsAPI transform subcommands to this format (at least 1 level deep, I haven't tested nested subcommands)

Another bug is that if any plugin requires SettingsAPI, then SettingsAPI is ran before plugin init, meaning that Plugins[] will be empty and the default settings merge will fail which leads to Settings.plugins.Foo being undefined. This pr also fixes that by moving that logic inside the Proxy


Vendicated:
Lol what this causes a WebContents crash

Vendicated:
Oh cause i shadowed patches and push to the array i'm looping over so it's an infinite loop

src/api/Commands/index.ts
```diff
@@ -88,11 +93,36 @@ function modifyOpt(opt: Option | Command) {
...
-        modifyOpt(opts[i]);
+        prepareOption(opts[i]);
+    });
+    return opt;
+}
+
+// Yes, Discord registers individual commands for each subcommand
+// TODO: This probably doesn't support nested subcommands. If that is ever needed,
+// investigate
+function registerSubCommands(cmd: Command, plugin: string) {
```
Vendicated, line 104: I just realised that those won't get unregistered correctly. I don't really care right now, that can be fixed later 🚎 :trollface:

MeguminSama (approved):

src/api/Commands/index.ts
MeguminSama: 🚎

PR #180, opened 2022-10-30
fix forceOwnerCrown Plugin Spamming Errors in Console

nmsturcke:
Fix the plugin spamming errors in console when checking group chats, instead, log them as debugging for people to see if they want to enable verbose console logs.

image

### DIFF
--- a/src/plugins/forceOwnerCrown.ts
+++ b/src/plugins/forceOwnerCrown.ts
@@ -26,7 +26,7 @@ waitFor(["getGuild"], m => GuildStore = m);
 export default definePlugin({
     name: "ForceOwnerCrown",
     description: "Force the owner crown next to usernames even if the server is large.",
-    authors: [Devs.D3SOX],
+    authors: [Devs.D3SOX, Devs.Nickyux],
     patches: [
         {
             // This is the logic where it decides whether to render the owner crown or not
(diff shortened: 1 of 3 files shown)


D3SOX:
I'd rather add a check if it's a guild than just adjusting the log level

Vendicated:
"no guildid or userid" yet userid is defined? this seems more like a bug than having to adjust log level

Vendicated:
uhh wrong button

D3SOX:
I've looked into this and it seems like it's using the same method for both cases. I'd just remove the else parte as I haven't seen any other good indicator to differentiate between groups and guilds here.

nmsturcke:
Another option is if the guild isn't in the GuildStore just ignore

D3SOX:
wdym? Checking the GuildStore wouldn't help as it we need to decide upfront from where the method got triggered.

nmsturcke:
@D3SOX

src/plugins/forceOwnerCrown.ts
```diff
@@ -38,6 +38,11 @@ export default definePlugin({
         },
     ],
     isGuildOwner(props) {
+        // Check if channel is a Group DM, if so return false
+        if (props.channel.type === 3) {
```
D3SOX, line 42:
```suggestion
        if (props?.channel?.type === 3) {
```

I've also thought about checking the type. Are you sure `3` is only used for group channels?

nmsturcke: image
According to docs yes

D3SOX (approved):
LGTM now

Vendicated (approved):

PR #184, opened 2022-10-31
feat(nitroBypass): add sticker bypass

D3SOX:
TODO
- [x] ~~Don't hardcode `https://cdn.discordapp.com`?~~
- [x] Send normally if allowed (from current guild)
- [x] Implement animated stickers (if possible) or add a hint about it
- [X] Fix glitching GIF
- [x] Allow sending Discord internal stickers (implement lottie to gif)


D3SOX:
I'll try to resolve the to-dos tomorrow, anybody's welcome to help out :)

D3SOX:
Can't find a way to fix the glitchy GIF for now. I'd leave it like this for now, it's better than nothing.

src/plugins/nitroBypass.ts
```diff
@@ -88,11 +168,23 @@ export default definePlugin({
         return Boolean(UserStore.getCurrentUser().premiumType);
     },
 
+    getStickerLink(stickerId: string) {
+        return `${location.origin}/stickers/${stickerId}.png`;
```
ArjixWasTaken, line 172: Discord has a map exported (just an object) with all the API endpoints as well as the asset endpoints.

You should probably look for that, just search through the sources for the discord assets link and you will find it.

ArjixWasTaken:
Otherwise this LGTM!

ArjixWasTaken:
Otherwise this LGTM!

src/api/MessageEvents.ts
```diff
@@ -46,16 +46,20 @@ const editListeners = new Set<EditListener>();
 export function _handlePreSend(channelId: string, messageObj: MessageObject, extra: any) {
     for (const listener of sendListeners) {
         try {
-            listener(channelId, messageObj, extra);
-        } catch (e) { MessageEventsLogger.error(`MessageSendHandler: Listener encoutered an unknown error. (${e})`); }
+            const result = listener(channelId, messageObj, extra);
+            if (result && result.cancel) {
```
Vendicated, line 50:
```suggestion
            if (result && result.cancel === true) {
```
just to make it more explicit

src/api/MessageEvents.ts
```diff
@@ -46,16 +46,20 @@ const editListeners = new Set<EditListener>();
...
+                return true;
+            }
+        } catch (e) { MessageEventsLogger.error(`MessageSendHandler: Listener encountered an unknown error. (${e})`); }
```
Vendicated, line 53: I know this is old code but
```suggestion
        } catch (e) { MessageEventsLogger.error("MessageSendHandler: Listener encountered an unknown error\n", e); }
```

src/api/MessageEvents.ts
```diff
@@ -90,7 +94,7 @@ export function _handleClick(message: Message, channel: Channel, event: MouseEve
     for (const listener of listeners) {
         try {
             listener(message, channel, event);
-        } catch (e) { MessageEventsLogger.error(`MessageClickHandler: Listener encoutered an unknown error. (${e})`); }
+        } catch (e) { MessageEventsLogger.error(`MessageClickHandler: Listener encountered an unknown error. (${e})`); }
```
Vendicated, line 97: same as above

src/plugins/nitroBypass.ts
```diff
@@ -16,17 +16,60 @@
...
+    Function("self", await fetch("https://cdnjs.cloudflare.com/ajax/libs/apng-canvas/2.1.1/apng-canvas.min.js").then(r => r.text()))(winProxy);
+    // @ts-ignore
+    return exports.APNG;
+});
+
+const DRAFT_TYPE = 0;
+// https://github.com/mattdesl/gifenc
+// this lib is way better than gif.js and all other libs, they're all so terrible but this one is nice
+// @ts-ignore ts mad
+const getGifEncoder = makeLazy(() => import("https://unpkg.com/gifenc@1.0.3/dist/gifenc.esm.js"));
```
Vendicated, line 40: since two plugins use this same package now, I think it would be good if we introduced a central file like utils/dependencies.ts that would export all the dependencies plugins might use

src/plugins/nitroBypass.ts
```diff
@@ -63,15 +114,44 @@ export default definePlugin({
...
+                replace: "$1=$2.sticker_packs;Vencord.Plugins.plugins.NitroBypass.savePacks($1);$3.$4.dispatch({type:\"STICKER_PACKS_FETCH_SUCCESS\""
+            },
+        },
     ],
+    savePacks(stickerPacks) {
+        this.stickerPacks = stickerPacks;
+    },
+    saveStickerMap(stickerMap) {
+        this.stickerMap = stickerMap;
+        // TODO: I think this could use something like const StickerStore = lazyWebpack(filters.byProps("getStickers??"));
```
Vendicated, line 140: so?

ArjixWasTaken (changes requested):
Oof, the review was pending

src/plugins/nitroBypass.ts
```diff
@@ -104,17 +196,107 @@ export default definePlugin({
...
+
+                                // if it's animated download it, convert to gif and send it
+                                if (isAnimated) {
+
+                                    if (!this.apng || !this.gif) {
+                                        console.error("[NitroBypass] Can't send animated sticker because the apng or gif library is not loaded");
+                                        return { cancel: true };
+                                    }
+
+                                    const { GIFEncoder, quantize, applyPalette } = this.gif;
```
Vendicated, line 239: since this is async anyway, please import the libraries on demand instead of unconditionally

Vendicated (changes requested):
Thanks! Some feedback

src/plugins/nitroBypass.ts
Vendicated: likely findByProps("STICKER")

D3SOX: > likely findByProps("STICKER")

Returns this
image

D3SOX:
Ok TIL the rebase on master option makes all commits unverified (for whatever reason) and now it looks like I pushed everything again

Vendicated:
lmao yeah, merge doesn't have this issue

src/plugins/nitroBypass.ts
Vendicated: find(m => typeof m.STICKER === "function")

D3SOX: You mean't like this? `STICKER` only returns `/stickers/INPUT` 🤔 

```suggestion
        const { STICKER } = find(m => typeof m.STICKER === "function");
        return `${location.origin}${STICKER(`${stickerId}.png`)}`;
```

Vendicated: Well there's not really any issue with hardcoding it, it's not like it's gonna change

ArjixWasTaken: You are getting the wrong webpack module, imma help when I'm on my pc

ArjixWasTaken:
@D3SOX 
`GLOBAL_ENV.CDN_HOST` gives `cdn.discordapp.com`

Vendicated:
cdn sticker endpoint has cors issues. why are u even making an issue out of a total non issue, the current implementation is fine

ArjixWasTaken:
i am not? the PR description itself has it as a goal to not hardcode the url

src/plugins/nitroBypass.ts
```diff
@@ -16,17 +16,31 @@
...
 import { Settings } from "../Vencord";
-import { findByProps } from "../webpack";
+import { filters, findByProps, waitFor } from "../webpack";
 import { UserStore } from "../webpack/common";
 
+const DRAFT_TYPE = 0;
+const promptToUpload = lazyWebpack(filters.byCode("UPLOAD_FILE_LIMIT_ERROR"));
+
+let ChannelStore;
+waitFor(["getChannel"], m => ChannelStore = m);
```
Vendicated, line 33: pretty sure this is a common module

src/plugins/nitroBypass.ts
```diff
@@ -104,17 +159,104 @@ export default definePlugin({
...
+
+                                        const canvas = document.createElement("canvas");
+                                        canvas.width = canvas.height = resolution;
+                                        const ctx = canvas.getContext("2d")!;
+
+                                        const { frames } = apng;
+
+                                        for (const frame of frames) {
+                                            ctx.clearRect(0, 0, canvas.width, canvas.height);
+                                            ctx.drawImage(frame.img, 0, 0, resolution, resolution);
```
Vendicated, line 214: instead of always reusing the same resolution, why not use each frame's height and width? maybe that will fix the weird glitching, who knows

src/plugins/nitroBypass.ts
```diff
@@ -104,17 +159,104 @@ export default definePlugin({
...
+                                    username: "Vencord"
+                                }
+                            });
+                            return { cancel: true };
+                        }
+
+                        const stickerLink = this.getStickerLink(stickerId);
+
+                        if (this.stickerMap) {
+                            // get guild id from sticker
```
Vendicated, line 186: it would be cool if you could make it also check if you can use that sticker, since then it would still be useful for nitro havers to be able to post unavailable stickers

src/plugins/nitroBypass.ts
```diff
@@ -88,8 +139,12 @@ export default definePlugin({
         return Boolean(UserStore.getCurrentUser().premiumType);
     },
 
+    getStickerLink(stickerId: string) {
+        return `${location.origin}/stickers/${stickerId}.png`;
```
Vendicated, line 143: consider using the ?size parameter to make the fake stickers be exactly the same size as real ones. Would be cool to make it a setting maybe?

Vendicated (changes requested):

D3SOX:
Thanks for the suggestions, will implement later

D3SOX:
> i am not? the PR description itself has it as a goal to not hardcode the url
> although it doesnt actually hardcode it

I've set it as TODO because I had previously hardcoded it but then it turned out that I need to use discord.com instead anyways because the cdn endpoint can't be fetched for animated sticker because of CORS

ArjixWasTaken:
naruhodo

src/plugins/nitroBypass.ts
D3SOX: I tried it right now and it makes it glitch even worse

src/plugins/nitroBypass.ts
D3SOX: Where to get the size from the real sticker from?

src/plugins/nitroBypass.ts
D3SOX: Can nitro users not use all stickers?

src/plugins/nitroBypass.ts
Vendicated: size 160 is exactly the same as real stickers 
image

https://media.discordapp.net/stickers/1026517526106087454.webp?size=160

ArjixWasTaken: image

looks like it is 160x160

D3SOX: I tried to add `?size=32`, `?size=64`, `?size=128` and it doesn't seem that it makes a difference.

src/plugins/nitroBypass.ts
Vendicated: Some may be unavailable due to the server hosting them losing a premium tier or similar

src/plugins/nitroBypass.ts
D3SOX: It seems like it only works for the format of `https://media.discordapp.net/stickers/1026517526106087454.webp?size=160` and not `https://discord.com/stickers/1026517526106087454.png?size=160`

src/plugins/nitroBypass.ts
D3SOX: Hm, how would I go and check if the sticker is currently available? Do you have a tip where I could look?

Vendicated: sticker.available = true

D3SOX: I'm literally blind, I looked inside the sticker object and overlooked that property 🤦

src/plugins/nitroBypass.ts
Vendicated: findByProps("getAllGuildStickers").getAllGuildStickers() => Map<GuildID, Sticker[]>

src/plugins/nitroBypass.ts
D3SOX: Nvm I think `available` is not correct as it's always `true` for me and I don't have Nitro

Vendicated: available means whether the sticker is available, not usable

D3SOX: Ah I see. So if the user has Nitro and it's unavailable it should use the bypass.

D3SOX: Turns out this needs another patch to make unavailable stickers clickable 😒

src/plugins/nitroBypass.ts
```diff
@@ -126,7 +272,7 @@ export default definePlugin({
                 if (emoji == null || (emoji.guildId === guildId && !emoji.animated)) continue;
                 if (!emoji.require_colons) continue;
 
-                const url = emoji.url.replace(/\?size=[0-9]+/, "?size=48");
+                const url = emoji.url.replace(/\?size=\d+/, "?size=48");
```
ArjixWasTaken, line 275: uhh....depending on the locale of the user, `\d` can match letters that represent numbers...
well, I guess there is no point worrying about that, since its kinda hard for those letters to be there in the first place

ArjixWasTaken: I just wanted to let you know that I didn't do it for no reason

ArjixWasTaken:
It would be nice if this worked when editing a message as well

but in that case instead of adding an attachment you'd have to append the link at the end, and that wouldn't work well for animated stickers, so nvm

src/plugins/nitroBypass.ts
D3SOX: > uhh....depending on the locale of the user, `\d` can match letters that represent numbers...

Oh, I didn't know about that. I think then I'll revert this change

ArjixWasTaken: nah its ok, the url is given by discord in the first place, so its kinda guaranteed that only european numerals (AKA [0-9]) will be there

ArjixWasTaken:
Question: Can ya also resize the animated stickers?
(I know its not possible via the url query parameters since you are converting them to gifs in the first place)

D3SOX:
Animated stickers are already created in the configured size

ArjixWasTaken:
oh yeah, i was on an older commit, nvm then

D3SOX:
I think now the glitching is the only thing that's left, any help is appreciated, otherwise I'd say we could merge it like this for now

src/plugins/nitroBypass.ts
```diff
@@ -63,15 +93,38 @@ export default definePlugin({
...
+            find: "ingestStickers:",
+            predicate: () => Settings.plugins.NitroBypass.enableStickerBypass === true,
+            replacement: {
+                match: /(\w+)=(\w+).sticker_packs;(\w+)\.(\w+)\.dispatch\(\{type:"STICKER_PACKS_FETCH_SUCCESS"/g,
+                replace: "$1=$2.sticker_packs;Vencord.Plugins.plugins.NitroBypass.savePacks($1);$3.$4.dispatch({type:\"STICKER_PACKS_FETCH_SUCCESS\""
+            },
+        },
     ],
+    savePacks(stickerPacks) {
+        this.stickerPacks = stickerPacks;
```
Vendicated, line 107: is this still needed? I told you how to get the stickers from Discord dynamically

src/plugins/nitroBypass.ts
```diff
@@ -88,50 +141,163 @@ export default definePlugin({
...
+    start() {
+        if (!Settings.plugins.NitroBypass.enableEmojiBypass && !Settings.plugins.NitroBypass.enableStickerBypass) {
             return;
         }
 
         const { getCustomEmojiById } = findByProps("getCustomEmojiById");
+        const { getAllGuildStickers } = findByProps("getAllGuildStickers");
+
+        // make all available to click
+        getAllGuildStickers().forEach((packs: Sticker[]) => {
```
Vendicated, line 157: why not via a patch? what if a new sticker is created after loading

D3SOX: Tried to find the place where it's queried but I didn't suceed

Vendicated (approved):

PR #185, opened 2022-11-01
fix renamed app.asar detection on Windows

DiamondMiner88:
guh

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,6 @@ If you're just a normal user, use [our simple gui installer!](https://github.com
 
 If you wanna make plugins or contribute or just want to build from source and install manually, read [Megu's Installation Guide!](docs/1_INSTALLING.md)
 
-### For Canary/PTB users
-
-Due to a recent electron update, you may have to go to `%localappdata%/app-1.0.51/resources` (or the equivalent install location)
-and rename `app.asar` to `_app.asar`, and `app` to `app.asar`
-
 ## Installing on Browser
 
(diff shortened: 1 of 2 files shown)


src/patcher.ts
```diff
@@ -30,7 +30,7 @@ console.log("[Vencord] Starting up...");
 const injectorPath = require.main!.filename;
 
 // special discord_arch_electron injection method
-const asarName = injectorPath.endsWith("app.asar/index.js") ? "_app.asar" : "app.asar";
+const asarName = require.main!.path.endsWith("app.asar") ? "_app.asar" : "app.asar";
```
Vendicated, line 33: out of curiosity, what is require.main!.filename on Windows?

README.md
```diff
@@ -18,6 +18,11 @@ If you're just a normal user, use [our simple gui installer!](https://github.com
 
 If you wanna make plugins or contribute or just want to build from source and install manually, read [Megu's Installation Guide!](docs/1_INSTALLING.md)
 
+### For Canary/PTB users
```
Vendicated, line 21: no, the installer should do this

src/patcher.ts
DiamondMiner88: ~AppData\\Local\\DiscordCanary\\app-1.0.51\\resources\\app.asar\\index.js

Vendicated: OH it just failed due to using \ instead of / lmao

PR #188, opened 2022-11-04
feat: make text selectable in PatchHelper

MeguminSama:


src/components/PatchHelper.tsx
```diff
@@ -85,16 +85,16 @@ function ReplacementComponent({ module, match, replacement, setReplacementError
 
         return (
             <>
-                {Parser.parse(fullMatch)}
-                {Parser.parse(groups)}
+                <span style={{ userSelect: "text" }}>{Parser.parse(fullMatch)}</span>
```
Vendicated, line 88: wouldn't a div make more sense since this is not supposed to be inline?

ArjixWasTaken: or just do [this](https://github.com/Vendicated/Vencord/pull/52/files#diff-92e7c1a997c8562475cf740f730ccfeeaa109111562dd1e9fb050e433f4851c4R55)

Vendicated: no, that's really ugly

PR #189, opened 2022-11-04
feat(plugins): add Startup Timings

MeguminSama:


src/plugins/startupTimings/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import definePlugin from "../../utils/types";
+
+export default definePlugin({
+    name: "StartupTimings",
+    description: "Adds Startup Timings to the Settings menu",
+    authors: [Devs.Megu],
+    patches: [{
+        find: "PAYMENT_FLOW_MODAL_TEST_PAGE,",
+        replacement: {
+            match: /({section:[\w.]+?\.PAYMENT_FLOW_MODAL_TEST_PAGE,)/,
```
Vendicated, line 31: isnt [\w.] just .? Also would be better to use {1,3} or similar to not match half the module (it will inevitably happen )

ArjixWasTaken: image

PR #191, opened 2022-11-05
feat: support CSS modules

MeguminSama:
This PR adds support for CSS modules to Vencord.

These can be used as follows:

```tsx
import styles from "./MyComponent.module.css";

function MyComponent() {
	return (
		<div className={styles.myClass}>Hello, World!</div>
	);
}
```

#### esbuild-css-modules-plugin@^2.6.2
Adds support for CSS & CSS Module importing. ~37.5kB build size.

#### string-hash@^1.1.1
String hashing module used to generate class hashes.
This comes as a subdependency of esbuild-css-modules-plugin, so has no extra cost.

### Recommended VSCode Extension (optional)

`clinyong.vscode-css-modules` provides QoL support for CTRL+Clicking CSS modules from inside TSX files.


scripts/build/common.mjs
```diff
@@ -150,7 +152,35 @@ export const commonOpts = {
...
     sourcemap: watch ? "inline" : "",
     legalComments: "linked",
-    plugins: [fileIncludePlugin, gitHashPlugin, gitRemotePlugin],
+    plugins: [
+        fileIncludePlugin,
+        gitHashPlugin,
+        gitRemotePlugin,
+        CssModulesPlugin({
+            // Sometimes CSS is injected before the DOM is ready, so some extra logic is needed.
+            inject: (cssContent, digest) => `
```
Vendicated, line 161: this injects the css text into the js?

Wouldn't it be better to generate a separate css bundle and load that with webFrame.insertCSS / a separate style sheet in browser?

Vendicated:
Can you use scss with this?

How about auto prefixer (on web builds)?

Vendicated:
Can probably close this?

The current css system is good enough imo, and css modules are counter intuitive since it's then impossible for users to style with custom css

PR #197, opened 2022-11-09
Spotify Premium plugin.

xPolar:
I've created a plugin to bypass Discord's clientside check for whether or not you have Spotify Premium. Thus, you can use Spotify Premium-only features such as group listening for free.

### DIFF
--- a/src/plugins/spotifyPremium.ts
+++ b/src/plugins/spotifyPremium.ts
@@ -1,0 +1,39 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


src/plugins/spotifyPremium.ts
```diff
@@ -0,0 +1,41 @@
...
+	authors: [
+		{
+			id: 619284841187246090n,
+			name: "Polar",
+		},
+		Devs.Megu
+	],
+	patches: [
+		{
+			find: "l.Z.getActiveSocketAndDevice();",
```
MeguminSama, line 34: while the `Z` should be safe as-is, please change the `l` to `\w{1,3}` or equivalent as this could change

MeguminSama (changes requested):

Vendicated:
As discussed in the Discord, already part of ify. thanks though!

PR #198, opened 2022-11-09
fix(spotifyControls): add album/cover null checks (local files)

D3SOX:
Fixes local files

image

after 

image

### DIFF
--- a/src/plugins/spotifyControls/PlayerComponent.tsx
+++ b/src/plugins/spotifyControls/PlayerComponent.tsx
@@ -248,7 +248,7 @@ function Info({ track }: { track: Track; }) {
         </>
     );
 
-    if (coverExpanded) return (
+    if (coverExpanded && img) return (
         <div id={cl("album-expanded-wrapper")}>
             {i}
         </div>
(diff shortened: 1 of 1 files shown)


D3SOX:
I just noticed this breaks when the user previously opened the cover view, need to push another fix

src/plugins/spotifyControls/PlayerComponent.tsx
```diff
@@ -233,18 +233,22 @@ function Info({ track }: { track: Track; }) {
...
-            onClick={() => setCoverExpanded(!coverExpanded)}
-            onContextMenu={e => {
-                ContextMenu.open(e, () => <AlbumContextMenu track={track} />);
-            }}
-        />
+        <>
+            {img && (
+                <img
+                    id={cl("album-image")}
+                    src={img?.url}
```
Vendicated, line 240:
```suggestion
                    src={img.url}
```

src/plugins/spotifyControls/PlayerComponent.tsx
```diff
@@ -233,18 +233,22 @@ function Info({ track }: { track: Track; }) {
...
+                    onContextMenu={e => {
+                        ContextMenu.open(e, () => <AlbumContextMenu track={track} />);
+                    }}
+                />
+            )}
+        </>
     );
 
-    if (coverExpanded) return (
+    if (coverExpanded && img) return (
```
Vendicated, line 251: uh how could u expand cover if img is null?

D3SOX: Expand cover, then play a local file

PR #202, opened 2022-11-10
Plugin to fix TikTok links

somerandomcloud:
Uses [TikTxk](https://github.com/Britmoji/tiktxk) to fix ugly/dysfunctional tiktok embeds.


Vendicated:
we already have enough text replace plugins

At this point someone should just make a generic text replace plugin with support for custom rules and the common ones as presets

somerandomcloud:
Alright, then I'll close this

PR #208, opened 2022-11-13
feat(plugin): FriendInvites

xafn:


Vendicated:
Thank you!! I think you were a little confused in the view invites command cause you were +ing on an array and typed it as any (got some errors huh?). Normally you would .push() to it or use a string instead

xafn:
im not a great javascripter tbh  
image

Vendicated:
haha don't worry about it 🚎

PR #218, opened 2022-11-19
feat(plugin): Quick mention button

kem0x:


Vendicated:
This is kindof obsolete since you can just shift click people's username

kem0x:
I've never heard of that feature till this second bro where are you coming with this stuff from 💀 

on another note tho, I still think this plugin is easier and more demanded atleast from everyone i know since it's just one click instead of holding keyboard, i believe all other mods had this plugin so why not vencord and my friends want it!!!!

src/plugins/quickMention.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    authors: [Devs.kemo],
+
+    description: "Adds a quick mention button to the message actions bar",
+
+    patches: [
+        {
+            find: "Messages.MESSAGE_UTILITIES_A11Y_LABEL",
+            replacement: {
+                match: /(null,)(.{1}&&!.{1}\?(.{2})\(\{.*?,channel:(.*?),message:(.*?),onClick:.*?\}\))/,
```
Vendicated, line 36: since this matches
```js
null,h&&!d?Vn({key:"reply",label:Mt.Z.Messages.MESSAGE_ACTION_REPLY,icon:jn.Z,channel:t,message:n,onClick:jt.HH})
```
why not include the key and the MESSAGE_ACTION_REPLY in your regex? that would make it a lot more robust

Vendicated: like
```js
/null,(.{1,2})&&.{0,10}key:"reply",label:.{0,20}MESSAGE_ACTION_REPLY.+?onClick.{0,10}\}\)/
```

kem0x: I'm pretty sure this regex bricked my discord 💀

Vendicated (approved):

PR #220, opened 2022-11-19
feat(plugin): Last.fm rich presence

dzshn:
adds a plugin for last.fm rpc, based on [dimdenGD/LastFMRichPresence](https://github.com/dimdenGD/LastFMRichPresence)

## Preview
image

### DIFF
--- a/src/plugins/lastfm.ts
+++ b/src/plugins/lastfm.ts
@@ -1,0 +1,185 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Sofia Lima
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { lazyWebpack } from "../utils/misc";
+import definePlugin, { OptionType } from "../utils/types";
+import { Settings, Webpack } from "../Vencord";
+
+interface ActivityAssets {
+    large_image?: string
+    large_text?: string
+    small_image?: string
+    small_text?: string
+}
+
+interface Activity {
+    state: string
+    details?: string
+    timestamps?: {
+        start?: Number
+    }
+    assets?: ActivityAssets
+    buttons?: Array<string>
+    name: string
+    application_id: string
+    metadata?: {
+        button_urls?: Array<string>
+    }
+}
+
+interface TrackData {
+    name: string
+    album: string
+    artist: string
+    url: string
+    imageUrl?: string
+}
+
+// only relevant enum values
+enum ActivityType {
+    PLAYING = 0,
+    LISTENING = 2,
+}
+
+enum ActivityFlag {
+    INSTANCE = 1 << 0,
+}
+
+const applicationId = "1043533871037284423";
+
+const rpcServer = lazyWebpack(Webpack.filters.byProps("dispatch", "_subscriptions"));
+const presenceStore = lazyWebpack(Webpack.filters.byProps("getLocalPresence"));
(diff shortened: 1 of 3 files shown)


dzshn:
additionally - I would like to note that connecting to the RPC api via `WebSocket` does not work because the `Origin` header is set automatically and there isn't really a way to disable that. if it makes sense to implement some sort of RPC api using the same functions I found I'd be glad to work on that too

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+    LISTENING = 2,
+}
+
+enum ActivityFlag {
+    INSTANCE = 1 << 0,
+}
+
+const applicationId = "1043533871037284423";
+
+const rpcServer = lazyWebpack(Webpack.filters.byProps("dispatch", "_subscriptions"));
```
Vendicated, line 65: this is just FluxDispatcher that u can import from Commons

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+}
+
+export default definePlugin({
+    name: "LastFMRichPresence",
+    description: "Little plugin for Last.fm rich presence",
+    authors: [
+        {
+            name: "dzshn",
+            id: 310449948011528192n,
```
Vendicated, line 93: please add yourself to the Devs object in constants instead

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+        },
+    ],
+    options: {
+        username: {
+            description: "last.fm username",
+            type: OptionType.STRING,
+            placeholder: "dimdendev",
```
Vendicated, line 100: I'd just not have a placeholder to not confuse people (same below), otherwise some might think the api key will already work or forget to change this

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+        hideWithSpotify: {
+            description: "hide last.fm presence if spotify is running",
+            type: OptionType.BOOLEAN,
+            default: true,
+        }
+    },
+    start() {
+        this.settings = Settings.plugins.LastFMRichPresence;
+
+        setInterval(() => { this.updatePresence(); }, 16000);
```
Vendicated, line 116: if you do `this.interval = setInterval(...)` you can then `clearInterval(this.interval)` in stop

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+    stop() {},
+    async fetchTrackData(): Promise<TrackData | null> {
+        if (!this.settings.username || !this.settings.apiKey) return null;
+
+        const response = await fetch(`https://ws.audioscrobbler.com/2.0/?method=user.getrecenttracks&api_key=${this.settings.apiKey}&user=${this.settings.username}&limit=1&format=json`);
+        const trackData = (await response.json()).recenttracks.track[0];
+
+        if (!trackData["@attr"]?.nowplaying) return null;
+
+        // why does the json api have xml structure
```
Vendicated, line 127: lastfm goofy

src/plugins/lastfm.ts
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        const hideAlbumName = !trackData.album || trackData.album === trackData.name;
+
+        let assets: ActivityAssets;
+        if (trackData.imageUrl) {
+            assets = {
+                large_image: await getApplicationAsset(trackData.imageUrl),
+                large_text: trackData.name,
+                small_image: await getApplicationAsset("lastfm-small"),
```
Vendicated, line 161: probably would be best to cache these
```suggestion
                small_image: await getSmallImage(),
```

then on top level add
```js
const getSmallImage = makeLazy(() => getApplicationAsset("lastfm-small"))
```

Vendicated (changes requested):

src/plugins/lastfm.ts
dzshn: leaving it as-is since it's already cached by that function

src/plugins/lastfm.tsx
```diff
@@ -0,0 +1,208 @@
...
+            description: "last.fm api key",
+            type: OptionType.STRING,
+        },
+        hideWithSpotify: {
+            description: "hide last.fm presence if spotify is running",
+            type: OptionType.BOOLEAN,
+            default: true,
+        },
+        useListeningStatus: {
+            description: "show \"Listening to\" status instead of \"Playing\"",
```
Vendicated, line 122:
```suggestion
            description: 'show "Listening to" status instead of "Playing"',
```

Vendicated (approved):
Awesome, thank yoU!!

PR #234, opened 2022-11-21
Add a "NEW" Badge for New Plugins (V2)!

nmsturcke:
This does the following:

- [x] Store creation time of existing plugins in datastore
- [x] Check for creation time and, if it's less than 2 days ago, mark as new
- [x] Add a badge for new plugins
- [x] Compare creation time with something to determine if the plugin is new, or they just install Vencord


Vendicated:
please fix conflicts

nmsturcke:
> please fix conflicts

Done

ArjixWasTaken:
Since you made a badge component, I'd suggest making the badge name and color be properties

So that the same component can be reused

nmsturcke:
> Since you made a badge component, I'd suggest making the badge name and color be properties
> 
> So that the same component can be reused

Changed ^^

src/components/PluginSettings/index.tsx
```diff
@@ -242,6 +246,24 @@ export default ErrorBoundary.wrap(function Settings() {
...
+            Vencord_existingPlugins[plugin.name] = Object.keys(existingPlugins || []).includes(plugin.name) ? existingPlugins[plugin.name] : dateNow;
+            if ((Vencord_existingPlugins[plugin.name] + 172800) > dateNow) {
+                newPlugins.push(plugin.name);
+            }
+        });
+        DataStore.set("Vencord_existingPlugins", Vencord_existingPlugins);
+        return newPlugins;
+    }));
+
+    if (JSON.stringify(newPlugins) === JSON.stringify(sortedPluginNames)) {
```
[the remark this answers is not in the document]
Vendicated: this should also be moved inside the useAwaiter

src/components/PluginSettings/styles.ts
```diff
@@ -48,3 +48,23 @@ export const SettingsIcon: React.CSSProperties = {
...
 };
+
+export const BadgeStyle: React.CSSProperties = {
+    paddingTop: "0",
+    paddingBottom: "0",
+    paddingRight: "6px",
+    paddingLeft: "6px",
+    fontFamily: "var(--font-display)",
+    fontWeight: "500px",
+    textTransform: "uppercase",
```
Vendicated, line 59: why not just make it upper case in js? xd

src/components/PluginSettings/styles.ts
```diff
@@ -48,3 +48,23 @@ export const SettingsIcon: React.CSSProperties = {
...
+    whiteSpace: "nowrap",
+    textOverflow: "ellipsis",
```
Vendicated, line 61: i dont think a badge should ever overflow and if it does thereshould definitely not be an ellipsis

src/components/PluginSettings/styles.ts
```diff
@@ -48,3 +48,23 @@ export const SettingsIcon: React.CSSProperties = {
...
+    overflow: "hidden",
+    borderRadius: "8px",
+    height: "16px",
+    minWidth: "16px",
+    minHeight: "16px",
+    fontSize: "12px",
+    lineHeight: "16px",
+    color: "white",
```
Vendicated, line 69: use a discord variable

src/components/PluginSettings/styles.ts
```diff
@@ -48,3 +48,23 @@ export const SettingsIcon: React.CSSProperties = {
...
+    textTransform: "uppercase",
+    whiteSpace: "nowrap",
+    textOverflow: "ellipsis",
+    overflow: "hidden",
+    borderRadius: "8px",
+    height: "16px",
+    minWidth: "16px",
+    minHeight: "16px",
+    fontSize: "12px",
+    lineHeight: "16px",
```
Vendicated, line 68: most of these should be obsolete

Vendicated (changes requested):

src/components/PluginSettings/index.tsx
```diff
@@ -242,6 +246,24 @@ export default ErrorBoundary.wrap(function Settings() {
         );
     };
 
+    let [newPlugins] = useAwaiter(() => DataStore.get("Vencord_existingPlugins").then((existingPlugins: Record<string, number>) => {
+        const dateNow: number = Date.now() / 1000;
+        const Vencord_existingPlugins: Record<string, number> = {};
+        const newPlugins: Array<string> = [];
+        sortedPlugins.forEach(plugin => {
+            Vencord_existingPlugins[plugin.name] = Object.keys(existingPlugins || []).includes(plugin.name) ? existingPlugins[plugin.name] : dateNow;
+            if ((Vencord_existingPlugins[plugin.name] + 172800) > dateNow) {
```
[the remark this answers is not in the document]
nmsturcke: Done

src/components/PluginSettings/index.tsx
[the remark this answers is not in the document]
nmsturcke: Done, not sure if I set the type correctly though. I set it to `LoDashWrapper<any>`, be sure to say if it needs changing

src/components/PluginSettings/styles.ts
```diff
@@ -48,3 +48,23 @@ export const SettingsIcon: React.CSSProperties = {
     background: "transparent",
     marginRight: 8
 };
+
+export const BadgeStyle: React.CSSProperties = {
+    paddingTop: "0",
+    paddingBottom: "0",
+    paddingRight: "6px",
+    paddingLeft: "6px",
```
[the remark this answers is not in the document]
nmsturcke: Done

src/components/PluginSettings/styles.ts
nmsturcke: Just copied badge css properties lol

src/components/PluginSettings/styles.ts
nmsturcke: Done

src/components/PluginSettings/styles.ts
nmsturcke: Done

src/components/PluginSettings/styles.ts
nmsturcke: Done

src/components/PluginSettings/index.tsx
```diff
@@ -242,6 +246,24 @@ export default ErrorBoundary.wrap(function Settings() {
         );
     };
 
+    const [newPlugins, newpluginsError, newPluginsLoading] = useAwaiter(() => DataStore.get("Vencord_existingPlugins").then((existingPlugins: Record<string, number>) => {
+        const dateNow: number = Date.now() / 1000;
+        const Vencord_existingPlugins: Record<string, number> = {};
```
Vendicated, line 251:
```suggestion
        const existingPlugins: Record<string, number> = {};
```

src/components/PluginSettings/index.tsx
```diff
@@ -242,6 +246,24 @@ export default ErrorBoundary.wrap(function Settings() {
...
+        let newPlugins: Array<string> = [];
+        sortedPlugins.map(plugin => {
+            Vencord_existingPlugins[plugin.name] = Object.keys(existingPlugins || []).includes(plugin.name) ? existingPlugins[plugin.name] : dateNow;
```
Vendicated, line 254:
```suggestion
            Vencord_existingPlugins[plugin.name] = existingPlugins[plugin.name] ?? dateNow;
```

src/components/PluginSettings/index.tsx
```diff
@@ -224,6 +226,7 @@ export default ErrorBoundary.wrap(function Settings() {
 
     const sortedPlugins = React.useMemo(() => Object.values(Plugins)
         .sort((a, b) => a.name.localeCompare(b.name)), []);
+    const sortedPluginNames = Object.values(sortedPlugins).map(plugin => plugin.name);
```
Vendicated, line 229: this should be inside the awaiter to avoid unnecessary work

src/components/PluginSettings/index.tsx
```diff
@@ -243,6 +246,24 @@ export default ErrorBoundary.wrap(function Settings() {
         );
     };
 
+    const [newPlugins, , newPluginsLoading] = useAwaiter(() => DataStore.get("Vencord_existingPlugins").then((cachedPlugins: Record<string, number>) => {
+        const dateNow: number = Date.now() / 1000;
+        const existingPlugins: Record<string, number> = {};
+        let newPlugins: Array<string> = [];
+        sortedPlugins.map(plugin => {
```
Vendicated, line 253: this is map misuse since you're not using the result at all

Vendicated (changes requested):

src/components/PluginSettings/index.tsx
```diff
@@ -243,6 +245,26 @@ export default ErrorBoundary.wrap(function Settings() {
         );
     };
 
+    const [newPlugins, , newPluginsLoading] = useAwaiter(() => DataStore.get("Vencord_existingPlugins").then((cachedPlugins: Record<string, number>) => {
```
Vendicated, line 248: this get will return undefined if there is no entry which will later on read to a read of undefined.foo which will lead to a promise rejection. Please handle this case

src/globals.d.ts
```diff
@@ -54,6 +56,7 @@ declare global {
             push(chunk: any): any;
             pop(): any;
         };
+        _: LoDashWrapper<any>;
```
Vendicated, line 59: shouldn't this be LodashStatic?

src/components/PluginSettings/index.tsx
Vendicated: why did you mark this as resolved? this is still an issue

src/components/PluginSettings/index.tsx
```diff
@@ -243,6 +245,26 @@ export default ErrorBoundary.wrap(function Settings() {
...
+        const dateNow: number = Date.now() / 1000;
+        const existingPlugins: Record<string, number> = {};
+        const sortedPluginNames = Object.values(sortedPlugins).map(plugin => plugin.name);
+
+        let newPlugins: Array<string> = [];
+        sortedPlugins.forEach(plugin => {
+            existingPlugins[plugin.name] = cachedPlugins[plugin.name] ?? dateNow;
```
ArjixWasTaken, line 255:
```suggestion
            existingPlugins[plugin.name] = cachedPlugins?.[plugin.name] ?? dateNow;
```

when indexing a nullable value, you can simply use `?.` before the brackets to ensure that it won't error

ArjixWasTaken (changes requested):

src/components/PluginSettings/index.tsx
ArjixWasTaken: since this is the only usage of `cachedPlugins` that I can see within this function, this should be enough to ensure no error happens

src/components/PluginSettings/index.tsx
ArjixWasTaken:
```suggestion
    const [newPlugins, , newPluginsLoading] = useAwaiter(() => DataStore.get("Vencord_existingPlugins").then((cachedPlugins: Record<string, number> | undefined) => {
```

if it is true that it can be undefined, then it should be typed accordingly

src/components/PluginSettings/index.tsx
ArjixWasTaken: (yes I know, this is weird syntax, but oh well)

Vendicated (approved):

PR #238, opened 2022-11-21
 UserPlugins: Add clickable badges to external user plugins

romangraef:
This patch adds a button to each user plugin (this is determined by the
plugin directory). Additionally user plugins can specify an
`externalLink` in their plugin definition to have their badge be
clickable and link to an external website.

Note that this incorporates one file from #234 for the badge colors, but i can also just copy those styles over manually.

### DIFF
--- a/scripts/build/common.mjs
+++ b/scripts/build/common.mjs
@@ -60,19 +60,23 @@ export const globPlugins = {
         });
 
         build.onLoad({ filter, namespace: "import-plugins" }, async () => {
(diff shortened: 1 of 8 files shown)


Vendicated:
Loading plugins from urls isn't really possible because that means they are loaded too late. You need some way of obtaining them synchronously (e.g. save to file ahead of time and read with fs.readFileSync)

romangraef:
Yes, right now i allow downloading plugins to localstorage / datastorage and then force a CTRL+R reload.

PR #239, opened 2022-11-21
feat(spotifyControls): support saving songs

D3SOX:
TODO
- [x] Implement removing from saved tracks
- [x] Maybe prompt the user to re-authenticate when a request fails with status 403
- [x] Show in album context menu

Possible improvements

- [ ] Sync when updated save status via Spotify client (We could remove `if (!this.savedTrackIds.has(track.id)) {` but `SPOTIFY_PLAYER_STATE` is not called when the user clicks on the icon only after a seek/play/pause/etc..)
- [ ] Use heart icon in context menu instead of checkbox
- [ ] Find a better spot to call `checkSaved`
- [ ] Improve the patch regex?
- [ ] Make it prettier when showing as icon (Icons are no longer centered now)

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -123,7 +123,6 @@ export default definePlugin({
         deleteStyle: {
             type: OptionType.SELECT,
             description: "The style of deleted messages",
-            default: "text",
             options: [
                 { label: "Red text", value: "text", default: true },
                 { label: "Red overlay", value: "overlay" }
--- a/src/plugins/spotifyControls/PlayerComponent.tsx
+++ b/src/plugins/spotifyControls/PlayerComponent.tsx
@@ -81,10 +81,10 @@ function Button(props: React.ButtonHTMLAttributes<HTMLButtonElement>) {
     );
 }
 
-function Controls() {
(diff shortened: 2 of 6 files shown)


D3SOX:
I think this is now at a state where it could get merged. Waiting for feedback on this

Vendicated:
>  Use heart icon in context menu instead of checkbox
Would be cool to have this before merge! You can do this by passing a Menu.MenuItem with render prop. This render prop can render any custom component

somerandomcloud:
As said above, would be nice to have the possible additions (and maybe even screenshots of final product?)

KraXen72:
@D3SOX sned screenshot kthx

D3SOX:
You can just build my branch but okay here is the current state
image
image

KraXen72:
swagger, can you hide the heart button that's next to repeat and only keep the context menu one?

D3SOX:
It's already a setting where you want it

KraXen72:
no no just asking if it's a setting okay

Vendicated:
Please fix conflicts. How about the todos? I think I remember you showing something about the heart button one

Vendicated:
> How about the todos? I think I remember you showing something about the heart button one

Vendicated:
closing as this has been unfinished for very long now and thus outdated. thank you for your effort!

PR #249, opened 2022-11-22
feat(plugins): Volume Booster plugin

Nuckyz:


D3SOX:
Can we add support for stream volume too?

Vendicated (approved):

PR #252, opened 2022-11-23
Badges api

SL4F:
added badges api to a folder and added extra folders in those for badges and icons. custom badges are in a user id format allowing for nice organization and knowing which belongs to who.


janisslsm:
explod

Vendicated:
As discussed on Discord, storing binaries like images in git isn't the best idea and doing it this way is silly anyway cause then i gotta push an update just to update badges

PR #254, opened 2022-11-23
add new quotes to loadingQuotes

KraXen72:
you've seen them in the dms ven :)))))


Vendicated:
that sounds kinda lewd

PR #257, opened 2022-11-24
WIP translation support

tbodt:


Vendicated:
Heya! This is a cool idea, but I don't think I will be accepting it, at least not in the current form. This would double the bundle size for a very niche language only few will use and potentially grow even bigger if more languages are added. Also if included like this, it would be pain to maintain since it would require an update whenever Discord adds a new string. Perhaps it could be remote fetched or similar?

Vendicated:
I'm closing this for now because it's stale and we already discussed this not being very mergeable at least in its current state. Thank you though!

PR #264, opened 2022-11-26

ActuallyTheSun:
message component embed:
image

with background:
image

automod embed:
image


Vendicated:
I've already asked this before, but why not render an actual message component instead of using embeds? Similar to cyns cumcord plugin

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,299 @@
...
+                replace: "function $1($2){var $3=$2.message,$4=$2.channel$5().embedCard$6})}\
+var messageEmbed={mle_AutomodEmbed:$1};"
+            }]
+        }
+    ],
+    options: {
+        messageBackgroundColor: {
+            description: "Background color for messages in rich embeds",
+            type: OptionType.BOOLEAN,
+            restartRequired: false
```
D3SOX, line 173: It's `restartNeeded` and I think this line is obsolete as it's false by default
```suggestion
```

castdrian:
I feel like the option to hide the URL and just display the component would be good as well

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+const noContent = (attachments: number, embeds: number): string => {
+    if (!attachments && !embeds) return "";
+    if (!attachments) return `[no content, ${embeds} embed${embeds !== 1 ? "s" : ""}]`;
+    if (!embeds) return `[no content, ${attachments} attachment${attachments !== 1 ? "s" : ""}]`;
+    return `[no content, ${attachments} attachment${attachments !== 1 ? "s" : ""} and ${embeds} embed${embeds !== 1 ? "s" : ""}]`;
+};
+
+function requiresRichEmbed(message: Message) {
+    console.log(message.attachments, message.embeds, message.components);
```
Vendicated, line 133:
```suggestion

```

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,325 @@
...
+            if (linkedMessage) messageCache[messageID] = { message: linkedMessage, fetched: true };
+            else {
+                getMessage(channelID, messageID, { channelID: message.channel_id, messageID: message.id });
+                return null;
+            }
+        }
+        const linkedChannel = ChannelStore.getChannel(channelID);
+        if (!linkedChannel) {
+            return null;
+        }
```
Vendicated, line 232: this should be above the message retrieval and also have a check for whether you can access the channel to avoid sending 4xx api requests

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,325 @@
...
+        const meProps: MessageEmbedProps = {
+            message: linkedMessage,
+            channel: linkedChannel,
+            guildID
+        };
+        setTimeout(() => dispatchBlankUpdate(channelID, messageID), 100);
```
Vendicated, line 238: is this needed? If so, it would probably make more sense to use setImmediate?

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,325 @@
...
+                        {[
+                            <span>{isDM ? "Direct Message - " : (guild as Guild).name + " - "}</span>,
+                            ...(isDM
+                                ? Parser.parse(`<@${dmReceiver.id}>`)
+                                : Parser.parse(`<#${channel.id}>`)
+                            )
+                        ]}
+                    </Text>,
+                    iconProxyURL: guild
+                        ? `https://${window.GLOBAL_ENV.CDN_HOST}/icons/${guild.id}/${guild.icon}.png?size=32`
```
Vendicated, line 275: why so small? I would use something like 512

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -0,0 +1,325 @@
...
+                        : `https://${window.GLOBAL_ENV.CDN_HOST}/avatars/${dmReceiver.id}/${dmReceiver.avatar}`
+                }
+            }}
+            renderDescription={() => {
+                return <div key={message.id} className={classes(...classNames)}>
+                    <ChannelMessage
+                        id={`message-link-embeds-${message.id}`}
+                        message={message}
+                        channel={channel}
+                        animateAvatar={false}
```
Vendicated, line 285:
```suggestion
                        animateAvatar={true}
```

Vendicated:
it also doesn't properly updated fetched messages, you have to switch channel and come back to render them 
image

CodeF53:
image

It doesn't properly load replied messages yet.

To recreate:
1. paste `https://discord.com/channels/1015060230222131221/1024286218801926184/1051238467318853662` into any server other than Vencord Server Server Server
2. ctrl+r to make sure your browser doesn't have the messages cached from before
3. observe as it fails to get the message it replied to

Vendicated:
that should work but I'm pretty sure this kills embeds, try that on a message with other embeds

honestly idk how to do it properly

Vendicated:
I guess it's not a big deal but would still be nice to do properly

PR #265, opened 2022-11-27
Add support for DVM installation path.

hanahaneull:
Support for DVM installation path (`~/.dvm`)


Vendicated:
Hi! This script is mostly obsolete now, you should pr to the [installer](https://github.com/Vendicated/VencordInstaller) as well

PR #266, opened 2022-11-27
fix: add predicate to updater menu item

MeguminSama:


src/components/VencordSettings/index.tsx
```diff
@@ -39,19 +39,25 @@ interface SettingsProps {
...
+const SettingsTabs: Record<string, SettingsTab> = {
     VencordSettings: { name: "Vencord", component: () => <VencordSettings /> },
     VencordPlugins: { name: "Plugins", component: () => <PluginsTab /> },
     VencordThemes: { name: "Themes", component: () => <Text variant="text-md/medium">Coming soon to a Vencord near you!</Text> },
-    VencordUpdater: { name: "Updater", component: () => Updater ? <Updater /> : null },
+    VencordUpdater: { name: "Updater" }, // Only show updater if IS_WEB is false
     VencordSettingsSync: { name: "Backup & Restore", component: () => <BackupRestoreTab /> },
 };
 
+if (!IS_WEB) SettingsTabs.VencordUpdater.component = () => Updater && <Updater />;
```
Vendicated, line 55: whaaa why not 
```suggestion
if (!IS_WEB) SettingsTabs.VencordUpdater = {} 
```
so u don't need check below

PR #267, opened 2022-11-28
feat(plugin): ShikiCodeblocks

Vap0r1ze:
some shiki codeblock stuff


Vap0r1ze:
Quick overview of out of scope changes:
  - added an optional `paths` param to [`useSettings`](https://github.com/Vendicated/Vencord/pull/267/files#diff-3c8501547f6f6c951239aeed2abcf2e47cae318a486be0154503b038bd0410b8) so u can subscribe to specific keys
  - [new csp parsing](https://github.com/Vendicated/Vencord/pull/267/files#diff-9a380b1e6cc28b9e0f9351e48ad74ec7ee7bff0388ec3bdaaaf2679848aa6cd3)
  - add [module type](https://github.com/Vendicated/Vencord/pull/267/files#diff-884a2fb17edc1f2ad8e6b5337ea995514b10644c35559cb5b6315f4ea8d589ff) for `.txt` files (esbuild does this by default)
  - general [url](https://github.com/Vendicated/Vencord/pull/267/files#diff-69fc08f3a6190fbb9be394c676bea64aad03c350e4729ede32428f71c569cbf5) and [text](https://github.com/Vendicated/Vencord/pull/267/files#diff-712ffc191bdeefcad981f5da7c2bb1550371e10c6854ed1b624ec4f415f7686e) utils
  - added [`hljs`](https://github.com/Vendicated/Vencord/pull/267/files#diff-d81e8c10cd7caf6ba37902248179ce745966ef5072095a0cbc9eb31c7075642c) to webpack common

Vap0r1ze:
yea so
image

Vap0r1ze:
New out of scope change:
  - added support for a "--sourcemap" flag in the [`buildWeb`](https://github.com/Vendicated/Vencord/blob/4fe8ac993d454c0cfb6983c62849df4c9ad1d58a/scripts/build/buildWeb.mjs#L59) script

Vendicated:
> New out of scope change:
> 
> * added support for a "--sourcemap" flag in the [`buildWeb`](https://github.com/Vendicated/Vencord/blob/4fe8ac993d454c0cfb6983c62849df4c9ad1d58a/scripts/build/buildWeb.mjs#L59) script

you can just run with the --watch flag to achieve the same result plus no minification

src/plugins/shikiCodeblocks/api/languages.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export const getGrammar = (lang: Language): Promise<NonNullable<ILanguageRegistration["grammar"]>> => {
+    if (lang.grammar) return Promise.resolve(lang.grammar);
+    return fetch(lang.grammarUrl).then(res => res.json());
+};
+
+const aliasCache = new Map<string, Language>();
+export const resolveLang = (idOrAlias: string) => {
+    if (Object.hasOwn(languages, idOrAlias)) return languages[idOrAlias];
```
ArjixWasTaken, line 66:
```suggestion
    if (Object.hasOwnProperty.call(languages, idOrAlias)) return languages[idOrAlias];
```

ArjixWasTaken (changes requested):
plz

src/modules.d.ts
```diff
@@ -20,7 +20,7 @@
 /// <reference types="standalone-electron-types"/>
 
 declare module "~plugins" {
-    const plugins: Record<string, import("@utils/types").Plugin>;
+    const plugins: Record<string, import("./utils/types").Plugin>;
```
Vendicated, line 23: huh? is the old one broken?

src/plugins/shikiCodeblocks/api/shiki.ts
```diff
@@ -0,0 +1,118 @@
...
+        const client = await shiki.clientPromise;
+        if (shiki.loadedThemes.has(themeUrl)) return;
+
+        await client.run("loadTheme", { theme: themeUrl });
+
+        shiki.loadedThemes.add(themeUrl);
+    },
+    setTheme: async (themeUrl: string) => {
+        await shiki.clientPromise;
+        if (!themeUrl) themeUrl = themeUrls[0];
```
Vendicated, line 80:
```suggestion
        themeUrl ||= themeUrls[0];
```

src/plugins/shikiCodeblocks/components/Code.tsx
```diff
@@ -0,0 +1,90 @@
...
+                            {content}
+                        </span>
+                    ))}
+                </>
+            );
+        });
+    }
+
+    const codeTableRows = lines.map((line, i) => (
+        <tr>
```
Vendicated, line 83: these should have a key to prevent bugs https://reactjs.org/docs/lists-and-keys.html

src/plugins/shikiCodeblocks/components/Code.tsx
```diff
@@ -0,0 +1,90 @@
...
+            // [Cynthia] this makes it so when you highlight the codeblock
+            // empty lines are also selected and copied when you Ctrl+C.
+            if (line.length === 0) {
+                return <span>{"\n"}</span>;
+            }
+
+            return (
+                <>
+                    {line.map(({ content, color, fontStyle }) => (
+                        <span
```
Vendicated, line 66: same as below, add a key please

src/plugins/shikiCodeblocks/components/Code.tsx
```diff
@@ -0,0 +1,90 @@
...
+    tokens,
+}: CodeProps) => {
+    let lines!: JSX.Element[];
+
+    if (useHljs) {
+        try {
+            const { value: hljsHtml } = hljs.highlight(lang!, content, true);
+            lines = hljsHtml
+                .split("\n")
+                .map(line => <span dangerouslySetInnerHTML={{ __html: line }} />);
```
Vendicated, line 45: also needs a key, also i really really hate the dangerouslySetInnerHTML is there no other way?

src/plugins/shikiCodeblocks/components/Highlighter.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const useHljs = shouldUseHljs({ lang, tryHljs });
+
+    const [preRef, isIntersecting] = useIntersection();
+
+    const tokens = useAsyncMemo<IThemedToken[][] | null>(async () => {
+        if (!shikiLang || useHljs || !isIntersecting) return null;
+        return await shiki.tokenizeCode(content, lang!);
+    }, [lang, content, currentThemeId, isIntersecting], null);
+
+    const themeBase: ThemeBase = React.useMemo(() => ({
```
Vendicated, line 66: could probably not memo this (unless object equality is needed) cause recomputing this is probably cheaper than the memo logic

src/plugins/shikiCodeblocks/hooks/useAsyncMemo.ts
```diff
@@ -0,0 +1,37 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { React } from "@webpack/common";
+
+export function useAsyncMemo<V>(factory: () => Promise<V>, deps: unknown[], initialValue: V) {
```
Vendicated, line 21: you can use the useAwaiter utility instead! (well it doesn't have the deps argument but feel free to add it

src/plugins/shikiCodeblocks/hooks/useIntersection.ts
```diff
@@ -0,0 +1,56 @@
...
+                    observerRef.current = null;
+                    return;
+                }
+            }
+        });
+        observerRef.current.observe(element);
+    };
+
+    return [refCallback, isIntersecting] as const;
+};
```
Vendicated, line 56: cool utilities, maybe add these to the general utils? At least quickReply can make use of the top function

src/plugins/shikiCodeblocks/index.ts
```diff
@@ -0,0 +1,157 @@
...
+import { shiki } from "./api/shiki";
+import { themes } from "./api/themes";
+import { createHighlighter } from "./components/Highlighter";
+import { DeviconSetting, HljsSetting, ShikiSettings } from "./types";
+
+const themeNames = Object.keys(themes);
+const mainStyle: HTMLStyleElement = document.createElement("style");
+const devIconStyle: HTMLStyleElement = document.createElement("style");
+mainStyle.innerText = cssText;
+devIconStyle.innerHTML = "@import url('https://cdn.jsdelivr.net/gh/devicons/devicon@v2.10.1/devicon.min.css');";
```
Vendicated, line 36: you should move these into the start method so we don't create pointless style elements when the plugin is off

src/utils/text.ts
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+// Utils for readable text transformations eg: `toTitle(fromKebab())`
+
+// Case style to words
+export const fromCamel = (text: string) => text.split(/(?=[A-Z])/).map(w => w.toLowerCase());
+export const fromConst = (text: string) => text.toLowerCase().split("_");
+export const fromKebab = (text: string) => text.toLowerCase().split("-");
+export const fromPascal = (text: string) => text.split(/(?=[A-Z])/).map(w => w.toLowerCase());
+export const fromTitle = (text: string) => text.toLowerCase().split(" ");
```
Vendicated, line 26: these names don't really say much, maybe wordsFromTitle or titleToWords or similar would make more sense?

src/utils/text.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+// Utils for readable text transformations eg: `toTitle(fromKebab())`
+
+// Case style to words
+export const fromCamel = (text: string) => text.split(/(?=[A-Z])/).map(w => w.toLowerCase());
+export const fromConst = (text: string) => text.toLowerCase().split("_");
```
Vendicated, line 23: const? this is snake_case

Vendicated (dismissed):
Thanks for the pr! Other than the feedback below, a bit nitpicky but could you please use interfaces instead of types (where applicable) and actual named functions over arrow function assignments for consistency?

ArjixWasTaken:
Just so you know, it appears that discord already has lodash bundled, so you can just do
image

Instead of the ugly `Object.prototype.hasOwnProperty.call({cat: 5}, "cat");`

#234 adds the types and declares that it is globally accessible (for typescript to not complain)

Vendicated:
that'd still be pretty ugly with `window._.isIn`. The cleanest solution would be to polyfill Object.hasOwn, I personally wouldn't mind adding super simple polyfills like this one

src/plugins/shikiCodeblocks/components/Code.tsx
ArjixWasTaken: how would it prevent bugs? keys are only for rendering optimization...
instead of re-rendering the entire list of elements, it just re-renders the specific items

src/utils/text.ts
ArjixWasTaken: `titleToWords` feels wrong, maybe `titleCaseToWords`, but that is too long

src/plugins/shikiCodeblocks/components/Code.tsx
Vendicated: this is not true. If you omit keys, React has no way of knowing which element exactly is which, which will in the worst case lead to mixed up states (as was the case in the old plugin page where if you searched something the switch state got messed up). if it works as expected, you were simply lucky. It also technically shouldn't matter if the rendered items are stable and never change but its still best practice to add them anyway

ArjixWasTaken: state? what state? they don't have individual state, unless you are referring to smth else

ArjixWasTaken:
> that'd still be pretty ugly with `window._.isIn`.

That is up for debate tbh

src/plugins/shikiCodeblocks/components/Code.tsx
MeguminSama: If the values in `lines` ever changes, the contents of the elements returned from the map can get mixed up badly. In this case "state" isn't referring to things like useState, but just the state of the components in general (props, children, etc)

ArjixWasTaken: from what I see, lines is assigned to only once during the function's lifetime
(well, the try/catch could count as two assignments)

the component would fully re-render if the lines change

PS: In case I am THAT blind, please do point it out

src/modules.d.ts
Vap0r1ze: not intentional, prob some git merge and eslint magic, will fix

src/plugins/shikiCodeblocks/components/Code.tsx
Vap0r1ze: when i made the first iteration of this plugin like 2y ago discord's hljs only had an html output, i'll see if it has an AST output in its current state

Vendicated: .highlight("js", "console.log('owo')", true).emitter.rootNode.children returns an AST like structure but you'd have to format that yourself so probably not really usable. It's fine to use dangerouslySetInnerHTML since hljs is safe but it's just a bit terrifying to look at

Vap0r1ze: afaict there is only an html output

Vap0r1ze:
Let me know if I missed anything in [`0cfa569`](https://github.com/Vendicated/Vencord/pull/267/commits/0cfa56985f388de446cbdb3a6797942bb2bb9254)

src/plugins/reviewDB/components/ReviewsView.tsx
```diff
@@ -27,7 +27,12 @@ import ReviewComponent from "./ReviewComponent";
 const Classes = findLazy(m => typeof m.textarea === "string");
 
 export default function ReviewsView({ userId }: { userId: string; }) {
-    const [reviews, _, isLoading, refetch] = useAwaiter(() => getReviews(userId), []);
+    const [refetchCount, setRefetchCount] = React.useState(0);
+    const [reviews, _, isLoading] = useAwaiter(() => getReviews(userId), {
+        fallbackValue: []
+    });
+
+    const dirtyRefetch = () => setRefetchCount(refetchCount + 1);
```
Vendicated, line 35: pretty sure this won't work? even if the state changes, useAwaiter won't rerun its body. Did you forget to add the refetch count as dependency?

Vap0r1ze: ah yep

src/plugins/shikiCodeblocks/index.ts
```diff
@@ -0,0 +1,153 @@
...
+                    default: true,
+                },
+                {
+                    label: "Primary",
+                    value: HljsSetting.Primary,
+                },
+                {
+                    label: "Always",
+                    value: HljsSetting.Always,
+                },
```
Vendicated, line 109: what exactly do these do?

src/plugins/shikiCodeblocks/index.ts
```diff
@@ -0,0 +1,153 @@
...
+            ],
+        },
+        useDevIcon: {
+            type: OptionType.SELECT,
+            description: "How to show devicons on codeblocks",
```
Vendicated, line 114:
```suggestion
            description: "How to show language icons on codeblocks",
```
(?) I'm not sure what devicons stands for

Vap0r1ze: its the library i use for lang icons, this is prob better since its user-facing text

src/utils/misc.tsx
```diff
@@ -28,34 +28,44 @@ export function makeLazy<T>(factory: () => T): () => T {
     return () => cache ?? (cache = factory());
 }
 
-type AwaiterRes<T> = [T, any, boolean, () => void];
+type AwaiterRes<T> = [T, any, boolean];
+interface AwaiterOpts<T> {
+    fallbackValue: T,
+    deps?: unknown[],
+    onError?: (e: unknown) => unknown,
```
Vendicated, line 35:
```suggestion
    onError?(e: any): void,
```
unknown is absolute pain to work with

Vap0r1ze: this was ur code lol

Vendicated: lol really

src/plugins/shikiCodeblocks/index.ts
Vap0r1ze: Decides when to use hljs highlighter rather than shiki
`Primary` uses hljs if the language exists in hljs, otherwise use shiki
`Secondary` uses shiki if the language exists in shiki, otherwise use hljs

Vendicated: Always is the same as disabling the plugin? If so, why even have that option

Would be nice to change the labels to a more descriptive text so it's easier to Understand. If it looks weird because too much text you could use the settingsAboutComponent prop

Vap0r1ze: `Always` still has the shiki theme background and the copy button, in the future i plan on adding a language switcher too

Vap0r1ze: and also line numbers

Vap0r1ze:
[`39e53c5`](https://github.com/Vendicated/Vencord/pull/267/commits/39e53c5d4f9d95c1ac965549df57f0e3700b20ba)

Vap0r1ze:
Done and merged

Vendicated:
uhhh 
image

image

Vendicated:
This crashed Discord btw - To prevent crashes please wrap components in an ErrorBoundary

Vendicated (approved):

PR #268, opened 2022-11-28
Path aliases, better lazyWebpack

Vendicated:


Vendicated:
I have no clue why I didn't do this right from the start and why this helper even was in the util folder instead of webpack

PR #269, opened 2022-11-28
feat: Proper CSS api & css bundle

Vendicated:
Need to figure out a good way to do this in the userscript and also let fridge know he needs to update the loader


Vendicated:
Issue: The updater will not download renderer.css. Need to find a fix for this / implement a temporary hack that will download the renderer.css file if it doesn't exist

src/api/Styles.ts
```diff
@@ -0,0 +1,70 @@
...
+    const style = ensureStyle(name);
+    style.vars = vars;
+    if (style.mount?.isConnected) compileStyle(style);
+};
+
+export const compileStyle = (style: Style) => {
+    if (!style.mount) throw new Error("Style is not mounted");
+
+    style.mount.textContent = style.source
+        .replace(/(?<=#\{)(\w+)(?=\})/g, (_, name) => style.vars[name] ?? "");
```
Vendicated, line 69: what exactly does this do?

Vendicated: it seems to enable some sort of variables? That should be documented

src/api/Styles.ts
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * // Inside some plugin method like "start()" or "[option].onChange()"
+ * enableStyle(pluginStyle);
+ */
+export const enableStyle = (name: string) => {
+    const style = ensureStyle(name);
+
+    if (style.dom?.isConnected) return false;
+
+    style.dom ??= document.createElement("style");
```
Vendicated, line 46: imo this should have some id so it's obvious where it comes from

Vendicated: like id = "vc-styles-NAME"

Vap0r1ze: true

src/api/Styles.ts
```diff
@@ -0,0 +1,158 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import type { MapValue } from "type-fest/source/entry";
+
+export type Style = MapValue<typeof VencordStyles>;
+
+export const styleMap = window.VencordStyles ??= new Map();
+
+export const ensureStyle = (name: string) => {
```
Vendicated, line 25: ~~bikeshedding but~~ requireStyle would probably be a better name

src/api/Styles.ts
```diff
@@ -0,0 +1,158 @@
...
+ * setStyleClassNames(pluginStyle, classNames);
+ * enableStyle(pluginStyle);
+ * ```
+ * ```scss
+ * // -- plugin.css --
+ * .plugin-root [--thin]::-webkit-scrollbar { ... }
+ * ```
+ * ```scss
+ * // -- final stylesheet --
+ * .plugin-root .thin-31rlnD.scrollerBase-_bVAAt::-webkit-scrollbar { ... }
```
Vendicated, line 102: pog

src/api/Styles.ts
Vap0r1ze: 🚴 F2 it u wont

scripts/build/common.mjs
```diff
@@ -147,6 +147,31 @@ export const fileIncludePlugin = {
     }
 };
 
+const styleModule = await readFile("./scripts/build/module/style.js", "utf-8");
```
Vendicated, line 150:
```suggestion
const styleModule = readFileSync("./scripts/build/module/style.js", "utf-8");
```

sync is way faster than async and the await makes it synchronous anyway

Vap0r1ze: i figured it would allow esbuild to read multiple files in parallel

scripts/build/common.mjs
```diff
@@ -147,6 +147,31 @@ export const fileIncludePlugin = {
...
+/**
+ * @type {import("esbuild").Plugin}
+ */
+export const stylePlugin = {
+    name: "style-plugin",
+    setup: ({ onResolve, onLoad }) => {
+        onResolve({ filter: /\.css\?managed$/, namespace: "file" }, ({ path, resolveDir }) => ({
+            path: relative(process.cwd(), join(resolveDir, path.replace("?managed", ""))),
```
Vap0r1ze, line 158: i willl figure this out later

Vendicated:
This is basically ready now, just need to wait for ArmCord to support it

PR #271, opened 2022-11-28
Fix Background Colour for Plugin Card & Settings

nmsturcke:
Fix background colour for plugin card & settings when having a custom theme.

### Before
image
image

## After
image
image

### DIFF
--- a/src/components/PluginSettings/styles.ts
+++ b/src/components/PluginSettings/styles.ts
@@ -24,7 +24,7 @@ export const PluginsGrid: React.CSSProperties = {
 };
 
 export const PluginsGridItem: React.CSSProperties = {
-    backgroundColor: "var(--background-modifier-selected)",
+    backgroundColor: "var(--primary-dark-530)",
     color: "var(--interactive-active)",
     borderRadius: 3,
     cursor: "pointer",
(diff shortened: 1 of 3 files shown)


Vendicated:
are you sure this isn't just your theme being bad? it works fine for me and the card is consistent with that in the streamer mode settings page

MeguminSama:
I suggest you modify your theme to support the missing css variables. The variables used have been kept consistent with other settings pages.

nmsturcke:
> are you sure this isn't just your theme being bad? it works fine for me and the card is consistent with that in the streamer mode settings page

I'm using Discord's theme experiment feature thing

nmsturcke:
image
this

MeguminSama:
The experiment is exactly that - an experiment. I doubt its complete yet.

nmsturcke:
Okay, I'll close then ig.

MeguminSama:
Thank you though, will keep this in mind to test once/if Discord chooses to release this feature.

Vendicated:
it works as expected for me, even with the experiment. the dono and button cards are themed correctly and the plugin cards while being ugly are themed the same as the highlight which is the expected behaviour 
image
image

nmsturcke:
> it works as expected for me, even with the experiment. the dono and button cards are themed correctly and the plugin cards while being ugly are themed the same as the highlight which is the expected behaviour image image

Not for me 🤷 
image
image

PR #274, opened 2022-11-30
Add support for Flatpak for Git updating 

abUwUser:
Yes, the way to verify if it is being ran on Flatpak looks bad and it requires more testing, specifically with updating it. Also, it needs checking to see if it works without Flatpak

Roadmap:
[ ] Runs with Discord not installed with Flatpak
[ ] Updates successfully with or without Flatpak


docs/1_INSTALLING.md
```diff
@@ -64,6 +64,8 @@ pnpm inject
 
 Then fully close Discord from your taskbar or task manager, and restart it. Vencord should be injected - you can check this by looking for the Vencord section in Discord settings.
 
+**NOTE:** If you installed Discord using Flatpak, run `sudo flatpak override com.discordapp.Discord --talk-name=org.freedesktop.Flatpak` to give permissions for Discord to use Portals.
```
Vendicated, line 67: 1) The installer should do this
2) I think this command straight up gives full host access to the host which some users might find undesirable. Would be useful to find out if there's a way to restrict it to git in a specific directory. Perhaps we could somehow expose a shell script to the flatpak
3) it appears flatpak-spawn is in the `flatpak-xdg-utils` package which not everyone might have installed (I don't)

abUwUser: 1. Going to add it if I can't find a way to realize 2
2. Yeah, looks like a better option, but i have no idea how to do that. Time to google
3. Huh, if I knew how to allow access to git (like what [com.visualstudio.code](https://github.com/flathub/com.visualstudio.code) does), I would do that

As of now, get this commit

abUwUser: According to [here](https://discourse.gnome.org/t/access-system-commands-in-flatpak/5682/2), flatpak apps are isolated to other apps, and they can't access anything other than what their package bundles provides. A.K.A. node and git would need to be provided by the flatpack package if you would like to make it 100% sandboxed. As I understood, 2 is impossible to do without literally slapping git in it

For 3, due to the restrictions of 2, there isn't much to do without using flatpak-spawn or slapping git inside the flatpak

abUwUser: As a way to kinda met the concerns with 2, maybe we could prompt that the installation will disable sandbox for Discord, and ask if he is fine with that

scripts/patcher/install.js
```diff
@@ -79,6 +79,26 @@ async function install(installations) {
             );
             process.exit(1);
         }
+
+        const answer = await question(
+            `The installer will give full host access to ${selected.branch} in order to enable updating using Git. Proceed? [Y/n]: `
```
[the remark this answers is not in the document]
abUwUser: gonna check, but thanks for the suggestion!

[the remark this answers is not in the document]
abUwUser: Apparently no. I wonder what talk-name
image

If I could take a wild guess it would be [org.freedesktop.portal.Flatpak](https://docs.flatpak.org/en/latest/libflatpak-api-reference.html#gdbus-org.freedesktop.portal.Flatpak)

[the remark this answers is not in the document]
abUwUser: Nope
image

Vendicated (approved):

abUwUser:
is it ready to merge?

abUwUser:
oh nice. pog

PR #277, opened 2022-12-01
Add Firefox extension build

romangraef:
Because there are so few differences between manifest v2 and v3 aside
from header rewriting i just wrote two different manifests. We also generate a .xpi file that gets added to the nightly releases on main pushes. I am aware of #69, however that PR does not support manifest v2 (yet?), and it is still a few months till firefox releases manifest v3, so I think until then this would be a worthwhile effort to maintain.


romangraef:
This needs setting `WEBEXT_USER` and `WEBEXT_SECRET` from  https://addons.mozilla.org/en-US/developers/addon/api/key in order to work with the signed releases

.github/workflows/build.yml
```diff
@@ -33,6 +33,12 @@ jobs:
             - name: Build web
               run: pnpm buildWeb --standalone
 
+            - name: Sign firefox extension
+              run: pnpm web-ext sign --api-key $WEBEXT_USER --api-secret $WEBEXT_SECRET
```
Vendicated, line 37: could this use pnpx? this way we don't have to depend on web-ext which is absolutely massive

browser/background.js
```diff
@@ -0,0 +1,45 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+function setContentTypeOnStylesheets(details) {
+    if (details.type === "stylesheet") {
+        details.responseHeaders.push({ name: "Content-Type", value: "text/css" });
```
Vendicated, line 21: shouldn't this try to find an existing one first? otherwise we'll have duplicate headers

browser/manifestv2.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "web_accessible_resources": [
+        "dist/Vencord.js"
+    ],
+    "background": {
+        "scripts": [
+            "background.js"
+        ]
+    }
+}
```
Vendicated, line 33: isn't this manifest missing the gecko specific id to even be accepted by web-ext?

PR #281, opened 2022-12-02
fix inconsistent margins & capitalisation

davidcralph:
CSS:
* Margins in Vencord settings are now consistent 😩 

English:
* Fancy link to the GitHub signup page in instructions for theme config is here
* QuickCss no longer exists and has been replaced with the superior QuickCSS, along with StartUp and Development
* Whether or not you like the change to the themes wording is up to you

Enjoy


src/components/VencordSettings/ThemesTab.tsx
```diff
@@ -52,12 +52,13 @@ function Validators({ themeLinks }: { themeLinks: string[]; }) {
     return (
         <>
             <Forms.FormTitle className={Margins.marginTop20} tag="h5">Validator</Forms.FormTitle>
-            <Forms.FormText>This section will tell you whether your themes can successfully be loaded</Forms.FormText>
+            <Forms.FormText>This section will tell you if your themes be loaded successfully</Forms.FormText>
```
MeguminSama, line 55: This is grammatically incorrect

MeguminSama (dismissed):

src/components/VencordSettings/ThemesTab.tsx
davidcralph: this was the result of a lack of sleep and will be remediated, thank you for your concern

PR #286, opened 2022-12-03
add commas in member count

Commandtechno:
a lot easier to read the member count, especially in large servers!

before

image

after

image


Commandtechno:
lint seems to work now 👍

image

not sure if any code formatting is needed...

PR #289, opened 2022-12-04
feat(plugin): ExternalFileUpload

somerandomcloud:
Upload files to an external file server via a simple slash command (/upload)


ArjixWasTaken:
Having progress indicators would be nice, although not currently supported by the `fetch` API, for progress callbacks you might want to use the `xmlhttprequest` API (although this would not work for the userscript build, you will have to use the fetch polyfill there, it has progress callbacks)

@Vendicated I think we should make the `fetch` polyfill accessible on other builds (as a util, not a polyfill)
(but it would use `xmlhttprequest` instead of `GM_xmlhttpRequest` when its not a userscript build)

somerandomcloud:
> Having progress indicators would be nice

I will include this if I continue working on this plugin, but this was really just a quick (5h) functional mock-up for a user in the discord.

castdrian:
If you do continue to work on it, maybe importing sharex configs could be done to directly upload to your own server and send the url it returns, just throwing it out here

somerandomcloud:
> If you do continue to work on it, maybe importing sharex configs could be done to directly upload to your own server and send the url it returns, just throwing it out here

You probably dont need a ShareX config, just the login kek. gonna start back up one of my sharex servers, and start testing in a bit

somerandomcloud:
Ok so scratch what I said before, you do need ShareX config. I implemented a pretty basic system for now, but there are a few issues that I will fix later tonight:

- No way to auth implemented yet
- Getting the final url for the file works on one json layer, but not more (stupid JSON in TS....)
- Generally janky with some uploading services (notably, uguu.se. I have been wrestling with this service since yesterday)

These issues should all be easily fixable, just gotta do it. Buuuut for now I am going to head out and get a kebab

Vendicated:
image

PR #290, opened 2022-12-04
Updates on PluginModal, making it more well integrated with discord's UI

abUwUser:
Some changes were made to make the PluginModal UI not look like a prototype version of itself, like moving the plugin authors to PluginCard and removing the redundant About Me section
image
image


MeguminSama:
image
"redundant about section"

abUwUser:
@MeguminSama I mean as this
image

MeguminSama:
Yes, you've removed the about section and not given another way to view it. The descriptions shown in the grid of plugins is not the full description.

abUwUser:
> Yes, you've removed the about section and not given another way to view it. The descriptions shown in the grid of plugins is not the full description.

Oh, okay. Will see about that

abUwUser:
Well, the only field of description that is used is the `description` object, so if you meant that plugins have a small description and a full description then that's false. Other than that, descriptions aren't cut midhalf, because I've removed the height limit, and well, it is more of a issue with the plugin developer if they put too big descriptions, plus descriptions cut in half aren't that great

x6r:
so instead of having a convenient way to view description as well as other slightly less relevant info about the plugin (like the authors) you invalidate long description with no other way to see them because uhhhhhh???

and this does not look good to me
image

MeguminSama (changes requested):
As much as I appreciate the effort, I just don't see how this is beneficial.

You're now sending API requests for tons of user profiles en-masse when you load the plugin page - this is not the kind of change I will accept.

You've also removed the ability to see a full description for plugins with longer or more detailed descriptions.

I can't accept this PR until these fundamental issues are resolved.

abUwUser:
Well, true, now I am sending multiple requests for tons of users, and I guess there isn't a good solution for that, so I will close this PR for that.

But really, could you give an example of a plugin with detailed description? I don't find any examples of it

PR #291, opened 2022-12-04
style: 🎨 make viewRaw 100% better

RadNotRed:


Vendicated:
show a screenshot please, and I don't like this keybind change but it could just be a setting

RadNotRed:
image

and the setting bit would work

x6r:
why is it blue

RadNotRed:
idk it just feels right having a copy button as blue

Vendicated:
the blue doesn't fit in at all and the key change isn't really desirable

PR #292, opened 2022-12-05
feat(plugin): noMoreEmbeds

castdrian:
Wraps all urls in chevrons to prevent embeds


Vendicated:
We already have enough plugins that edit text so I will not be accepting any more. Instead, someone should make a more generic plugin that allows you to specify own replacements. Thank you anyway

PR #297, opened 2022-12-07
feat(ShikiCodeblocks): Settings preview

Vap0r1ze:
This modifies the `PluginModal` component to pass down an optional `tempSettings` prop to `settingsAboutComponent`.
The prop works because each setting will create a new object onChange

### DIFF
--- a/src/plugins/shikiCodeblocks/components/Highlighter.tsx
+++ b/src/plugins/shikiCodeblocks/components/Highlighter.tsx
@@ -42,6 +42,7 @@ export interface HighlighterProps {
     lang?: string;
     content: string;
     isPreview: boolean;
+    tempSettings?: Record<string, any>;
 }
 
 export const createHighlighter = (props: HighlighterProps) => (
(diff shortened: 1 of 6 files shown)


Vendicated:
Good idea, I didn't have in mind that the about component will have outdated settings

src/plugins/shikiCodeblocks/components/Highlighter.tsx
```diff
@@ -42,6 +42,7 @@ export interface HighlighterProps {
     lang?: string;
     content: string;
     isPreview: boolean;
+    tempSettings?: Record<string, any>;
```
Vendicated, line 45: this could technically be typed strongly via generics based on the options object of your plugin, but that's probably out of scope of this pr

Vap0r1ze: Yea that would be good if there was a settings rework for strong typings

Issue #303, opened 2022-12-08
[Feature Request] Plugin to replace all notification sounds with .mp3's


Luna-devv:
Just as the title says. A plugin with what you can change all notification sounds in discord using .mp3 files. ([discord://-/settings/notifications](discord://-/settings/notifications))

ArjixWasTaken:
already requested in the discord, and I said that I am working on it
it was also already made in the past, but the implementation was lacking so it was closed

ArjixWasTaken:
btw the title `Plugin to replace all notification sounds with .mp3's` doesn't say much
they are already mp3s to begin with...

Luna-devv:
with custom mp3 files

PR #305, opened 2022-12-09
feat: Sound Changer Plugin

ArjixWasTaken:
Allows you to set a custom link for each discord sound that exists.

Closes: #303, #322

### DIFF
--- a/src/plugins/soundChanger.tsx
+++ b/src/plugins/soundChanger.tsx
@@ -1,0 +1,153 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Settings } from "@api/settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import { LazyComponent, makeLazy } from "@utils/misc";
+import definePlugin, { OptionType } from "@utils/types";
+import { findLazy, findModuleId, wreq } from "@webpack";
+import { React, Select, Text, TextArea } from "@webpack/common";
+
+const useEffect = makeLazy(() => React.useEffect);
+const useState = makeLazy(() => React.useState);
+const TextAreaProps = findLazy(m => typeof m.textarea === "string");
+const DeleteIcon = LazyComponent(() => wreq(findModuleId("M15 3.999V2H9V3.999H3V5.999H21V3.999H15Z")!!).Z);
+
+type SoundsChanged = {
+    name: string;
+    original_link: string;
+    new_link: string;
+}[];
+
+const SoundChangerSettings = ({ setValue }: { setValue: (newValue: any) => void; }) => {
+    const [sounds, setSounds] = useState()({});
+    const [soundsChanged, setSoundsChanged] = useState()<SoundsChanged>([]);
+
+    useEffect()(() => {
+        const modId = findModuleId("call_ringing.mp3");
+        if (modId == null) return;
+
+        setSounds(JSON.parse(`${wreq.m[modId]}`.match(/\w=(\{(?:.|\n)*?\})/)!![1]));
+        setSoundsChanged((Settings.plugins.SoundChanger.soundsChanged || []) as SoundsChanged);
+    }, []);
+
+    const save = (a: SoundsChanged) => {
+        setValue(a.filter(s => s.new_link.trim().length > 0));
+    };
+
+    return (
+        <ErrorBoundary>
+            <table>
+                <thead>
+                    <th><Text variant="text-md/bold" style={{ textAlign: "left" }}>Name</Text></th>
+                    <th><Text variant="text-md/bold" style={{ textAlign: "left" }}>Custom sound</Text></th>
+                    <th></th>
+                </thead>
+                <tbody>
+                    {
+                        soundsChanged.map(sound => {
+                            return (
+                                <tr key={sound.name} style={{ marginTop: "5%" }}>
+                                    <td style={{ textAlign: "left", top: "50%", transform: "translateY(-50%)", width: "50%" }} >
+                                        <Text variant="text-sm/normal">{sound.name.replace("./", "")}</Text>
+                                    </td>
+                                    <td style={{ width: "45%" }}>
+                                        <TextArea
+                                            className={TextAreaProps.textarea}
+                                            placeholder="Link to an mp3"
+                                            spellCheck={false}
+                                            onChange={e => {
+                                                for (const s of soundsChanged) {
+                                                    if (s.name === sound.name) {
+                                                        s.new_link = e.nativeEvent.target.value;
+                                                        setSoundsChanged(old => { save(soundsChanged); return soundsChanged; });
+                                                        break;
+                                                    }
+                                                }
+                                            }}
+                                            value={soundsChanged.find(s => s.name === sound.name)?.new_link}
+                                        />
+                                    </td>
+                                    <td
+                                        style={{ transform: "translateY(-40%)", width: "5%" }}
+                                        onClick={() => {
+                                            const newSounds = soundsChanged.filter(s => s.name !== sound.name);
+                                            setSoundsChanged(old => { save(newSounds); return newSounds; });
+                                        }}>
+                                        <span style={{ cursor: "pointer" }}>
+                                            <DeleteIcon width="24" height="24" color="var(--status-danger)" />
+                                        </span>
+                                    </td>
+                                </tr>
+                            );
+                        })
+                    }
+                    {!soundsChanged.length &&
+                        <Text style={{
+                            transform: "translateX(50%)",
+                            marginTop: "10%",
+                            marginBottom: "25%"
+                        }} variant="text-sm/normal">No sounds have been changed.</Text>
+                    }
+                </tbody>
+            </table>
+
+            <div style={{ marginBottom: "5%" }}>
+                <Select
+                    options={
+                        Object.keys(sounds)
+                            .filter(sound => soundsChanged.find(a => a.name === sound) === undefined)
+                            .map(sound => ({ label: sound.replace("./", ""), value: sound }))
+                    }
+                    placeholder="Change a sound"
+                    maxVisibleItems={5}
+                    closeOnSelect={true}
+                    isSelected={() => false}
+                    serialize={a => String(a)}
+                    select={(sound: string) => {
+                        const filename = `${wreq.m[sounds[sound]]}`.match(/"(.*?\.mp3)"/)!![1];
+                        const newSounds = soundsChanged.concat([{ name: sound, original_link: filename, new_link: "" }]);
+                        setSoundsChanged(old => { save(newSounds); return newSounds; });
+                    }}
+                />
+            </div>
+        </ErrorBoundary>
+    );
+};
+
+export default definePlugin({
+    name: "SoundChanger",
+    authors: [Devs.Arjix],
+    description: "Allows you to modify any discord sound.",
+    options: {
+        soundsChanged: {
+            description: "",
+            type: OptionType.COMPONENT,
+            component: SoundChangerSettings,
+            restartNeeded: true
+        }
+    },
+    patches: [{
+        find: "call_calling.mp3",
+        replacement: {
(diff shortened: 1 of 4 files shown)


ArjixWasTaken:
I'd love some help with making the settings prettier, but they are fine me thinks.

MeguminSama:
I don't think checking link validity is very important for this

ArjixWasTaken:
I'd love some code reviews @Vendicated 
I know that the settings UI sucks, I just can't see how to improve it, since I lack the talent.

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                    placeholder="Change a sound"
+                    maxVisibleItems={5}
+                    closeOnSelect={true}
+                    isSelected={() => false}
+                    serialize={a => String(a)}
+                    select={(sound: string) => {
+                        const filename = `${wreq.m[sounds[sound]]}`.match(/"(.*?\.mp3)"/)!![1];
+                        const newSounds = soundsChanged.concat([{ name: sound, original_link: filename, new_link: "" }]);
+                        setSoundsChanged(_ => { save(newSounds); return newSounds; });
```
ArjixWasTaken, line 128: should prob just do
```suggestion
                        setSoundsChanged(newSounds);
```

for this one, cause `save(...)` would be pointless since it filters out any empty sounds, and it is bound to be empty at this point in time

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                                                for (const s of soundsChanged) {
+                                                    if (s.name === sound.name) {
+                                                        s.new_link = e.nativeEvent.target.value;
+                                                        setSoundsChanged(_ => { save(soundsChanged); return soundsChanged; });
+                                                        break;
+                                                    }
+                                                }
+                                            }}
+                                            value={soundsChanged.find(s => s.name === sound.name)?.new_link}
+                                        />
```
ArjixWasTaken, line 87: Would be nice if I could highlight the text area in red to show that a link is invalid.
Maybe doing a `HEAD` request to the link to check if it fails or not?

(by invalid link I am not only referring to it being malformed, but also non-existent (or it 404ing), if I only wanted to check for it being valid per the URL spec, I could just construct a URL instance and try/catch it)

ArjixWasTaken (commented):
These are my personal thoughts on this

src/plugins/soundChanger.tsx
ArjixWasTaken: I will push a commit for this btw, its just that I noticed it whilst making my review

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+    new_link: string;
+}[];
+
+const SoundChangerSettings = ({ setValue }: { setValue: (newValue: any) => void; }) => {
+    const [sounds, setSounds] = useState()({});
+    const [soundsChanged, setSoundsChanged] = useState()<SoundsChanged>([]);
+
+    useEffect()(() => {
+        const modId = soundsModule();
+        if (modId == null) return; // Should never happen, but oh well
```
ArjixWasTaken, line 45: I could probably show a notification saying that the module was not found and to contact the dev[s]

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    <th><Text variant="text-md/bold" style={{ textAlign: "left" }}>Custom sound</Text></th>
+                    <th></th>
+                </thead>
+                <tbody>
+                    {
+                        soundsChanged.map(sound => {
+                            return (
+                                <tr key={sound.name} style={{ marginTop: "5%" }}>
+                                    <td style={{ textAlign: "left", top: "50%", transform: "translateY(-50%)", width: "50%" }} >
+                                        <Text variant="text-sm/normal">{sound.name.replace("./", "")}</Text>
```
ArjixWasTaken, line 70: Should I parse and display the names in a humanly way?
Like instead of `call_ringing.mp3` to display `Call Ringing`?

ArjixWasTaken: It also would be awesome if I added a play button after the name, similarly to how discord does in the notification settings.
image

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                                    <td
+                                        style={{ transform: "translateY(-40%)", width: "5%" }}
+                                        onClick={() => {
+                                            const newSounds = soundsChanged.filter(s => s.name !== sound.name);
+                                            setSoundsChanged(_ => { save(newSounds); return newSounds; });
+                                        }}>
+                                        <span style={{ cursor: "pointer" }}>
+                                            <DeleteIcon width="24" height="24" color="var(--status-danger)" />
+                                        </span>
+                                    </td>
```
ArjixWasTaken, line 98: I should move the click listener to the element with the mouse pointer, so that no accidental deletions happen.

```suggestion
                                    <td style={{ transform: "translateY(-40%)", width: "5%" }}>
                                        <span
                                            style={{ cursor: "pointer" }}
                                            onClick={() => {
                                                const newSounds = soundsChanged.filter(s => s.name !== sound.name);
                                                setSoundsChanged(_ => { save(newSounds); return newSounds; });
                                        }}>
                                            <DeleteIcon width="24" height="24" color="var(--status-danger)" />
                                        </span>
                                    </td>
```

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                            );
+                        })
+                    }
+                    {!soundsChanged.length &&
+                        <Text style={{
+                            transform: "translateX(50%)",
+                            marginTop: "10%",
+                            marginBottom: "25%"
+                        }} variant="text-sm/normal">No sounds have been changed.</Text>
+                    }
```
ArjixWasTaken, line 109: Could prob improve this, maybe add the discord artwork for when smth is missing?

src/plugins/soundChanger.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    maxVisibleItems={5}
+                    closeOnSelect={true}
+                    isSelected={() => false}
+                    serialize={a => String(a)}
+                    select={(sound: string) => {
+                        const filename = `${wreq.m[sounds[sound]]}`.match(/"(.*?\.mp3)"/)!![1];
+                        const newSounds = soundsChanged.concat([{ name: sound, original_link: filename, new_link: "" }]);
+                        setSoundsChanged(newSounds);
+                    }}
+                />
```
ArjixWasTaken, line 130: I really love how well this works, but it feels off that a select element is used like this, I just don't know how to have the same kind of dropdown with a more intuitive approach.

ArjixWasTaken (commented):
some more notes

src/plugins/soundChanger.tsx
ArjixWasTaken: Would this be better if I added a label to the left?
But I like how wide it is right now.

src/plugins/soundChanger.tsx
ArjixWasTaken: Ok, added a toast for now, any better ideas?

ArjixWasTaken:
@Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+    }],
+    start() {
+        this.style = document.createElement("style");
+        this.style.id = "SoundChanger";
+        this.style.innerHTML = cssText;
+        document.head.appendChild(this.style);
+    },
+    stop() {
+        document.head.removeChild(this.style);
+    }
```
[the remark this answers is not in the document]
ArjixWasTaken: nvm, it aint invisible, it actually still kinda works, it just isnt fancy

ArjixWasTaken:
image

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+        a.load();
+    });
+};
+
+const SoundChangerSettings = ({ setValue }: { setValue: (newValue: any) => void; }) => {
+    const [sounds, setSounds] = useState()({});
+    const [soundsChanged, setSoundsChanged] = useState()<SoundsChanged>([]);
+    const [audioPreviews, setAudioPreviews] = useState()<{ [key: string]: HTMLAudioElement; }>({});
+
+    useEffect()(() => {
```
Vendicated, line 64: this looks super strange, why not use proxyLazy or make a wrapper instead?
```js
const useEffect= (...args: Parameters<typeof React.useEffect>) => React.useEffects(args)
```

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    {
+                        soundsChanged.map(sound => {
+                            return (
+                                <tr key={sound.name} style={{ marginTop: "5%" }}>
+                                    <td style={{ top: "50%", transform: "translateY(-50%)", width: "50%" }} >
+                                        <div className="SC sound-label" style={{ display: "flex", flexDirection: "row", textAlign: "left", rowGap: "10px", columnGap: "10px", justifyItems: "center" }}>
+                                            <Text style={{ marginTop: "0.8%" }} variant="text-sm/normal">{filenameToTitleCase(sound.name)}</Text>
+                                            <span
+                                                onClick={e => {
+                                                    const revTreePath = (e as any).nativeEvent.path as HTMLElement[];
```
Vendicated, line 105: what in the world does this span and the entire onclick do? You also should not be using a span, use a proper button or at least use the role attribute

ArjixWasTaken: ...have you actually used the plugin?
It is the audio icon next to the sound name

You click it to preview the audio, you can click it again to stop.
The icon is only visible when you hover over the sound

Similar to discord's notification settings

ArjixWasTaken: I used a span cause by default it is only as big as the content within it. And it doesn't force other content to be below it.

I'll add the role tomo

ArjixWasTaken: looks like tomo meant 3 days later

ArjixWasTaken:
@Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated @Vendicated

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+    original_link: string;
+    new_link: string;
+}[];
+
+const snakeToTitleCase = txt => txt.split(/_+/).map((w: string) => {
+    const word = w.trim();
+    return word[0].toUpperCase() + word.slice(1).toLowerCase();
+}).join(" ");
+
+const filenameToTitleCase = txt => snakeToTitleCase(txt.replace(".mp3", "").replace("./", ""));
```
Vendicated, line 47: utils/text has this sort of stuff

src/plugins/soundChanger/soundChangerStyles.css
```diff
@@ -0,0 +1,22 @@
+.SC.sound-label svg:not(.audio-preview-playing) {
```
Vendicated, line 1: please use a more descriptive class to avoid potential conflicts

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+            // Should never happen, but you never know...
+            Toasts.show({
+                type: Toasts.Type.FAILURE,
+                message: "Couldn't find the module responsible for managing the sound links.",
+                id: Toasts.genId()
+            });
+            return;
+        }
+
+        setSounds(JSON.parse(`${wreq.m[modId]}`.match(/\w=(\{(?:.|\n)*?\})/)!![1]));
```
Vendicated, line 78: this should just be a memo

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+        setSoundsChanged((Settings.plugins.SoundChanger.soundsChanged || []) as SoundsChanged);
```
Vendicated, line 79: why does this setting need to be a state?

src/plugins/soundChanger/index.tsx
Vendicated: what I mean by what does this do is that this is dom manipulation instead of using proper React features. You really should avoid that at all cost, it's an anti pattern and might cause crashes

src/plugins/soundChanger/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+                                                        setSoundsChanged(_ => { save(soundsChanged); return soundsChanged; });
+                                                        break;
+                                                    }
+                                                }
+                                            }}
+                                            onBlur={async e => {
+                                                const link = e.nativeEvent.target.value.trim();
+                                                if (!link.length) return;
+
+                                                // TODO: Do more than just toasting an error message...
```
Vendicated, line 151: Please complete or remove this todo

PR #306, opened 2022-12-10
feat(plugin): add OpenInApp

paintingofblue:
Only supports Spotify and Steam, but can add others if requested


src/plugins/openInApp.ts
```diff
@@ -0,0 +1,44 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Sofia Lima
```
Vendicated, line 3: did u forget to update the copyright header?

PR #308, opened 2022-12-10
bypasser 2: electric boogaloo

exhq:
Screenshot_20220415_175630

### DIFF
--- a/src/plugins/bypasser.ts
+++ b/src/plugins/bypasser.ts
@@ -1,6 +1,6 @@
 /*
  * Vencord, a modification for Discord's desktop app
- * Copyright (c) 2022 Vendicated and contributors
+ * Copyright (c) 2022 exhq
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
should rename it to `Automod Bypass[er]`
~~and also, how does this work? you replace every character with itself, can't see the bypass~~

EDIT: wow, github does not show zero width characters, lmao, I thought there was an update that highlighted them, did they undo it?

src/plugins/bypasser.ts
```diff
@@ -0,0 +1,73 @@
...
+    },
+    commands: [
+        {
+            name: "bypass",
+            description: "bypass automod",
+            options: [RequiredMessageOption],
+            execute: opts => ({
+                content: findOption(opts, "message", "").replaceAll(
+                    /(.)/g,
+                    "$1​‍‍⁤"
```
Vap0r1ze, line 52: You should use unicode escapes rather than putting zero-width characters inside code.
```js
"$1\u200b\u200d\u200d\u2064"
```

src/plugins/bypasser.ts
```diff
@@ -0,0 +1,73 @@
...
+                ),
+            }),
+        },
+    ],
+
+    addPrefix(msg: MessageObject) {
+        msg.content = msg.content.replaceAll(/(.)/g, "$1​‍‍⁤");
```
Vap0r1ze, line 59: Same here
```js
"$1\u200b\u200d\u200d\u2064"
```

Vendicated:
<:husk:859796756111294474>

PR #332, opened 2022-12-12
Improve Settings UI & View Raw Modal

abUwUser:
This PR adds some visual tweaks to some parts of Vencord, such as:
- Plugin Settings
    - Adding a close button
    - Removes the footer if the plugin does not have settings
    - Puts the Authors on its separated section
    - Fixes some styling to fit with discord's UI
image
- View Raw
    - Fixes some styling issues
    - Puts the copy JSON and copy message content buttons on a footer
image
- VSCode config changes
    - Makes so that VSC uses the TS version from node_modules


Vap0r1ze:
i like the copy raw style change, did you add any bottom padding the inner container, it used to touch the edge

abUwUser:
> i like the copy raw style change, did you add any bottom padding the inner container, it used to touch the edge

yep i did

src/components/PluginSettings/PluginModal.tsx
```diff
@@ -121,33 +123,33 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
     }
 
     function renderSettings() {
-        if (!pluginSettings || !plugin.options) {
+        if (!hasSettings || !plugin.options) {
```
abUwUser, line 126: had to put another `!plugin.options` because typescript went dumb and didn't noticed that hasSettings already covers plugin.options

.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
     "javascript.format.semicolons": "insert",
     "typescript.format.semicolons": "insert",
     "typescript.preferences.quoteStyle": "double",
-    "javascript.preferences.quoteStyle": "double"
-}
+    "javascript.preferences.quoteStyle": "double",
+    "typescript.tsdk": "node_modules/typescript/lib"
```
Vendicated, line 16: why? what's the benefit of this?

src/plugins/viewRaw.tsx
```diff
@@ -89,30 +89,33 @@ function openViewRawModal(msg: Message) {
         <ErrorBoundary>
             <ModalRoot {...props} size={ModalSize.LARGE}>
                 <ModalHeader>
-                    <Forms.FormTitle tag="h1">View Raw</Forms.FormTitle>
+                    <Text variant="heading-lg/medium" style={{ flexGrow: 1 }}>View Raw</Text>
```
Vendicated, line 92: why? a heading makes more sense here semantically

abUwUser: what do you mean by heading?

Vendicated: you changed it from a heading (h1) to a normal text which breaks semantics and makes no sense

.vscode/settings.json
Vendicated: Is there any issue with the vscode supplied one?

abUwUser: nope, but rarely the installed version on vscode can be different from the repo one, making vscode to think that features exists in where it doesn't. e.g. an project uses version 4.6 and vscode uses 4.9

ArjixWasTaken: you would assume that such errors are caught when building the project since the builder does not use vscode's typescript...
so does it really matter?

src/plugins/viewRaw.tsx
abUwUser: the Forms.FormTitle adds additional margin to the bottom, which makes ModalHeader to stretch
image
image
the change to Text with a heading-lg/medium is to fit better with discord's modals
image
also checking now, it is "semibold", not "medium". gonna make a commit changing that

.vscode/settings.json
abUwUser: > you would assume that such errors (from the difference in TS versions) are caught when building the project since the builder does not use vscode's typescript...

i mean, it would be annoying to see an eternal red underlined text because of VSCode not having the correct TS version, or worse and more probably, TS saying that a feature does not exist while VSCode says it does

PR #333, opened 2022-12-12
Rename the plugin "Oneko" to "Neko"

abUwUser:
The program "Oneko" is a fork of "Xneko", that in itself it is a port of "Neko" (made originally for the NEC PC-9801) for the X windowing system. Since the plugin "Oneko" does not have any affiliation with program Oneko, but has with Neko, it is better to rename it. [Information at Wikipedia](https://en.wikipedia.org/wiki/Neko_(software))

It is better to still name internally as "Oneko" internally [since adryd has made that mistake](https://github.com/adryd325/oneko.js) and that it would require a PR in that repository to rename everything


Vendicated:
it doesn't really matter, oneko is just the Linux port of the original program that people will be familiar with the most likely

abUwUser:
it matters, as people will eventually associate the plugin to "oneko", a program that it is a port of the original program. it makes more sense to associate the plugin with the original program than with the port

abUwUser:
plus it is just a renaming to better match with history of the program, not that it wouldn't break vencord or anything

ArjixWasTaken:
first of all, neko means cat, and it aint a unique program name...changing it to neko wouldn't make people think that it is related to another project named neko.

Besides, the plugin is using assets directly from the oneko repository, so its not as if the plugin is not related to oneko.

abUwUser:
> first of all, neko means cat, and it aint a unique program name...changing it to neko wouldn't make people think that it is related to another project named neko.

yes true, but this does not explain the sense to make it related to oneko, since it is a port of a desktop application. using ProtonGE as an example, if you made a script that is compatible with Proton, but you thought that Proton does not exist and that ProtonGE exists, and then named your script as "ProtonGE-script", would you rename to "Proton-script" if you wanted to also support Proton?

> Besides, the plugin is using assets directly from the oneko repository, so its not as if the plugin is not related to oneko.

theorically, it does use the ones from oneko.js, but that's still correct

ArjixWasTaken:
> would you rename to "Proton-script" if you wanted to also support Proton?

nope, I'd pick a more descriptive name, lol, but I get what you are trying to say

> theorically, it does use the ones from oneko.js, but that's still correct

When I say oneko I am referring to oneko.js, I didn't even know of another oneko.

abUwUser:
> When I say oneko I am referring to oneko.js, I didn't even know of another oneko.

oh lol

PR #337, opened 2022-12-15
feat(NoReplyMention): exempt list support

Axyriah:
Mentions automatically users specified on the plugin settings


src/utils/constants.ts
```diff
@@ -14,163 +14,169 @@
  *
```
MeguminSama, line 14: Why did you reformat the entire file?

MeguminSama (changes requested):

src/utils/constants.ts
Axyriah: my bad, prettierd automatically did it

Vendicated: you should use the vscode formatter, not prettier

Axyriah: ayo what for some reason it applied things that shouldn't have been aplied

MeguminSama: No worries :)

Vendicated:
I think you accidently pushed green text? it doesn't look finished at least

Axyriah:
> I think you accidently pushed green text? it doesn't look finished at least

yep, it just got saved from another branch and git didn't remove it
about the vscode formatter, i can't use it since i'm using neovim

src/plugins/noReplyMention.tsx
```diff
@@ -14,22 +14,45 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
-*/
+ */
 
+// eslint-disable-next-line header/header
```
MeguminSama, line 19: What's this needed for?

src/plugins/noReplyMention.tsx
```diff
@@ -14,22 +14,45 @@
...
+import { Settings } from "@api/settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
+
+type Reply = {
```
MeguminSama, line 24: For consistency, could you make this an interface instead of a type?

src/plugins/noReplyMention.tsx
Axyriah: That `.eslintrc.json` on the root seems to get me errors in both neovim and vscode on that header for some reason

MeguminSama: The last `*/` in the header shouldn't have a space in front of it - maybe it'll stop erroring if you remove it?
Also the header is above the next-line comment so that shouldn't be doing anything? Unless I'm misunderstanding something.

src/plugins/noReplyMention.tsx
Axyriah: Sure

src/plugins/noReplyMention.tsx
```diff
@@ -16,20 +16,44 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+
```
MeguminSama, line 19: Just to be nit-picky, could you remove the extra newline? :P

Axyriah: done 🗿🍷

MeguminSama (approved):

Vendicated (approved):

PR #339, opened 2022-12-16
refactor: identifier escapes + "self" group

Vap0r1ze:
Additions:
  - A **custom regex escape** for `PatchReplacement["match"]`, `\i`, that aliases the SWC base54 identifier pattern (`[A-Za-z_$][\w$]*`).
  	- ~~Does not work in getters~~
  - **Eslint support** for the above
  - Adds a **`$self` alias** for `PatchReplacement["replace"]` (eg: `"$self.run();"` => `"Vencord.Plugins.plugins.ShikiCodeblocks.run();"`)
    - ~~Does not work in functions or getters~~

### DIFF
--- a/src/plugins/shikiCodeblocks/index.ts
+++ b/src/plugins/shikiCodeblocks/index.ts
@@ -22,7 +22,7 @@ import { wordsFromPascal, wordsToTitle } from "@utils/text";
 import definePlugin, { OptionType } from "@utils/types";
 
 import previewExampleText from "~fileContent/previewExample.tsx";
-import cssText from "~fileContent/style.css";
+import cssText from "~fileContent/shiki.css";
 
 import { Settings } from "../../Vencord";
 import { shiki } from "./api/shiki";
@@ -44,8 +44,8 @@ export default definePlugin({
         {
             find: "codeBlock:{react:function",
             replacement: {
-                match: /codeBlock:\{react:function\((.),(.),(.)\)\{/,
-                replace: "$&return Vencord.Plugins.plugins.ShikiCodeblocks.renderHighlighter($1,$2,$3);",
+                match: /codeBlock:\{react:function\((\i),(\i),(\i)\)\{/,
+                replace: "$&return $self.renderHighlighter($1,$2,$3);",
             },
         },
     ],
--- a/src/utils/patches.ts
+++ b/src/utils/patches.ts
@@ -21,16 +21,13 @@ export type ReplaceFn = (match: string, ...groups: string[]) => string;
 export function canonicalizeMatch(match: RegExp | string) {
     if (typeof match === "string") return match;
     const canonSource = match.source
-        .replace(/(?<=(?:^|[^\\])(?:\\\\)*)\\i/g, "[A-Za-z_$][\\w$]*");
+        .replaceAll("\\i", "[A-Za-z_$][\\w$]*");
     return new RegExp(canonSource, match.flags);
 }
 
 export function canonicalizeReplace(replace: string | ReplaceFn, pluginName: string) {
     if (typeof replace === "function") return replace;
-    return replace.replace(
-        /(?<=(?:^|[^$])(?:\$\$)*)\$self/gi,
-        `Vencord.Plugins.plugins.${pluginName}`,
-    );
+    return replace.replaceAll("$self", `Vencord.Plugins.plugins.${pluginName}`);
 }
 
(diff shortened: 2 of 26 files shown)


src/plugins/shikiCodeblocks/index.ts
```diff
@@ -44,8 +44,8 @@ export default definePlugin({
         {
             find: "codeBlock:{react:function",
             replacement: {
-                match: /codeBlock:\{react:function\((.),(.),(.)\)\{/,
-                replace: "$&return Vencord.Plugins.plugins.ShikiCodeblocks.renderHighlighter($1,$2,$3);",
+                match: /codeBlock:\{react:function\((\i),(\i),(\i)\)\{/,
+                replace: "$&return $self.renderHighlighter($1,$2,$3);",
             },
```
Vap0r1ze, line 49: Usage example

MeguminSama:
Is there a particular reason you chose not to add support for `$self` in function replacements?

Vap0r1ze:
> Is there a particular reason you chose not to add support for `$self` in function replacements?

I can probably do this via Proxies along with the getters tbh

Vap0r1ze:
Now works with functions and getters

src/webpack/patchWebpack.ts
```diff
@@ -135,15 +135,16 @@ function patchPush() {
...
-                        // @ts-ignore we change all patch.replacement to array in plugins/index
+                        if (!Array.isArray(patch.replacement)) continue;
                         for (const replacement of patch.replacement) {
                             if (replacement.predicate && !replacement.predicate()) continue;
                             const lastMod = mod;
                             const lastCode = code;
 
                             try {
+                                // @ts-ignore String#replace has no [string, function] overload
                                 const newCode = code.replace(replacement.match, replacement.replace);
```
Vendicated, line 146: huh?

Vap0r1ze: Since `.match` is `string | RegExp`, and `.replace` is `string | ReplaceFn`, TypeScript will make sure the `String#replace` signature has assignable params for every combo of those unions. There'll be an error since there is no overload for `string.replace(string, match => replacement)`

Vendicated: there is though

Vap0r1ze: oh

Vap0r1ze: not sure why i thought that, but i'll remove the false explanation (it does error if u remove it)

Vendicated: why? it didn't before

Vap0r1ze: it did, you were just `ts-ignore`ing the whole for loop block rather than that line explicitly

src/utils/patches.ts
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+export type ReplaceFn = (match: string, ...groups: string[]) => string;
+
+export function canonicalizeMatch(match: RegExp | string) {
+    if (typeof match === "string") return match;
+    const canonSource = match.source
+        .replace(/(?<=(?:^|[^\\])(?:\\\\)*)\\i/g, "[A-Za-z_$][\\w$]*");
```
Vendicated, line 24: why is such a complex regex needed?

src/utils/patches.ts
```diff
@@ -0,0 +1,46 @@
...
+    return new RegExp(canonSource, match.flags);
+}
+
+export function canonicalizeReplace(replace: string | ReplaceFn, pluginName: string) {
+    if (typeof replace === "function") return replace;
+    return replace.replace(
+        /(?<=(?:^|[^$])(?:\$\$)*)\$self/gi,
```
Vendicated, line 31: same here

src/utils/patches.ts
Vap0r1ze: as to not replace `/\\i/`

src/utils/patches.ts
Vap0r1ze: same here but with the edge case `"$$self"`

Vap0r1ze:
tested on these plugins and all of the patches were successful
image

Vendicated (approved):

MeguminSama (approved):

PR #340, opened 2022-12-16
WIP - youtuber

exhq:
Screenshot_20220511_182544


ArjixWasTaken:
You check for `open.spotify.com`
And then proceed to assume that it is a track...

It could be an album, you don't check

PR #343, opened 2022-12-16
feat(PlatformIndicators): add indicator to channel messages

ActuallyTheSun:
image

this adds a `MessageDecorationsAPI` so other plugins can add their own stuff, and the indicators on the member list were moved to a new `MemberListDecoratorsAPI` (both api's are basically clones of `MessageAccessoriesAPI` :trollface:)

profile badges were not switched to `BadgesAPI` as that requires an image link. is it worth making the api allow passing your own component as a badge?


Vendicated:
I think allowing custom components would definitely be useful, would for example allow lazy fetching badges

src/api/MemberListDecorators.ts
```diff
@@ -0,0 +1,65 @@
...
+    decorators.set(identifier, { decorator, onlyIn });
+}
+
+export function removeDecorator(identifier: string) {
+    decorators.delete(identifier);
+}
+
+export function __addDecoratorsToList(props: DecoratorProps): (JSX.Element | null)[] {
+    const isInGuild = !!(props.guildId);
+    return [...decorators].map(([_, decoratorObj]) => {
```
Vendicated, line 57:
```suggestion
    return decorators.values().map((decoratorObj) => {
```

src/api/MessageDecorations.ts
```diff
@@ -0,0 +1,63 @@
...
+export function addDecoration(identifier: string, decoration: Decoration) {
+    decorations.set(identifier, decoration);
+}
+
+export function removeDecoration(identifier: string) {
+    decorations.delete(identifier);
+}
+
+export function __addDecorationsToMessage(props: DecorationProps): (JSX.Element | null)[] {
+    return [...decorations].map(([_, decoration]) => {
```
Vendicated, line 60:
```suggestion
    return [...decorations.values()].map(decoration => {
```

src/plugins/platformIndicators.tsx
```diff
@@ -16,13 +16,16 @@
...
+import { addDecorator, removeDecorator } from "@api/MemberListDecorators";
+import { addDecoration, removeDecoration } from "@api/MessageDecorations";
 import { Settings } from "@api/settings";
 import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByCodeLazy } from "@webpack";
 import { PresenceStore, Tooltip } from "@webpack/common";
-import { User } from "discord-types/general";
+import User from "discord-types/general/User.js";
```
Vendicated, line 28: whaaa

src/plugins/platformIndicators.tsx
```diff
@@ -75,79 +78,90 @@ const PlatformIndicator = ({ user }: { user: User; }) => {
...
-            }}
+            style={{ marginLeft: "4px", gap: "4px" }}
         >
             {icons}
-        </div>
-    );
+        </span>;
+
+    if (span) return indicator;
+    return <div style={{ display: "flex", alignItems: "center" }}>{indicator}</div>;
```
Vendicated, line 90: whats this for? the element shouldnt matter

src/plugins/platformIndicators.tsx
ActuallyTheSun: idk vscode just felt like it

src/plugins/platformIndicators.tsx
ActuallyTheSun: if a div is user for the indicator in messages it goes funny
image

i just left the div from the originan plugin but i think simply not using it anywhere works

src/plugins/apiBadges.tsx
```diff
@@ -66,11 +66,18 @@ export default definePlugin({
...
+            replacement: [
+                {
+                    match: /src:(\w{1,3})\[(\w{1,3})\.key\],/,
+                    // <img src={badge.image ?? imageMap[badge.key]} {...badge.props} />
+                    replace: (_, imageMap, badge) => `src: ${badge}.image ?? ${imageMap}[${badge}.key], ...${badge}.props,`
+                },
+                {
+                    match: /spacing:(\d{1,2}),children:(.{1,40}(.{1,2})\.jsx.+(.{1,2})\.onClick.+\)})},/,
+                    replace: (_, s, origBadgeComponent, React, badge) =>
+                        `spacing:${s},children:${badge}.component?()=>(0,${React}.jsx)(${badge}.component,{...${badge}}):${origBadgeComponent}},`
```
Vendicated, line 78: what exactly does this do? Please add a comment

src/plugins/platformIndicators.tsx
Vendicated: >simply not using it anywhere works
as in, always use span? If so, just remove the div and always do span

src/plugins/platformIndicators.tsx
```diff
@@ -75,79 +78,90 @@ const PlatformIndicator = ({ user }: { user: User; }) => {
...
+        description: "In user profiles, as badges",
+        onEnable: () => addBadge(badge),
+        onDisable: () => removeBadge(badge)
+    },
+    messages: {
+        description: "Inside messages",
+        onEnable: () => addDecoration("platform-indicator", props =>
+            <ErrorBoundary noop><PlatformIndicator user={
+                props.decorations[1]?.find(i => i.key === "new-member")?.props.message?.author
+            } span /></ErrorBoundary>
```
Vendicated, line 117: sorry but could you please format these components on newlines, something like
```jsx
<ErrorBoundary>
    <PlatformIndicator />
</ErrorBoundary>

File: src/plugins/apiMemberListDecorators.ts [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "MemberListDecoratorsAPI",
+    description: "API to add decorators to member list (both in servers and DM's)",
```
Vendicated, line 24:
```suggestion
    description: "API to add decorators to member list (both in servers and DMs)",
```

src/plugins/platformIndicators.tsx
```diff
@@ -75,79 +80,95 @@ const PlatformIndicator = ({ user }: { user: User; }) => {
...
+        description: "In user profiles, as badges",
+        onEnable: () => addBadge(badge),
+        onDisable: () => removeBadge(badge)
+    },
+    messages: {
+        description: "Inside messages",
+        onEnable: () => addDecoration("platform-indicator", props =>
+            <ErrorBoundary noop>
+                <PlatformIndicator user={
+                    props.decorations[1]?.find(i => i.key === "new-member")?.props.message?.author
```
Vendicated, line 121: whats decorations[1]? Is hardcoding this index safe? What if a plugin adds its own decoration at the start, would this shift to 2?

Vendicated (approved):
other than those small things, looks good!

src/plugins/platformIndicators.tsx
ActuallyTheSun: > whats decorations[1]?
[it has comments](https://github.com/Vendicated/Vencord/blob/7393ccd44af724b0a6724198ef4975217b8af2dc/src/api/MessageDecorations.ts#L34-L43)

> Is hardcoding this index safe?
yes

> What if a plugin adds its own decoration at the start, would this shift to 2?
no

PR #344, opened 2022-12-16
Also push builds to https://github.com/Vencord/builds

pointydev:
First attempt, probably doesn't work. Have fun fixing it <3

Thanks,
Elliott

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,10 +55,21 @@ jobs:
               run: |
                 echo "release_tag=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
 
-            - name: Upload Devbuild
+            - name: Upload Devbuild as release
               run: |
                 gh release upload devbuild --clobber dist/*
                 gh release edit devbuild --title "DevBuild $RELEASE_TAG"
               env:
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                 RELEASE_TAG: ${{ env.release_tag }}
+
+            - name: Upload Devbuild to builds repo
+              run: |
+                gh repo clone $GH_REPO upload
+                cp -fr dist/ upload/
+                cd upload
+                git add -A
+                git commit -a -m "Upload devbuild"
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                GH_REPO: Vencord/builds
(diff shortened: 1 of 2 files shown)


ArjixWasTaken:
any reason as to why this is useful in any kind of way?

pointydev:
> any reason as to why this is useful in any kind of way?

because i was asked to in order to fix my own problem :husk:
https://discord.com/channels/1015060230222131221/1015060231060983891/1052726531182563408

ArjixWasTaken:
:husk:

bad Ven

.github/workflows/build.yml
```diff
@@ -55,10 +55,22 @@ jobs:
...
                 gh release upload devbuild --clobber dist/*
                 gh release edit devbuild --title "DevBuild $RELEASE_TAG"
               env:
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                 RELEASE_TAG: ${{ env.release_tag }}
+
+            - name: Upload Devbuild to builds repo
+              run: |
+                gh repo clone $GH_REPO upload
+                cp -fr dist/ upload/
```
Vendicated, line 69: this creates the folder dist into upload? Wouldn't it make more sense to copy the contents of it only?

Vendicated: i also think deleting all files in the repo beforehand would be good to clean up possibly outdated files

something like (untested)
```sh
rm -rf upload/*
GLOBIGNORE=.git:.gitignore
rm -rf upload/.*
cp -r dist/* upload
```

.github/workflows/build.yml
```diff
@@ -55,10 +55,22 @@ jobs:
...
+                cd upload
+                git add -A
+                git commit -a -m "Upload devbuild"
+                git push
```
Vendicated, line 73: this should probably be a force push to prevent potential issues
```suggestion
                git push --force origin main
```

.github/workflows/build.yml
```diff
@@ -55,10 +55,22 @@ jobs:
...
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Vendicated, line 75: i think this secret only has permissions to the current repo, so I'll have to add my own to the repo secrets

pointydev: Changed to `secrets.BUILDS_TOKEN` so you can set that.

Vendicated (approved):

PR #352, opened 2022-12-20
[SpotifyControls] Add option to show Controls on hover

HypedDomi:
7dMuU8


src/plugins/spotifyControls/index.tsx
```diff
@@ -17,15 +17,38 @@
 */
 
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
+import { Settings } from "Vencord";
```
Vendicated, line 21:
```suggestion
import { Settings } from "@api/settings";
```

src/plugins/spotifyControls/index.tsx
```diff
@@ -17,15 +17,38 @@
...
 
 import { Player } from "./PlayerComponent";
 
+const toggleHoverControls = (value: boolean) => {
```
Vendicated, line 25:
```suggestion
function toggleHoverControls(value: boolean) {
```

Vendicated:
You should avoid transitioning things like height, width, left/right/... This is because they force the browser to relayout and can't be gpu accelerated which makes them feel choppy, especially on older hardware. It would be much nicer to use translateY to achieve this

Good read: https://medium.com/outsystems-experts/how-to-achieve-60-fps-animations-with-css3-db7b98610108

PR #354, opened 2022-12-20
feat(plugin) KeyboardClicks

wotanut:
Adds the option to turn on "Keyboard Clicks" Similar to Opera where when you type a click sound is played.


Vendicated:
i don't think this is very useful. Why not just do this on a system level?

wotanut:
You might want this plug-in if:

- you only want keyboard clicks *in* discord.
- your operating system doesn't support click sounds (at least not easily) I believe ChromeOS does not support these as an example.
- you're an idiot and have no idea how to enable them in settings.

Whilst I agree this isn't *the* most useful/coolest plug-in in the repo, it's not exactly the most useless.

paintingofblue:
just use mechakeys

wotanut:
From what I can tell about mechakeys is that
- it does support Windows
- According to their website they only support Windows but on their GitHub they support Linux, MacOS and Windows (confusing but something I cba to verify so i'll assume they support all three)
- In any case this still does not support users running Vanilla ChromeOS with Discord Web and Vencord Web, nor does it account for keyclicks in discord (and probably not for the less technologically capable people who don't know how to run a Linux VM on ChromeOS)

And at the end of the day, running a couple of lines of a plugin in a browser is probably going to be easier on a crappy computer (which most Chromebooks are) than an entirely separate app doing it's own thing.

ArjixWasTaken:
> i don't think this is very useful. Why not just do this on a system level?

90% of vencord plugins aren't objectively useful (im looking at you fart2)

src/plugins/keyboardClicks.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+const keydown = (e: KeyboardEvent) => {
+    const audio = new Audio("https://cdn.discordapp.com/attachments/1019011011786833921/1054854818109345913/click2.wav");
+    audio.play();
+};
+
+export default definePlugin({
+    name: "keyboardClicks",
```
Vendicated, line 28:
```suggestion
    name: "KeyboardClicks",
```

Vendicated:
Hi! I'm closing this for inactivity, feel free to reopen if you're still interested in getting this merged

PR #355, opened 2022-12-20
feat(plugins): DisableDMCallIdle

Nuckyz:


ArjixWasTaken:
I didn't know that js regex supported named groups.

I thought that was a python only feature.

PR #361, opened 2022-12-23
css-import: style api

Vap0r1ze:
This targets the `css-import` branch


Vap0r1ze:
First commit is a merge from main, review here: [`6b415f3..HEAD`](https://github.com/Vendicated/Vencord/pull/361/files/6b415f3..HEAD)
Alongside the pre-existing css stuff, there is a working style manager API with variable support (useful for loading in classnames from modules).

One issue atm is that the injected stylesheet's precedence exists between user-agent and DOM styles.
Also, what are your thoughts on Sass support?

PR #369, opened 2022-12-25
feat(plugin): TimedLightTheme

sunnniee:
automatically toggles light or dark theme based on time of day

closes #363


MeguminSama:
using a two-point slider instead of text fields could look pretty cool...

ActuallyTheSun:
since a two point slider is not happening, would something like this work instead? (clicking the buttons switches the selected slider)
image
image
i think it would be better to just stick to text inputs though

ArjixWasTaken:
why is it not happening?

src/plugins/timedLightTheme.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { Settings } from "@api/settings";
+import { Devs } from "@utils/constants.js";
+import Logger from "@utils/Logger";
+import definePlugin, { OptionType } from "@utils/types";
+import { waitFor } from "@webpack";
+import { UserSettingsProtoStore } from "@webpack/common";
+
+let updateTheme: (theme: "light" | "dark") => Promise<void>;
+waitFor(["updateTheme"], m => ({ updateTheme } = m));
```
Vendicated, line 27: please use lazyWebpack to avoid unnecessary searches if the plugin is disabled

src/plugins/timedLightTheme.ts
```diff
@@ -0,0 +1,89 @@
...
+        end: {
+            description: "When to enter dark mode (24-hour time)",
+            type: OptionType.STRING,
+            default: "20:00",
+            placeholder: "xx:xx",
+            isValid: t => /^\d{0,2}(?::\d{0,2})?$/.test(t),
+        },
+    },
+
+    // eslint-disable-next-line consistent-return
```
Vendicated, line 53: i think i disabled this rule so this ignore should be obsolete (?) 

the rule is a good idea to catch bugs but annoying often so I think I did

src/plugins/timedLightTheme.ts
```diff
@@ -0,0 +1,89 @@
...
+        const theme = getTheme();
+        // do not remove
+        if (!theme) return null;
+
+        if ((now >= endTimestamp || now <= startTimestamp) && theme === "light") return updateTheme("dark");
+        else if ((now <= endTimestamp && now >= startTimestamp) && theme === "dark") return updateTheme("light");
+    },
+
+    start() {
+        interval = setInterval(() => this.checkForUpdate(), 60_000);
```
Vendicated, line 82: Why use an interval? I would instead calculate the time until next theme shift and use setTimeout instead. If you extract it into a method, it should be fairly simple to restart the timer on settings changes

src/plugins/timedLightTheme.ts
ActuallyTheSun: it's still enabled

Vendicated: i see, feel free to disable it then!

ActuallyTheSun:
in settings, after an error is thrown by `isValid`, nothing checks to remove the error once the input is valid again
image

is there any issue if i add a check or would that break something?
image

ActuallyTheSun:
review?

ArjixWasTaken:
never

ArjixWasTaken:
sad to see this go

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+const Themes: Record<number, string> = findLazy(m => m[16] === "EASTER_EGG");
+const updateTheme: (gah: { theme: "light" | "dark", backgroundGradientPresetId?: number; }) => Promise<void> = findByCodeLazy("clientThemeSettings:{");
+
+const getBasicTheme = () =>
+    PreloadedUserSettings.getCurrentValue().appearance.theme === 1 ? "dark" : "light";
+const getClientThemeId = () =>
+    PreloadedUserSettings.getCurrentValue().appearance.clientThemeSettings?.backgroundGradientPresetId?.value as number ?? -1;
+
+const canUseClientThemes = () =>
+    Vencord.Plugins.isPluginEnabled("FakeNitro") || (UserStore.getCurrentUser().premiumType ?? 0) === 2;
```
Vendicated, line 43:
```suggestion
    isPluginEnabled("FakeNitro") || (UserStore.getCurrentUser().premiumType ?? 0) === 2;
```

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+            updateThemeIfNecessary("dark");
+            nextChange = setTimeout(() => this.checkForUpdate(), startTimestamp - now);
+        }
+        else if (now >= startTimestamp && now <= endTimestamp) {
+            updateThemeIfNecessary("light");
+            nextChange = setTimeout(() => this.checkForUpdate(), endTimestamp - now);
+        }
+        else if (now > endTimestamp) {
+            updateThemeIfNecessary("dark");
+            nextChange = setTimeout(() => this.checkForUpdate(), (startTimestamp + 86400_000) - now);
```
Vendicated, line 171: could you add a comment stating how much time this is? 
```suggestion
            nextChange = setTimeout(() => this.checkForUpdate(), (startTimestamp + 86_400_000) - now);
```

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+// hh:mm to unix timestamp
+function toAdjustedTimestamp(t: string): number {
+    const [hours, minutes] = t.split(":").map(i => i && parseInt(i, 10));
+    return new Date().setHours(hours as number, minutes || 0, 0, 0);
+}
+function showError() {
+    Toasts.show({
+        message: "TimedLightTheme - Invalid settings",
```
Vendicated, line 71: this is not very informative. you should advise the user how to solve the issue

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+interface PreferredTheme {
+    light: number;
+    dark: number;
+}
+
+let nextChange: NodeJS.Timeout;
+
+const PreloadedUserSettings = findLazy(m => m.ProtoClass?.typeName?.includes("PreloadedUserSettings"));
+const Themes = findLazy(m => m[16] === "EASTER_EGG") as Record<number, string>;
+const updateTheme = findByCodeLazy("clientThemeSettings:{") as ({ }: { theme: "light" | "dark", backgroundGradientPresetId?: number; }) => Promise<void>;
```
Vendicated, line 35:
```suggestion
const updateTheme = findByCodeLazy("clientThemeSettings:{") as (data: { theme: "light" | "dark", backgroundGradientPresetId?: number; }) => Promise<void>;
```

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+// the plugin uses -1 as a fallback for no client theme. discord doesn't
+interface PreferredTheme {
+    light: number;
+    dark: number;
+}
+
+let nextChange: NodeJS.Timeout;
+
+const PreloadedUserSettings = findLazy(m => m.ProtoClass?.typeName?.includes("PreloadedUserSettings"));
```
Vendicated, line 33: pretty sure this is a webpack common now? If not, please make it one, sine @utils/discord also uses it

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { definePluginSettings } from "@api/Settings.js";
+import { Devs } from "@utils/constants.js";
```
Vendicated, line 20:
```suggestion
import { Devs } from "@utils/constants";
```

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+const Themes = findLazy(m => m[16] === "EASTER_EGG") as Record<number, string>;
+const updateTheme = findByCodeLazy("clientThemeSettings:{") as ({ }: { theme: "light" | "dark", backgroundGradientPresetId?: number; }) => Promise<void>;
+
+const getBasicTheme = () =>
+    PreloadedUserSettings.getCurrentValue().appearance.theme === 1 ? "dark" : "light";
+const getClientThemeId = () =>
+    PreloadedUserSettings.getCurrentValue().appearance.clientThemeSettings?.backgroundGradientPresetId?.value as number ?? -1;
+
+const canActuallyUseClientThemes = () =>
+    (UserStore.getCurrentUser().premiumType ?? 0) === 2;
```
Vendicated, line 43:
```suggestion
    UserStore.getCurrentUser().premiumType === 2;
```

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+        message: "TimedLightTheme - Invalid settings",
+        id: Toasts.genId(),
+        type: Toasts.Type.FAILURE,
+        options: {
+            duration: 3000,
+            position: Toasts.Position.BOTTOM
+        }
+    });
+}
+const snakeToTitleCase = (text: string) =>
```
Vendicated, line 80: this might or might not exist in @utils/text

src/plugins/timedLightTheme.tsx
```diff
@@ -0,0 +1,186 @@
...
+}
+
+const settings = definePluginSettings({
+    start: {
+        description: "When to enter light mode (24-hour time)",
+        type: OptionType.STRING,
+        default: "08:00",
+        placeholder: "xx:xx",
+        isValid: t => /^\d{0,2}(?::\d{0,2})?$/.test(t),
+        onChange: () => Vencord.Plugins.isPluginEnabled("TimedLightTheme") && (Vencord.Plugins.plugins.TimedLightTheme as any).checkForUpdate(),
```
Vendicated, line 117: if you make checkForUpdates a top level function, you can avoid this Vencord.Plugins.plugins.Plugin as any hack

additionally, just import isPluginEnabled instead of using the global!

PR #373, opened 2022-12-29
feat(messageActions): make features toggleable

favna:
This changes the `messageAction` plugin to make it configurable as to whether the "delete on click" and "double click to edit" features should be enabled or not.

The motivation for this change is that as a server moderator I've had it happen many times that `isDeletePressed` was incorrectly set to `true` and I ended up deleting messages of other users incorrectly which is extreme frustration. Because I want to keep using "double click to edit" I do not want to disable the plugin outright, so the best next option is to make both features configurable.


Vendicated:
> The motivation for this change is that as a server moderator I've had it happen many times that isDeletePressed was incorrectly set to true and I ended up deleting messages of other users incorrectly which is extreme frustration

I don't understand that, why not fix the actual underlying bug then instead of adding a workaround?

I also don't like that the way you implemented it means that options require a restart. Why not do the check inside the handler instead so it immediately works after toggling?

favna:
> I don't understand that, why not fix the actual underlying bug then instead of adding a workaround?

If I knew what the bug was, sure. But I have absolutely no idea how to debug the issue either because it doesn't always happen and I cannot reproduce it actively. 

> I also don't like that the way you implemented it means that options require a restart. Why not do the check inside the handler instead so it immediately works after toggling?

I tried it without restarting and that didn't work. I'm not sure how Vencord's `start` is handled internally but toggling `Vencord.Settings.plugins.MessageClickActions.enableDoubleClickToEdit` and _not_ restarting ends up not updating the handler because `start` doesn't get called anew. I couldn't find anything like a `settingsUpdate` method.

Vendicated:
> I tried it without restarting and that didn't work. I'm not sure how Vencord's start is handled internally but toggling Vencord.Settings.plugins.MessageClickActions.enableDoubleClickToEdit and not restarting ends up not updating the handler because start doesn't get called anew. I couldn't find anything like a settingsUpdate method.

Well you already check the setting inside the handler, so you could just remove the condition and **always** register the handler!

Vendicated:
Just did the change myself since it was simple. Thank you!

favna:
Coulda sworn when I tested it, it didn't work but yeah.. it does now. Thanks for resolving it!

Vendicated:
maybe forgot to restart or something 😄

PR #386, opened 2023-01-08
add @someone command

C10udburst:


SammCheese:
:trollface:

Nuckyz (changes requested):
Should probably properly handle when the command is used on a DM

Vendicated:
I don't think I want this to be part of the mod since it has the potential to be abused a lot and annoy a lot of people. I also don't want random children pinging random people for no reason on the discord

PR #391, opened 2023-01-09
git updater: Fix macOS

HypedDomi:
Fixes this issue:
image


Vendicated:
I don't like this since this will break things if someone has their node in /usr/bin or similar. Why not add /usr/local/bin to PATH instead and let the system handle the lookup?

Vendicated:
like
```js
if (process.platform === "darwin") process.env.PATH = `/usr/local/bin:${process.env.PATH}`
```
then just use node like before

PR #400, opened 2023-01-11
feat: Add option to disable the window frame

zt64:


AAGaming00:
guhhh useless almost all wm can do this

insane

src/patcher.ts
```diff
@@ -65,12 +65,15 @@ if (!process.argv.includes("--vanilla")) {
         };
     }
 
+    const settings = JSON.parse(readSettings());
```
Vendicated, line 68: should be in try catch in case user corrupts settings, like from where u removed this

zt64: Should I remove the other try catch or keep that too

zt64: I left it

src/patcher.ts
```diff
@@ -118,7 +124,6 @@ if (!process.argv.includes("--vanilla")) {
         });
 
         try {
-            const settings = JSON.parse(readSettings());
             if (settings.enableReactDevtools)
```
Vendicated, line 127:
```suggestion
            if (settings?.enableReactDevtools)
```

PR #403, opened 2023-01-12
feat: Typesafe Settings Definitions

Vap0r1ze:
This PR creates an alternative to the `options` field that is more typesafe: a new `settings` plugin field that takes in settings, which can be defined using `definePluginSettings`

You can check the changes to see shiki as a usage example, but a [^1]basic overview:
```ts
  someText: { type: OptionType.STRING },
  amount: { type: OptionType.NUMBER },
}, {
  // value is implicitly a string
  someText: {
    isValid(value) { return value.length > 5 && this.store.amount > 0 }
  },
})

// After plugin initialization
settings.store.someText; // string
settings.store.someText = "hello"; // will send out updates to existing react hooks

// React hook (optional filter to avoid needless updates)
settings.use() // { someText: string, amount: number }
settings.use(["amount"]) // { amount: number }
```

[^1]: Some fields were left out for simplicity, I think only `description`

`definePluginSettings(defs[, checks])`: this uses essentially the same format as the `options` property, except the option properties that are (what I'm calling) "checks". A "check" is a option property that:
   - Is a method
   - Has a return value that may rely on other plugin settings

At the moment the only "check" fields that exist are `disabled` and `isValid`


src/api/settings.ts
```diff
@@ -198,3 +198,19 @@ export function migratePluginSettings(name: string, ...oldNames: string[]) {
         }
     }
 }
+
+export function definePluginSettings<D extends SettingsDefinition, C extends SettingsChecks<D>>(def: D, checks?: C) {
+    const definedSettings: DefinedSettings<D> = {
+        get store() {
```
MeguminSama, line 204: Could you add some jsdoc to these properties?

MeguminSama (approved):
There's a lot going on here, but code works & looks fine to me.

src/api/settings.ts
Vap0r1ze: yop

Vap0r1ze:
> There's a lot going on here, but code works & looks fine to me.

I tried to make it completely opt-in and compatible with the `options` field type-wise. Runtime-wise the `settings` field is just [an alias for `options`](https://github.com/Vendicated/Vencord/pull/403/files#diff-66ace9db4d3f14bc7ff52498147b0a9ef6e33070073803893afdaee7be8e3198R63-R72) at the moment.

src/utils/types.ts
```diff
@@ -206,12 +212,67 @@ interface IPluginOptionComponentProps {
...
+
+/** An instance of defined plugin settings */
+export interface DefinedSettings<D extends SettingsDefinition = SettingsDefinition, C extends SettingsChecks<D> = {}> {
+    /** Shorthand for `Vencord.Settings.plugins.PluginName`, but with typings */
+    store: SettingsStore<D>;
+    /**
+     * React hook for getting the settings for this plugin
+     * @param filter optional filter to avoid rerenders for irrelavent settings
+     */
+    use<F extends Extract<keyof D, string>>(filter?: F[]): F extends `${any}` ? Pick<SettingsStore<D>, F> : never;
```
Vendicated, line 245: image

Vendicated (changes requested):

Vendicated (approved):

Vendicated:
this is epic btw, now migrate all plugins and I'll buy you a pizza 🚎 💯

PR #405, opened 2023-01-12
feat(plugin): SpeechBubble

Rapougnac:
Basically, it adds a way to put speech-bubbles on top of the provided image


src/webpack/common.tsx
```diff
@@ -64,6 +65,9 @@ export let RelationshipStore: Stores.RelationshipStore & {
     getSince(userId: string): string;
 };
 
+export const fetchUser = findByCodeLazy(".USER(");
```
Vendicated, line 68: could you please update other plugins using these modules (petpet for example)

src/plugins/speechBubble.ts
```diff
@@ -0,0 +1,179 @@
...
+                    throw "Failed to fetch user. Check the console for more info.";
+                }
+        }
+    }
+    return null;
+}
+
+export default definePlugin({
+    name: "SpeechBubble",
+    description: "Adds a speech bubble in the top of an image",
```
Vendicated, line 84:
```suggestion
    description: "Slash command to create speechbubble memes",
```

Vendicated:
Sorry for only looking at this now!
Why not make the speechbubble part transparent?
It would also be cool to have the option to add it on the bottom, similar to esmBot

Rapougnac:
> Why not make the speechbubble part transparent?
 It's on my plan, I just got busy last week, but I'll do this, sure

PR #407, opened 2023-01-13
Use GUI installer for pnpm inject/uninject

Vendicated:
This replaces the node inject scripts with a small wrapper that downloads the latest Installer and runs it with the current working directory as Vencord dir

This way, the dev installer is never outdated.

TODO:
- [x] The installer will currently force update Vencord before patching, need to add an environment variable to disable this
- [ ] ~~Need to add a way to run as root~~


abUwUser:
i am worried that this will remove the ability to use the local version of Vencord for local development. is that the case?

also, it is better to include switches to Vencord's GUI installer so automation tasks can be perfomed

Captain8771:
> i am worried that this will remove the ability to use the local version of Vencord for local development. is that the case?

I mean, if it is the case, you can always manually inject it. probably a pain, but possible.

MeguminSama:
> i am worried that this will remove the ability to use the local version of Vencord for local development. is that the case?
> 
> also, it is better to include switches to Vencord's GUI installer so automation tasks can be perfomed

it does not. it runs the installer with the `VENCORD_USER_DATA_DIR` envvar pointing to the build folder.

abUwUser:
oh okay. what about the GUI switchers?

Vendicated:
> i am worried that this will remove the ability to use the local version of Vencord for local development. is that the case?

have you read the pr description?
"runs it with the current working directory as Vencord dir"

Vendicated:
the only limitation of this change would be that windows 7 or 32 bit users could not use it anymore. But quite frankly, I do not care

abUwUser:
> have you read the pr description?
> "runs it with the current working directory as Vencord dir" 

oh yeah. i didn't get it well enough at first, that's why i've asked about

abUwUser:
> the only limitation of this change would be that windows 7 or 32 bit users could not use it anymore. But quite frankly, I do not care 

is it a limitation of the UI lib used on the installer?

PR #409, opened 2023-01-13
added customizibility for parameter names

exhq:
WHY THE FUCK DOES RENAMING BRANCH CLOSE PRS THE FUCK

### DIFF
--- a/src/api/Commands/index.ts
+++ b/src/api/Commands/index.ts
@@ -1,6 +1,6 @@
 /*
  * Vencord, a modification for Discord's desktop app
- * Copyright (c) 2022 Vendicated and contributors
+ * Copyright (c) exhq
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -19,7 +19,8 @@
 import { makeCodeblock } from "@utils/misc";
 
 import { sendBotMessage } from "./commandHelpers";
-import { ApplicationCommandInputType, ApplicationCommandOptionType, ApplicationCommandType, Argument, Command, CommandContext, Option } from "./types";
+import { ApplicationCommandInputType, ApplicationCommandOptionType, ApplicationCommandType, Argument, Command, CommandContext, Option } from './types';
+import { OptionType } from '../../utils/types';
 
 export * from "./commandHelpers";
 export * from "./types";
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Confusion 100

exhq:
sorry for the stupid naming.
`OptionalMessageOption` and  `RequiredMessageOption` have a hardcoded name and description of "message"

however, the function i added gives full customizibility.

Vendicated:
this is pointless though, why not just specify the option as an object instead of using this function

exhq:
uh i just thought it'd be nice if we had something like this

romangraef (changes requested):
You probably want to update the copyright header, and maybe add type annotations.

romangraef:
> this is pointless though, why not just specify the option as an object instead of using this function

Dont really disagree, but like its also pointless to grab the Required and Optional from discord code instead of just using objects for that.

Vendicated:
No it's not, by grabbing them from Discord they're localised

src/api/Commands/index.ts
```diff
@@ -36,6 +36,16 @@ const ReqPlaceholder = Symbol("RequiredMessageOption") as any as Option;
  * @see {@link RequiredMessageOption}
  */
 export let OptionalMessageOption: Option = OptPlaceholder;
+
+export const CustomMessageOption = (name, type, description, required) => ({
+    name,
+    displayName: name,
+    type,
```
romangraef, line 43: Isn't CustomMessageOption intended to be text, in which case the type is always 3. Although a better option in my opinion would be just to expose all of the Application Command Option Types and allow end users to specify the type themselves.

romangraef:
Oh, the localization is nice.

src/api/Commands/index.ts
romangraef: Actually, i feel like `CustomMessageOption` is also a wrong name, since this isn't a message option, just a generic Option constructor with some defaults?

Vendicated:
I still don't really get the point of this, why would I want to use
```js
options: [CustomMessageOption("option", "cool", false)]
```
instead of
```js
options: [{
    name: "option",
    description: "cool",
    required: false
}]
```

exhq:
first time i tried to add a parameter it took me a while to find the correct object. you could call this skill issue but i feel like its nice to have a way for beginners to have an easy function to work with

Vendicated:
if they don't find the object (I assume you mean the option types) they wouldn't find this function either, no? Also if you can't figure out how to do something it usually helps a lot to look at other plugins

romangraef:
I think the better option is just to have the application command option type enum exposed in some way (like OptionType.STRING, OptionType.ROLE, etc.). the `Option` type is already enforced im pretty sure.

exhq:
fair enough, i will now disappear

PR #410, opened 2023-01-13
fix(vcDoubleClick): fix functionality

D3SOX:
It broke for stage channels. This makes it work for both types on the latest stable release again.

(dunno why I had to remove `(.*)` but it didn't replace the second spot with it)


Vendicated:
does this fix everything showing as vcs

PR #413, opened 2023-01-14
Reimplement Discord's Switch to fix performance

Vendicated:
Also makes the Reload card always show (with info when there's nothing to reload) to fix the entire UI shifting when you toggle a plugin and fixes the css watcher


Vendicated:
This is basically ready now, please test and let me know if I should change anything

MeguminSama (approved):

PR #416, opened 2023-01-14
fix(showHiddenChannels): remove obsolete icons patch

### LINKED ISSUE #415: [Bug] ShowHiddenChannels: Hidden channels showing as voice channels
Mentioned a number of times in Discord but nobody seems to have picked it up yet.

image

D3SOX:
resolves #415

I first modified it to not change the channel type inside the switch

```ts
{
    // This is where the logic that chooses the icon is, we override it to be locked if it's hidden
    find: ".rulesChannelId))",
    replacement: {
        match: /(\w+)\.locked(.*?)switch\((\w+)\.type\)/g,
        replace: "Vencord.Plugins.plugins.ShowHiddenChannels.isHiddenChannel($3)||$1.locked$2switch($3.type)"
    }
},
```

but then noticed it already looks good without that patch, opinions?

### DIFF
--- a/src/plugins/showHiddenChannels.tsx
+++ b/src/plugins/showHiddenChannels.tsx
@@ -53,14 +53,6 @@ export default definePlugin({
             }
         },
         {
-            // This is where the logic that chooses the icon is, we override it to be a locked voice channel if it's hidden
-            find: ".rulesChannelId))",
-            replacement: {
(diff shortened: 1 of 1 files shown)


Vendicated:
who even changed it to locked vc lol

D3SOX:
Yeah i don't see why that patch was necessary in the first place from todays perspective

D3SOX:
@sneksnake initially pushed this code and may know more .-.

PR #424, opened 2023-01-16
feat(PronounDB): Add local pronoun overrides

romangraef:
I originally planned to add a right click context menu option to change this, but the patches only ever worked in like one thing and I just couldn't be bothered anymore.

### DIFF
--- a/src/plugins/pronoundb/index.ts
+++ b/src/plugins/pronoundb/index.ts
@@ -1,6 +1,6 @@
 /*
  * Vencord, a modification for Discord's desktop app
- * Copyright (c) 2022 Vendicated and contributors
+ * Copyright (c) 2022-2023 Vendicated and contributors
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


Vendicated:
have u looked into context menu commands for this?

Vendicated:
would be way cooler than manual command

romangraef:
> have u looked into context menu commands for this?

Yes:

> I originally planned to add a right click context menu option to change this, but the patches only ever worked in like one location and I just couldn't be bothered anymore.

Vendicated:
context menu commands, not patches.

romangraef:
That seems like it could work, ill check it out!

src/plugins/pronoundb/index.ts
```diff
@@ -81,6 +84,60 @@ export default definePlugin({
...
+                        })),
+                        {
+                            label: "Clear local override",
+                            name: "Clear local override",
+                            value: "clear",
+                        }
+                    ],
+                }
+            ],
+            inputType: ApplicationCommandInputType.BOT,
```
Vendicated, line 120:
```suggestion
            inputType: ApplicationCommandInputType.BOT,
            type: ApplicationCommandType.USER
```

basidally just this and then it would show up in context menu > apps
but I remember trying this once and it not working. The fix might or might be trivial, I haven't bothered looking into it

romangraef: This still does not work in all contexts (for example it does not work in servers), and i could not get the arguments to work to figure out what to change the pronouns to.
image

romangraef:
Also when testing this again, it seems like the datastore seems to crash every now and then with: 

```
Error: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.
    at VencordRenderer:4145:27
    at async getLocalPronounOverride (VencordRenderer:7202:24)
    at async Object.execute (VencordRenderer:7453:39)
```

Am I doing something wrong with creating this data store or should i instead just be using the regular vencord data store with something like `Pronouns_${id}` as key?

Vendicated:
Now that we have a nice API for interacting with context menus, perhaps you might want to revisit this? I'm closing this for now to have slightly tidier PRs but feel free to reopen if you're interested in doing so (or maybe make a new PR, as this one has many conflicts and stuff)

PR #429, opened 2023-01-18
feat(Plugin): LurkGuild

s809:
Allows to preview public guilds from invite


Vendicated:
isn't this already a default feature? I'm also really worried about joining guilds via plugins since it's very easy to get banned

s809:
It's built into discovery, and it's same as searching a server there and clicking on it (except some analytics will be missing obv). So as long as you don't spam it i think it should be good

Vendicated:
Please add all the analytics parameters

src/plugins/lurkGuild.tsx
```diff
@@ -0,0 +1,112 @@
...
+        name: "пшш",
+        id: 559800250924007434n
+    }],
+
+    patches: [{
+        // Grab invite guild
+        find: ".GuildSplash,{",
+        replacement: {
+            match: /(\.GuildSplash,{guild:)([a-z]+)(})/,
+            replace: "$1Vencord.Plugins.plugins.LurkGuild.setContext($2)$3"
```
Vendicated, line 94:
```suggestion
            replace: "$1$self.setContext($2)$3"
```

src/plugins/lurkGuild.tsx
```diff
@@ -0,0 +1,112 @@
...
+        }
+    },{
+        // Render after normal button
+        find: ".GuildSplash,{",
+        replacement: {
+            match: /(Messages\.JOINED_GUILD:[a-zA-Z.]+?\.Messages\.JOIN_GUILD}\))/,
+            replace: "$1,Vencord.Plugins.plugins.LurkGuild.render()"
```
Vendicated, line 101:
```suggestion
            replace: "$1,$self.render()"
```

src/plugins/lurkGuild.tsx
```diff
@@ -0,0 +1,112 @@
...
+        .match(/(function ([a-zA-Z_]+)(?:(?!function ).)+\.GUILD_DISCOVERY,object:.*?\.apply\(this,arguments\)})/s)[2];
+    return Object.values(wreq(id)).find(v =>
+        typeof v === "function" &&
+        v.toString().includes(`return ${funcName}.apply(this,arguments)`)) as typeof lurkGuild;
+});
+
+var context: Guild;
+function LurkGuildButton() {
+    // Outside variable can be changed by render of another invite
+    const [guild] = useState(context);
```
Vendicated, line 55: what's the point of this state? This does nothing

s809: Ok state itself was useless, but invite guild needs to be captured for button to open the correct guild instead of whatever was rendered last

src/plugins/lurkGuild.tsx
```diff
@@ -0,0 +1,113 @@
...
+        replacement: {
+            match: /(\.GuildSplash,{guild:)([a-z]+)(})/,
+            replace: "$1$self.setContext($2)$3"
+        }
+    },{
+        // Render after normal button
+        find: ".GuildSplash,{",
+        replacement: {
+            match: /(Messages\.JOINED_GUILD:[a-zA-Z.]+?\.Messages\.JOIN_GUILD}\))/,
+            replace: "$1,$self.render()"
```
Vendicated, line 102: instead of using the context variable, doesn't this patch have access to the guild object? If so, you should pass it as prop, global variables should be a last resort

PR #432, opened 2023-01-19
feat(plugin): TypingTweaks

zt64:
Adds avatars and role colors to typing users


src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findByCodeLazy } from "@webpack";
+import { GuildMemberStore, React } from "@webpack/common";
+
+const Avatar = findByCodeLazy(".Positions.TOP,spacing:");
+
+export default definePlugin({
+    name: "TypingTweaks",
+    description: "Adds some tweaks to typing indicators",
```
Vendicated, line 28:
```suggestion
    description: "Show avatars and role colours in the typing indicator",
```

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,82 @@
...
+        for (const child of children) {
+            const i = children.indexOf(child);
+
+            if (child.type !== "strong") continue;
+
+            children[i] = this.TypingUser({
+                ...props,
+                user: users[element++]
+            });
+        }
```
Vendicated, line 59: you can use a .map for this like
```js
return children.map(c => c.type === "strong" ? c : <this.TypingUser {...props} user={users[element++]} />)
```

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+        return children;
+    },
+
+    TypingUser({ user, guildId }) {
```
Vendicated, line 64:
```suggestion
    TypingUser: ErrorBoundary.wrap(({ user, guildId }) => {
```

(You of course also need to close the function call at the end of the function and should also pass { noop: true } as option to .wrap()

ErrorBoundaries are component wrappers that will catch errors thrown by the wrapped component (instead of crashing) and optionally render fallback ui (by default a fat error card so thats why noop, to render nothing instead)
Just good practice to prevent crashes if your plugin ever breaks

Vendicated:
other than that epic, I'm surprised it's this simple

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export default definePlugin({
+    name: "TypingTweaks",
+    description: "Adds some tweaks to typing indicators",
+    authors: [Devs.zt],
+    patches: [
+        {
+            find: "getCooldownTextStyle",
+            replacement: {
+                match: /=(.{1,2})\[2];(.+)"aria-atomic":!0,children:(.{1,2})}\)/,
```
Vendicated, line 34:
```suggestion
                match: /=(\i)\[2];(.+)"aria-atomic":!0,children:(\i)}\)/,
```
you can use \i to match identifiers instead of .{1,2}, recently added patcher feature

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,82 @@
...
+                display: "grid",
+                gridAutoFlow: "column",
+                gap: "4px",
+                color: GuildMemberStore.getMember(guildId, user.id)?.colorString
+            }
+        }>
+            <div style={{ marginTop: "4px" }}>
+                <Avatar
+                    size={Avatar.Sizes.SIZE_16}
+                    src={user.getAvatarURL(guildId, 16)}
```
Vendicated, line 76: SO SMALL WHAT

i would do 256 so it's high res and downscaled

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,107 @@
...
+    },
+    showRoleColors: {
+        type: OptionType.BOOLEAN,
+        default: true,
+        description: "Show role colors in the typing indicator"
+    },
+    alternativeFormatting: {
+        type: OptionType.BOOLEAN,
+        default: true,
+        description: "Use alternative formatting for when several users are typing"
```
Vendicated, line 42:
```suggestion
        description: "Show a more useful message when several users are typing"
```

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,107 @@
...
+        },
+        {
+            find: "getCooldownTextStyle",
+            replacement: {
+                match: /(\i)\.length\?.\..\.Messages\.THREE_USERS_TYPING.format\(\{a:(\i),b:(\i),c:.}\).+?SEVERAL_USERS_TYPING/,
+                replace: "$&.format({a:$2,b:$3,c:$1.length})"
+            },
+            predicate: () => settings.store.alternativeFormatting
+        }
+    ],
```
Vendicated, line 81: could you please add comments explaining what each individual patch does?

Vendicated (changes requested):

src/plugins/typingTweaks.tsx
zt64: Guhhh fine

src/plugins/typingTweaks.tsx
```diff
@@ -0,0 +1,111 @@
...
+        {
+            find: "getCooldownTextStyle",
+            replacement: {
+                match: /return \i\.Z\.getName\(.,.\.props\.channel\.id,(.)\)/,
+                replace: "return $1"
+            }
+        },
+        // Changes indicator to format message with the typing users
+        {
+            find: "t(n,\"SEVERAL_USERS_TYPING\"",
```
Vendicated, line 69:
```suggestion
            find: ',"SEVERAL_USERS_TYPING","',
```

Vendicated:
good

PR #444, opened 2023-01-21
feat(Plugin): tiktxk

My-Name-Is-Jeff:
This plugin aims to improve embeds for TikTok by using [Britmoji/tiktxk](https://github.com/Britmoji/tiktxk), a CloudFlare worker which creates embeds that actually contain the video.

Works on the following URL formats:

- https://www.tiktok.com/@{username}/video/{videoId}
- https://www.tiktok.com/t/{shareId}
- https://vm.tiktok.com/{shareId}/

### DIFF
--- a/src/plugins/tiktxk.ts
+++ b/src/plugins/tiktxk.ts
@@ -1,0 +1,48 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


Vendicated:
Thanks for the effort, but I'm not accepting any more text replace plugins! We already have too many and there's way too many use cases for wanting to replace text like this. Instead, someone should make a generic text replace plugin where users can manually add rules

My-Name-Is-Jeff:
> Thanks for the effort, but I'm not accepting any more specific text replace plugins! We already have too many and there's way too many use cases for wanting to replace text like this. Instead, someone should make a generic text replace plugin where users can manually add rules

Would you be open to merging this plugin with the existing FxTwitter plugin by renaming the plugin and adding each site as its own boolean setting? That way a user wouldn't need to create/find their own patterns.

Vendicated:
That doesn't fix the issue at hand: opiniated repetitive plugins

A generic text replace plugin would let users input any amount of matchers and their replacements. We could then have a channel in the Discord where people can post their rules for all sorts of things and people could just add the ones they want

abUwUser:
i guess a separated git repo would be better to handle it instead of regex snippets on a discord channel. the plugin would fetch a list of regex matches from that repo, and then the user could choose which ones they want. the reason as this is better is because bad regex expressions could be sent by users, and without a good way to properly review them, maybe it could bring down user experience. other than that, the plugin could sort that list by popularity and by release, things not possible in a discord channel, plus it would stay the server clean c;

note that the option to make your own regex snippets could stay as is.

Vendicated:
that's overly complicated and I don't think that would be relevant

abUwUser:
that makes sense

PR #447, opened 2023-01-22
Ignore dotfiles in plugin dirs

hunt-g:
Skip dot files in root of plugin dirs.

Which previously would cause build to fail if say, `.git/` existed in `src/userplugins/`.


Vendicated:
cool thanks!

PR #456, opened 2023-01-26
feat: make message pre send/edit listeners asynchronous

hunt-g:
Makes listeners async, so pre event listeners can wait for async operations to complete before sending messages.

### DIFF
--- a/src/plugins/reviewDB/components/ReviewsView.tsx
+++ b/src/plugins/reviewDB/components/ReviewsView.tsx
@@ -32,6 +32,7 @@ export default function ReviewsView({ userId }: { userId: string; }) {
         fallbackValue: [],
         deps: [refetchCount],
     });
+    const username = UserStore.getUser(userId)?.username ?? "";
 
     const dirtyRefetch = () => setRefetchCount(refetchCount + 1);
 
@@ -79,7 +80,7 @@ export default function ReviewsView({ userId }: { userId: string; }) {
             <textarea
                 className={classes(Classes.textarea.replace("textarea", ""), "enter-comment")}
                 // this produces something like '-_59yqs ...' but since no class exists with that name its fine
-                placeholder={"Review @" + UserStore.getUser(userId)?.username ?? ""}
+                placeholder={reviews?.some(r => r.senderdiscordid === UserStore.getCurrentUser().id) ? `Update review for @${username}` : `Review @${username}`}
                 onKeyDown={onKeyPress}
                 style={{
                     marginTop: "6px",


src/plugins/apiMessageEvents.ts
```diff
@@ -27,11 +27,11 @@ export default definePlugin({
         {
             find: "sendMessage:function",
             replacement: [{
-                match: /(?<=_sendMessage:function\([^)]+\)){/,
-                replace: "{if(Vencord.Api.MessageEvents._handlePreSend(...arguments)){return;};"
+                match: /(_sendMessage:function.*?{)([\w\W]*)(\},startEditMessage:function)/,
```
Vendicated, line 30: Please do not use [\w\W] just use `.`, also always use the non greedy modifier. You also seem to just match to the next function, hardcoding that function's name. this is a bad idea and you should find a better way, what if they change order of their methods?

src/plugins/apiMessageEvents.ts
```diff
@@ -27,11 +27,11 @@ export default definePlugin({
...
+                replace: "$1return Vencord.Api.MessageEvents._handlePreSend(...arguments).then(()=>{$2})$3"
             }, {
-                match: /(?<=\beditMessage:function\([^)]+\)){/,
-                replace: "{Vencord.Api.MessageEvents._handlePreEdit(...arguments);"
+                match: /(editMessage:function.*?{)([\w\W]*)(\},suppressEmbeds:function)/,
```
Vendicated, line 33: same as above

src/api/MessageEvents.ts
```diff
@@ -41,16 +41,16 @@ export interface MessageExtra {
     stickerIds?: string[];
 }
 
-export type SendListener = (channelId: string, messageObj: MessageObject, extra: MessageExtra) => void | { cancel: boolean; };
-export type EditListener = (channelId: string, messageId: string, messageObj: MessageObject) => void;
+export type SendListener = (channelId: string, messageObj: MessageObject, extra: MessageExtra) => Promise<void | { cancel: boolean; }>;
```
Vendicated, line 44:
```suggestion
export type SendListener = (channelId: string, messageObj: MessageObject, extra: MessageExtra) => Promiseable<void | { cancel: boolean; }>;
```

Just make the promise aspect optional instead of forcing plugins to make their listeners async

Vendicated:
Hi! Are you still working on this?

hunt-g:
Yes, I have just been busy recently.

hunt-g:
Done 👍🏼

PR #459, opened 2023-01-27
feat(ShowHiddenChannels): fix channel switch keybinds jumping to hiddens

dzshn:
adds two patches so <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>Up</kbd>/<kbd>Down</kbd> and <kbd>Alt</kbd>+<kbd>Up</kbd>/<kbd>Down</kbd> don't teleport you to the void accidentally

### DIFF
--- a/src/plugins/showHiddenChannels.tsx
+++ b/src/plugins/showHiddenChannels.tsx
@@ -183,15 +183,15 @@ export default definePlugin({
         {
             find: '"alt+shift+down"',
             replacement: {
-                match: /\(\(0,\i\.\i\)\((?<channel>\i)\.type\)\?/,
-                replace: "!$self.isHiddenChannel($<channel>)&&$&"
+                match: /(?<=getChannel\(\i\);return null!=(?<channel>\i))(?=.{1,130}hasRelevantUnread\(\i\))/,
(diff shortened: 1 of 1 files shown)


src/plugins/showHiddenChannels.tsx
```diff
@@ -178,7 +178,22 @@ export default definePlugin({
...
                 replace: "&&!$self.isHiddenChannel($<channel>)"
             }
-        }
+        },
+        // Patch keybind handlers so you can't accidentally jump to hidden channels
+        {
+            find: '"alt+shift+down"',
+            replacement: {
+                match: /\(\(0,\i\.\i\)\((?<channel>\i)\.type\)\?/,
+                replace: "!$self.isHiddenChannel($<channel>)&&$&"
```
Nuckyz, line 187:
```suggestion
                match: /(?<=getChannel\(\i\);return null!=(?<channel>\i))(?=.{1,130}hasRelevantUnread\(\i\))/,
                replace: "&&!$self.isHiddenChannel($<channel>)"
```

Nuckyz (changes requested):
Makes the patches a bit harder to fail or patch the wrong thing

Nuckyz (approved):

PR #460, opened 2023-01-28
ShowHiddenChannels: New screen for showing hidden channels

Nuckyz:
image
image


src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+}));
+
+const ForumLayoutTypesToNames = proxyLazy(() => ({
+    [ForumLayoutTypes.Default]: "Not set",
+    [ForumLayoutTypes.LIST]: "List view",
+    [ForumLayoutTypes.GRID]: "Gallery view"
+}));
+
+// Icon from the modal when clicking a message link you don't have access to view
+const HiddenChannelLogo = "/assets/433e3ec4319a9d11b0cbe39342614982.svg";
```
Vendicated, line 56: is hardcoding this safe?

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+function HiddenChannelLockScreen(channel: any) {
```
Vendicated, line 58: why not use the channel type from discord-types?

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+                            )}
+                        </Tooltip>
+                    }
+                </div>
+
+                <Text variant="text-lg/normal">You can not see the {channel.isForumChannel() ? "posts" : "messages"} of this channel. {channel.isForumChannel() && channel.topic?.length > 0 && "However you may see its guidelines:"}</Text >
+
+                {channel.isForumChannel() && channel.topic?.length > 0 && <div className="shc-lock-screen-topic-container">
+                    {Parser.parseTopic(channel.topic, false, { channelId: channel.id })}
+                </div>}
```
Vendicated, line 90: why do you only show the topic of forum channels and not regular channels?

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
Nuckyz: I would have to the match the url from the module which isn't exported and is lazy loaded

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
Nuckyz: misses some types related to forums

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
Nuckyz: Regular channels already show the topic on the header, I'm not removing that component. However forum channels do not show it there

Vendicated: I see

src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
Vendicated: could just `Channel & { myChannelExtension?(): void }`

Nuckyz: sure I will do it later

Nuckyz: Done, also fixed a typo I found in the process lol

Nuckyz:
Aight this is for sure ready now, there is nothing else to add

src/plugins/showHiddenChannels/util.ts
```diff
@@ -0,0 +1,54 @@
...
+    years: { value: 31536000, unit: "year" },
+    months: { value: 2592000, unit: "month" },
+    days: { value: 86400, unit: "day" },
+    hours: { value: 3600, unit: "hour" },
+    minutes: { value: 60, unit: "minute" },
+    seconds: { value: 1, unit: "second" },
+    milliseconds: { value: 0.001, unit: "millisecond" }
+};
+
+export function humanizeTime(time: number, unit: Units): string {
```
Vendicated, line 36: i don't like this, why not use moment for this?

Vendicated: or Intl https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

Nuckyz: moment seems to follow Discord thresholds and refuses to change its settings (it turns 30 seconds into a few seconds)

And I couldn't find a way to do it with Intl

PR #469, opened 2023-02-02
ShowHiddenChannels: Stage and voice channels support

Nuckyz:


Vendicated:
why in the world did u add milliseconds who cares about that

src/utils/text.ts
```diff
@@ -37,25 +37,60 @@ export const wordsToPascal = (words: string[]) =>
...
     words.map(w => w[0].toUpperCase() + w.slice(1)).join(" ");
 
+const units = ["years", "months", "weeks", "days", "hours", "minutes", "seconds"] as const;
+type Units = typeof units[number];
+
+function getUnitStr(unit: Units, isOne: boolean, short: boolean) {
+    if (short === false) return isOne ? unit.slice(0, -1) : unit;
+
+    if (unit === "months") return "mon";
+    if (unit === "minutes") return "min";
```
Vendicated, line 47: wouldn't this always use the singular? should use plural here if applicable imo. Why not just use the single letter form, i doubt there will ever be a case where it only says month or mins

Nuckyz: done

Vendicated (approved):

PR #471, opened 2023-02-06
feat(plugin): richerCider

cryptofyre:
This plugin addition adds additional capabilities to our music app Cider by allowing it to use the "Listening" activity type, more will likely be added down the road as Discord matures or we deem more RPC features suitable.

This does not interfere with other plugins/apps outside of potentially Cider Rich Presence when combined with other global modifiers.

[Cider](https://cider.sh)
[GitHub](https://github.com/ciderapp/Cider)

It was fun collaborating with you all in the Discord server on this and I hope to contribute more in the future!


src/plugins/richerCider.ts
```diff
@@ -0,0 +1,50 @@
...
+const appIds = [
+    "911790844204437504",
+    "886578863147192350",
+    "1020414178047041627",
+    "1032800329332445255"
+]
+
+export default definePlugin({
+    name: "richerCider", // why must you make me run in circles for 2 hours.
+    description: "Enhances Cider by adding the \"Listening to\" to the user's rich presence when applicable ID is found.",
```
Vendicated, line 30: maybe explain more thoroughly what cider even is because I assume most people don't know

cryptofyre: Yeah fair, ill update the description.

cryptofyre: Should be good now, take a look!

d3rpp:
seems legit

somerandomcloud:
y'all are impossible to please wtf

Vendicated (changes requested):
pls fix lint issues, otherwise looks fine

Vendicated (approved):

cryptofyre:
Should be nice and tidy now, thanks for the pointers!

PR #476, opened 2023-02-09
make proxies enumerable

Vap0r1ze:
so you can do the following things
```ts
const x = findLazy(...)
Object.keys(x)
Object.entries(x)
// maybe others
```

### DIFF
--- a/src/utils/proxyLazy.ts
+++ b/src/utils/proxyLazy.ts
@@ -22,6 +22,9 @@ const unconfigurable = ["arguments", "caller", "prototype"];
 
 const handler: ProxyHandler<any> = {};
 
+const GET_KEY = Symbol.for("vencord.lazy.get");
+const CACHED_KEY = Symbol.for("vencord.lazy.cached");
+
 for (const method of [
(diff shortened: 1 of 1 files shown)


src/utils/proxyLazy.ts
```diff
@@ -54,7 +57,12 @@ handler.getOwnPropertyDescriptor = (target, p) => {
     if (typeof p === "string" && unconfigurable.includes(p))
         return Reflect.getOwnPropertyDescriptor(target, p);
 
-    return Reflect.getOwnPropertyDescriptor(target.get(), p);
+    if (p !== GET_KEY && p !== CACHED_KEY) return;
```
Vendicated, line 60:
```suggestion
    if (p === GET_KEY || p === CACHED_KEY) throw "NO";
```

is this what u meant? Otherwise it makes no sense to me

Vap0r1ze: it was to prevent it from breaking is someone defined those symbol props on a module, but i make the keys unique symbols to make it not possible

PR #477, opened 2023-02-10
feat(plugins): Permissions Viewer

Nuckyz:
image
image
image


src/plugins/permissionsViewer/components/RolesAndUsersPermissions.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+export interface RoleOrUserPermission {
+    type: PermissionType;
+    id: string;
+    permissions?: bigint;
+    overwriteAllow?: bigint;
+    overwriteDeny?: bigint;
+}
+
+const Permissions: Record<string, bigint> = findLazy(m => typeof m.ADMINISTRATOR === "bigint");
```
Vendicated, line 42: pretty sure this is in multiple plugins now (?), if so please move to commons

src/plugins/permissionsViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+});
+
+function MenuItem(guildId: string, id?: string, type?: MenuItemParentType) {
+    const guild = GuildStore.getGuild(guildId);
+
+    const permissions = [] as RoleOrUserPermission[];
+    let header: string;
+
+    switch (type) {
```
Vendicated, line 64: why is this done here? This item will be rendered many times when the user might not even use it. Move this logic into the onClick handler to avoid unnecessary work

src/plugins/permissionsViewer/components/RolesAndUsersPermissions.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    useStateFromStores(
+        [GuildMemberStore],
+        () => GuildMemberStore.getMemberIds(guild.id),
+        null,
+        (old, current) => old.length === current.length
+    );
+
+    const usersToRequest = permissions.filter(permission => permission.type === PermissionType.User)
+        .map(({ id }) => id)
```
Vendicated, line 56: filter.map.filter? why not merge the second filter in the first?`

Also this would better be inside the effect

src/plugins/permissionsViewer/components/RolesAndUsersPermissions.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <ModalRoot
+            {...modalProps}
+            size={ModalSize.MEDIUM}
+        >
+            <ModalHeader>
+                <Text className="permviewer-perms-title" variant="heading-lg/semibold">{header} permissions:</Text>
+                <ModalCloseButton onClick={modalProps.onClose} />
+            </ModalHeader>
+            <ModalContent>
+                {selectedItem === undefined && (
```
Vendicated, line 80:
```suggestion
                {!selectedItem && (
```

src/plugins/permissionsViewer/components/RolesAndUsersPermissions.tsx
```diff
@@ -0,0 +1,229 @@
...
+                    <div className="permviewer-perms-no-perms">
+                        <Text variant="heading-lg/normal">No permissions to display!</Text>
+                    </div>
+                )}
+
+                {selectedItem !== undefined && (
```
Vendicated, line 86:
```suggestion
                {selectedItem && (
```

src/plugins/permissionsViewer/components/RolesAndUsersPermissions.tsx
```diff
@@ -0,0 +1,229 @@
...
+                                                        if ((overwriteDeny & bit) > 0n) {
+                                                            permissionState = false;
+                                                            break outer;
+                                                        }
+                                                    }
+                                                    permissionState = null;
+                                                }
+                                            }
+
+                                            switch (permissionState) {
```
Vendicated, line 150: - what is the purpose of these svgs? it's hard to tell what each svg represents. It would be helpful to make these svgs separate components so it's more readable and not such a massive function
- you should probably either add a title explaining what the svg is or add aria-hidden in case it's not of relevance

Vendicated (approved):

PR #273, opened 2022-11-30
feat(plugin): RoleColorEverywhere (RCE)

12944qwerty:
currently only user mentions and typing colored

todo:
 - make typing users work with i18n
 - vc users
 - slate mentions
 - statuses
 - popouts

PR #482, opened 2023-02-11
feat(plugin): RoleColorEverywhere

lewisakura:
supersedes #273

colors chat mentions (including in slate), member list role titles, and voice users. typing colors are handled by TypingTweaks so it's not included here to prevent duplicate functionality.

### DIFF
--- a/src/plugins/roleColorEverywhere.tsx
+++ b/src/plugins/roleColorEverywhere.tsx
@@ -40,7 +40,7 @@ const settings = definePluginSettings({
 
 export default definePlugin({
     name: "RoleColorEverywhere",
-    authors: [Devs.KingFish],
+    authors: [Devs.KingFish, Devs.lewisakura],
     description: "Adds the top role color anywhere possible",
     patches: [
         // Chat Mentions
--- a/src/plugins/roleColorEverywhere.tsx
+++ b/src/plugins/roleColorEverywhere.tsx
@@ -58,10 +58,6 @@ export default definePlugin({
                     match: /(function\((.)\).+?roleIcon.{5,20}null,).," \u2014 ",.\]/,
                     replace: "$1$self.roleGroupColor(e)]"
                 },
-                {
-                    match: /n\.isShown;/,
-                    replace: "$&$self.testing(e,n,f,h,c,s);"
-                }
             ],
             predicate: () => settings.store.memberList,
         },
--- a/src/plugins/roleColorEverywhere.tsx
+++ b/src/plugins/roleColorEverywhere.tsx
@@ -55,7 +55,7 @@ export default definePlugin({
             find: ".memberGroupsPlaceholder",
             replacement: [
                 {
-                    match: /(memo\(\(function\((.)\).{0,500}CHANNEL_MEMBERS_A11Y_LABEL.+?roleIcon.{5,20}null,).," \u2014 ",.\]/,
+                    match: /(memo\(\(function\((.)\).{300,500}CHANNEL_MEMBERS_A11Y_LABEL.{100,200}roleIcon.{5,20}null,).," \u2014 ",.\]/,
                     replace: "$1$self.roleGroupColor($2)]"
                 },
             ],
--- a/src/plugins/roleColorEverywhere.tsx
+++ b/src/plugins/roleColorEverywhere.tsx
@@ -70,7 +70,7 @@ export default definePlugin({
             find: ".memberGroupsPlaceholder",
             replacement: [
                 {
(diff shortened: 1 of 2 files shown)


src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,94 @@
...
+    name: "RoleColorEverywhere",
+    authors: [Devs.KingFish, Devs.lewisakura],
+    description: "Adds the top role color anywhere possible",
+    patches: [
+        // Chat Mentions
+        {
+            find: 'className:"mention"',
+            replacement: [
+                {
+                    match: /user:(.),channelId:(.).{0,300}?"@".concat\(.+?\)/,
```
Vendicated, line 47:
```suggestion
                    match: /user:(\i),channelId:(\i).{0,300}?"@"\.concat\(.+?\)/,
```

src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,94 @@
...
+                }
+            ],
+            predicate: () => settings.store.chatMentions,
+        },
+        // Member List Role Names
+        {
+            find: ".memberGroupsPlaceholder",
+            replacement: [
+                {
+                    match: /(function\((.)\).+?roleIcon.{5,20}null,).," \u2014 ",.\]/,
```
Vendicated, line 58: this is extremely dangerous. This will match everything starting from the first instance of (function(a) up to roleIcon, which could be thousands of characters! Please rewrite this in a different way or alternatively change the .+? to .{0,100} or whatever is appropriate

src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,94 @@
...
+                    replace: "$1$self.roleGroupColor(e)]"
+                },
+            ],
+            predicate: () => settings.store.memberList,
+        },
+    ],
+    settings,
+
+    getColor(userId, channelId) {
+        const channel = Vencord.Webpack.Common.ChannelStore.getChannel(channelId);
```
Vendicated, line 68:
```suggestion
        const channel = ChannelStore.getChannel(channelId);
```
Just import it!!

Vendicated (changes requested):

src/plugins/roleColorEverywhere.tsx
lewisakura: this is how it was in the original patch too which is scary

lewisakura:
@Vendicated im happy that im not going to add anything or change this now so i think it's ready for a final review

src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,142 @@
...
+        if (!guildId) {
+            const channel = ChannelStore.getChannel(channelId!);
+            if (!channel) {
+                return null;
+            }
+
+            guildId = channel.guild_id;
+        }
+
+        logger.info(guildId);
```
Vendicated, line 113:
```suggestion
```

:shipit:

lewisakura: oops

src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    replace: "$&...$self.getVoiceProps(this.props),"
+                }
+            ],
+            predicate: () => settings.store.voiceUsers,
+        }
+    ],
+    settings,
+
+    getColor(userId, { channelId, guildId }: { channelId?: string; guildId?: string; }) {
+        if (!guildId && !channelId) return null;
```
Vendicated, line 102: u can do all the below via
```suggestion
        if (!(guildId ??= ChannelStore.getChannel(channelId!)?.guild_id)) return null;
```
(there's no need to check channelId, getChannel handles null case)

src/plugins/roleColorEverywhere.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+            guildId = channel.guild_id;
+        }
+
+        const member = GuildMemberStore.getMember(guildId, userId);
+        if (!member) {
+            return null;
+        }
+
+        return member?.colorString;
```
Vendicated, line 118:
```suggestion
        return GuildMemberStore.getMember(guildId, userId)?.colorString ?? null;
```

Vendicated (changes requested):
Also could you please type the function params? Helps to catch silly bugs like typos / partial refactors later down the road

Vendicated (approved):
okay looks good, thank you!

Vendicated:
:shipit:

PR #483, opened 2023-02-11
Update LastFM plugin

RuiNtD:
Reformat the results
Simplify some code
Add some sanity checks


src/plugins/lastfm.tsx
```diff
@@ -30,6 +30,12 @@ interface ActivityAssets {
     small_text?: string;
 }
 
+
+type ActivityButton = {
```
Vendicated, line 34:
```suggestion
interface ActivityButton {
```

src/plugins/lastfm.tsx
```diff
@@ -135,74 +146,91 @@ export default definePlugin({
...
+        if (!this.settings.username || !this.settings.apiKey) return;
+
+        try {
+            const params = new URLSearchParams({
+                method: "user.getrecenttracks",
+                api_key: this.settings.apiKey,
+                user: this.settings.username,
+                limit: "1",
+                format: "json"
+            }).toString();
```
Vendicated, line 159:
```suggestion
            });
```

src/plugins/lastfm.tsx
```diff
@@ -135,74 +146,91 @@ export default definePlugin({
...
+            if (!trackData) return;
+            if (!trackData["@attr"]?.nowplaying) return;
+
+            // why does the json api have xml structure
+            return {
+                name: trackData.name || "Unknown",
+                album: trackData.album["#text"],
+                artist: trackData.artist["#text"] || "Unknown",
+                url: trackData.url,
+                imageUrl: (trackData.image || []).filter(x => x.size === "large")[0]?.["#text"]
```
Vendicated, line 173:
```suggestion
                imageUrl: trackData.image?.find(x => x.size === "large")?.["#text"]
```

src/plugins/lastfm.tsx
```diff
@@ -135,74 +146,91 @@ export default definePlugin({
...
+            };
+        } catch (e) {
+            console.log("Failed to query Last.fm API", e);
```
Vendicated, line 176:
```suggestion
            new Logger("LastFMRichPresence").error("Failed to query Last.fm API", e);
```

src/plugins/lastfm.tsx
```diff
@@ -135,74 +146,91 @@ export default definePlugin({
...
-        }
+        const assets: ActivityAssets = {
+            large_image: await getApplicationAsset(trackData.imageUrl || "lastfm-large"),
+            large_text: trackData.album,
+            small_image: await getApplicationAsset("lastfm-small"),
+            small_text: "Last.fm",
+        };
 
-        const hideAlbumName = !trackData.album || trackData.album === trackData.name;
+        let buttons: ActivityButton[] = [
```
Vendicated, line 206:
```suggestion
        const buttons: ActivityButton[] = [
```

src/plugins/lastfm.tsx
```diff
@@ -135,74 +146,91 @@ export default definePlugin({
...
-        } else {
-            assets = {
-                large_image: await getApplicationAsset("lastfm-large"),
-                large_text: "Last.fm",
-            };
-        }
+        if (this.settings.shareUsername)
+            buttons.push({
+                label: "Last.fm Profile",
+                url: `https://www.last.fm/user/${this.settings.username}`,
```
Vendicated, line 216: you should probably urlencode this

Vendicated (changes requested):

src/plugins/lastfm.tsx
RuiNtD: The code won't get this far unless the API returns something (which assumes the username is already valid).

src/plugins/lastfm.tsx
RuiNtD: This requires an additional import. I'll add that in a separate commit.

RuiNtD:
Feel free to merge the commits btw :P

Vendicated:
@dzshn look

RuiNtD:
My changes also recommending replacing the `lastfm-large` asset in the Discord application with either [Last.fm's placeholder song icon](https://lastfm.freetls.fastly.net/i/u/200x200/2a96cbd8b46e442fc41c2b86b821562f.png) (which is what the API incorrectly provides if the album has no art) or [Last.fm's placeholder album icon](https://lastfm.freetls.fastly.net/i/u/300x300/c6f59c1e5e7240a4c0d427abd71f3dbb.jpg) for consistency, because TBH [this](https://user-images.githubusercontent.com/1149870/219843897-2944753c-b2ec-4c2f-9d9a-80374e4bd96c.png) looks ugly AF. Am I trying to advertise Last.fm or something? I just want people to know what I'm listening to. lol

RuiNtD:
image
image

I literally could've just added a "Hide album art if empty" toggle. Apparently nobody felt the desire to work with me to get it done.

RuiNtD:
You brought it up that *one* time, and all you said was to add an option for custom application ID, which would do nothing to solve the problem. I was willing to collaborate and all I got was "my way is the only way" (paraphrase). You can even see a 😴 reaction in the screenshot you sent. I even made a thread in the Discord that nobody ever talked in.

src/plugins/lastfm.tsx
```diff
@@ -79,14 +89,68 @@ async function getApplicationAsset(key: string): Promise<string> {
...
+        description: "text shown in status",
+        type: OptionType.STRING,
+        default: "some music",
+    },
+    useListeningStatus: {
+        description: 'show "Listening to" status instead of "Playing"',
+        type: OptionType.BOOLEAN,
+        default: false,
+    },
+    useLogo: {
```
Vendicated, line 129: i think the way this setting is worded is confusing, why not do something like this instead?

## Fallback Logo
- Use last.fm logo when album is unknown or has no album art
- Use last.fm logo when album is unknown, use last.fm placeholder logo when album has no album art
- Use last.fm placeholder logo when album is unknown or has no album art

src/plugins/lastfm.tsx
```diff
@@ -136,73 +179,109 @@ export default definePlugin({
...
+            const res = await fetch(`https://ws.audioscrobbler.com/2.0/?${params}`);
+            let json: any | undefined;
+            try {
+                json = await res.json();
+                if (json.error) {
+                    logger.error("Error from Last.fm API", `${json.error}: ${json.message}`);
+                    return null;
+                }
+            } catch {
+                throw new Error(`${res.status} ${res.statusText}`);
```
Vendicated, line 203: this code is weird, I would do something like this instead
```js
try {
	const res = await fetch(`https://ws.audioscrobbler.com/2.0/?${params}`);
	if (!res.ok) throw `${res.status} ${res.statusText}`;

	var json = await res.json();
	if (json.error) throw `${json.error}: ${json.message}`;
} catch (err) {
	logger.error("Error fetching last.fm data", err);
	return null;
}
const trackData = json.recenttracks?.track[0];
...
```

src/plugins/lastfm.tsx
```diff
@@ -136,73 +179,109 @@ export default definePlugin({
...
+                label: "View Song",
+                url: trackData.url,
+            },
+        ];
+
+        if (settings.store.shareUsername)
+            buttons.push({
+                label: "Last.fm Profile",
+                url: `https://www.last.fm/user/${settings.store.username}`,
+            });
```
Vendicated, line 268: instead of conditionally pushing, you could
```js
const buttons: ActivityButton[] = [
    {
        label: "View Song",
        url: trackData.url,
    },
	settings.store.shareUsername && {
        label: "Last.fm Profile",
        url: `https://www.last.fm/user/${settings.store.username}`,
    }
].filter(isTruthy);
```

src/plugins/lastfm.tsx
Vendicated: I just realised you're nesting try catches here. Why? One is enough

RuiNtD: The inner try catch is used just to properly format the error message provided to the outer catch, so it would say something like "Failed to query Last.fm API 500 Internal Server Error" instead of returning a JSON error, but yeah, I agree with the above code. Will commit that in a bit.

src/plugins/lastfm.tsx
RuiNtD: ...I mean, I *could*, but this code is more readable. Besides, it's only one conditional. Maybe if there were more conditional buttons, it would make more sense, but I don't think rich presence even allows more than two buttons?

src/plugins/lastfm.tsx
RuiNtD: Seems a bit too wordy, tho. It might make more sense to remove the middle option, for something like this:

## When album is unknown or album art is missing
- Prefer generic placeholder
- Prefer Last.fm logo

Feedback? Could also open up potential in the future for more options; like trying to look up album art on another service (like Google Images) or [Identicons](http://identicon.net/) (unlikely, but possible).

src/plugins/lastfm.tsx
```diff
@@ -136,73 +175,112 @@ export default definePlugin({
...
-        return {
-            name: trackData.name || "Unknown",
-            album: trackData.album["#text"],
-            artist: trackData.artist["#text"] || "Unknown",
-            url: trackData.url,
-            imageUrl: (trackData.image || []).filter(x => x.size === "large")[0]?.["#text"]
-        };
+            const json = await res.json();
+            if (json.error) {
+                logger.error("Error from Last.fm API", `${json.error}: ${json.message}`);
```
Vendicated, line 195: very small nitpick: you could just make this
```suggestion
                throw `${json.error}: ${json.message}`;
```
to have slightly less duplicated code (the catch already logs and returns null)

Vendicated (approved):
other than the small nitpick LGTM

src/plugins/lastfm.tsx
RuiNtD: This is an error from the API itself, rather than an error parsing the response.

PR #485, opened 2023-02-12
fix client hanging up indefinitely on recursive links in MessageLinkEmbeds

dzshn:
prevents recursion by limiting depth to 16 messages and checking if any parent message is being linked inside it at any point

### DIFF
--- a/src/plugins/messageLinkEmbeds.tsx
+++ b/src/plugins/messageLinkEmbeds.tsx
@@ -37,6 +37,7 @@ import {
 import { Channel, Guild, Message } from "discord-types/general";
 
 let messageCache: { [id: string]: { message?: Message, fetched: boolean; }; } = {};
+const messageLinkHierarchy: { [id: string]: string[] } = {};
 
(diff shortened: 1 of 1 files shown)


dzshn:
uhm this may or may not be causing the client to randomly hang up for a while (but not indefinitely) when switching channels. I'll take a look tomorrow but for now I sleep

ActuallyTheSun:
while this pr is open, might as well fix the guild name showing in both the embed and channel mention
image

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -133,6 +134,19 @@ const computeWidthAndHeight = (width: number, height: number) => {
     return { width: newWidth, height: newHeight };
 };
 
+function checkRecursion(id: string, rootId: string | null = null, depth = 0): boolean {
+    rootId ??= id;
```
Vap0r1ze, line 138: btw u can
```suggestion
function checkRecursion(id: string, rootId = id, depth = 0): boolean {
```

PR #490, opened 2023-02-12
Feat(InvisibleChat): Add Autodecryption

SammCheese:
Self explanatory, pls merg

### DIFF
--- a/src/plugins/invisibleChat/index.tsx
+++ b/src/plugins/invisibleChat/index.tsx
@@ -239,7 +239,7 @@ export function isCorrectPassword(result: string): boolean {
 export async function iteratePasswords(message: Message): Promise<string | false> {
     const passwords = settings.store.savedPasswords.split(",").map(s => s.trim());
 
-    if (!message || !message?.content || !passwords || !passwords.length) return false;
+    if (!message?.content || !passwords?.length) return false;
 
     let { content } = message;
 
--- a/src/plugins/invisibleChat/index.tsx
+++ b/src/plugins/invisibleChat/index.tsx
@@ -17,11 +17,13 @@
 */
 
 import { addButton, removeButton } from "@api/MessagePopover";
+import { definePluginSettings } from "@api/settings";
 import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants";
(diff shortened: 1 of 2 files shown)


src/plugins/invisibleChat/index.tsx
```diff
@@ -213,7 +227,31 @@ export function encrypt(secret: string, password: string, cover: string): string
...
+    const decrypted = steggo.reveal(secret, password);
+    return removeIndicator ? decrypted.replace("\u200b", "") : decrypted;
 }
 
+export function isCorrectPassword(result: string): boolean {
+    return result.endsWith("\u200b");
+}
+
+export async function iteratePasswords(message: Message): Promise<string | false> {
+    const passwords = settings.store.savedPasswords.split(", ");
```
Vendicated, line 240: this is a bit wonky and requires exactly 1 space, why not support any amount of whitespace?
```suggestion
    const passwords = settings.store.savedPasswords.split(",").map(s => s.trim());
```

src/plugins/invisibleChat/index.tsx
```diff
@@ -213,7 +227,31 @@ export function encrypt(secret: string, password: string, cover: string): string
...
+
+    if (!message || !message?.content || !passwords || !passwords.length) return false;
```
Vendicated, line 242:
```suggestion
    if (!message?.content || !passwords?.length) return false;
```

way way cooler

src/plugins/invisibleChat/index.tsx
```diff
@@ -213,7 +227,31 @@ export function encrypt(secret: string, password: string, cover: string): string
...
+
+    let { content } = message;
+
+    // we use an extra variable so we dont have to edit the message content directly
+    if (message.content.match(/^\W/)) content = `d ${message.content}d`;
```
Vendicated, line 247: wait whats going on with these `d`s? Also regex.test is the way to go when you want a boolean!
```suggestion
    if (/^\W/.test(message.content)) content = `d ${message.content}d`;
```

SammCheese: The d's simulate a normal message with 2 words. This is necessary to avoid stegcloak mangling the decrypted message

PR #492, opened 2023-02-13
Fix Who Reacted eventually erroring

Nuckyz:


src/plugins/fakeNitro.ts
```diff
@@ -64,17 +75,26 @@ export default definePlugin({
...
-                "canUseAnimatedEmojis",
-                "canUseEmojisEverywhere"
-            ].map(func => {
-                return {
-                    match: new RegExp(`${func}:function\\(.+?\\{`),
-                    replace: "$&return true;"
-                };
-            })
+                {
+                    match: /(?<=(?<intention>\i)=\i\.intention.+?\.(?:canUseEmojisEverywhere|canUseAnimatedEmojis)\(\i)(?=\))/g,
```
Vendicated, line 82: isn't non fixed size lookbehind bad bad

Nuckyz: It doesn't turn bad bad on that case, patch is super quick

src/plugins/typingTweaks.tsx
```diff
@@ -82,7 +82,7 @@ export default definePlugin({
             <strong key="0">{a}</strong>,
             ", ",
             <strong key="2">{b}</strong>,
-            `, and ${c} others are typing...`
+            `, and ${c} other${c !== 1 ? "s" : ""} are typing...`
```
Vendicated, line 85: this is an impossible case, there's a different message for 3 users

PR #496, opened 2023-02-15
feat: Context Menu API

Nuckyz:
You can see it being used here https://github.com/Nuckyz/Vencord/blob/feat/permissions-viewer-context-menu-api/src/plugins/permissionsViewer/index.tsx

### DIFF
--- a/src/api/ContextMenu.ts
+++ b/src/api/ContextMenu.ts
@@ -98,16 +98,11 @@ export function _patchContextMenu(props: ContextMenuProps) {
     const contextMenuPatches = patches.get(props.navId);
 
     if (contextMenuPatches) {
-        let previousChildren: Array<React.ReactElement> = [...props.children];
-
         for (const patch of contextMenuPatches) {
             try {
                 patch(props.children, props.contextMenuAPIArguments);
-                previousChildren = [...props.children];
             } catch (err) {
                 ContextMenuLogger.error(`Patch for ${props.navId} errored,`, err);
-                ContextMenuLogger.info(`Restoring previous children for ${props.navId}...`, previousChildren);
-                props.children = previousChildren;
             }
         }
     }
--- a/src/api/ContextMenu.ts
+++ b/src/api/ContextMenu.ts
@@ -17,9 +17,18 @@
 */
 
 import Logger from "@utils/Logger";
-import React from "react";
+import type { ReactElement } from "react";
 
+/**
+ * @param children The rendered context menu elements
+ * @param args Any arguments passed into making the context menu, like the guild, channel, user or message for example
+ */
 export type NavContextMenuPatchCallback = (children: Array<React.ReactElement>, args?: Array<any>) => void;
+/**
+ * @param The navId of the context menu being patched
+ * @param children The rendered context menu elements
+ * @param args Any arguments passed into making the context menu, like the guild, channel, user or message for example
+ */
 export type GlobalContextMenuPatchCallback = (navId: string, children: Array<React.ReactElement>, args?: Array<any>) => void;
 
 const ContextMenuLogger = new Logger("ContextMenu");
@@ -62,10 +71,7 @@ export function addGlobalContextMenuPatch(patch: GlobalContextMenuPatchCallback)
 export function removeContextMenuPatch<T extends string | Array<string>>(navId: T, patch: NavContextMenuPatchCallback): T extends string ? boolean : Array<boolean> {
     const navIds = Array.isArray(navId) ? navId : [navId as string];
 
-    const results: Array<boolean> = [];
-    for (const id of navIds) {
-        results.push(navPatches.get(id)?.delete(patch) ?? false);
-    }
+    const results = navIds.map(id => navPatches.get(id)?.delete(patch) ?? false);
 
     return (Array.isArray(navId) ? results : results[0]) as T extends string ? boolean : Array<boolean>;
 }
@@ -104,9 +110,9 @@ export function findGroupChildrenByChildId(id: string, children: Array<React.Rea
 }
 
 interface ContextMenuProps {
-    contextMenuAPIArguments?: Array<any>;
+    contextMenuApiArguments?: Array<any>;
     navId: string;
-    children: Array<React.ReactElement>;
+    children: Array<ReactElement>;
     "aria-label": string;
     onSelect: (() => void) | undefined;
     onClose: (callback: (...args: Array<any>) => any) => void;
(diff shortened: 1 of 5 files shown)


src/api/ContextMenu.ts
```diff
@@ -0,0 +1,114 @@
...
+    contextMenuAPIArguments?: Array<any>;
+    navId: string;
+    children: Array<React.ReactElement>;
+    "aria-label": string;
+    onSelect: (() => void) | undefined;
+    onClose: (callback: (...args: Array<any>) => any) => void;
+}
+
+export function _patchContextMenu(props: ContextMenuProps) {
+    const contextMenuPatches = patches.get(props.navId);
```
Vendicated, line 98: i'm not a fan of doing it per navId, what if someone wants to patch multiple navIds? We could just call the callbacks unconditionally and let the consumer be responsible of checking the navId and or children

src/api/ContextMenu.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    if (contextMenuPatches) {
+        let previousChildren: Array<React.ReactElement> = [...props.children];
```
Vendicated, line 101: I'm not a fan of this spread, not only will it unnecessarily make and destruct a lot of arrays, it also doesn't cover the case of modifying something like children[0].props.children. Also are you sure children is always an array? The name is a bit misleading, in the case of a single child it will be a flat ReactNode. If you really want to do this, wrapping it with a Proxy would be more robust, but I personally don't think it's necessary to even do this in the first place

src/plugins/apiContextMenu.ts
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }]
+            });
+
+            removeListener(listener);
+        }
+    }
+}
+
+addListener(listener);
```
Vendicated, line 51: couldn't this be a waitFor?

src/webpack/webpack.ts
```diff
@@ -112,7 +126,7 @@ export const find = traceFunction("find", function find(filter: FilterFn, getDef
         }
     }
 
-    return null;
+    return isWaitFor ? [found ?? null, foundId ?? null] : found ?? null;
```
Vendicated, line 129: i would prefer inlining this over the break since it likely performs better and the breaks with nested loops are a bti confusing to follow. Also on a different note, all waitFors ( I think ) across Vencord never find anything here because they are ran too early, they just call this function so it also works fine when ran later

Vendicated (changes requested):

src/plugins/apiContextMenu.ts
Nuckyz: makes no difference

Vendicated: the difference is that you then repeat less code cause waitFor handles the looping and stuff for u :P

src/plugins/apiContextMenu.ts
```diff
@@ -28,7 +28,7 @@ function listener(exports: any, id: number) {
         if (typeof prop !== "function") continue;
 
         const str = Function.prototype.toString.call(prop);
-        if (str.includes("♫ ⊂(｡◕‿‿◕｡⊂) ♪")) {
```
Vendicated: NOOOOO I MISS HIM ALREADY

Nuckyz: 😭

Vendicated (approved):

Vendicated:
Looks good but i want to test this more because I'm worried it could break stuff

src/webpack/webpack.ts
Vendicated: (thats not how i meant it )

src/webpack/webpack.ts
```diff
@@ -82,26 +82,40 @@ export const find = traceFunction("find", function find(filter: FilterFn, getDef
...
+        }
 
         // the length check makes search about 20% faster
         for (const nestedMod in mod.exports) if (nestedMod.length <= 3) {
             const nested = mod.exports[nestedMod];
-            if (nested && filter(nested)) return nested;
+            if (nested && filter(nested)) {
+                found = nested;
+                foundId = Number(key);
+                break outer;
```
Vendicated, line 113: i meant inlining it here and in the other places
```js
return isWaitFor ? [nested, Number(key)] : nested;
```

Nuckyz: oh

src/api/ContextMenu.ts
```diff
@@ -0,0 +1,135 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import Logger from "@utils/Logger";
+import React from "react";
```
Vendicated, line 20:
```suggestion
import type { ReactElement } from "react";
```

src/api/ContextMenu.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Remove a context menu patch
+ * @param navId The navId(s) for the context menu(s) to remove the patch
+ * @param patch The patch to be removed
+ * @returns Wheter the patch was sucessfully removed from the context menu(s)
+ */
+export function removeContextMenuPatch<T extends string | Array<string>>(navId: T, patch: NavContextMenuPatchCallback): T extends string ? boolean : Array<boolean> {
+    const navIds = Array.isArray(navId) ? navId : [navId as string];
+
+    const results: Array<boolean> = [];
```
Vendicated, line 65:
```suggestion
    const results = navIds.map(id => navPatches.get(id)?.delete(patch) ?? false);
```

src/api/ContextMenu.ts
```diff
@@ -0,0 +1,135 @@
...
+            const found = findGroupChildrenByChildId(id, nextChildren, nextChildren);
+            if (found !== null) return found;
+        }
+    }
+
+    return null;
+}
+
+interface ContextMenuProps {
+    contextMenuAPIArguments?: Array<any>;
```
Vendicated, line 107:
```suggestion
    contextMenuApiArguments?: Array<any>;
```

src/plugins/apiContextMenu.ts
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }]
+            });
+
+            removeListener(listener);
+        }
+    }
+}
+
+addListener(listener);
```
Vendicated, line 51: this should probably be done either conditionally (only when the plugin is enabled), or the plugin should be marked as required

PR #498, opened 2023-02-15
feat(plugin): MentionCacheFix

My-Name-Is-Jeff:
This plugin fetches uncached mentioned users when hovering over messages.

- [ ] get react to update the message without needing the user to unhover


Vendicated:
this is now available as ValidUser!

PR #505, opened 2023-02-17
feat: Cloud settings sync

lewisakura:
#313

At present, you can authorize the Vencord OAuth application (in app) and sync settings to and from the server, with timestamp validation. Next steps would be to sync on startup if necessary, encryption(?), actually check if settings sync is enabled for certain operations, and then finally prettying up the UI before shipping.

The backend server (https://github.com/Vencord/Backend) is a running project alongside this PR and nothing is final as of writing.


lewisakura:
changes of note:
- this now creates a new settings entry (`settingsSync`) to control whether it's turned on or off, instead of relying on the presence of the secret purely
- cloud settings now has a separate tab to hold its settings, and in the event that we extend the backend more, we can shove all of this under a generic "cloud" tab, or.. "Vencloud" if you will :^)
- uses a lot less IDB entries (in fact, just two - one to store the secret and one to store when settings were last written to)
- it can now synchronize your settings on startup, but currently requires vencord to be restarted after - i am not sure of a better workaround besides putting the settings sync before plugin loading, but i'm not sure of the consequences (@Vendicated you'll have to tell me if this is okay to do or not)
- backend is now go

lewisakura:
whilst not fully finished i will now accept reviews for improvements and changes - just bare in mind some of this may not be final still and therefore may be invalidated at a later date, even if the change is made.

src/Vencord.ts
```diff
@@ -30,17 +30,34 @@ import "./webpack/patchWebpack";
...
 import { onceReady } from "./webpack";
 import { SettingsRouter } from "./webpack/common";
 
 export let Components: any;
 
+async function syncSettings() {
+    if (Settings.backend.settingsSync && await shouldCloudSync()) {
+        await getCloudSettings(false);
+
+        showNotice(
```
Vendicated, line 44: Maybe use a notification here? A notice would be a pretty big nuisance imo

src/Vencord.ts
```diff
@@ -30,17 +30,34 @@ import "./webpack/patchWebpack";
...
+        );
+    }
+}
+
 async function init() {
     await onceReady;
     startAllPlugins();
     Components = await import("./components");
 
+    await syncSettings();
```
Vendicated, line 59: I don't think blocking here is relevant
```suggestion
    syncSettings();
```

src/api/settings.ts
```diff
@@ -136,6 +147,7 @@ function makeProxy(settings: any, root = settings, path = ""): Settings {
             }
             // And don't forget to persist the settings!
             VencordNative.ipc.invoke(IpcEvents.SET_SETTINGS, JSON.stringify(root, null, 4));
+            DataStore.set("Vencord_settingsLastSaved", Date.now());
```
Vendicated, line 150: This might or might not be problematic and cause corruption due to concurrent writes, I think this would be better to store alongside settings

src/components/VencordSettings/index.tsx
```diff
@@ -48,6 +49,7 @@ const SettingsTabs: Record<string, SettingsTab> = {
     VencordThemes: { name: "Themes", component: () => <ThemesTab /> },
     VencordUpdater: { name: "Updater" }, // Only show updater if IS_WEB is false
     VencordSettingsSync: { name: "Backup & Restore", component: () => <BackupRestoreTab /> },
+    VencordCloudSettings: { name: "Cloud Settings", component: () => <CloudSettingsTab /> }
```
Vendicated, line 52: any reason this isn't just in the settings sync tab?

src/utils/cloud.tsx
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import * as DataStore from "@api/DataStore";
+import { findByProps } from "@webpack";
+import { Toasts, UserStore } from "@webpack/common";
+
+import { Settings } from "../Vencord";
```
Vendicated, line 23:
```suggestion
import { Settings } from "@api/Settings";
```

src/utils/cloud.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    openModal((props: any) => <OAuth2AuthorizeModal
+        {...props}
+        scopes={["identify"]}
+        responseType="code"
+        redirectUri="https://vencord.vendicated.dev/api/v1/callback"
+        permissions={0n}
+        clientId="1075583776979169361"
+        cancelCompletesFlow={false}
+        callback={async (u: string) => {
```
Vendicated, line 56:
```suggestion
        callback={async (callbackUrl: string) => {
```

src/utils/cloud.tsx
```diff
@@ -0,0 +1,89 @@
...
+                if (secret) {
+                    cloudLogger.info("Authorized with secret");
+                    await DataStore.set("Vencord_cloudSecret", secret);
+                    toast(Toasts.Type.SUCCESS, "Cloud enabled!");
+                } else {
+                    toast(Toasts.Type.FAILURE, "Setup failed (no secret returned?).");
+                }
+            } catch (e: any) {
+                cloudLogger.error("Failed to authorize", e);
+                toast(Toasts.Type.FAILURE, `Setup failed (${e.toString()}).`);
```
Vendicated, line 76: I'd use Notifications here instead of Toasts as Toasts are awkward for this

src/utils/cloud.tsx
```diff
@@ -0,0 +1,89 @@
...
+        }
+        }
+    />);
+}
+
+export async function getCloudAuth() {
+    const userId = UserStore.getCurrentUser().id;
+    const secret = await DataStore.get("Vencord_cloudSecret");
+
+    return btoa(`${userId}:${secret}`);
```
Vendicated, line 88:
```suggestion
    return window.btoa(`${userId}:${secret}`);
```
Just to avoid the deprecation warning

src/utils/settingsSync.ts
```diff
@@ -16,8 +16,11 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import * as DataStore from "@api/DataStore";
 import { Toasts } from "@webpack/common";
+import * as fflate from "fflate";
```
Vendicated, line 21: Only import the named members you need so unused code can be treeshaked

src/utils/settingsSync.ts
```diff
@@ -121,3 +125,137 @@ export async function uploadSettingsBackup(showToast = true): Promise<void> {
...
+
+        cloudSettingsLogger.info("Settings deleted from cloud successfully");
+        toast(Toasts.Type.SUCCESS, "Deleted your settings from the cloud!");
+    } catch (e: any) {
+        cloudSettingsLogger.error("Failed to delete", e);
+        toast(Toasts.Type.FAILURE, `Deletion failed (${e.toString()}).`);
+    }
+}
+
+export async function shouldCloudSync() {
```
Vendicated, line 238: This screams race condition. In IO you should never first check availablity then execute, since the state might change inbetween those two actions. Instead, you should just execute the actual operation then handle errors (or in this case you should use If-None-Match and check for 304 status code)

Vendicated (changes requested):

src/Vencord.ts
lewisakura: my personal justification for this was that a settings sync should be a big deal since it changes all of the settings and the standard importSettings() requires you to reload the client too, but i can switch to a notif

Vendicated: That may be true for the majority of users, but imagine dual booting and regularely changing ur settings so u get the notice every time u open discord essentially

src/api/settings.ts
lewisakura: it shouldn't be too big of a deal, it just mainly needs to be different to whatever it was before so even a corrupted value is okay, but to be fair i have actually hammered this and it doesn't seem to ever get corrupted. storing alongside settings wouldn't be too bad of an idea though

src/components/VencordSettings/index.tsx
lewisakura: clutter and futureproofing - if we decide to have more backend integrations it'd be best for a separate tab. also the reason why it's now the generic "cloud integrations" when enabling

src/utils/settingsSync.ts
lewisakura: oops, this was meant to be temporary

src/utils/settingsSync.ts
lewisakura: that's a new header i've learnt of today, will implement and change

src/api/settings.ts
Vendicated: yeah I'm not sure how much race conditions are of an issue for idb, i assume they already have locking mechanisms or similar, so this might just be fine. But i also think since this is just a number and makes no sense outside of the context of the settings object it's best stored alongside them

lewisakura:
can someone check if they can repro a strange bug where plugins don't enable when settings are synchronized, even after a full restart? not sure if it's a me thing or just some weird side effect of the way i import settings, but i experienced it earlier and couldn't replicate it after, so i can't give steps 😓

lewisakura:
> can someone check if they can repro a strange bug where plugins don't enable when settings are synchronized, even after a full restart? not sure if it's a me thing or just some weird side effect of the way i import settings, but i experienced it earlier and couldn't replicate it after, so i can't give steps 😓

I haven't been able to replicate this so I assume it's been fixed in some passing Vencord update either by accident or the bug was never a real thing in the first place and I'm just stupid.

package.json
```diff
@@ -65,7 +65,7 @@
         "type-fest": "^3.5.3",
         "typescript": "^4.9.4"
     },
-    "packageManager": "pnpm@7.13.4",
+    "packageManager": "pnpm@7.30.5",
```
Vendicated, line 68: is this just latest @7? Maybe we could just @8

lewisakura: It is. I'll bump it to 8.

src/components/VencordSettings/CloudTab.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    showNotification({
+        title: "Cloud Integrations",
+        body: "Successfully erased all data.",
+        color: "var(--green-360)"
+    });
+}
+
+function SettingsSyncSection() {
+    const settings = useSettings(["cloud.authenticated", "cloud.settingsSync"]);
```
Vendicated, line 67: could destructure
```suggestion
    const { cloud: { authenticated, settingsSync } } = useSettings(["cloud.authenticated", "cloud.settingsSync"]);
```

src/components/VencordSettings/CloudTab.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const sectionEnabled = useMemo(
```
Vendicated, line 68: memo is pointless here
```suggestion
    const sectionEnabled = authenticated && settingsSync;
```

src/utils/settingsSync.ts
```diff
@@ -121,3 +126,169 @@ export async function uploadSettingsBackup(showToast = true): Promise<void> {
...
+}
+
+export async function getCloudSettings(shouldNotify = true, force = false) {
+    try {
+        const res = await fetch(new URL("/v1/settings", getCloudUrl()), {
+            method: "GET",
+            headers: new Headers({
+                Authorization: await getCloudAuth(),
+                Accept: "application/octet-stream",
+                "if-none-match": Settings.cloud.settingsSyncVersion.toString()
```
Vendicated, line 183: soooo inconsistent :(
```suggestion
                "If-None-Match": Settings.cloud.settingsSyncVersion.toString()
```

src/utils/settingsSync.ts
```diff
@@ -121,3 +126,169 @@ export async function uploadSettingsBackup(showToast = true): Promise<void> {
...
+            cloudSettingsLogger.error(`Failed to sync down, API returned ${res.status}`);
+            showNotification({
+                title: "Cloud Settings",
+                body: `Could not synchronize settings (API returned ${res.status}).`,
+                color: "var(--red-360)"
+            });
+            return false;
+        }
+
+        const written = parseInt(res.headers.get("etag")!);
```
Vendicated, line 217:
```suggestion
        const written = Number(res.headers.get("etag")!);
```

Vendicated (changes requested):

Vendicated (approved):

MeguminSama (approved):

PR #508, opened 2023-02-17
feat(plugin): add support for VxTwitter in FxTwitter plugin

TurboCheetah:


Vendicated:
As mentioned in a similar pr, fx is simply better due to being more feature rich so the addition of this is pointless. I'm also not accepting any new text replace plugins since they're too repetitive and opiniated, I will only accept a generic text replace plugin that allows the user to specify custom rules as that will cover all cases

PR #512, opened 2023-02-19
typings for defaultless settings

Vap0r1ze:
this adds real typings for settings definitions that dont have a `default` field:
- `COMPONENT` types are still always `any`
- `SELECT` types are `T | undefined` if there are no `{ default: boolean }` elements
- everything else is `T | undefined` if there is no default field

### DIFF
--- a/src/plugins/customRPC.tsx
+++ b/src/plugins/customRPC.tsx
@@ -19,6 +19,7 @@
 import { definePluginSettings } from "@api/settings";
 import { Link } from "@components/Link";
 import { Devs } from "@utils/constants";
+import { isTruthy } from "@utils/guards";
 import { useAwaiter } from "@utils/misc";
 import definePlugin, { OptionType } from "@utils/types";
 import { filters, findByCodeLazy, findByPropsLazy, mapMangledModuleLazy } from "@webpack";
@@ -173,13 +174,13 @@ async function createActivity(): Promise<Activity | undefined> {
         activity.buttons = [
             buttonOneText,
             buttonTwoText
-        ].filter((s): s is string => !!s);
+        ].filter(isTruthy);
 
         activity.metadata = {
             button_urls: [
                 buttonOneURL,
                 buttonTwoURL
-            ].filter((s): s is string => !!s)
+            ].filter(isTruthy)
         };
     }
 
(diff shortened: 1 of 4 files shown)


Vap0r1ze:
@Captain8771 can you review as well

Captain8771:
(i have no clue what is good practice in typing)

src/plugins/customRPC.tsx
```diff
@@ -93,13 +93,13 @@ const numOpt = (description: string) => ({
...
 
-const choice = (label: string, value: any, _default?: Boolean) => ({
+const choice = (label: string, value: any, _default?: boolean) => ({
     label,
     value,
     default: _default
 }) as const;
 
-const choiceOpt = (description: string, options) => ({
+const choiceOpt = <T,>(description: string, options: T) => ({
```
Vendicated, line 102: what the comma do

src/plugins/customRPC.tsx
```diff
@@ -173,13 +173,13 @@ async function createActivity(): Promise<Activity | undefined> {
         activity.buttons = [
             buttonOneText,
             buttonTwoText
-        ].filter(Boolean);
+        ].filter((s): s is string => !!s);
```
Vendicated, line 176: maybe can make this a util for reusability
```ts
function filterTruthy<T>(item: T): item is Exclude<T, 0 | "" | false | null | undefined> {
	return Boolean(s);
}

// usage
[...].filter(filterTruthy);
```

src/plugins/customRPC.tsx
Vap0r1ze: removes ambiguity in tsx files

src/plugins/customRPC.tsx
Vap0r1ze: REAL

src/plugins/customRPC.tsx
Vendicated: ah, i think just dont use arrow func and u wont need to

Vap0r1ze: true but the return is just an object kinda cringe to

Vap0r1ze:
yop

PR #517, opened 2023-02-21
fix: vcDoubleClick and revealAllSpoilers patch

D3SOX:
~~Currently, that's the only `onClick` in that module. I could add back the `className: "channelMention"` check if so desired.~~
See second commit

### DIFF
--- a/src/plugins/vcDoubleClick.ts
+++ b/src/plugins/vcDoubleClick.ts
@@ -50,8 +50,8 @@ export default definePlugin({
             // channel mentions
             find: ".shouldCloseDefaultModals",
             replacement: {
-                match: /onClick:(\w+),/,
-                replace: "onClick:(_vcEv)=>(_vcEv.detail>=2)&&($1)(_vcEv),"
+                match: /onClick:(\w+),(.{0,30}className:"channelMention",)/,
+                replace: "onClick:(_vcEv)=>(_vcEv.detail>=2)&&($1)(_vcEv),$2"
             }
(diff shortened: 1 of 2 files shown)


Nuckyz:
I am pretty sure the match and replace was correct, the issue was the find

src/plugins/vcDoubleClick.ts
```diff
@@ -48,10 +48,10 @@ export default definePlugin({
         },
         {
             // channel mentions
-            find: ".EMOJI_IN_MESSAGE_HOVER",
+            find: ".shouldCloseDefaultModals",
```
Nuckyz, line 51:
```suggestion
            find: "[[2*",
```

D3SOX:
> I am pretty sure the match and replace was correct, the issue was the find 

Oh didn't put that into consideration I just recreated it from scratch and that is what I've come up with

src/plugins/vcDoubleClick.ts
D3SOX: `[[2*` doesn't work for me. That targets a different module. On Discord `Stable 175517 (1fc3105)`

D3SOX:
@Nuckyz Okay it only works with how it's currently pushed for me. Can you double check and send your Discord version?

Nuckyz (approved):

D3SOX:
Ven merge this pls

Vendicated:
thank you!

PR #530, opened 2023-02-25
Add BetterFolders plugin

Juby210:


scrazzz (approved):

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+            type: OptionType.BOOLEAN,
+            description: "Force a folder to open when switching to a server of that folder",
+            default: false,
+        },
+    },
+
+    start() {
+        const GuildFolderStore = findByProps("getSortedGuilds");
+        const ExpandedFolderStore = findByProps("getExpandedFolders");
+        const { toggleGuildFolderExpand } = findByProps("move", "toggleGuildFolderExpand");
```
Vendicated, line 129: you can just move these to the top and use findByPropsLazy, then you also don't need to assign to this.closeFolders here!

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        });
+    },
+    stop() {
+        if (this.onSwitch) FluxDispatcher.unsubscribe("CHANNEL_SELECT", this.onSwitch);
+        if (this.onToggleFolder) FluxDispatcher.unsubscribe("TOGGLE_GUILD_FOLDER_EXPAND", this.onToggleFolder);
+    },
+
+    FolderSideBar: FolderSideBar,
+    getGuildsTree: function (folders, oldTree) {
+        if (!this.GuildsTree) this.GuildsTree = find(m => m.prototype?.convertToFolder);
```
Vendicated, line 167: same here

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+import FolderSideBar from "./FolderSideBar";
+
+export default definePlugin({
+    name: "BetterFolders",
+    description: "Shows server folders on dedicated sidebar and adds folder related improvements",
+    authors: [Devs.juby],
+    patches: [
+        {
+            find: "(\"guildsnav\")",
```
Vendicated, line 35:
```suggestion
            find: '("guildsnav")',
```

src/plugins/betterFolders/betterFolders.css
```diff
@@ -0,0 +1,16 @@
+.bf-folder-sidebar [class*="wrapper-"] > div:first-of-type {
```
Vendicated, line 1: for consistency and possibly avoiding clashes with other mods, can you please make classes start with .vc-?
```suggestion
.vc-bf-folder-sidebar [class*="wrapper-"] > div:first-of-type {
```

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+        const res = self.Guilds(props);
+        const scrollerProps = res.props.children?.props?.children?.[1]?.props;
+        if (scrollerProps?.children) {
+            const servers = scrollerProps.children.find(c => c?.props?.["aria-label"] === Messages.SERVERS);
+            if (servers) scrollerProps.children = servers;
+        }
+        return res;
+    };
+
+    return self._FolderSideBar = () => {
```
Vendicated, line 43: I don't really follow what's going on here, is this just to create a closure and do the webpack searches lazily? If so, same thing applies, use find(byX)Lazy!!

Vendicated (changes requested):

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+            predicate: () => Settings.plugins.BetterFolders.closeAllHomeButton,
+            replacement: {
+                match: ",onClick:function(){if(!__OVERLAY__){",
+                replace: "$&$self.closeFolders();"
+            }
+        }
+    ],
+
+    options: {
```
Vendicated, line 98: options is deprecated, instead you should use
```js
const settings = definePluginSettings(OptionsObject);

console.log(settings.store.sidebar)

definePlugin({
	...,
	settings
})

```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+import { Settings } from "@api/settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy, findLazy } from "@webpack";
+import { FluxDispatcher } from "@webpack/common";
+
+import FolderSideBar from "./FolderSideBar";
+
+const GuildsTree = findLazy(m => m.prototype?.convertToFolder);
+const GuildFolderStore = findByPropsLazy("getSortedGuilds");
```
Vendicated, line 30:
```suggestion
const GuildFolderStore = findStoreLazy("SortedGuildStore");
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+const ExpandedFolderStore = findByPropsLazy("getExpandedFolders");
```
Vendicated, line 31:
```suggestion
const ExpandedFolderStore = findStoreLazy("ExpandedGuildFolderStore");
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        forceOpen: {
+            type: OptionType.BOOLEAN,
+            description: "Force a folder to open when switching to a server of that folder",
+            default: false,
+        },
+    },
+
+    start() {
+        const getGuildFolderIdx = id => GuildFolderStore.guildFolders.findIndex(e => e.guildIds.indexOf(id) !== -1);
+        const getGuildFolder = id => GuildFolderStore.guildFolders[getGuildFolderIdx(id)];
```
Vendicated, line 133: why not just
```suggestion
        const getGuildFolder = id => GuildFolderStore.guildFolders.find(f => f.guildIds.includes(id));
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        const { closeAllFolders, closeOthers, forceOpen } = Settings.plugins.BetterFolders;
+        if (closeAllFolders || forceOpen) FluxDispatcher.subscribe("CHANNEL_SELECT", this.onSwitch = data => {
```
Vendicated, line 136: I would just subscribe unconditionally here and instead check the settings inside the handler so that people don't need to restart when changing settings

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+    name: "BetterFolders",
+    description: "Shows server folders on dedicated sidebar and adds folder related improvements",
+    authors: [Devs.juby],
+    patches: [
+        {
+            find: '("guildsnav")',
+            predicate: () => Settings.plugins.BetterFolders.sidebar,
+            replacement: [
+                {
+                    match: /(\w{1,3})\(\){return \w{1,3}\(\(0,\i\.jsx\)\("div",{className:\w{1,3}\(\)\.guildSeparator}\)\)}/,
```
Vendicated, line 44:
```suggestion
                    match: /(\i)\(\){return \i\(\(0,\i\.jsx\)\("div",{className:\i\(\)\.guildSeparator}\)\)}/,
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                    replace: "$&$self.Separator=$1;"
+                },
+
+                // BEGIN Folder component patch
+                {
+                    match: /\i=(\i)\.folderIconContent,\i=\i\.id,\i=\i\.name,/,
+                    replace: "$&bfHideServers=$1.bfHideServers,"
+                },
+                {
+                    match: /\w{1,3}\(\(function\(\i,\i,\i\){var \i=\i\.key;return.+\(\i\)},\i\)}\)\)/,
```
Vendicated, line 54:
```suggestion
                    match: /\i\(\(function\(\i,\i,\i\){var \i=\i\.key;return.+\(\i\)},\i\)}\)\)/,
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                    replace: "bfHideServers?null:$&"
+                },
+                // END
+
+                // BEGIN Guilds component patch
+                {
+                    match: /(\i)\.themeOverride,(.{15,25}\(function\(\){var \i=)(\w{1,3}\.\i\.getGuildsTree\(\))/,
```
Vendicated, line 61:
```suggestion
                    match: /(\i)\.themeOverride,(.{15,25}\(function\(\){var \i=)(\i\.\i\.getGuildsTree\(\))/,
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                    replace: "$1.themeOverride,bfPatch=$1.bfGuildFolders,$2bfPatch?$self.getGuildsTree(bfPatch,$3):$3"
+                },
+                {
+                    match: /\((\i)\)({switch\(\i\.type\){case \w{1,3}\.\w{1,3}\.FOLDER)/,
```
Vendicated, line 65:
```suggestion
                    match: /\((\i)\)({switch\(\i\.type\){case \i\.\i\.FOLDER)/,
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                    replace: "($1,bfIndex)$2"
+                },
+                {
+                    match: /return(\(0,\i\.jsx\))(\(\w{1,3},{)(folderNode:\i,setNodeRef:\i\.setNodeRef,draggable:!0,.+},\i\.id\));case/,
```
Vendicated, line 69:
```suggestion
                    match: /return(\(0,\i\.jsx\))(\(\i,{)(folderNode:\i,setNodeRef:\i\.setNodeRef,draggable:!0,.+},\i\.id\));case/,
```

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,173 @@
...
+                },
+
+                // BEGIN Folder component patch
+                {
+                    match: /\i=(\i)\.folderIconContent,\i=\i\.id,\i=\i\.name,/,
+                    replace: "$&bfHideServers=$1.bfHideServers,"
+                },
+                {
+                    match: /\w{1,3}\(\(function\(\i,\i,\i\){var \i=\i\.key;return.+\(\i\)},\i\)}\)\)/,
+                    replace: "bfHideServers?null:$&"
```
Vendicated, line 55: this is really bad because you use a variable declared in a different patch here. If the first patch fails but this one still works, you will now use an undeclared variable which will instantly make this code throw and likely cause a crash

Vendicated (changes requested):

src/plugins/betterFolders/index.ts
Juby210: fixed by using arguments[0] to get props

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+*/
+
+import { Settings } from "@api/settings";
+import { findByPropsLazy, findLazy } from "@webpack";
+import { React, useStateFromStores } from "@webpack/common";
+
+const classes = findByPropsLazy("sidebar", "guilds");
+const i18n = findLazy(m => m.Messages?.["en-US"]);
+
+const Animations = findByPropsLazy("useTransition");
```
Vendicated, line 26: image

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Settings } from "@api/settings";
+import { findByPropsLazy, findLazy } from "@webpack";
+import { React, useStateFromStores } from "@webpack/common";
+
+const classes = findByPropsLazy("sidebar", "guilds");
```
Vendicated, line 23: image

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,165 @@
...
+                },
+
+                // BEGIN Guilds component patch
+                {
+                    match: /(\i)\.themeOverride,(.{15,25}\(function\(\){var \i=)(\i\.\i\.getGuildsTree\(\))/,
+                    replace: "$1.themeOverride,bfPatch=$1.bfGuildFolders,$2bfPatch?$self.getGuildsTree(bfPatch,$3):$3"
+                },
+                {
+                    match: /return(\(0,\i\.jsx\))(\(\i,{)(folderNode:\i,setNodeRef:\i\.setNodeRef,draggable:!0,.+},\i\.id\));case/,
+                    replace: "var folder=$1$2bfHideServers:!bfPatch,$3;return bfPatch&&arguments[1]?[$1($self.Separator,{}),folder]:folder;case"
```
Vendicated, line 94: this relies on a variable from an earlier patch which is a really bad idea. Please either find a better way to do it or at least check that that variable exists first using `typeof bfPatch !== "undefined"`

src/plugins/betterFolders/index.ts
```diff
@@ -0,0 +1,165 @@
...
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy, findLazy, findStoreLazy } from "@webpack";
+import { FluxDispatcher } from "@webpack/common";
+
+import FolderSideBar from "./FolderSideBar";
+
+const GuildsTree = findLazy(m => m.prototype?.convertToFolder);
+const GuildFolderStore = findStoreLazy("SortedGuildStore");
+const ExpandedFolderStore = findStoreLazy("ExpandedGuildFolderStore");
+const FolderUtils = findByPropsLazy("move", "toggleGuildFolderExpand");
```
Vendicated, line 32: image

Vendicated (approved):

PR #534, opened 2023-02-26
Add MoreSearchOperators plugin

Davr1:
This plugin adds/fixes experimental search filters. Some of these are already available in the mobile app TabsV2 redesign as a part of the gui.

- fileType - already implemented in the desktop app, but it doesn't work
- fileName - should be implemented soon
- embedType
- sortBy / sortOrder - same thing as the buttons below the search bar

image
image


somerandomcloud (approved):

src/plugins/MoreSearchOperators.ts
```diff
@@ -0,0 +1,213 @@
...
+export default definePlugin({
+    name: "MoreSearchOperators",
+    description: "Adds experimental search operators.",
+    authors: [Devs.Davri],
+
+    patches: [{
+        find: "Messages.SEARCH_SHORTCUT",
+        replacement: {
+            match: /(.)\((.),.\..{1,3}\.ANSWER_PINNED/,
+            replace: Object.keys(searchOperators).map(a => `$1($2,"${a}",$self.searchOperators.${a}),`).join("") + Object.keys(searchAnswers).map(a => `$1($2,"${a}",$self.searchAnswers.${a}),`).join("") + "$&"
```
Vap0r1ze, line 180:
```suggestion
            match: /(\i)\((\i),\i\.\i\.ANSWER_PINNED/,
            replace: Object.keys(searchOperators).map(a => `$1($2,"${a}",$self.searchOperators.${a}),`).join("") + Object.keys(searchAnswers).map(a => `$1($2,"${a}",$self.searchAnswers.${a}),`).join("") + "$&"
```
Vencord has a custom regex escape `\i` that matches SWC's mangled variable names

src/plugins/MoreSearchOperators.ts
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }, {
+        find: "Messages.SEARCH_ANSWER_FROM",
+        replacement: {
+            match: /var .=.\[.\];switch\(.\)\{/,
+            replace: "$&" + Object.entries(searchAnswers).map(([k, v]) => `case "${k}":a.add(e.getData("${v['_dataKey']}"));break;`).join("")
```
Vap0r1ze, line 186:
```suggestion
            match: /(?<=\i\.forEach\(\(function\((\i)\).{300,500})var (\i)=\i\[\i\];switch\(\i\)\{/,
            replace: "$&" + Object.entries(searchAnswers).map(([k, v]) => `case "${k}":$2.add($1.getData("${v['_dataKey']}"));break;`).join("")
```
You shouldn't hard code variable names in replaces. I used a lookbehind here to capture the search answer variable, and to not mess with the match replacement. Also using a limited wildcard match is kinda epic for things like this (eg: `.{200,500}`)

src/plugins/MoreSearchOperators.ts
```diff
@@ -0,0 +1,213 @@
...
+        }
+    },
+    {
+        find: "Messages.SEARCH_ANSWER_FROM",
+        replacement: {
+            match: /function .\(.\)\{switch\(.\)\{/,
```
Vap0r1ze, line 192:
```suggestion
            match: /function \i\(\i\)\{switch\(\i\)\{/,
```

src/plugins/MoreSearchOperators.ts
```diff
@@ -0,0 +1,213 @@
...
+            replace: "$&" + Object.entries(searchOperators).map(([k, v]) => `case \"${k}\":return \"${v['_options']}\";`).join("")
+        }
+    },
+    {
+        find: "Messages.SEARCH_GROUP_HEADER",
+        replacement: {
+            match: /(.{2})\((.{2}),.\..{3}\.FILTER_HAS,\{titleText:function\(\)\{return .\..\.Messages.SEARCH_GROUP_HEADER_HAS\}\}\)/,
```
Vap0r1ze, line 199:
```suggestion
            match: /(\i)\((\i),\i\.\i\.FILTER_HAS,\{titleText:function\(\)\{return \i\.\i\.Messages.SEARCH_GROUP_HEADER_HAS\}\}\)/,
```
@Vendicated we need brace balancing so baddd

src/plugins/MoreSearchOperators.ts
```diff
@@ -0,0 +1,213 @@
...
+            replace: Object.entries(searchOperators).map(([k, v]) => `$1($2,\"${k}\",{titleText:()=>\"${v['_title']}\"}),`).join("") + "$&"
+        }
+    },
+    {
+        find: "displayName=\"SearchAutocompleteStore\"",
+        replacement: {
+            match: /(.{2})\((.),.\..{3}\.FILTER_PINNED,!0\),/,
```
Vap0r1ze, line 206:
```suggestion
            match: /(\i)\((\i),\i\.\i\.FILTER_PINNED,!0\),/,
```

Vap0r1ze (changes requested):

src/plugins/MoreSearchOperators.ts
Davr1: Didn't know about that, it should probably be added to the plugin guide

Vendicated:
Sorry for not looking at this in so long. I really like this but like I mentioned the first time this was brought up I am still worried about the risks involved with sending "bad" requests like this or whether using this is safe. Have you been using this for a while? Do you know of others who used this?

ArjixWasTaken:
Is your concern that discord would think of the user as a self-bot user?

Vendicated:
Other than those concerns and the fact that there are many lint errors (pnpm lint:fix to the rescue!), it still works now, which is neat

Following the patches is a bit hard though. Please comment them so we can understand what's going on

Also you seem to be generating a lot of code as string with your patches, which tends to not be as robust and not very maintainable, better style is to insert a function call to $self.doFoo() and then implement the code in an actual function doFoo() on your plugin, passing arguments to it if necessary

Davr1:
> Sorry for not looking at this in so long. I really like this but like I mentioned the first time this was brought up I am still worried about the risks involved with sending "bad" requests like this or whether using this is safe. Have you been using this for a while? Do you know of others who used this?

I mainly use the filetype option and have not had any issues with it, and I doubt that discord even cares about additional search parameters. I briefly mentioned this plugin in a private server with ~1 vencord user, tho I'm not sure if they use it.

Since my original pull request they have rewritten the backend to further optimize stuff before it's released on mobile, so there were some minor changes in how it works.
EmbedType now only works when searching in a channel (for example in the favorites server where channels are not directly related) and not the entire server. I couldn't think of a good way to address this issue, it's possible to conditionally switch to channel-only search with the source being the current channel (or in:channel), but that's inconsistent with the rest of the operators. 
image

Davr1:
Some patches affect switch statements so it's probably better to keep them as is

Vendicated:
using embedType gives me a 500 Internal Server error lol

Vendicated:
lol what
image

Vendicated:
> lol what image

fileName seems to just not work at all and for some reason always only returns this message. That leaves only one relevant filter fileType

Davr1:
> using embedType gives me a 500 Internal Server error lol

it should now work when searching in a single channel (either the currently selected channel or using in:channel)

Davr1:
kinda wish that the other embed types were also searchable
I don't even know where post_preview or application_news is used, would be helpful if someone could test that
image

Davr1:
> > lol what image
> 
> fileName seems to just not work at all and for some reason always only returns this message. That leaves only one relevant filter fileType

this is so confusing lol
it only finds files where at least one letter from the query matches the entire filename (can't be more than one letter long)

Vendicated:
lol average discord code

Vendicated:
i was kinda expecing you to push more changes. Should we not remove the filename param? seeing how useless it is

Davr1:
tbh I was hoping they would fix it but they probably changed their decision, they're still working on a new search ui so I will add it back if it gets fixed at some point

Pierre2004:
Too bad it's only for mobile

Vendicated:
how's the status on it, did they fix it now?

Davr1:
I've lost interest, will not update this further

PR #544, opened 2023-02-28
feat(plugin): PreserveToken

NurMarvin:
Discord hides your user token in the local storage of an iframe, whenever you open the devtools. This is mostly meant to help prevent token grabbing through social engineering but has the side effect that you sometimes get logged out of Discord when Discord gets closed whilst you still have your dev tools open. This plugin basically just early-returns the `hideToken` function that usually takes care of the hiding.


NurMarvin:
Apparently already covered by the `NoDevtoolsWarning` plugin

PR #551, opened 2023-03-01
feat(plugin): SearchReply

Syncxv:
simple plugin that adds a reply button to search results

before:
before

after:
after


Vendicated:
wait for context menu api

Vendicated:
Context menu api is here now so you can use it instead!

Syncxv:
ready

src/plugins/searchReply.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { LazyComponent } from "@utils/misc";
+import definePlugin from "@utils/types";
+import { findByCode, findByCodeLazy, findLazy } from "@webpack";
+import { ChannelStore, Menu, SelectedChannelStore } from "@webpack/common";
+import { Message } from "discord-types/general";
+
+const ReplyIcon = LazyComponent(() => findByCode("M10 8.26667V4L3 11.4667L10 18.9333V14.56C15 14.56 18.5 16.2667 21 20C20 14.6667 17 9.33333 10 8.26667Z"));
+
+// for people who speak different languages
+const i18n = findLazy(m => m.Messages?.MESSAGE_ACTION_REPLY);
```
Vendicated, line 30: Could you move this to Webpack commons please? I feel like many more plugins would want to use it, and I'm pretty sure MessageLogegr already does!

src/plugins/searchReply.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const replyFn = findByCodeLazy("showMentionToggle", "TEXTAREA_FOCUS", "shiftKey");
+
+const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) => {
+    if (!props) return;
```
Vendicated, line 35: i don't think this is a relevant case to check?

src/plugins/searchReply.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const { message } = props as { message: Message; };
```
Vendicated, line 36: you could just type the props argument up above I'm fairly certain

src/plugins/searchReply.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    // make sure they are in the same channel as the message
+    if (!message || SelectedChannelStore.getChannelId() !== (message.channel_id ?? message.getChannelId())) return;
```
Vendicated, line 39:
```suggestion
    if (!message || SelectedChannelStore.getChannelId() !== message.channel_id)) return;
```

src/plugins/searchReply.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const channel = ChannelStore.getChannel(message.channel_id ?? message.getChannelId());
```
Vendicated, line 41: same as above!

Vendicated (changes requested):
Looks good other than the feedback!

Vendicated (approved):

PR #552, opened 2023-03-02
feat(plugins): add FixInbox plugin

MeguminSama:
By default, when you open the Unread tab in the Inbox, discord spams a ton of GUILD_SUBSCRIPTIONS events to the gateway. If you're in a lot of servers, this can cause the gateway client to get ratelimited, leaving your client in an unfortunate crash-loop as the gateway attempts to reconnect but just gets ratelimited every time, until you CTRL+R or restart the client.

This plugin stops the client from emitting GUILD_SUBSCRIPTIONS events when you open the Unreads tab. This solves the crashes, but means only already-subscribed guilds will be shown. It seems that high-affinity guilds are already subscribed most of the time, so this doesn't really affect UX much. I think it's a good tradeoff compared to crashing your client :trollface:

image

### DIFF
--- a/src/plugins/FixInbox.tsx
+++ b/src/plugins/FixInbox.tsx
@@ -30,8 +30,8 @@ export default definePlugin({
         replacement: {
             // This function normally dispatches a subscribe event to every guild.
             // this is badbadbadbadbad so we just get rid of it.
-            match: /INBOX_OPEN:function.+?return!1},/,
-            replace: "INBOX_OPEN:function(){return true;},"
+            match: /INBOX_OPEN:function.+?\{/,
+            replace: "$&return true;"
         }
     }],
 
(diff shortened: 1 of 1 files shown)


src/plugins/FixInbox.tsx
```diff
@@ -0,0 +1,56 @@
...
+    name: "FixInbox",
+    description: "Fixes the Unreads Inbox from crashing Discord when you're in lots of guilds.",
+    authors: [Devs.Megu],
+
+    patches: [{
+        find: "INBOX_OPEN:function",
+        replacement: {
+            // This function normally dispatches a subscribe event to every guild.
+            // this is badbadbadbadbad so we just get rid of it.
+            match: /INBOX_OPEN:function.+?return!1},/,
```
Vendicated, line 33: i don't like this regex for the simple reason that if that function ever changes to not return !1 anymore u might eat tons of code and cause issues. Why not go the much simpler and safer route of
```js
match: /INBOX_OPEN:function.+?\{/
replace: "$&return true;"
```

PR #554, opened 2023-03-02
firefox: Fix csp (QuickCss, themes, some plugins)

Vendicated:
This code just doesn't work and i don't know why


TheKodeToad:
Should probably also allow pronoundb api?

PR #556, opened 2023-03-03
Add a setting to disable the restart prompt after auto update

X1nto:
Disables the annoying green restart prompt on the top after Vencord auto-updates.

### DIFF
--- a/src/Vencord.ts
+++ b/src/Vencord.ts
@@ -49,18 +49,20 @@ async function init() {
             if (Settings.autoUpdate) {
                 await update();
                 const needsFullRestart = await rebuild();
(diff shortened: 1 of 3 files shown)


Vendicated:
I thought about this a while ago, but maybe instead we could change it to be a Notification? Those are far less intrusive while still giving users this info

Vendicated:
Vencord Notifications! it's in the api namespace. They're very similar to Windows notifications but inside Discord (or native, configurable for the user)

X1nto:
> (or native, configurable for the user)
That's super nice! I'll close this PR and open a new one.

PR #540, opened 2023-02-27
feat(plugin): TextReplace

AutumnVN:
### Things i did:

- [x] replace message before send
- [x] better gui 
- [x] use DataStore
- [x] support both string and regex
- [x] editable
- [x] auto create new empty line

### How does this shit currently look:

image

PR #557, opened 2023-03-03
feat(plugin): FxAll to fix Instagram, Reddit, Twitter and TikTok links!

eggsy:
This is a plugin inspired from the existing [`FxTwitter`](https://github.com/Vendicated/Vencord/blob/main/src/plugins/fxTwitter.ts) plugin by @somerandomcloud and @CanCodes' [`LinkFixer`](https://github.com/CanCodes/LinkFixer) bot. It replaces the following services with proxy services:

- [x] Twitter: `fxtwitter.com`
- [x] Reddit: `rxyddit.com`
- [x] Instagram: `ddinstagram.com`
- [x] TikTok: `tiktxk.com`

so you get better Discord embeds, check out the GIF to see it working 🔥 

fxall-vencord

### DIFF
--- a/src/plugins/fxAll.ts
+++ b/src/plugins/fxAll.ts
@@ -1,0 +1,66 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { addPreSendListener, MessageObject, removePreSendListener } from "@api/MessageEvents";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+const regexList = [
+    {
+        test: /https?:\/\/twitter\.com(?=\/\w+?\/status\/)/g,
+        replace: "https://fxtwitter.com"
+    },
+    {
+        test: /https?:\/\/(?:www|vm)\.tiktok\.com/g,
+        replace: "https://vm.tiktxk.com"
+    },
+    {
+        test: /https?:\/\/(?:www\.)?instagram\.com(?=\/p\/)/g,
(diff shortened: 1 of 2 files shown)


somerandomcloud:
Hey, sorry but we aren't accepting any more text replacement plugins. Only kind we will probably accept is a generic text replacement plugin. There is an open pr for it already at #540 . Maybe you can help out there?

eggsy:
> Hey, sorry but we aren't accepting any more text replacement plugins. Only kind we will probably accept is a generic text replacement plugin. There is an open pr for it already at #540 . Maybe you can help out there?

Oh, that's fine, though it might be cool to have this one as it is preconfigured to match the most known and used services out there. Most people might not be really familiar with regex and it'll be hard for them to type the regexes for all these services. I could keep maintaining this plugin to add new services and update the existing ones casually. The other plugin is by far better than this for power users, but for the end user, they might need something simple like this... If this is not enough to convince you guys, it's fine to close the issue 🥲

somerandomcloud:
Thing with someone maintaining the regexes is that they will probably have a bias towards which service to use. For example, some prefer using fxtwitter instead of vxtwitter for whatever reason. Plus there is already a plan to make the other plugin easier for end users.

eggsy:
Alright then, I'm leaving a suggestion to the issue you linked. I don't have time to contribute to it as of now but maybe later I could take a look into it. For now I'll just leave what I have in mind so you guys could maybe work it out! Kudos 🥰

AutumnVN:
> Most people might not be really familiar with regex and it'll be hard for them to type the regexes for all these services.
The other plugin replace string to string, with an optional 3rd string to make it will only run if 3rd string also match, so user dont need to know regex.

eggsy:
> The other plugin replace string to string, with an optional 3rd string to make it will only run if 3rd string also match, so user dont need to know regex.

That's true, but regex is not just a "string to string" replace. Check the code, you'll see it checks whether user posted a video by looking at the following keywords in the link: `vm`, `status`, `p` etc. if it doesn't, there's no need to replace them. If users were to replace all Instagram, Twitter and TikTok links to proxy, it won't change anything for example user profiles, hashtags, search pages, and many more because there is no need to display an embed for those. Regex comes in handy in this case. It will just replace when needed and it will leave the other links so there is no need for an additional redirection.

AutumnVN:
The "optional 3rd string" will prevent that, like i said above it will only run that "replace task" if the message also contains the "optional 3rd string".

eggsy:
> The "optional 3rd string" will prevent that, like i said above it will only run that "replace task" if the message also contains the "optional 3rd string".

Oh sorry, missed that part, alright then. Still, I trust in regex more but that'll do the job as well.

Vendicated:
See #540

Once that plugin is merged, we can make a channel on the Discord where people can share their rules for tasks like this

Vendicated:
TextReplace is now a thing, feel free to post your replacement in the #textreplace-rules channel on Discord!

PR #566, opened 2023-03-04
Fix(InvisibleChat) Fix chatbar icon patch (closes #560)

### LINKED ISSUE #560: [Bug] Invisible Chat not working
### Discord Account

Tami#7071

There is no Invisible Chat Icon to use (and the plugin has no other options than setting a password so it is enabled)

Show the Invisible Chat icon ^^

1. Install Vencord
2. Enable Invisible Chat
3. No Button

### Errors

```
No crash
```

### Request Agreement

SammCheese:


src/plugins/invisibleChat/index.tsx
```diff
@@ -131,8 +131,8 @@ export default definePlugin({
         {
             find: ".activeCommandOption",
             replacement: {
-                match: /.=.\.activeCommand,.=.\.activeCommandOption,.{1,133}(.)=\[\];/,
-                replace: "$&;$1.push($self.chatBarIcon());",
+                match: /(.)\.push.{1,50}\(.{1,3},\{.{1,30}\},"gift"\)\)/,
```
Vendicated, line 134:
```suggestion
                match: /(.)\.push.{1,50}\(\i,\{.{1,30}\},"gift"\)\)/,
```

PR #581, opened 2023-03-07
feat(plugin): F8Break

lewisakura:
Simplistic plugin that makes it easier to register an F8-to-debug handler. Vap brought it up in #general so I made it.


Vap0r1ze:
REAL

src/plugins/f8break.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+
+    stop() {
+        window.removeEventListener("keydown", this.event);
+    },
+
+    event(e: KeyboardEvent) {
+        if (e.code === "F8") {
+            // Hi! You've just paused the client. Pressing F8 in DevTools or in the main window will unpause it again.
+            // It's up to you on what to do, friend. Happy travels!
```
Vap0r1ze, line 38: Worth noting that this will only show up in dev builds

Vap0r1ze (approved):

src/plugins/f8break.ts
lewisakura: i think it's fine, mainly just a meme note anyway

PR #585, opened 2023-03-08
feat(plugin): Wikisearch

somerandomcloud:
Description: `Searches Wikipedia for your requested query. (/wikisearch)`
(If you have any better description ideas please comment I beg of you)

image


src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                if (!word) {
+                    return sendBotMessage(ctx.channel.id, {
+                        content: "No word was defined!"
+                    });
+                }
+
+                fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${word}`)
+                    .then(response => response.json())
+                    .then(async data => {
```
lewisakura, line 53: why are you doing this if your execute is async?
```typescript
const data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${word}`)
    .then(response => response.json())
```

also you need to handle errors correctly in case this fails

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,95 @@
...
+                        if (!data.query.search[0]) return sendBotMessage(ctx.channel.id, { content: "No results :(" });
+
+                        const firstres = data.query.search[0];
+
+                        let alt_data;
+                        let thumbnail;
+
+                        await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&prop=info%7Cdescription%7Cimages%7Cimageinfo%7Cpageimages&list=&meta=&indexpageids=1&pageids=${firstres.pageid}&formatversion=2&origin=*`)
+                            .then(data => data.json())
+                            .then(data => alt_data = data.query.pages[0] || null);
```
lewisakura, line 65:
```typescript
const alt_data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&prop=info%7Cdescription%7Cimages%7Cimageinfo%7Cpageimages&list=&meta=&indexpageids=1&pageids=${firstres.pageid}&formatversion=2&origin=*`)
    .then(data => data.json())
    .then(data => data.query.pages[0] ?? null);
```

and same issue with error handling

lewisakura (changes requested):

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+                    });
+                }
+
+                const data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${word}`).then(response => response.json())
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Please report this issue on the Vencord repository." });
+                    });
+
+                if (!data.query) {
```
Vendicated, line 57:
```suggestion
                if (!data.query?.search?.length) {
```
then u can just merge this if with the one below

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+                if (!word) {
+                    return sendBotMessage(ctx.channel.id, {
+                        content: "No word was defined!"
+                    });
+                }
+
+                const data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${word}`).then(response => response.json())
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Please report this issue on the Vencord repository." });
```
Vendicated, line 54:
```suggestion
                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Check the console for more info" });
```

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+
+                if (!data.query.search[0]) return sendBotMessage(ctx.channel.id, { content: "No results :(" });
+
+                let thumbnail;
+
+                let alt_data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&prop=info%7Cdescription%7Cimages%7Cimageinfo%7Cpageimages&list=&meta=&indexpageids=1&pageids=${data.query.search[0].pageid}&formatversion=2&origin=*`)
+                    .then(data => data.json())
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Please report this issue on the Vencord repository." });
```
Vendicated, line 70: this return returns from the catch, not from the surrounding function. Instead, you need to return something like null here then check for that condition after the fetch

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+                    console.log(data);
+                    return sendBotMessage(ctx.channel.id, { content: "No reply was given from Wikipedia. Check console for errors" });
+                }
+
+                if (!data.query.search[0]) return sendBotMessage(ctx.channel.id, { content: "No results :(" });
+
+                let thumbnail;
+
+                let alt_data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&prop=info%7Cdescription%7Cimages%7Cimageinfo%7Cpageimages&list=&meta=&indexpageids=1&pageids=${data.query.search[0].pageid}&formatversion=2&origin=*`)
+                    .then(data => data.json())
```
Vendicated, line 67:
```suggestion
                    .then(res => res.json())
                    .then(data => data.query.pages[0])
```

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Please report this issue on the Vencord repository." });
+                    });
+
+                alt_data = alt_data.query.pages[0] ?? null;
+
+                const thumbnail_data = alt_data.thumbnail;
+
+                if (!thumbnail_data) thumbnail = null;
```
Vendicated, line 77:
```suggestion
                const thumbnail = thumbnail_data && { url: .... };
```

Vendicated (changes requested):
Other than the feedback, don't use snake_case this isn't python. Use camelCase

somerandomcloud:
Should have resolved everything

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,96 @@
...
+                if (!word) {
+                    return sendBotMessage(ctx.channel.id, {
+                        content: "No word was defined!"
+                    });
+                }
+
+                const data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${word}`).then(response => response.json())
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Check the console for more info" });
```
Vendicated, line 54: as I said, return inside catch and similar returns from THAT function (the one passed to catch), so you need to return some value from it like null and check outside of the catch

like
```ts
const res = await fetch(....).catch(err => { ...; return null });
if (!res) return;
```

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,96 @@
...
+                    console.log(data);
+                    return sendBotMessage(ctx.channel.id, { content: "No results given" });
+                }
+
+                const altData = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&prop=info%7Cdescription%7Cimages%7Cimageinfo%7Cpageimages&list=&meta=&indexpageids=1&pageids=${data.query.search[0].pageid}&formatversion=2&origin=*`)
+                    .then(res => res.json())
+                    .then(data => data.query.pages[0])
+                    .catch(err => {
+                        console.log(err);
+                        return sendBotMessage(ctx.channel.id, { content: "There was an error. Check the console for more info" });
```
Vendicated, line 67: same as above

somerandomcloud:
Resolved all conversations! Sorry for inactivity <:trolley:1024751352028602449>

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,102 @@
...
+            execute: async (_, ctx) => {
+                const word = findOption(_, "search");
+
+                if (!word) {
+                    return sendBotMessage(ctx.channel.id, {
+                        content: "No word was defined!"
+                    });
+                }
+
+                const data = await fetch(`https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&formatversion=2&origin=*&srsearch=${encodeURIComponent(word.toString())}`).then(response => response.json())
```
Vendicated, line 51: these hardcoded query params are kind of painful to read & maintain, you can make it more readable by using URLSearchParams!
```suggestion
                const data = await fetch("https://en.wikipedia.org/w/api.php?" + new URLSearchParams({
                	action: "query",
                	format: "json",
                	list: "search",
                	formatversion: 2,
                	origin: "*",
                	srsearch: word
                }).then(response => response.json())
```

somerandomcloud: thought it looked good an autocommitted. oops

somerandomcloud: fixed

src/plugins/wikisearch.ts
```diff
@@ -0,0 +1,110 @@
...
+                }
+
+                const dataSearchParams = new URLSearchParams({
+                    action: "query",
+                    format: "json",
+                    list: "search",
+                    formatversion: "2",
+                    origin: "*",
+                    srsearch: word
+                } as unknown as URLSearchParams); // URLSearchParams doesn't take obj for some reason???
```
Vendicated, line 57:
```suggestion
                });
```

it very much does, I just tested it

somerandomcloud: image

Vendicated: because the type of srsearch is not string

Vendicated: add a third Parameter that's just an empty string to findOption

somerandomcloud: Well, fixed I guess...

Vendicated (approved):

PR #586, opened 2023-03-08
feat(plugins): SilentMessageToggle

Nuckyz:


src/plugins/silentMessageToggle.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Button, ButtonLooks, ButtonWrapperClasses, React, Tooltip } from "@webpack/common";
+
+function SilentMessageToggle() {
+    const [enabled, setEnabled] = React.useState(false);
+
+    React.useEffect(() => {
+        const listener: SendListener = (_, message) => {
+            if (enabled) {
+                setEnabled(false);
+                if (message.content.slice(0, 8) !== "@silent ") message.content = "@silent " + message.content;
```
Vendicated, line 32:
```suggestion
                if (!message.content.startsWith("@silent ")) message.content = "@silent " + message.content;
```

src/plugins/silentMessageToggle.tsx
```diff
@@ -0,0 +1,85 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
```
Vendicated, line 3:
```suggestion
 * Copyright (c) 2023 Vendicated and contributors
```

src/plugins/silentMessageToggle.tsx
```diff
@@ -0,0 +1,85 @@
...
+        const listener: SendListener = (_, message) => {
+            if (enabled) {
+                setEnabled(false);
+                if (!message.content.startsWith("@silent ")) message.content = "@silent " + message.content;
+            }
+        };
+
+        addPreSendListener(listener);
+        return () => void removePreSendListener(listener);
+    }, [enabled]);
```
Vendicated, line 38: Don't think this dependency is needed? All it does is needlessly remove and readd the listener upon toggling

```suggestion
    }, []);
```

src/plugins/silentMessageToggle.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    return (
+        <Tooltip text="Toggle Silent Message">
+            {tooltipProps => (
+                <Button
+                    {...tooltipProps}
+                    onClick={() => setEnabled(!enabled)}
```
Vendicated, line 45:
```suggestion
                    onClick={() => setEnabled(prev => !prev)}
```

when using the old state you should always use the function update style instead of using the variable directly as it's not guaranteed the variable is up to date

Vendicated: could also useReducer here
```ts
const [enabled, toggleEnabled] = React.useReducer(x => !x, false);
```

Vendicated (approved):

SylveonBottle:
https://github.com/Nuckyz/Vencord/blob/6492908a62cbb8e56be6b28e574a3e9e33cc5051/src/plugins/silentMessageToggle.tsx#L35

hey so uh can we not
or at the very least, have an option to disable this?

cappscolton:
@SylveonBottle you may be interested in tracking my fix for this feature:
https://github.com/Vendicated/Vencord/issues/1032

PR #589, opened 2023-03-10
feat(plugins): FriendsOnServer

Nuckyz:
image


dolfies:
👋 Hello

The current implementation of this is not only spamming the API, but spamming a particularly touchy endpoint. This makes it slow and highly likely to get users banned (especially with people that have 1k friends!). Instead, you can gather a list of IDs that you don't have a cached member for in the guild, and use OP 8 Request Guild Members' `user_ids` parameter with chunks of 100 IDs. This is significantly faster and safer.

MeguminSama:
funny method for the above
image

Nuckyz:
> 👋 Hello
> 
> The current implementation of this is not only spamming the API, but spamming a particularly touchy endpoint. This makes it slow and highly likely to get users banned (especially with people that have 1k friends!). Instead, you can gather a list of IDs that you don't have a cached member for in the guild, and use OP 8 Request Guild Members' `user_ids` parameter with chunks of 100 IDs. This is significantly faster and safer.

Yep, I will use that, I'm just waiting for another pull request to be merged, since it has a wrapper around using that which makes it better

src/plugins/friendsOnServer/style.css
```diff
@@ -0,0 +1,26 @@
...
+.frds-on-svr-open-modal-btn {
+    all: unset;
+    cursor: pointer;
+}
+
+.frds-on-svr-avatar-img {
+    border-radius: 50%;
+    width: 50px;
+    height: 50px;
+}
```
Vendicated, line 26: classnames should start with `.vc-` prefix to make it obvious they're vencord classes! Also I would prefer using em/rem over px for everything except border-radius here

src/plugins/friendsOnServer/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+function FriendsOnServer({ guildId, modalProps }: { guildId: string; modalProps: ModalProps; }) {
+    const friendIds = useStateFromStores(
+        [RelationshipStore],
+        () => RelationshipStore.getFriendIDs(),
+        null,
+        (old, current) => JSON.stringify(old) === JSON.stringify(current)
+    );
+
+    useStateFromStores(
```
Vendicated, line 36: this should probably be removed?

src/plugins/friendsOnServer/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const friendsInGuild = friendIds.filter(friendId => GuildMemberStore.isMember(guildId, friendId));
+
+    return (
+        <ErrorBoundary>
+            <ModalRoot
+                {...modalProps}
+                size={ModalSize.MEDIUM}
+            >
+                <ModalHeader>
+                    <Text className="frds-on-svr-title" variant="heading-lg/semibold">Friends on this server:</Text>
```
Vendicated, line 62: these classes should use classNameFactory to make it more maintainable!

src/plugins/friendsOnServer/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+                    <ModalCloseButton onClick={modalProps.onClose} />
+                </ModalHeader>
+                <ModalContent>
+                    <div className="frds-on-svr-frds-container">
+                        {friendsInGuild.length === 0
+                            ? <Text variant="text-md/normal">There are no friends on this server :/</Text>
```
Vendicated, line 68: maybe it would be useful to check here whether it is currently still fetching members and if so display something like "Loading..."

Because it would be weird to see this message then suddenly have people appear

src/plugins/friendsOnServer/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+                action={() => openModal(modalProps => <FriendsOnServer guildId={props.guild.id} modalProps={modalProps} />)}
+            />
+        ));
+    }
+};
+
+export default definePlugin({
+    name: "FriendsOnServer",
+    description: "Adds an option to server context menus to show what friends are on that server.",
+    authors: [Devs.Nuckyz],
```
Vendicated, line 115:
```suggestion
    authors: [Devs.Nuckyz],
    dependencies: ["ContextMenuAPI", "MenuItemDeobfuscatorAPI"],
```

Vendicated:
I'm not sure I like the only pfp approach. I don't know most of my friends by pfp. Why not go for something similar to the mutual friends section instead?
image

Nuckyz:
I think megu is taking over this with GuildProfiles, can you confirm? @MeguminSama

MeguminSama:
Yup. Making progress.

Vendicated:
Let's close this then?

PR #603, opened 2023-03-14
messageLogger deleted attachment hover transition now smoother

LordEliasTM:
Moved the transition from :hover to the actual class, 
because if you moved your mouse off the attachment the transition would be instant.
Additionally did it a 150ms ease-in-out, because it looks smoother than linear.


Henry-Hiles:
Seems good to me.

PR #605, opened 2023-03-15
fix(RoleColorEverywhere): chat mentions

lewisakura:
Seems Discord are now using a full channel object here. Patch changed to accomodate.


lewisakura:
Stable works with the current RCE patch, PTB/canary/future stable need the new patch. I'll keep both in - they're no-op if it isn't the right version and don't collide with anything else.

lewisakura:
Updated on stable, old patch no longer needed

PR #607, opened 2023-03-18
messageLogger: fix edited timestamp styling & add i18n

LordEliasTM:
Title says it i think


src/plugins/messageLogger/index.tsx
```diff
@@ -25,10 +25,14 @@ import { Devs } from "@utils/constants";
 import Logger from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
 import { moment, Parser, Timestamp, UserStore } from "@webpack/common";
+import { Webpack } from "Vencord";
```
Vendicated, line 28: bad

PR #613, opened 2023-03-20
Fixes and make guild tooltip show users inside hidden voice channels

Nuckyz:


src/plugins/messageLogger/index.tsx
```diff
@@ -174,6 +174,7 @@ export default definePlugin({
                     match: /(MESSAGE_UPDATE:function\((\w)\).+?)\.update\((\w)/,
                     replace: "$1" +
                         ".update($3,m =>" +
+                        "   (($2.message.flags & 64) === 64 || (Vencord.Settings.plugins.MessageLogger.ignoreBots && $2.message.author?.bot) || (Vencord.Settings.plugins.MessageLogger.ignoreSelf && $2.message.author?.id === Vencord.Webpack.Common.UserStore.getCurrentUser().id)) ? m :" +
```
Vendicated, line 177: I don't like this much logic being here, would be nicer to make it a method that uses try catch just to make sure

Nuckyz: honestly message logger is a mess as a whole, that can be changed later which maybe a full rewrite

PR #641, opened 2023-03-21
Polish

3kh0:
Here is what I did:
- Remove random space on `19`
- Covert browser download to HTML markdown with set height to make it more uniform
- Add user script button
- Redid the note for user scripts
- Linked docs folder
- Removed `contribute` from being said 72 times
- Removed `join` from being said 395 times
- Added discord invite embed

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,31 +16,30 @@ The cutest Discord client mod
 
 ## Installing / Uninstalling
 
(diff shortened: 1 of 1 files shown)


Vendicated:
this is a big downgrade, the join join join is important ☹️💔

PR #642, opened 2023-03-21
fix messagelogger deleted styles

iwa:
hello! little pr that adds more styles to elements inside deleted messages:
- shows links in darker red (with underline to see they're links), instead of plain blue
- applies grayscale filter to iframes too (for example, spotify embeds)

### DIFF
--- a/src/plugins/messageLogger/deleteStyleText.css
+++ b/src/plugins/messageLogger/deleteStyleText.css
@@ -2,7 +2,7 @@
     color: #f04747;
 }
 
-.messagelogger-deleted div a {
+.messagelogger-deleted a {
     color: #be3535;
     text-decoration: underline;
 }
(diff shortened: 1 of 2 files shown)


src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -1,3 +1,8 @@
 .messagelogger-deleted div {
     color: #f04747;
 }
+
+.messagelogger-deleted div a {
```
Vendicated, line 5:
```suggestion
.messagelogger-deleted a {
```

Vendicated: unless there's a reason to have div too?

iwa: it was mainly to confine the style to the message content, in case discord change its one day html struct and adds others hyperlink in the message component, but well it currently doesn't seem to have any impact so yeah we can get rid of it

Vendicated:
thank you!

PR #681, opened 2023-03-24
feat(plugin): PrettierCodeblocks

DuroCodes:
**Description:** Formats code inside of code blocks with Prettier, also automatically formats upon sending/editing a message if setting is enabled.

**Demo:**

https://user-images.githubusercontent.com/47304910/227414794-dfd4d8a3-d56e-4d7f-8106-e4c59ee69c5d.mov

### DIFF
--- a/src/plugins/prettierCodeblocks.ts
+++ b/src/plugins/prettierCodeblocks.ts
@@ -38,7 +38,7 @@ const cache = new Map<string, string>();
 
 export default definePlugin({
     name: "PrettierCodeblocks",
-    description: "Formats code in codeblock with Prettier.",
+    description: "Formats code in codeblocks with Prettier.",
     dependencies: ["MessageEventsAPI"],
     authors: [Devs.Duro],
     settings,
--- a/src/plugins/prettierCodeblocks.ts
+++ b/src/plugins/prettierCodeblocks.ts
@@ -73,7 +73,7 @@ export default definePlugin({
             const lang = m.split("\n")[0]?.replace("```", "");
             const lines = m.split("\n").filter(l => !l.trim().includes("```"));
 
-            return makeCodeblock(this.format(lines.join("\n")), lang);
+            return makeCodeblock(this.format(lines.join("\n"), lang), lang);
         });
     },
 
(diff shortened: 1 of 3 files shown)


Vendicated:
hmm, I'm not sure how useful this is, considering it only supports js and css. Perhaps there is a formatter that supports more languages? I also feel like a lot of times people will post only a part of the code, and often times this will not be fully valid (since there's parts missing) which means this will fail, so I think Unindent is more useful?

Vendicated:
(I'm not saying this won't be accepted, just trying to maybe make this more useful and generalised)

DuroCodes:
> hmm, I'm not sure how useful this is, considering it only supports js and css. Perhaps there is a formatter that supports more languages? I also feel like a lot of times people will post only a part of the code, and often times this will not be fully valid (since there's parts missing) which means this will fail, so I think Unindent is more useful?

I'll try to find another formatted that would support more languages! I've only found Prettier so far but I'll look into some more

Henry-Hiles:
There is rome, but I thibk thats supports the same or less than prettier.

socram03:
> (I'm not saying this won't be accepted, just trying to maybe make this more useful and generalised)

This looks like something that can solve this problem.

https://github.com/unibeautify/unibeautify

Gabe616:
> hmm, I'm not sure how useful this is, considering it only supports js and css. Perhaps there is a formatter that supports more languages? I also feel like a lot of times people will post only a part of the code, and often times this will not be fully valid (since there's parts missing) which means this will fail, so I think Unindent is more useful?

prettier supports a lot more languages
> Prettier is an opinionated code formatter with support for:
> - JavaScript (including experimental features)
> - [JSX](https://facebook.github.io/jsx/)
> - [Angular](https://angular.io/)
> - [Vue](https://vuejs.org/)
> - [Flow](https://flow.org/)
> - [TypeScript](https://www.typescriptlang.org/)
> - CSS, [Less](http://lesscss.org/), and [SCSS](https://sass-lang.com/)
> - [HTML](https://en.wikipedia.org/wiki/HTML)
> - [Ember/Handlebars](https://handlebarsjs.com/)
> - [JSON](https://json.org/)
> - [GraphQL](https://graphql.org/)
> - [Markdown](https://commonmark.org/), including [GFM](https://github.github.com/gfm/) and [MDX](https://mdxjs.com/)
> - [YAML](https://yaml.org/)

castdrian:
all of these are just markup, none of those an actual coding lang

Vendicated:
thank you a lot for your contribution, but this has limited use in its current state and has gone very stale

so i'll be closing this, sorry!

PR #682, opened 2023-03-24
fix: Check if custom theme link ends in `.css`

ihyajb:
image

Better safe then sorry I guess lol


Vendicated:
the extension doesn't actually have any meaning, it's possible to host a css file without the .css extension. Thus I don't really like this because

ihyajb:
ah true, understood

PR #683, opened 2023-03-24
Fake Nitro Transform Stickers option and other stuff

Nuckyz:


Nuckyz:
I'm making this a draft until I make things like this not transform cuz they look hella bad 
image

Nuckyz:
I did the things

src/plugins/fakeNitro.tsx
```diff
@@ -86,6 +95,56 @@ interface StickerPack {
     stickers: Sticker[];
 }
 
+const fakeNitroEmojiRegex = /\/emojis\/(\d+?)\.(png|webp|gif)/;
+const fakeNitroStickerRegex = /\/stickers\/(\d+?)\./;
+const fakeNitroGifStickerRegex = /\/attachments\/\d+?\/\d+?\/(\d+?)\.gif/;
+const fakeNitroNameRegex = /(?:\?|&)name=(\w+)/;
```
Vendicated, line 101: Parsing url stuff with regex like this is meh, it's way more robust to use the URL api (this will throw on invalid urls so might wanna try catch)
```js
new URL(url).searchParams.get("name")
```

src/plugins/fakeNitro.tsx
```diff
@@ -344,26 +430,125 @@ export default definePlugin({
...
-            const fakeNitroMatch = element.props.href.match(/https:\/\/cdn\.discordapp\.com\/emojis\/(\d+?)\.(png|webp|gif).+?(?=\s|$)/);
-            if (!fakeNitroMatch) {
-                newContent.push(element);
-                continue;
+            if (settings.store.transformEmojis) {
+                const fakeNitroMatch = element.props.href.match(fakeNitroEmojiRegex);
+                if (fakeNitroMatch) {
+                    const emojiName = EmojiStore.getCustomEmojiById(fakeNitroMatch[1])?.name ?? element.props.href.match(fakeNitroNameRegex)?.[1] ?? "FakeNitroEmoji";
+
+                    newContent.push(MessageElementsParser.customEmoji.react({
```
Vendicated, line 438: the way u get this is really cursed, you can just use the Parser from webpack commons

```suggestion
                    newContent.push(Parser.defaultRules.customEmoji.react({
```

src/plugins/fakeNitro.tsx
```diff
@@ -344,26 +430,125 @@ export default definePlugin({
...
+                        jumboable: !inline,
+                        animated: fakeNitroMatch[2] === "gif",
+                        emojiId: fakeNitroMatch[1],
+                        name: emojiName,
+                        fake: true
+                    }, void 0, { key: "0" }));
```
Vendicated, line 444: this is likely not the key you should be using

Nuckyz: I don't know the right key lol, from my tests it was sometimes "1"

Nuckyz:
I did the le things

src/plugins/fakeNitro.tsx
```diff
@@ -513,7 +716,8 @@ export default definePlugin({
                 if (emoji == null || (emoji.guildId === guildId && !emoji.animated)) continue;
                 if (!emoji.require_colons) continue;
 
-                const url = emoji.url.replace(/\?size=\d+/, `?size=${Settings.plugins.FakeNitro.emojiSize}`);
+                const url = emoji.url.replace(/\?size=\d+/, `?size=${Settings.plugins.FakeNitro.emojiSize}`)
+                    + `&name=${emoji.name}`;
```
Vendicated, line 720: It would probably be better to use URLSearchParams here as well
```js
emoji.url.replace(/\?size=\d+/, "?" + new URLSearchParams({
	size: Settings.plugins.FakeNitro.emojiSize,
	name: emoji.name
}))
```

src/plugins/fakeNitro.tsx
```diff
@@ -483,7 +685,7 @@ export default definePlugin({
                     }
 
                     delete extra.stickerIds;
-                    messageObj.content += " " + link;
+                    messageObj.content += " " + link + `&name=${sticker.name}`;
```
Vendicated, line 688: Not sure if names can contain non urlsafe characters but better be safe than sorry

```suggestion
                    messageObj.content += " " + link + `&name=${encodeURIComponent(sticker.name)}`;
```

src/plugins/fakeNitro.tsx
```diff
@@ -407,8 +602,29 @@ export default definePlugin({
...
 
-        let lastImg: HTMLImageElement | null = null;
-        for (const { left, top, width, height, disposeOp, img, delay } of frames) {
+        let previousFrame: ApngFrame | null = null;
+        let previousFrameData: ImageData | null = null;
+
+        for (const frame of frames) {
+            const { left, top, width, height, img, delay, blendOp, disposeOp } = frame;
+
+            if (previousFrame?.disposeOp === ApngDisposeOp.BACKGROUND) {
```
Vendicated, line 611: why not keep this logic at the end of the loop like previously? Would eliminate the need for keeping some variables

src/plugins/fakeNitro.tsx
```diff
@@ -331,39 +417,148 @@ export default definePlugin({
...
+            if (match) {
+                if (StickerStore.getStickerById(match[1])) return false;
+            }
+
+            return true;
+        });
+    },
+
+    shouldKeepEmojiLink(link: any) {
+        return link.target?.match(fakeNitroEmojiRegex);
```
Vendicated, line 559:
```suggestion
        return link.target && fakeNitroEmojiRegex.test(link.target);
```

src/plugins/fakeNitro.tsx
```diff
@@ -331,39 +417,148 @@ export default definePlugin({
...
+        return false;
+    },
+
+    filterAttachments(attachments: Message["attachments"]) {
+        return attachments.filter(attachment => {
+            if (attachment.content_type !== "image/gif") return true;
+
+            const match = attachment.url.match(fakeNitroGifStickerRegex);
+            if (match) {
+                if (StickerStore.getStickerById(match[1])) return false;
```
Vendicated, line 551: Why is this here? Maybe add a comment explaining what's going on cause i have no clue

src/plugins/fakeNitro.tsx
```diff
@@ -331,39 +417,148 @@ export default definePlugin({
...
+        return stickers;
+    },
+
+    shouldIgnoreEmbed(embed: Message["embeds"][number], message: Message) {
+        if (message.content.split(/\s/).length > 1) return false;
+
+        switch (embed.type) {
+            case "image": {
+                if (settings.store.transformEmojis) {
+                    const match = embed.url!.match(fakeNitroEmojiRegex);
```
Vendicated, line 524: same as below and same for everywhere else, if you just need a boolean whether it matches or not, use regex.test() instead (has some weird behaviour on global regexes but that isn't the case here)

Vendicated (changes requested):

Vendicated (approved):

Issue #649, opened 2023-03-21
[Feature Request] Make Vencord section title look consistent with Discord


3kh0:
### Discord Account

3kh0_#6969

### What is it that you'd like to see?

![2023_03_21_0u4_Kleki](https://user-images.githubusercontent.com/58097612/226701381-af472856-d619-4122-a751-b87498e11226.png)

![image](https://user-images.githubusercontent.com/58097612/226701455-e745e816-76eb-47cf-94b5-65f4c6a6518a.png)

The Vencord section title seems off from the other discord sections. So having the vencord title to match the others would make it just that little bit better

### Request Agreement

- [X] This is not a plugin request

PR #685, opened 2023-03-24
Make Vencord title look consistent with Discord (#649)

AutumnVN:


Vendicated:
Hardcoding the colour like this is bad. You should use the correct css variable. But this is probably already somehow incorporated into something like the FormTitle component

dossy:
Discord uses the CSS variable `--header-primary`:

image

dossy:
resolves: #649

PR #693, opened 2023-03-24
MessageLogger: Add context menu entry to remove history

Vendicated:
If it wasn't obvious, these buttons remove the edit history and fully remove deleted messages

image
image

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -18,6 +18,7 @@
 
 import "./messageLogger.css";
 
+import { addContextMenuPatch, NavContextMenuPatchCallback, removeContextMenuPatch } from "@api/ContextMenu";
 import { Settings } from "@api/settings";
 import { disableStyle, enableStyle } from "@api/Styles";
 import ErrorBoundary from "@components/ErrorBoundary";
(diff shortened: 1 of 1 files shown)


src/plugins/messageLogger/index.tsx
```diff
@@ -43,20 +44,48 @@ function addDeleteStyle() {
...
+    dependencies: ["ContextMenuAPI", "MenuItemDeobfuscatorAPI"],
 
     start() {
         addDeleteStyle();
+        addContextMenuPatch("message", patchMessageContextMenu);
     },
 
     stop() {
-        document.querySelectorAll(".messagelogger-deleted").forEach(e => e.remove());
-        document.querySelectorAll(".messagelogger-edited").forEach(e => e.remove());
```
Vendicated, line 57: removed these because doing this actually causes crashes. Imo cleanup isn't that important, people can always just restart

src/plugins/messageLogger/index.tsx
```diff
@@ -271,15 +301,10 @@ export default definePlugin({
             // Module 748241
             find: "Message must not be a thread starter message",
             replacement: [
-                {
-                    // Write message.deleted to deleted var
-                    match: /var (\w)=(\w).id,(?=\w=\w.message)/,
-                    replace: "var $1=$2.id,deleted=$2.message.deleted,"
-                },
```
rushiiMachine, line 278: is this no longer needed?

Vendicated: See my change right below, instead of doing this you can just use arguments[0].message.deleted

PR #695, opened 2023-03-24
Use HTTP HEAD for validating themes to avoid unnecessary data traffic

LordEliasTM:
Title says it


Vendicated:
Thank you but I deliberately didn't do this because some servers might not support HEAD. This also shouldn't really matter at all because the theme needs to be fetched anyway to apply it, assuming it correctly sets cache headers it will just reuse the result of the fetch here

LordEliasTM:
Ah right, didn't think of that

PR #696, opened 2023-03-25
feat(settings): OptionType.ARRAY | OptionType.REGEX

### LINKED ISSUE #615: [Feature Request] Array Settings type
### Discord Account

a vending machine

Currently, to allow one or more inputs to a single setting, you have to make users enter a comma separated list which is error prone and not very user friendly.

It would be really nice to instead have an array option. One way this could be rendered is that there's a single input with an X button behind it and a plus button above or below that adds a new similar input, so you'd have something like

##### Some Title
[ input box 1 ] :x:
[ input box 1 ] :x:
➕

### Request Agreement

Vap0r1ze:
Closes #615

> **Warning**
> Before you test this PR, make a backup of your settings file. This PR is WIP and changes the way settings are (de)serialized.

Adds future support for "impure" types (types aren't JSON values). Right now it adds Arrays, RegExps, and Maps; the example impl is a TextReplace plugin. Not sure what use-case map could have, but I added it before so bleh.

TODO:
- [ ] PluginModal components for each setting type
- [ ] Testing the (de)serialization


Mylloon:
Any update on this?

Vendicated:
this is very stale now and i don't quite see the use case for wanting to store maps or regexes. those would better be stored in indexeddb. it seems a bit overengineered to need this serialisation step instead of just continuing to use basic json

as for arrays, definitely useful but better implemented like described in https://github.com/Vendicated/Vencord/issues/2210

thanks for your efforts regardless <3

PR #702, opened 2023-03-25
Updated 2_PLUGINS.md

PierreV23:
Updated 2_PLUGINS.md to also say to use `pnpm inject` once.


MeguminSama:
This is already stated in the install guide.

PR #704, opened 2023-03-25
fix: API plugins being force enabled

lewisakura:
Adds an extra check into the plugin loader to fix a logic error where API plugins were force enabled even if their dependents weren't.

### DIFF
--- a/src/plugins/index.ts
+++ b/src/plugins/index.ts
@@ -43,8 +43,8 @@ export function isPluginEnabled(p: string) {
 
 const pluginsValues = Object.values(Plugins);
 
-// First roundtrip to mark and force enable dependencies
-for (const p of pluginsValues) {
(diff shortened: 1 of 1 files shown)


Vendicated:
I think this has the issue where if plugin b depends on plugin a and plugin a depends on plugin c, plugin c will not be correctly enabled, because at the time plugin a is visited it's not enabled yet, but afterwards once b is visited, a is enabled. But as of now no apis have any dependencies so it's fine

lewisakura:
I can see that being an issue, but we do have the `startDependenciesRecursive` method so if we ever need APIs to depend on APIs for whatever eldritch reason we can fix up the logic to do so.

Vendicated:
maybe add a comment on it for now? Something like `FIXME: might need to revisit this if there's ever nested (dependencies of dependencies) dependencies`

PR #710, opened 2023-03-26
feat(plugin): FakeProfileThemes

remty5:
This plugin allows users to theme their profile by storing the colors in an invisible string in their bio.
This is a port of [Alyxia's Vendetta plugin](https://github.com/lexisother/vendetta-plugins/tree/master/plugins/hiddenProfileColours)

https://user-images.githubusercontent.com/42155711/227810001-bf04f0e5-1f9a-4356-9bcd-0feb987e1263.mp4


src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,120 @@
...
+    ],
+    settings,
+    colorDecodeHook(user: UserProfile) {
+        if (user) {
+            // don't replace colors if already set with nitro
+            if (settings.store.nitroFirst && user.themeColors) return user;
+            const colors = decode(user.bio);
+            if (colors) {
+                user.premiumType = 2;
+                user.themeColors = colors;
```
Vendicated, line 103: mutating the user in place like this is meh, maybe you could make this pure?
```js
return new Proxy(user, {
	get(target, prop) {
		switch (prop) {
			case "premiumType": return 2;
			case "themeColors": colors;
		}
		return target[prop];
	}
});
```

src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,120 @@
...
+    },
+    addCopy3y3Button(primary: number, accent: number) {
+        return <Button
+            onClick={() => {
+                const colorString = encode(primary, accent);
+                copyWithToast(colorString);
+            }}
+            color={Button.Colors.PRIMARY}
+            size={Button.Sizes.XLARGE}
+            style={{ marginLeft: "16px" }}
```
Vendicated, line 116:
```suggestion
            className={Margins.left16}
```

src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,120 @@
...
+            if (settings.store.nitroFirst && user.themeColors) return user;
+            const colors = decode(user.bio);
+            if (colors) {
+                user.premiumType = 2;
+                user.themeColors = colors;
+            }
+        }
+        return user;
+    },
+    addCopy3y3Button(primary: number, accent: number) {
```
Vendicated, line 108: Always wrap react stuffs in an ErrorBoundary to catch errors
```suggestion
    addCopy3y3Button: ErrorBoundary.wrap(function (primary: number, accent: number) {
```
(and close it at the end)

Vendicated (changes requested):

Vendicated:
Also for future reference, you should never pr from the main branch! Make a feature branch for that

src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+export default definePlugin({
+    name: "FakeProfileThemes",
+    description: "Allows profile theming by hiding the colors in your bio thanks to invisible 3y3 encoding.",
+    authors: [Devs.Alyxia, Devs.Remty],
+    patches: [
+        {
+            find: "getUserProfile=",
+            replacement: {
+                match: /(getUserProfile=function\(\i\){return )(\i\[\i\])/,
```
Vendicated, line 92:
```suggestion
                match: /(?<=getUserProfile=function\(\i\){return )(\i\[\i\])/,
```

No reason to match unchanged code

src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,130 @@
...
+                replace: "$1$self.colorDecodeHook($2)"
```
Vendicated, line 93:
```suggestion
                replace: "$self.colorDecodeHook($1)"
```

followup of above change

src/plugins/fakeProfileThemes.tsx
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }, {
+            find: ".USER_SETTINGS_PROFILE_THEME_ACCENT",
+            replacement: {
+                match: /ACCENT}\)}\)}\)(?<=(?<=},color:(\i).+)},color:(\i).+)/,
```
Vendicated, line 98: this is kind of messy, matching specific brackets is always only a last resort as it's not very robust, I will look more at this tomorrow

Vendicated (changes requested):

Pixel556:
How to install?

TheKodeToad:
If you wait it will probably be available in Vencord at some point unless there are issues.

Vendicated (approved):

Vendicated:
Thank you!

Im-IntoDev:
Can other people see it?

Vendicated:
Anyone who has the plugin enabled will see it. Everyone else who doesn't will not see it

Im-IntoDev:
Well then it's not so usefull, it's extremely cool and well thought out tho!

Vendicated:
The other alternative would be to hack into Discord's servers and add the features to you but I fear that is not possible and also not very legal, so this is the best we can do

somerandomcloud:
common bestdiscord L

Im-IntoDev:
Well there's the better quality and emojis working without hacking into the servers.

somerandomcloud:
Sounds like you arent actually into dev

TheKodeToad:
xD it works by sending images

Yudin101:
Umm....How do i enable it??
like i have the plugin enabled but it is not doing anything i can see a copy button beside it but idk

PR #724, opened 2023-03-29
fix(plugins): update sentry patch

Vap0r1ze:
fear
this is a good example of when AST patching would be useful, also maybe a plugin option to disable the enforcing of concurrent patches


Vendicated:
Superseded by https://github.com/Vendicated/Vencord/commit/9d41b360c9190b8c62061b302155cfb6742044b2

PR #765, opened 2023-04-03
Add additional build flavours for Vencord Desktop

Vendicated:
(https://github.com/Vencord/Electron)


somerandomcloud:
Seems like everything is working as it should be on Linux

PR #776, opened 2023-04-04
Fix things being display twice with BetterFolders

AutumnVN:
Before:
image

After:
image

### DIFF
--- a/src/plugins/readAllNotificationsButton.tsx
+++ b/src/plugins/readAllNotificationsButton.tsx
@@ -62,10 +62,10 @@ export default definePlugin({
     renderReadAllButton: () => <ReadAllButton />,
 
     start() {
-        addServerListElement(ServerListRenderPosition.In, this.renderReadAllButton);
+        addServerListElement(ServerListRenderPosition.Above, this.renderReadAllButton);
     },
 
     stop() {
-        removeServerListElement(ServerListRenderPosition.In, this.renderReadAllButton);
+        removeServerListElement(ServerListRenderPosition.Above, this.renderReadAllButton);
     }
 });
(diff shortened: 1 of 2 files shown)


src/plugins/readAllNotificationsButton.tsx
```diff
@@ -62,10 +62,10 @@ export default definePlugin({
     renderReadAllButton: () => <ReadAllButton />,
 
     start() {
-        addServerListElement(ServerListRenderPosition.In, this.renderReadAllButton);
+        addServerListElement(ServerListRenderPosition.Above, this.renderReadAllButton);
```
Vendicated, line 65: what's this change for?

AutumnVN: i see ServerListIndicator doesnt have this "display twice" issue and it use `ServerListRenderPosition.Above`, so i change this to above too to make read all button doesnt display twice

Vendicated: then what's the css change for?

Vendicated: oh to hide the NEW, I see! I'm sorry, I was confused because I thought you for some reason changed both the css and the position

Vendicated:
looks good but please fix lint!

src/plugins/readAllNotificationsButton.tsx
AutumnVN: the red "new" bar
it doesnt do anything in BetterFolder

AutumnVN:
fixed

Vendicated:
thank you :3

PR #790, opened 2023-04-04
fix: tweak notification related ui

xafn:
image
image
image


src/components/VencordSettings/VencordTab.tsx
```diff
@@ -126,6 +125,12 @@ function VencordSettings() {
                             </Button>
                         </React.Fragment>
                     )}
+                    <Button
```
Vendicated, line 128: this is only on web, on pc this section is already full

xafn: guhuehrfuiersd insan

Vendicated: note how this is inside an IS_WEB

PR #792, opened 2023-04-05
Permission Viewer

peyton2465:


ArjixWasTaken:
This does indeed work, but I am 100% sure a lot of people will complain or say the plugin does not work, just because it is not identical to the BD and powercord counterparts.

Are you willing to add the context menu and popup or at least would you accept a PR on your PR that adds that?

peyton2465:
> This does indeed work, but I am 100% sure a lot of people will complain or say the plugin does not work, just because it is not identical to the BD and powercord counterparts.
> 
> Are you willing to add the context menu and popup or at least would you accept a PR on your PR that adds that?

I haven't used either of these other options, but I don't see why users would complain. This plugin enables the functionality to view permissions and settings the way discord intended.

ArjixWasTaken:
> > This does indeed work, but I am 100% sure a lot of people will complain or say the plugin does not work, just because it is not identical to the BD and powercord counterparts.
> > 
> > Are you willing to add the context menu and popup or at least would you accept a PR on your PR that adds that?
> 
> I haven't used either of these other options, but I don't see why users would complain. This plugin enables the functionality to view permissions and settings the way discord intended. 

Because users are dumb and annoying.

Vendicated:
Hi! This is now obsolete since the other PermissionViewer plugin has been merged. Thanks anway

PR #800, opened 2023-04-05
Split Large Messages plugin

PierreV23:
The well known Split Large Messages plugin, but now in Vencord. The regex pattern is subject to change, but I do not have time nor motivation to add or change anything to this plugin currently.


src/plugins/splitLargeMessages.ts
```diff
@@ -0,0 +1,167 @@
...
+        {
+            find: ".PREMIUM_MESSAGE_LENGTH_CHATBOX_FLAIR.format",
+            replacement: {
+                match: /,{children:\w+}\)\)}}\)]}\)(?<=function\s\w+\((\w+)\){var \w+,(\w+)=\w+\.type,(\w+)=\w+.textValue[.\s\S]+?,{children:(\w+)}\)\)}}\)]}\))/,
+                replace: ",{children:$2.analyticsName===\"normal\"?$3.length:$4}))}})]})"
+            }
+        }
+    ],
+    start() {
+        Vencord.Api.MessageEvents.addPreSendListener((channelId, message, _extra) => {
```
Vendicated, line 155: avoid using the Vencord global, import the functions you need
```suggestion
        addPreSendListener((channelId, message, _extra) => {
```

src/plugins/splitLargeMessages.ts
```diff
@@ -0,0 +1,167 @@
...
+});
+
+export default definePlugin({
+    name: "Split Large Messages",
+    description: "Splits larger than allowed messages into chunks and sends them in seperate messages.",
+    authors: [
+        {
+            id: 174134334628823041n,
+            name: "Pièrre",
+        }
```
Vendicated, line 135: instead of adding a plain object here, add yourself to the Devs constant and reference that

src/plugins/splitLargeMessages.ts
```diff
@@ -0,0 +1,167 @@
...
+                value: SplitPlace.Newline,
+                default: true,
+            },
+            {
+                label: "On last possible character, crams everything into smallest amount of possible messages.",
+                value: SplitPlace.LastCharacter,
+            }
+        ],
+    },
+    delay: {
```
Vendicated, line 119: remove this option. Have a constant unchangeable delay that is guaranteed to not get you ratelimited

src/plugins/splitLargeMessages.ts
```diff
@@ -0,0 +1,167 @@
...
+        case PremiumType.Nitro:
+            return 4000;
+        default:
+            return 2000;
+    }
+}
+
+function intoChunks(messageContent: string, maxMessageSize: number, splitPlace: SplitPlace): string[] {
+    if (messageContent.length <= maxMessageSize) return [messageContent];
+    var chunks: string[];
```
Vendicated, line 53: please avoid using var unless you have a good reason to. this should be a const

src/plugins/splitLargeMessages.ts
```diff
@@ -0,0 +1,167 @@
...
+import { definePluginSettings } from "@api/settings";
+import { makeRange } from "@components/PluginSettings/components/SettingSliderComponent";
+import { getCurrentChannel } from "@utils/discord";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+import { UserStore } from "@webpack/common";
+
+const MessageCreator = findByPropsLazy("getSendMessageOptionsForReply", "sendMessage");
+
+const wait = (s: number) => new Promise(res => setTimeout(res, s * 1000.0));
```
Vendicated, line 28: you can use sleep from utils instead

Vendicated:
Please limit the amount of messages that can be sent. For example, limit it at 5 messages. That is plenty and anything higher has high spam potential

ArjixWasTaken:
I honestly don't see the point of this plugin.
2 years ago it would make sense but, nowadays, discord prompts you to send a huge message as a file

Yeah, it might be inconvenient for mobile users, but I doubt anyone on mobile is willing to read a message longer than 2000/4000 characters.

Plus sending a giant blob of text as a file makes the chat look cleaner as well.

Vendicated:
^ I do agree with this, why would you not just send your message as a text file?

PR #803, opened 2023-04-05
fix

exitss:


Captain8771:
Why exactly is this needed?

exitss:
> Why exactly is this needed?

if you just do "npm i -g pnpm" it installs the wrong version and you will get the error: "ERR_PNPM_FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE  Cannot perform a frozen installation because the version of the lockfile is incompatible with this version of pnpm"

Vendicated:
https://github.com/Vendicated/Vencord/commit/63f5b0a6639bc0bbedd4420193a3772269db25aa

PR #804, opened 2023-04-05
Add debugging info to CustomRPC, and a link to the docs.

Captain8771:
I sure hope I didn't fuck up in making this pull request.

I wrote this code like a week or two ago and I forgot to push it :skull:


Captain8771:
> Error: Invalid plugin src/plugins/customRPC.tsx, because no default export called 'definePlugin' found

ven what the hell is this and why can't we 
```ts
const plugin = definePlugin(...);

export default plugin;
```

Vendicated:
> > Error: Invalid plugin src/plugins/customRPC.tsx, because no default export called 'definePlugin' found
> 
> ven what the hell is this and why can't we
> 
> ```ts
> const plugin = definePlugin(...);
> 
> export default plugin;
> ```

there's no reason to need to do that

Vendicated:
this is really bad because it doesn't actually address the root of the problem, which is the fact that rpc has 0 validation. it would be much better to add proper validation so we don't even have to help people with it

"missing large image tooltip"
"-1 is not a valid timestamp"
"blah is not a valid url"

PR #807, opened 2023-04-06
feat(plugin): ClientTheme

CodeF53:
### About:
A recreation of the original client theme discord experiment, before they ruined it with gradient overlays.

What the old client theme experiment looked like:
what the old client theme experiment looked like

### Demos:
image

https://user-images.githubusercontent.com/37855219/230238798-2284cb50-3409-45e3-9c4a-17e6f57d0581.mp4

### Limitations:
The original experiment automatically switched between a base of light/dark mode depending on the color selected. This helped guarantee that the resulting theme never looked bad. 

Since the murder of the experiment, discord has changed the way they do light theme. Making recreating what it once looked like no longer as simple as adjusting `--primary-xxx-hsl` vars.

Because of this, I can't support light theme. This leads to users being able to select colors that look like dogshit.
<details>
<preview>Example of dogshit color being selected</preview>

image
image
</details>
To address this, I have added a easy to see `Reset theme color` button, and warnings about light colors and light theme.

### Why isn't this just a theme?
1. ColorPicker go BRRRRRRR (distributing it as a theme makes would require users to calculate their own HSL values, including decimal precision, which isn't easily accessible anywhere, as all the HEX->HSL converters out there round their values, restricting freedom)
2. It dynamically generates the css lightness offsets for the `primary-xxx-hsl` vars. This makes it so when discord decides to add a new number to their primary color variables, this instantly has support for it, as opposed to the couple of days it may take for a theme to be updated.


iosrouter:
Preach. Me need this

CodeF53:
https://twitter.com/DiscordPreviews/status/1595569677316636675/video/1

video demo of what this was based off of

CodeF53:
This still works. Would love a review.

Wikinger8:
Really nice pull request. :)

CodeF53:
@Wikinger8 this pull request now lives in #third-party-plugins of the vencord discord. Check its thread for the latest code updates.
https://discord.com/channels/1015060230222131221/1032200195582197831/1166593154082947105

PR #813, opened 2023-04-06
feat(MessageLinkEmbeds): add whitelist/blacklist modes

### LINKED ISSUE #678: [Feature Request] MessageLinkEmbeds: blacklist servers
### Discord Account

nearcatch#3721

I'd like to blacklist certain servers from the MessageLinkEmbeds plugin, because these servers already have a bot that embeds previous messages. Right now those servers look like this:
image

### Request Agreement

ryanccn:
Sometimes servers may have their own bots that send embeds, or you just need the embeds for some specific servers. This adds a blacklist/whitelist system to do so.

By default it is in blacklist mode and the list is empty, so the original global functionality is retained.

Closes #678 apparently (yay)


src/plugins/messageLinkEmbeds.tsx
```diff
@@ -360,7 +382,11 @@ export default definePlugin({
 
             return (
                 <ErrorBoundary>
-                    <MessageEmbedAccessory message={props.message} />
+                    <MessageEmbedAccessory
+                        message={props.message}
+                        listMode={this.settings.store.listMode}
+                        serverList={this.settings.store.serverList?.split(",") ?? []}
```
Vendicated, line 388:
```suggestion
                        serverList={this.settings.store.serverList?.split(",").map(s => s.trim()) ?? []}
```

ryanccn:
Are we waiting for the list settings type

Vendicated:
Looks good for the most part. Could we also add author and channel (black|white)list? The same setting could be used, just dump all ids into one and check whether any of those 3 ids are included

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -360,7 +382,11 @@ export default definePlugin({
 
             return (
                 <ErrorBoundary>
-                    <MessageEmbedAccessory message={props.message} />
+                    <MessageEmbedAccessory
+                        message={props.message}
+                        listMode={this.settings.store.listMode}
+                        serverList={this.settings.store.serverList?.split(",").map(s => s.trim()) ?? []}
```
Vendicated, line 388: this split is obsolete. You can just .includes() on the raw string

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -360,7 +382,11 @@ export default definePlugin({
 
             return (
                 <ErrorBoundary>
-                    <MessageEmbedAccessory message={props.message} />
+                    <MessageEmbedAccessory
+                        message={props.message}
+                        listMode={this.settings.store.listMode}
```
Vendicated, line 387: there is no reason to pass this as a prop. settings should already be a global variable - if not, make it one - so just use it directly

ryanccn:
> Looks good for the most part. Could we also add author and channel (black|white)list? The same setting could be used, just dump all ids into one and check whether any of those 3 ids are included

@Vendicated done!

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -93,6 +93,25 @@ const settings = definePluginSettings({
...
+                default: true
+            },
+            {
+                label: "Whitelist",
+                value: "whitelist"
+            }
+        ]
+    },
+    idList: {
+        description: "Guild/channel/user IDs to embed/ignore message links in (separate with comma)",
```
Vendicated, line 112:
```suggestion
        description: "Guild/channel/user IDs to blacklist or whitelist (separate with comma)",
```

Vendicated (approved):

Vendicated:
thank you!

PR #817, opened 2023-04-07
InvisibleChat: Add missing dependency on MessagePopoverAPI

exitss:


Vendicated:
maybe use slightly more informative titles in the future lmao

PR #833, opened 2023-04-07
fix(FakeNitro): fix [object Object] when adding the Fake Nitro notice to emojis

whqwert:
before
after


Vendicated:
i don't think this is a fix, react nodes won't have concat either

whqwert:
> i don't think this is a fix, react nodes won't have concat either

it does fix the problem though?

image

Vendicated:
SUperseded by https://github.com/Vendicated/Vencord/commit/65f7cf9503a0c95affb3b6c2b694c6d2d26b90eb

PR #836, opened 2023-04-07
feat(plugin): NoUpsell

dzshn:
Suppresses all dismissible contents (aka Nitro and feature upsell popups and banners)

### DIFF
--- a/src/plugins/noDismissibleContents.ts
+++ b/src/plugins/noDismissibleContents.ts
@@ -27,7 +27,22 @@ export default definePlugin({
     description: "Dismiss all upsell popups about Discord features or Nitro ads before you ever see them",
     authors: [Devs.dzshn],
 
+    patches: [
+        {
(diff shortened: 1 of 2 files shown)


dzshn:
Race conditions should be fixed, it's really weird how this value gets initialised but this workaround should be good. Basically: even though `dismissibleContents` is as far as I know completely independent per-client it still attempts to be sent to Discord's API along with other user settings, and to future-proof for more prompts I made the array larger than it should be, so it fails the server-side schema. By setting it to an empty array (no contents dismissed, the initial value) and making the deserialiser (??????) always parse `dismissibleContents` as our array, it gets cached properly and doesn't fail server-side.

Henry-Hiles:
Couldn't this be done with CSS?

TheKodeToad:
I think with popups it would be buggy. 

Also removing this would be nice :p
image

PR #856, opened 2023-04-10
InvisibleChat: Add default encryption password setting + Replace Insert Text with Send Message

exitss:


src/plugins/invisibleChat/components/EncryptionModal.tsx
```diff
@@ -87,9 +85,7 @@ function EncModal(props: ModalProps) {
                         const toSend = noCover ? encrypted.replaceAll("d", "") : encrypted;
                         if (!toSend) return;
 
-                        ComponentDispatch.dispatchToLastSubscribed("INSERT_TEXT", {
-                            rawText: `${toSend}`
-                        });
+                        findByProps("editMessage").sendMessage(SelectedChannelStore.getChannelId(), { content: toSend });
```
Vendicated, line 88: why?

exitss: Currently, the plugin functions by inserting the encrypted text into the message box, and then the user must press "enter" to send the message. However, it could be made more efficient by having the plugin send the encrypted message directly to the server, thus eliminating the extra step. Additionally, the button in the modal window is labelled "send," which may mislead users into believing that the message is being sent to the server, even though it is not.

image

src/plugins/invisibleChat/components/EncryptionModal.tsx
```diff
@@ -85,7 +85,7 @@ function EncModal(props: ModalProps) {
                         const toSend = noCover ? encrypted.replaceAll("d", "") : encrypted;
                         if (!toSend) return;
 
-                        insertTextIntoChatInputBox(toSend);
+                        findByProps("editMessage").sendMessage(SelectedChannelStore.getChannelId(), { content: toSend });
```
Vendicated, line 88: this causes errors because your message body is missing required fields

Vendicated: I just added sendMessage in @utils/discord, please use it:
```suggestion
                        sendMessage(SelectedChannelStore.getChannelId(), { content: toSend });
```

Vendicated:
obsolete now, thankyou anyway!

PR #857, opened 2023-04-10
ImageZoom: Fix not being able to close carousels on sides

Syncxv:
man naming things is hard.

basically allows you to close the image when you click the backdrop. before you could only do that if you click top or bottom parts of the backdrop but now you can click any part of the backdrop and it will close it
it fixes this:
https://discord.com/channels/1015060230222131221/1015063227299811479/1094745672944726107

thanks aaaz#0001

### DIFF
--- a/src/plugins/imageZoom/styles.css
+++ b/src/plugins/imageZoom/styles.css
@@ -30,7 +30,7 @@
     transform: translateY(-50%);
 }
 
-[class^="focusLock"] > [class^="carouselModal"] > [class*="modalCarouselWrapper"] > [class^="wrapper"] {
+#magnify-modal {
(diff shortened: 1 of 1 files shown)


src/plugins/imageZoom/styles.css
```diff
@@ -29,3 +29,10 @@
     top: 50%;
     transform: translateY(-50%);
 }
+
+[class^="focusLock"] > [class^="carouselModal"] > [class*="modalCarouselWrapper"] > [class^="wrapper"] {
```
Vendicated, line 33: this seems to work too?
```suggestion
#magnify-modal {
```

PR #861, opened 2023-04-10
feat(plugin): LockServerLeave

TheKodeToad:
Still really not sure whether we/you need this!
merge if cute :3


TheKodeToad:
After implementing it, it feels even more useless! this should probably be prefered https://discord.com/channels/1015060230222131221/1015063227299811479/1095014211215704075

TheKodeToad:
Yes, I feel like it would also need other features but they're not really important on Discord since they can already be achieved in other ways (such as moving the server to the top or adding a star)

PR #862, opened 2023-04-10
PronounDB: Fix pronouns flickering

TheKodeToad:
When I used useAwaiter, it gave me some idea how it worked and therefore why this bug occurs.
It worries me how quick i did this :p

### DIFF
--- a/src/plugins/pronoundb/pronoundbUtils.ts
+++ b/src/plugins/pronoundb/pronoundbUtils.ts
@@ -41,16 +41,27 @@ const bulkFetch = debounce(async () => {
 });
 
 export function awaitAndFormatPronouns(id: string): string | null {
-    const [result, , isPending] = useAwaiter(() => fetchPronouns(id), {
-        fallbackValue: null,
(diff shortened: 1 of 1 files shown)


src/plugins/pronoundb/pronoundbUtils.ts
```diff
@@ -41,18 +41,29 @@ const bulkFetch = debounce(async () => {
...
 
 export function awaitAndFormatPronouns(id: string): string | null {
-    const [result, , isPending] = useAwaiter(() => fetchPronouns(id), {
-        fallbackValue: null,
-        onError: e => console.error("Fetching pronouns failed: ", e)
-    });
+    let result = getCachedPronouns(id);
+
+    if (!result) {
+        const [fetched, , isPending] = useAwaiter(() => fetchPronouns(id), {
```
Vendicated, line 47: hooks cannot be called conditionally, this will cause ugly errors

TheKodeToad: Ohh, would explain what was happening before :P

Vendicated (approved):

PR #873, opened 2023-04-11
UserVoiceShow: Properly fix lack of bottom padding for popout on current user

LordEliasTM:
Proper fix, instead of https://github.com/Vendicated/Vencord/pull/853


LordEliasTM:
merge if cute :3

SaeidNH:
Also, would be nice to see an icon in front of users when they are in a voice channel. When hover it shows what voice channel and server. and when clicked, go to the voice channel (not join)

PR #876, opened 2023-04-12
Option to transform emotes/stickers in compound messages

Nuckyz:


Vendicated (approved):

Vendicated (changes requested):
Causes a second "Reply" entry in the message context menu

Vendicated (approved):

PR #892, opened 2023-04-14
feat(PlatformIndicator): add icon size setting

hapaxant:
allows adjusting of indicator icon size


Vendicated:
I'm not a big fan of adding customisation options like this, there are too many options users might want. A better solution is to use css. You can select the icons with
```css
.vc-platform-indicator svg {}
```

PR #895, opened 2023-04-14
SHC: Fix emoji rendering & allowed users/roles edge cases

Nuckyz:


src/plugins/showHiddenChannels/components/HiddenChannelLockScreen.tsx
```diff
@@ -242,9 +245,15 @@ function HiddenChannelLockScreen({ channel }: { channel: ExtendedChannel; }) {
...
-                        })}
+                            name: defaultReactionEmoji.emojiName
+                                ? EmojiParser.convertSurrogateToName(defaultReactionEmoji.emojiName)
+                                : EmojiStore.getCustomEmojiById(defaultReactionEmoji.emojiId)?.name ?? "",
+                            emojiId: defaultReactionEmoji.emojiId ?? void 0,
+                            surrogate: defaultReactionEmoji.emojiName ?? void 0,
+                            src: defaultReactionEmoji.emojiName
+                                ? EmojiUtils.getURL(defaultReactionEmoji.emojiName)
+                                : void 0
+                        }, void 0, { key: "0" })}
```
Vendicated, line 256: is this key safe? do we even need to pass it? Maybe an empty object would work

PR #922, opened 2023-04-16
messageTags: Fix duplicate entries & replies

xMufaro:
Fixed so it doesn't always make a second reply with "Invalid sub-command"

Screenshot 2023-04-16 132239

### DIFF
--- a/src/api/Commands/index.ts
+++ b/src/api/Commands/index.ts
@@ -111,6 +111,7 @@ function registerSubCommands(cmd: Command, plugin: string) {
             ...o,
             type: ApplicationCommandType.CHAT_INPUT,
             name: `${cmd.name} ${o.name}`,
+            id: o.name,
             displayName: `${cmd.name} ${o.name}`,
             subCommandPath: [{
                 name: o.name,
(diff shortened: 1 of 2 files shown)


Luna-devv:
Woops, will look into it

Luna-devv:
nvm this is a PR not a issue

Luna-devv:
Not sure how I've never noticed that before. Question, does anyone have the issue too that `/tags ...` commands are there multiple times? Cuz I do not know how to fix that 
image

somerandomcloud:
> Not sure how I've never noticed that before. Question, does anyone have the issue too that `/tags ...` commands are there multiple times? Cuz I do not know how to fix that 
> image
> 

This has existed for awhile, thought there was a fix already...

Vendicated:
that likely happens due to duplicate ids, each subcommand likely needs its own id

xMufaro:
Wouldn't converting everything to a seperate command fix it?

image

xMufaro:
i think that fixes the problem

Vendicated:
does it work fine with the name as id? discord uses negative numbers, so I just followed suite with normal commands

xMufaro:
yes it works with the name as id. I also included previously given id to the string just in case there would be some compatibility issues in the future. Without that, the subcommands would just duplicate

Vendicated:
Thank you!

PR #923, opened 2023-04-16
ShowConnections plugin

TheKodeToad:
Adds connections to the profile as clickable icons with tooltips! Inspiration: https://github.com/Vendicated/Vencord/discussions/871 and devilbro's betterdiscord plugin

Here are some screenshots: 

Bad showcase

I also added some "helpful" text

TODO: maybe add to the DM profile card thing too and fix the icon spacing a bit


TheKodeToad:
Kind of useable, I'm having trouble finding the dm widget :p

Vendicated:
>  I'm having trouble finding the dm widget :p

Did you find it? And by that, do you mean the profile bar on the right?

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+*/
+
+import "./styles.css";
+
+import { Settings } from "@api/settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy, findStoreLazy } from "@webpack";
+import { Tooltip } from "webpack/common/components";
```
Vendicated, line 26:
```suggestion
import { Tooltip } from "@webpack/common";
```

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+interface ConnectionPlatform {
+    getPlatformUserUrl(connection: Connection): string;
+    icon: { lightSVG: string, darkSVG: string; };
+}
+
+function CompactConnectionComponent({ connection, theme }: { connection: Connection, theme: string; }) {
+    const platform = platforms.get(connection.type);
+    const url = platform.getPlatformUserUrl && platform.getPlatformUserUrl(connection);
```
Vendicated, line 47:
```suggestion
    const url = platform.getPlatformUserUrl?.(connection);
```

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        },
+        {
+            find: "name:\"Twitch\"",
+            replacement: {
+                match: /const (\i)={get:/,
+                replace: "const $1=$self.platforms={get:"
+            }
+        }
+    ],
+    options: {
```
Vendicated, line 117: you should use definePluginSettings() instead of the deprecated options

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                match: /function (\i)\(e\)/,
+                replace: "$self.Section=$1;$&"
+            }
+        },
+        {
+            find: "name:\"Twitch\"",
+            replacement: {
+                match: /const (\i)={get:/,
+                replace: "const $1=$self.platforms={get:"
+            }
```
Vendicated, line 114: using a patch to obtain this is unnecessary, a simple `findByProps("isSupported", "getByUrl")` does it

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    patches: [
+        {
+            find: ".Messages.BOT_PROFILE_SLASH_COMMANDS",
+            replacement: {
+                match: /,hideNote:\i\|\|\i}\)(?<=(\i)=\i\.user,.+?)(?<=(\i)=\(0,\i\.\i\)\(\i,\i\)\.profileTheme.+?)/,
+                replace: "$&,$self.component($1.id,$2)"
+            }
+        },
+        {
+            find: "\"lastSection\",",
```
Vendicated, line 103: same as below, you can get this via `findByCode("().lastSection")`

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export default definePlugin({
+    name: "QuickConnections",
+    description: "Show connected accounts in user popouts",
+    authors: [Devs.TheKodeToad],
+    patches: [
+        {
+            find: ".Messages.BOT_PROFILE_SLASH_COMMANDS",
+            replacement: {
+                match: /,hideNote:\i\|\|\i}\)(?<=(\i)=\i\.user,.+?)(?<=(\i)=\(0,\i\.\i\)\(\i,\i\)\.profileTheme.+?)/,
```
Vendicated, line 98: this matcher is very specific, are you sure this is necessary? You seem to be matching the profile theme, why not use the function call you match manually? It takes the user and their profile, both of which are passed as an argument to the function your code is inside (accessible via arguments[0].user and argument[0].profileTheme)

Vendicated (changes requested):

src/plugins/quickConnections/index.tsx
TheKodeToad: I don't think that would account for this option
Sync profile themes

Vendicated: image

What I meant is you're matching this, but it would be way simpler to just call p.Z yourself

Vendicated: image
image

you can use a findByCode to find this function

Vendicated: This way you can make your patch way way simpler and don't have to match any variables and instead only use arguments[0]

TheKodeToad: Wouldn't it require calling it with "o" and "y"? tbh i don't really understand what's going on with (0, p.Z) as it looks like a call with no method.

TheKodeToad: Ohh, right I read it again

src/plugins/quickConnections/index.tsx
```diff
@@ -54,6 +57,32 @@ interface ConnectionPlatform {
     icon: { lightSVG: string, darkSVG: string; };
 }
 
+function profilePopoutComponent(e: any) {
+    return component(e.user.id, getTheme(e.user, e.displayProfile).profileTheme);
```
Vendicated, line 61: please properly use react:
```jsx
return <Component id={e.user.id} theme={getTheme(e.user, e.displayProfile).profileTheme} />

File: src/plugins/quickConnections/index.tsx [hunk 8]
```diff
@@ -54,6 +57,32 @@ interface ConnectionPlatform {
     icon: { lightSVG: string, darkSVG: string; };
 }
 
+function profilePopoutComponent(e: any) {
```
Vendicated, line 60: Instead of using the ErrorBoundary in the below component (where it cannot catch all errors that might happen), you should instead wrap the outer function:
```js
const profilePopoutComponent = ErrorBoundary.wrap((e: any) => {
	...
});
```

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import "./styles.css";
+
+import { definePluginSettings } from "@api/settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByCodeLazy, findByPropsLazy, findStoreLazy } from "@webpack";
+import { User } from "discord-types/general";
+import { Tooltip } from "webpack/common";
```
Vendicated, line 27:
```suggestion
import { Tooltip } from "@webpack/common";
```

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const Section: React.ComponentType<any> = findByCodeLazy("().lastSection");
```
Vendicated, line 29:
```suggestion
const Section = LazyComponent(() => findByCode("().lastSection"));
```

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+const UserProfileStore = findStoreLazy("UserProfileStore");
+const ThemeStore = findStoreLazy("ThemeStore");
+const platforms: { get(type: string): ConnectionPlatform; } = findByPropsLazy("isSupported", "getByUrl");
+const getTheme: (user: User, displayProfile: any) => any = findByCodeLazy(",\"--profile-gradient-primary-color\"");
+
+const settings = definePluginSettings({
+    iconSize: {
+        type: OptionType.NUMBER,
+        description: "Icon size",
```
Vendicated, line 38: you should specify the unit here, i assume px?

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+        default: 32
+    },
+    iconSpacing: {
+        type: OptionType.NUMBER,
+        description: "Icon spacing",
```
Vendicated, line 43: what's spacing? Do you mean padding/margin? If so, please change it to that

Vendicated: but also the average user doesn't want to manually juggle px values. Maybe it would be better to use different presets like cozy & compact

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <QuickConnectionsComponent id={e.channel.recipients[0]} theme={ThemeStore.theme} />
+);
+
+function QuickConnectionsComponent({ id, theme }: { id: string, theme: string; }) {
+    const profile = UserProfileStore.getUserProfile(id);
+    if (!profile)
+        return null;
+
+    const connections: Connection[] = profile.connectedAccounts;
+    if (connections && connections.length !== 0) {
```
Vendicated, line 74: the happy path should come last imo, so I think it would be nicer (and shorter) to write this as `if (!connections?.length) return null`

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+function CompactConnectionComponent({ connection, theme }: { connection: Connection, theme: string; }) {
+    const platform = platforms.get(connection.type);
+    const url = platform.getPlatformUserUrl?.(connection);
+
+    return (
+        <Tooltip text={connection.name + (!connection.verified ? " (unverified)" : "") + (!url ? " (copy)" : "")} key={connection.id}>
+            {({ onMouseLeave, onMouseEnter }) =>
+                <a
+                    href={url ?? "javascript:void(0)"}
```
Vendicated, line 93: i think this might throw an ugly csp error because Discord doesn't allow inline handler script, please verify if this is the case by checking the console for errors

src/plugins/quickConnections/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    target="_blank"
+                    style={{
+                        backgroundImage: "url(" + (theme === "light" ? platform.icon.lightSVG : platform.icon.darkSVG) + ")",
+                        marginTop: settings.store.iconSpacing,
+                        marginRight: settings.store.iconSpacing,
+                        width: settings.store.iconSize,
+                        height: settings.store.iconSize
+                    }}
+                    className="vc-user-connection"
+                    onClick={() => !url && navigator.clipboard.writeText(connection.name)}
```
Vendicated, line 103: this should use the clipboard api from webpack commons instead

Vendicated (changes requested):

src/plugins/quickConnections/index.tsx
TheKodeToad: Ah, probably an artifact of refactoring :P

src/plugins/quickConnections/index.tsx
TheKodeToad: I suppose I could have presets.

src/plugins/quickConnections/index.tsx
TheKodeToad: I can't see anything

Vendicated:
Your screenshots show it above roles but for me it's below the notes box for some reason

src/plugins/quickConnections/index.tsx
Vendicated: !!
image

Vendicated: i fixed

Vendicated (approved):

Vendicated:
I went ahead and renamed the plugin from QuickConnections to ShowConnections to have the same name as the BD plugin (and imo be more self explanatory)

Vendicated:
awesome thank you, very cool plugin :3

TheKodeToad:
> Your screenshots show it above roles but for me it's below the notes box for some reason

I changed it intentionally.

Vendicated:
> I changed it intentionally.

Oh I see, I'm sorry for changing it then (though I personally like it better where it is now)! :(

We could perhaps add a setting to change their location if you care?

TheKodeToad:
It works both ways! I thought it was more consistent without the heading.

PR #924, opened 2023-04-16
Improve FakeNitro emoji popup message

yodaluca23:
Changed the message that'll pop up when clicking on a fake emoji, to be shorter, to make the popup smaller. I also changed it so it doesn't sound like it will only look like actual emojis for Vencord users, instead of anyone any modded Discord using similar plugins.

### DIFF
--- a/src/plugins/fakeNitro.tsx
+++ b/src/plugins/fakeNitro.tsx
@@ -534,7 +534,7 @@ export default definePlugin({
 
         switch (type) {
             case "STICKER": {
-                node.push(" Fake Nitro emoji renders like a real emoji only for you, appears as a link to non-plugin users.");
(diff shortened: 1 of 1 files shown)


src/plugins/fakeNitro.tsx
```diff
@@ -534,12 +534,12 @@ export default definePlugin({
 
         switch (type) {
             case "STICKER": {
-                node.push(" This is a Fake Nitro sticker. Only you can see it rendered like a real one, for non Vencord users it will show as a link.");
+                node.push(" Fake Nitro emoji renders like a real emoji only for you, appears as a link to non-plugin users.");
```
Nuckyz, line 537:
```suggestion
                node.push(" Fake Nitro sticker renders like a real sticker only for you, appears as a link to non-plugin users.");
```

PR #935, opened 2023-04-17
fix(ReviewDB): fix margin in self-profile preview

amsyarasyiq:


Vendicated:
Okay thank you! I changed it to not be a managed style

PR #938, opened 2023-04-17
feat(messageTags): Ability to send a tag with a hotkey

mufaroxyz:
Basically a plugin that allows you to bind a text to a specified key on your keyboard (macro) so you can send it anytime you want with a single press of a key

Bindable keys are hard coded since some of them were creating an unexpected behaviour


ArjixWasTaken:
Would be awesome if this was combined with [messageTags](https://github.com/Vendicated/Vencord/blob/main/src/plugins/messageTags.ts) instead.

Not much is added other than the keyboard shortcuts.

xMufaro:
> Would be awesome if this was combined with [messageTags](https://github.com/Vendicated/Vencord/blob/main/src/plugins/messageTags.ts) instead.
> 
> Not much is added other than the keyboard shortcuts.

Yeah that sounds well, I think I'll do that later this day

> PS: Adding a popup that opens with a keybind, that allows you to search for and select a saved message would be nice as well, maybe on another PR?

I could work on that as well

xMufaro:
it's combined now

PR #944, opened 2023-04-18
RelationShipNotifier: Support multiple users

Vendicated:


src/plugins/relationshipNotifier/utils.ts
```diff
@@ -30,11 +30,20 @@ const friends = {
     requests: [] as string[]
 };
 
+const guildsKey = () => `relationship-notifier-guilds-${UserStore.getCurrentUser().id}`;
+const groupsKey = () => `relationship-notifier-groups-${UserStore.getCurrentUser().id}`;
+const friendsKey = () => `relationship-notifier-friends-${UserStore.getCurrentUser().id}`;
+
+async function runMigrations() {
+    DataStore.delMany(["relationship-notifier-guilds", "relationship-notifier-groups", "relationship-notifier-friends"]);
```
Vendicated, line 38: I was originally gonna migrate these but figured it wouldn't be worth the trouble and also not even guaranteed they're from the current user

PR #947, opened 2023-04-18
feat(plugin): ImplicitRelationships

dolfies:
### Summary
Adds a section to the Friends tab that shows your implicit relationships.

### Limitations
~~The tab will be empty until implicit relationships are calculated, after which you have to reopen the Friends tab to see them. No idea how to go about fixing this.~~ Fixed now thanks to Arjix!


src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,132 @@
...
+        // Piggyback relationship fetch
+        {
+            find: "FriendsEmptyState: Invalid empty state",
+            replacement: {
+                match: /([^;]+)\.([^;]+)\.fetchRelationships\(\)/,
+                // This relationship fetch is actually completely useless, but whatevs
+                replace: "$1.$2.fetchRelationships(),Vencord.Plugins.plugins.ImplicitRelationships.fetchImplicitRelationships()"
+            },
+        },
+    ],
```
ArjixWasTaken, line 72: Since all the patches use the same `find`, that means they are in the same file.
One optimization is to have one patch with multiple replacements.

Yes it is valid, the type is [here](https://github.com/Vendicated/Vencord/blob/main/src/utils/types.ts#L38).

TL;DR you can do

```js
{
    find: "...",
    replacement: [
        {  },
        {  }
    ]
}
```

ArjixWasTaken (commented):
looks good to me, just one small thing

src/plugins/implicitRelationships.ts
dolfies: Didn't know this was possible, thanks!

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,122 @@
...
+                // Sections content
+                {
+                    match: /case ([^;]+)\.([^;]+)\.BLOCKED:return ([^;]+)\.type===([^;]+)\.([^;]+)\.BLOCKED/,
+                    replace: "case $1.$2.BLOCKED:return $3.type===$4.$5.BLOCKED;case $1.$2.IMPLICIT:return $3.type===5"
+                },
+                // Piggyback relationship fetch
+                {
+                    match: /([^;]+)\.([^;]+)\.fetchRelationships\(\)/,
+                    // This relationship fetch is actually completely useless, but whatevs
+                    replace: "$1.$2.fetchRelationships(),Vencord.Plugins.plugins.ImplicitRelationships.fetchImplicitRelationships()"
```
ArjixWasTaken, line 58:
```suggestion
                    replace: "$1.$2.fetchRelationships(),$self.fetchImplicitRelationships()"
```

you should use `$self` instead of the full path

ArjixWasTaken (commented):
oops, one more thing!

src/plugins/implicitRelationships.ts
dolfies: Another thing I didn't know existed, thanks again!

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,122 @@
...
+    name: "ImplicitRelationships",
+    description: "Shows your implicit relationships in the Friends tab.",
+    authors: [Devs.Dolfies],
+    patches: [
+        {
+            find: "FriendsEmptyState: Invalid empty state",
+            replacement: [
+                // Counts header
+                {
+                    match: /toString\(\)\}\);case ([^;]+)\.([^;]+)\.BLOCKED/,
```
Vendicated, line 36: are these identifiers? aka `case foo.bar.BLOCKED`

If so:
```suggestion
                    match: /toString\(\)\}\);case (\i\.\i\.BLOCKED)/,
```

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,122 @@
...
+                    replace: "toString()});case $1.$2.IMPLICIT:return \"Implicit — \"+t.toString();case $1.$2.BLOCKED"
```
Vendicated, line 37: don't hardcode minified variable names like "t" here

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,122 @@
...
+                    // This relationship fetch is actually completely useless, but whatevs
+                    replace: "$1.$2.fetchRelationships(),$self.fetchImplicitRelationships()"
+                },
+            ],
+        },
+    ],
+
+    hasDM(userId: string): boolean {
+        const privateChannels = ChannelStore.getSortedPrivateChannels();
+        return privateChannels.some(channel => channel.recipients.includes(userId));
```
Vendicated, line 66: avoid pointless variables
```suggestion
        return ChannelStore.getSortedPrivateChannels().some(channel => channel.recipients.includes(userId));
```

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+
+    async fetchImplicitRelationships() {
+        // Implicit relationships are defined as users that you:
+        // 1. Have an affinity for
+        // 2. Do not have a relationship with // TODO: Check how this works with pending/blocked relationships
+        // 3. Have a mutual guild with
+        const userAffinities: Set<string> = UserAffinitiesStore.getUserAffinitiesUserIds();
+        const nonFriendAffinities = Array.from(userAffinities).filter(
+            id => !RelationshipStore.getRelationshipType(id)
```
Vendicated, line 76: shouldn't this be
```suggestion
            id => RelationshipStore.getRelationshipType(id) === 5 /* IMPLICIT */
```
image

Vendicated (changes requested):

src/plugins/implicitRelationships.ts
dolfies: No, the whole point of this function is to figure out what users Discord treats as implicit relationships and insert them into cache. `nonFriendAffinities` is just user affinities that you do not have an existing *real* relationship with.

Implicit relationships aren't actually stored as relationships on Discord's end, so they aren't sent in READY or normal `GET /users/@me/relationships` calls. Implicit relationships themselves are a GameSDK artifact that was kept alive when they added the `relationships.read` OAuth2 scope (if you `GET /users/@me/relationships?with_implicit=true` with a bearer token with said scope, Discord calculates your implicit relationships and returns them as type 5).

The client backfills type 5 relationships when it detects them from PRESENCE_UPDATE events, and also caches all of them when the GameSDK's GET_RELATIONSHIPS is called (though it has the logic slightly wrong 😭).

src/plugins/implicitRelationships.ts
dolfies: Oopsies, forgot to change that after testing

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,121 @@
...
+    description: "Shows your implicit relationships in the Friends tab.",
+    authors: [Devs.Dolfies],
+    patches: [
+        {
+            find: "FriendsEmptyState: Invalid empty state",
+            replacement: [
+                // Counts header
+                {
+                    match: /toString\(\)\}\);case (\i\.\i)\.BLOCKED/,
+                    replace: "toString()});case $1.IMPLICIT:return \"Implicit — \"+arguments[1].toString();case $1.BLOCKED"
```
Vendicated, line 37:
```suggestion
                    replace: 'toString()});case $1.IMPLICIT:return "Implicit — "+arguments[1];case $1.BLOCKED'
```

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,121 @@
...
+                    replace: "case $1.ONLINE:case $1.IMPLICIT:return $2.SECTION_ONLINE"
+                },
+                // Sections header
+                {
+                    match: /\(0,(\i)\.jsx\)\((\i)\.TabBar\.Item,\{id:(\i)\.(\i)\.BLOCKED,className:([^\s]+?)\.item,children:\i\.\i\.Messages\.BLOCKED\}\)/,
+                    replace: "(0,$1.jsx)($2.TabBar.Item,{id:$3.$4.IMPLICIT,className:$5.item,children:\"Implicit\"}),$&"
+                },
+                // Sections content
+                {
+                    match: /case (\i\.\i)\.BLOCKED:return (\i)\.type===(\i\.\i)\.BLOCKED/,
```
Vendicated, line 51: using lookarounds, you can avoid just replacing Discord's code with itself
```suggestion
                    match: /(?<=case (\i\.\i)\.BLOCKED:return (\i)\.type===\i\.\i\.BLOCKED)/,
```

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,121 @@
...
+                    replace: "case $1.BLOCKED:return $2.type===$3.BLOCKED;case $1.IMPLICIT:return $2.type===5"
```
Vendicated, line 52: now you don't need any of their jazz
```suggestion
                    replace: ";case $1.IMPLICIT:return $2.type===5"
```

Vendicated, line 52: now you don't need any of their jazz
```suggestion
                    replace: ";case $1.IMPLICIT:return $2.type===5"
```

Vendicated (changes requested):

src/plugins/implicitRelationships.ts
dolfies: one day I will learn how these work 🙏

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,121 @@
...
+                    match: /(\i\.\i)\.fetchRelationships\(\)/,
+                    // This relationship fetch is actually completely useless, but whatevs
+                    replace: "$1.fetchRelationships(),$self.fetchImplicitRelationships()"
+                },
+            ],
+        },
+    ],
+
+    hasDM(userId: string): boolean {
+        return ChannelStore.getSortedPrivateChannels().some(channel => channel.recipients.includes(userId));
```
Vendicated, line 65:
```suggestion
        return userId in ChannelStore.getMutablePrivateChannels();
```

dolfies: I don't think this accounts for group DMs

Vendicated: it does! getSortedPrivateChannels just does Object.values(mutablePrivateChannels).sort().reverse()

Vendicated: my bad, i didn#t read the original code correctly
```suggestion
        return Object.values(ChannelStore.getSortedPrivateChannels()).some(channel => channel.recipients.includes(userId));
```

TheKodeToad:
> ### Limitations
> 
> The tab will be empty until implicit relationships are calculated, after which you have to reopen the Friends tab to see them. No idea how to go about fixing this.

Maybe with `useState` or `useForceUpdater` hackily put in a global?

dolfies:
> > ### Limitations
> > The tab will be empty until implicit relationships are calculated, after which you have to reopen the Friends tab to see them. No idea how to go about fixing this.
> 
> Maybe with `useState` or `useForceUpdater` possibly hackily put in a global?

No idea how to do that :)

ArjixWasTaken:
> > ### Limitations
> > The tab will be empty until implicit relationships are calculated, after which you have to reopen the Friends tab to see them. No idea how to go about fixing this.
> 
> Maybe with `useState` or `useForceUpdater` possibly hackily put in a global?

Tried it, doesn't work.
It clearly re-renders, but I think it has a cache of the data to render.

TheKodeToad:
Maybe make the data use `useState`?

ArjixWasTaken:
> Maybe make the data use `useState`?

please no more suggestions, they are dumb, I'm already investigating this

src/plugins/implicitRelationships.ts
```diff
@@ -0,0 +1,121 @@
...
+                userIds: toRequest.slice(i, i + 100),
+            });
+        }
+        for (let i = 0; i < 50 && count > 0; i++) {
+            await new Promise(r => setTimeout(r, 100));
+        }
+
+        const implicitRelationships = nonFriendAffinities.map(id => UserStore.getUser(id)).filter(user => user && !ignore.has(user.id));
+        const { relationships } = RelationshipStore.__getLocalVars();
+        implicitRelationships.forEach(user => relationships[user.id] = 5);
```
ArjixWasTaken, line 114:
```suggestion
        implicitRelationships.forEach(user => relationships[user.id] = 5);
        
        RelationshipStore.emitChange();
```

ArjixWasTaken: 🙏

ArjixWasTaken:
> Maybe make the data use `useState`?

Why did I call this dumb?
Well, this plugin does not inject a component or anything really.
It does not have its own data to render.

It inserts a relationship type, and users with that type, into an already existing component.
It then adds a label so that you can view that relationship type.

All the rendering and stuff is done by discord's code.

Event w/o that in mind, `useState` is not different in behavior from `useForceUpdater` when it comes to triggering a re-render of the component.

The issue was that discord listens for changes in the RelationshipStore with `useStateFromStores`, and the state was directly modified w/o React being aware of it.

So the `useStateFromStores` did not find a change.
Calling `RelationshipStore.emitChange();` is the fix, as it signals that the store has a new state.

src/plugins/implicitRelationships.ts
dolfies: very nice ty

src/plugins/implicitRelationships/index.ts
```diff
@@ -0,0 +1,187 @@
...
+            replacement: {
+                match: /toString\(\)\}\);case (\i\.\i)\.BLOCKED/,
+                replace: 'toString()});case $1.IMPLICIT:return "Implicit — "+arguments[1];case $1.BLOCKED'
+            },
+        },
+        // No friends page
+        {
+            find: "FriendsEmptyState: Invalid empty state",
+            replacement: {
+                match: /case (\i\.\i)\.ONLINE:return (\i)\.SECTION_ONLINE/,
```
Vendicated, line 51: if you don't actually need that part for anything, you can just slap it in a lookaround so you dont match it
```suggestion
                match: /case (\i\.\i)\.ONLINE:(?=return (\i)\.SECTION_ONLINE)/,
```

src/plugins/implicitRelationships/index.ts
```diff
@@ -0,0 +1,187 @@
...
+                replace: "case $1.ONLINE:case $1.IMPLICIT:return $2.SECTION_ONLINE"
```
Vendicated, line 52:
```suggestion
                replace: "case $1.ONLINE:case $1.IMPLICIT:"
```

src/plugins/implicitRelationships/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        }
+    ),
+
+    getAffinity(user: User): number {
+        const affinities: UserAffinity[] = UserAffinitiesStore.getUserAffinities();
+        return affinities.find(affinity => affinity.user_id === user.id)?.affinity ?? 0;
+    },
+
+    hasDM(userId: string): boolean {
+        return Object.values(ChannelStore.getSortedPrivateChannels()).some(channel => channel.recipients.includes(userId));
```
Vendicated, line 134: i see that you're calling this from an array loop and this is also a loop, so this could potentially add up and do a lot of looping

you could do it way more performant by doing something like
```js
const dmUserIds = new Set(
	Object.values(ChannelStore.getSortedPrivateChannels()).flatMap(c => c.recipients)
);

const toRequest = nonFriendAffinities.filter(id => !UserStore.getUser(id) || dmUserIds.has(id));
```

dolfies: fixed!

src/plugins/implicitRelationships/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        // Piggyback relationship fetch
+        {
+            find: ".fetchRelationships()",
+            replacement: {
+                match: /(\i\.\i)\.fetchRelationships\(\)/,
+                // This relationship fetch is actually completely useless, but whatevs
+                replace: "$1.fetchRelationships(),$self.fetchImplicitRelationships()"
+            },
+        },
+        // Modify sort -- thanks megu for the patch (from sortFriendRequests)
```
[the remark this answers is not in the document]
dolfies: they are now!!

src/plugins/implicitRelationships/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    sortList(row: any) {
+        return row.type === 5
+            ? -this.getAffinity(row.user)
+            : row.comparator;
+    }
+
+    getAffinity(user: User): number {
+        const affinities: UserAffinity[] = UserAffinitiesStore.getUserAffinities();
+        return affinities.find(affinity => affinity.user_id === user.id)?.affinity ?? 0;
```
Vendicated, line 132: can't you just use `findStore("UserAffinitiesStore").getUserAffinity(user.id)`? i dont have any affinities so cant test but seems to be what u want :p

dolfies: i swear that's new lol

Vendicated:
thank you!! yet another tab to show me i have no friends :blobcatcozy:

PR #958, opened 2023-04-21
Plugin: Activities

ArjixWasTaken:
This plugin will basically be a combination of [ActivityIcons](https://betterdiscord.app/plugin/ActivityIcons) and [ShowAllActivities](https://betterdiscord.app/plugin/ShowAllActivities).

Why a combination and not two plugins?
Well...that is just unnecessary, when we can just have a setting to toggle one feature on/off.

## Screenshots

- ActivityIcons
image
image
image

- ShowAllActivities
image


ArjixWasTaken:
The omega in question
image

:wheeze:

Vendicated:
I like ActivityIcons! But what's the point of ShowAllActivities? this info is already available from the Activities tab in the profile. Is that just for the mini popout?

Vendicated:
Why not make these separate plugins like the BD plugin? I think i would prefer that

If it's about reusing state, you can import files from other plugins just fine

ArjixWasTaken:
> I like ActivityIcons! But what's the point of ShowAllActivities? this info is already available from the Activities tab in the profile. Is that just for the mini popout?

yeah it is just for the mini popout

ArjixWasTaken:
> Why not make these separate plugins like the BD plugin? I think i would prefer that
> 
> If it's about reusing state, you can import files from other plugins just fine

sure thing, in one PR or two?

AutumnVN:
About ActivityIcons:

1. It shouldn't show the rpc icon at first. Example if user is only listening to spotify, shows only the headphone icon, not rpc icon & headphone icon.

2. Use rpc icon for rpc activity instead of controller icon.

3. If user are only playing a detected game, it should show controller icon instead of nothing. (Currently it only shows that when you have another activity at the same time)

ArjixWasTaken:
> It shouldn't show the rpc icon at first. Example if user is only listening to spotify, shows only the headphone icon, not rpc icon & headphone icon.

Sure

> Use rpc icon for rpc activity instead of controller icon.

Huh?
I never used the controller icon for rpc activity.
I only use it when an activity is of the type "playing".

Don't blame my plugin if people have "playing" as the activity type in their rich presence.

> If user are only playing a detected game, it should show controller icon instead of nothing. (Currently it only shows that when you have another activity at the same time)

I can't replicate what you are saying.

ArjixWasTaken:
PS: Doesn't work well with nitro profiles that are in white mode.

PR #966, opened 2023-04-22
fakeProfileThemes: settings UI improvements

lexisother:
As it says on the tin.


src/plugins/fakeProfileThemes.tsx
```diff
@@ -16,28 +16,27 @@
...
-import { Margins } from "@utils/margins";
 import { copyWithToast } from "@utils/misc";
 import definePlugin, { OptionType } from "@utils/types";
-import { Button, Forms } from "@webpack/common";
+import { findByPropsLazy } from "@webpack";
+import { Button, Flex, Forms, React, Text, UserStore } from "@webpack/common";
 import { User } from "discord-types/general";
 import virtualMerge from "virtual-merge";
 
+const UserProfileStore = findByPropsLazy("getUserProfile");
```
Vendicated, line 30:
```suggestion
const UserProfileStore = findStoreLazy("UserProfileStore");
```

src/plugins/fakeProfileThemes.tsx
```diff
@@ -77,44 +76,149 @@ const settings = definePluginSettings({
...
+                            }}
+                        />
+                        <Button
+                            onClick={() => {
+                                const colorString = encode(c1, c2);
+                                copyWithToast(colorString);
+                            }}
+                            color={Button.Colors.PRIMARY}
+                            size={Button.Sizes.XLARGE}
+                            // Why the fuck is this needed again?
```
Vendicated, line 191: ?

src/plugins/fakeProfileThemes.tsx
```diff
@@ -77,44 +76,149 @@ const settings = definePluginSettings({
...
+                        >
+                            Copy 3y3
+                        </Button>
+                    </Flex>
+                    <Forms.FormDivider
+                        style={{ marginTop: "8px", marginBottom: "8px" }}
+                    />
+                    <Forms.FormTitle tag="h3">Preview</Forms.FormTitle>
+                    <div
+                        onClick={e => {
```
Vendicated, line 202: i'm fairly certain you can achieve the same unclickability with
```jsx
style={{ pointerEvents: "none" }}
```

lexisother: No.

src/plugins/fakeProfileThemes.tsx
lexisother: Related to below `style` prop.

Vendicated: which is entirely unnecessary, hence why I'm confused

src/plugins/fakeProfileThemes.tsx
Vendicated: the banner and pfp are still clickable even with your solution. I think this is because clicks start in the deepest child then propagate up instead of it being the other way round

I think the only proper solution would be
```html
<div class="vc-fpt-preview">
	<Profile />
</div>

<style>
	.vc-fpt-preview * {
		pointer-events: none;
	}
</style>
```

src/plugins/fakeProfileThemes.tsx
lexisother: Not unnecessary. Solution doesn't work but I prefer text in buttons to not be selectable.

Vendicated: it's not selectable tho, you shouldn't need that

lexisother: Is on my end, fresh install. The old button before my PR was selectable as well.

Vendicated: i think it might have been a while ago, but fixed in https://github.com/Vendicated/Vencord/commit/b0caa6f4db7b6219f0a07dfb5089ba7767070328#diff-973dffee2779a8d441cedda9210ea3ae6f6a7f8a7b28825e5e3988d7c3c81192R62

your PR might not have that change, but since I just updated it, try pulling and see if it works without that css then

Vendicated:
very outdated now, my apologies for never getting to this :(

thank you a lot regardless!

lexisother:
...can't I just update it? I'll have to figure out what broke again but I don't see the problem with that.

Vendicated:
sure, if you'd like to! i figured you lost interest due to the long wait

lexisother:
I'm always up to fix up PRs if they regain attention. I'll see if I can get to it tomorrow. :)

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -101,21 +132,98 @@ export default definePlugin({
...
-                    <li>• go to your profile settings</li>
-                    <li>• choose your own colors in the Nitro preview</li>
-                    <li>• click the "Copy 3y3" button</li>
-                    <li>• paste the invisible text anywhere in your bio</li>
-                </ul><br />
-                <b>Please note:</b> if you are using a theme which hides nitro ads, you should disable it temporarily to set colors.
-            </Forms.FormText>
-        </Forms.FormSection>),
+    start: async () => {
+        await requireColorPicker();
```
Vendicated, line 136: i don't like this... this will unnecessarily load this chunk even if users never enter the settings

i would instead suggest creating a component like the following:
```jsx
function ColorPickers() {
	const [,,loadingColorPickerChunk] = useAwaiter(requireColorPicker);
	
	if (loadingColorPickerChunk)
		return <Forms.FormText>Loading Colorpicker chunk...</Forms.FormText>;
	
	...
}
```

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -101,21 +132,98 @@ export default definePlugin({
...
+    },
+    settingsAboutComponent: () => {
+        const existingColors = decode(
+            UserProfileStore.getUserProfile(UserStore.getCurrentUser().id).bio
+        ) ?? [0, 0];
+        const [c1, setc1] = React.useState(existingColors[0]);
```
Vendicated, line 142:
```suggestion
        const [color1, setColor1] = useState(existingColors[0]);
```

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -101,21 +132,98 @@ export default definePlugin({
...
+                    <br />
+                    To set your own colors:
+                    <ul>
+                        <li>
+                            • use the color pickers below to choose your colors
+                        </li>
+                        <li>• click the "Copy 3y3" button</li>
+                        <li>• paste the invisible text anywhere in your bio</li>
+                    </ul><br />
+                    <b>Please note:</b> if you are using a theme which hides nitro ads, you should disable it temporarily to set colors.
```
Vendicated, line 160: maybe instead of having this note, you could do something like the following to override nitro themes?

```css
.vc-fpt-color-pickers * {
	display: unset !important;
}
```

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -101,21 +132,98 @@ export default definePlugin({
...
+                    <Forms.FormDivider
+                        style={{ marginTop: "8px", marginBottom: "8px" }}
```
Vendicated, line 162:
```suggestion
                        className={classes(Margins.top8, Margins.bottom8)}
```

Vendicated (commented):
thanks for reviving this! looks good for the most part, just a few minor things

src/plugins/fakeProfileThemes/index.tsx
lexisother: Not sure what you mean here. This breaks the color pickers:
image

lexisother: Frankly, I don't remember why I wrote this... I don't know which "anti nitro upsell" theme/plugin I was referencing here and why it's even relevant, because I'm not doing anything upsell related and those themes don't exactly brick the functionality of the plugin either. Unless they force profile colours off, at which point I don't really see it as my problem anymore. (why set your profile colours if you're so against them?)

Vendicated: im pretty sure it's because they fully `display: none` this section which also hides our colour pickers

image

that's why i meant you could see which element they hide and force `display: block/grid/flex/...` it. hope you understand what i'm trying to say?

lexisother: They hide that *section*, but the color pickers have nothing to do with the nitro stuff, as they're also used in the server settings to set a role's colour. If those themes were to hide colour pickers, they'd shoot themselves in the foot, because they'd be hiding the ones in the role settings too.

Vendicated: alright got it, thank you!

Vendicated:
thank u!!

PR #967, opened 2023-04-22
Plugin: ReturnUnicodeURLs

EinTim23:
Recently discord pushed a new update to the discord app which urlencodes unicode urls before displaying them which destroys a lot of the custom url types of image uploading services like upload.systems. This plugin reverts this change(only visually, the link that is opened in the browser gets encoded as it should be).

Without plugin:
grafik

With plugin(and how it was before the discord app update):
grafik

### DIFF
--- a/src/plugins/returnUnicodeURLs.ts
+++ b/src/plugins/returnUnicodeURLs.ts
@@ -1,0 +1,30 @@
+import definePlugin from "@utils/types";
+export default definePlugin({
+    name: "ReturnUnicodeURLs",
+    description: "This plugin brings back the old rendering of URLs which allows proper displaying of unicode URLs",
+    authors: [
+        {
+            id: 1011864559168012301n,
+            name: "EinTim",
+        },
+    ],
+    start() { 
+        globalThis.originalCreateElement = document.createElement;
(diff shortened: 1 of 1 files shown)


Vendicated:
this is a good update by discord actually

Vendicated:
Monkey patching dom methods is also an extremely poor way of doing this, please don't

Vendicated:
this plugin also has a XSS vulnerability, for example the following link will execute alert("oopsie") (but it will thankfully be blocked by Discord's content security policy

`https://banana.com/%3Cimg%20src="a"%20onerror="alert('oopsie')"%3E`

ArjixWasTaken:
its only an XSS vuln if `.innerHTML` is used instead of `.innerText`

correct me if I'm wrong

ArjixWasTaken:
can we please revive this?
(ofc w/o the `.innerHTML` bit)

sending non-latin characters is annoying as the url gets huge after the unicode escape

ArjixWasTaken:
`https://google.com/search?q=γογλε` is escaped to `https://www.google.com/search?q=%CE%B3%CE%BF%CE%B3%CE%BB%CE%B5`

Issue #868, opened 2023-04-11
[Bug] Clicking on WhoReacted avatars also adds reaction


ryanccn:
### Discord Account

RyanCaoDev#0428

### What happens when the bug or crash occurs?

When you click on an avatar added by the WhoReacted plugin to see the user popout, it also triggers the click and adds a reaction for yourself.

### What is the expected behaviour?

The expected behavior is probably to only add the reaction when you click in the box somewhere that isn't a user avatar.

### How do you recreate this bug or crash?

1. Use the WhoReacted plugin
2. Click an avatar in a reaction

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well

KannaDev:
Made a pull request [here](https://github.com/Vendicated/Vencord/pull/977) that solves this issue.

ryanccn:
> Made a pull request [here](https://github.com/Vendicated/Vencord/pull/977) that solves this issue.

Thanks!

PR #977, opened 2023-04-23
WhoReacted: Fix clicking user pfps also adding reaction

KannaDev:
Solves the [following issue](https://github.com/Vendicated/Vencord/issues/868)

Have linted and tested on Discord Canary, has no more issues.


Vendicated:
doesn't this make the individual avatars not clickable?

KannaDev:
Clicking on the avatars opens the profile without it reacting, and reacting works fine if not clicking on the avatars. So not entirely sure what your question is, apologies.

PR #979, opened 2023-04-24
MuteNewGuilds: add toggles for guild, everyone, and roles.

carince:
Gives the user the ability to toggle which notification should be suppressed or not.


src/plugins/muteNewGuild.tsx
```diff
@@ -49,10 +50,28 @@ function NoDMNotificationsModal({ modalProps }: { modalProps: ModalProps; }) {
     );
 }
 
+const settings = definePluginSettings({
+    guild: {
+        description: "Suppress Guild",
```
Vendicated, line 55:
```suggestion
        description: "Mute Guild",
```

Vendicated (approved):
looks good other than that small nit

PR #994, opened 2023-04-27
feat(plugin): TextReplace

AutumnVN:
reopen of #540


Vendicated:
- As mentioned on discord, please fix the alignment image
- Adding a replacement for "ass" will also replace "bass". It would be good to have a way to make it only replace full words (this is super easy with [regex \b](https://www.regular-expressions.info/wordboundaries.html))
- Why do regex replacements not support "only if includes"? You can do it with lookarounds but that's very ugly and not many people will know about it
- Imo it should automatically delete multiple empty rules, currently if you empty all rules it will have multiple empty rules

Vendicated:
other than that, looks good!

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,278 @@
...
+                }
+            ];
+            await DataStore.set("TextReplace_rulesRegex", rulesRegex);
+        }
+        this.preSend = addPreSendListener((_, msg) => {
+            msg.content = " " + msg.content + " ";
+            for (const rule of rulesString) {
+                if (!rule.find || !rule.replace) continue;
+                if (rule.onlyIfIncludes && !msg.content.includes(rule.onlyIfIncludes)) continue;
+                msg.content = msg.content.replace(new RegExp(this.escapeRegex(rule.find), "g"), rule.replace);
```
Vendicated, line 260: why do you turn it into a regex? just use msg.content.replaceAll(rule.find, rule.replace)

Vendicated:
maybe hide the delete button the last (blank) entry cause u can't delete it anyway

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+        }
+        await DataStore.set("TextReplace_rulesString", rulesString);
+        update();
+    }
+
+    return (
+        <>
+            <h2 style={{
+                color: "#fff"
+            }}>Using string</h2>
```
Vendicated, line 85: this is bad because you're harcoding white. What if the user is using light theme? Just use Discord's Form components
```jsx
<Forms.FormTitle tag="h2">Using String</Forms.FormTitle>
```

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+                            </td>
+                            <Button
+                                size={Button.Sizes.MIN}
+                                onClick={() => onClickRemoveString(index)}
+                                style={{
+                                    background: "none",
+                                    bottom: "-7px"
+                                }}
+                            >
+                                <svg width="24" height="24" viewBox="0 0 24 24">
```
Vendicated, line 122: Currently people using a screenreader cannot tell what this button does, you should add a title to the svg 
```suggestion
                                <svg width="24" height="24" viewBox="0 0 24 24">
                                	<title>Delete Rule</title>
```

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+                                    <path fill="#f04747" d="M15 3.999V2H9V3.999H3V5.999H21V3.999H15Z"></path>
+                                    <path fill="#f04747" d="M5 6.99902V18.999C5 20.101 5.897 20.999 7 20.999H17C18.103 20.999 19 20.101 19 18.999V6.99902H5ZM11 17H9V11H11V17ZM15 17H13V11H15V17Z"></path>
+                                </svg>
+                            </Button>
+                        </tr>
+                    )
+                }
+            </table>
```
Vendicated, line 210: this is almost the same as above, you should extract it into a component to not duplicate the code

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+export default definePlugin({
+    name: "TextReplace",
+    description: "Replace text in your messages",
+    authors: [Devs.Samu, Devs.AutumnVN],
+    dependencies: ["MessageEventsAPI"],
+
+    settings,
+
+    stringToRegex(str: string) {
+        const m = str.match(/^([/~@;%#'])(.*?)\1([gimsuy]*)$/);
```
Vendicated, line 224: no idea what this does, please add a comment

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+        return m ? new RegExp(m[2], m[3]) : new RegExp(str);
+    },
+
+    async start() {
+        rulesString = await DataStore.get("TextReplace_rulesString");
```
Vendicated, line 229: since you're repeating these strings multiple times, it would be best to move them to a constant and use that to make sure the same string is used everywhere

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,274 @@
...
+import definePlugin, { OptionType } from "@utils/types";
+import { Button, TextInput } from "@webpack/common";
+
+let rulesString = [
+    {
+        find: "",
+        replace: "",
+        onlyIfIncludes: ""
+    }
+] as any;
```
Vendicated, line 33: you should make a type for these instead of using any
```ts
type Rules = Record<"find" | "replace" | "onlyIfIncludes", string>[];
```

Vendicated (changes requested):

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,209 @@
...
+                                    size={Button.Sizes.MIN}
+                                    onClick={() => onClickRemove(index)}
+                                    style={{
+                                        background: "none",
+                                        bottom: "-7px"
+                                    }}
+                                >
+                                    <svg width="24" height="24" viewBox="0 0 24 24">
+                                        <title>Delete Rule</title>
+                                        <path fill="#f04747" d="M15 3.999V2H9V3.999H3V5.999H21V3.999H15Z"></path>
```
Vendicated, line 106:
```suggestion
                                        <path fill="var(--status-danger)" d="M15 3.999V2H9V3.999H3V5.999H21V3.999H15Z"></path>
```

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,209 @@
...
+                                        <path fill="#f04747" d="M5 6.99902V18.999C5 20.101 5.897 20.999 7 20.999H17C18.103 20.999 19 20.101 19 18.999V6.99902H5ZM11 17H9V11H11V17ZM15 17H13V11H15V17Z"></path>
```
Vendicated, line 107:
```suggestion
                                        <path fill="var(--status-danger)" d="M5 6.99902V18.999C5 20.101 5.897 20.999 7 20.999H17C18.103 20.999 19 20.101 19 18.999V6.99902H5ZM11 17H9V11H11V17ZM15 17H13V11H15V17Z"></path>
```

Vendicated (changes requested):

src/plugins/textReplace.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    settings,
+
+    async start() {
+        stringRules = await DataStore.get(STRING_RULES_KEY) ?? makeEmptyRuleArray();
+        regexRules = await DataStore.get(REGEX_RULES_KEY) ?? makeEmptyRuleArray();
+
+        this.preSend = addPreSendListener((_, msg) => {
+            // pad so that rules can use " word " to only match whole "word"
+            msg.content = " " + msg.content + " ";
```
Vendicated, line 197: this makes every message shift to the left shortly after sending, you should trim at the end

Vendicated:
I'm still not sure if adding javascript rules is a good idea, seems like it could lead to security issues somehow

Vendicated (approved):

PR #999, opened 2023-04-28
betterFolders: option to close all folders when one server is selected

Redundakitties:
kind of jank animation for some reason probably my fault. thank you nookies for basically doing this for me :)
works well if you have the floating folders css where folders float over the rest of the window. 
https://github.com/Redundakitties/quickCSS-snippets/blob/master/snippets/floatingfolders.css

OR 

if you turn on dedicated sidebar for folders + no animation in the Betterfolders plugin


Vendicated:
This is cool but the very janky animation is kiond of terrible, we should get that fixed before merge

Vendicated (changes requested):
j

PR #1023, opened 2023-05-01
MessageEvents: Support sendMessage with attachments

Vendicated:


Vendicated:
marking this as draft because I want to update the types before merge to add uploads and some more stuff, but it's already ready, only ambient changes necessary

PR #1031, opened 2023-05-02
Plugin/friend notifications

axu5:
I have worked on a plugin called Friend Notifications, it's like the steam feature that when a friend comes online you get a little pop up that tells you they came online.

Similar (but not exactly) to request 10 in discussion https://github.com/Vendicated/Vencord/discussions/653.

Details
- A friend comes online, notifications should come up (on by default)
- A friend goes offline (this should be turned off by default as it could be annoying, but the option should be there)
- Add users/ remove users to track (I don't care about some people)
  - Context menu when right clicking a user to quickly add to/ remove from this "tracked" list
image
  - Settings menu to quickly see everyone you are tracking
image


src/plugins/friendNotifications/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    // invoked at all.  The presence of the key in the object alone is
+    // enough to trigger this behavior, even if the value is an empty array.
+    patches: [],
+
+    // Delete these two below if you are only using code patches
+    async start() {
+        await init();
+        addContextMenuPatch("user-context", UserContext);
+
+        FluxDispatcher.subscribe("PRESENCE_UPDATES", presenceUpdate);
```
ActuallyTheSun, line 53: You should use the `flux` property within the plugin schema to subscribe to events

```ts
export default definePlugin({
    // ...
    flux: {
        PRESENCE_UPDATES: presenceUpdate
    }
```

src/plugins/friendNotifications/settings.tsx
```diff
@@ -0,0 +1,114 @@
...
+        description: "Notifies you when a friend comes online",
+        default: true
+    },
+    tracking: {
+        type: OptionType.COMPONENT,
+        description: "People that should be tracked",
+        component: () => {
+            // Use useState to force a re-render on update to underlying data
+            // TODO find something more elegant
+            const [_count, setCount] = useState(0);
```
ActuallyTheSun, line 48: Use the `useForceUpdater` hook from utils

src/plugins/friendNotifications/settings.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+            if (users.length === 0) {
+                return <span style={{
+                    color: "var(--text-muted)"
+                }}>You don't have anyone added to your friend notifications</span>;
+            }
+
+            return <div> {
+                users.map(user => {
+                    return <div key={user.id} style={{
```
ActuallyTheSun, line 74: Using a .css file would be preferred over style tags. You can make a css file in the plugin's directory, import it here (only needs to be imported once per plugin) and it should work fine

axu5:
What should I do with the patches? For now I just removed them but I'm not exactly sure what it does/ or how to make them "robust."

Vendicated:
if your plugin doesn't have patches then u don't have to worry about it

x3rt:
Just wanted to add that at the moment pressing `Delete` to remove a tracked user will immediately remove them regardless of whether or not you decide to `Save & Close`. Making it respect Save & Close / Cancel is not currently planned for this PR though 😅

PR #1036, opened 2023-05-02
Uwufy: Add option to uwufy all messages

pandaninjas:


TheKodeToad:
i wanted to make this... but i could've sworn this was an option!

PR #1051, opened 2023-05-04
add a plugin that opens spotify links in the client

crwn1337:
i miss this plugin from betterdiscord


Vendicated:
this will only work for messages, not other links like bios. I think a click listener on the window that checks if the target is an a would be better as it covers all links

crwn1337:
im guessing you mean something like this?

also, [SpotifyStore](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/SpotifyStore.ts#L91) would need to be updated, maybe [somehow use an href here](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/PlayerComponent.tsx#L288) instead of an onclick?

i tried changing it but i hate working with js/ts in general.

Vendicated:
> also, [SpotifyStore](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/SpotifyStore.ts#L91) would need to be updated, maybe [somehow use an href here](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/PlayerComponent.tsx#L288) instead of an onclick?

image

crwn1337:
> > also, [SpotifyStore](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/SpotifyStore.ts#L91) would need to be updated, maybe [somehow use an href here](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/PlayerComponent.tsx#L288) instead of an onclick?
> 
> image

i meant that option **should** become obsolete after this plugin, but the spotifystore plugin would need to be updated to use hyperlinks instead of `openExternal`

Vendicated:
> > > also, [SpotifyStore](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/SpotifyStore.ts#L91) would need to be updated, maybe [somehow use an href here](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/PlayerComponent.tsx#L288) instead of an onclick?
> > 
> > 
> > image
> 
> i meant that option **should** become obsolete after this plugin, but the SpotifyControls plugin would need to be updated to use hyperlinks instead of `openExternal`

would you like to make this change?

src/plugins/openSpotifyLinks.ts
```diff
@@ -0,0 +1,47 @@
...
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "OpenSpotifyLinks",
+    description: "Opens Spotify links in the Spotify client.",
+    authors: [Devs.crwn],
+    patches: [],
+
+    start() {
+        this.clickListener = (event: MouseEvent) => {
```
Vendicated, line 29: this assignment is obsolete, why not just make it a proper instance method? As in
```ts
onClick(event: MouseEvent) {
...
},

start() {
...
}
```

src/plugins/openSpotifyLinks.ts
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "OpenSpotifyLinks",
+    description: "Opens Spotify links in the Spotify client.",
+    authors: [Devs.crwn],
+    patches: [],
```
Vendicated, line 26:
```suggestion
```

src/plugins/openSpotifyLinks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    start() {
+        this.clickListener = (event: MouseEvent) => {
+            const target = event.target as HTMLElement;
+            if (!(target instanceof HTMLAnchorElement)) return;
+
+            const spotifyLinkRegex = /^https?:\/\/open\.spotify\.com\/(track|album|playlist|artist)\/([a-zA-Z0-9]+)(\?.*)?$/;
```
Vendicated, line 33: please move this regex to the top level

src/plugins/openSpotifyLinks.ts
crwn1337: I'm guessing you mean like this? Since onClick isn't a method that's called by Vencord

crwn1337:
> > > > also, [SpotifyStore](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/SpotifyStore.ts#L91) would need to be updated, maybe [somehow use an href here](https://github.com/Vendicated/Vencord/blob/0b2c3c834a74fb84dc8cb6b22589530f3420828b/src/plugins/spotifyControls/PlayerComponent.tsx#L288) instead of an onclick?
> > > 
> > > 
> > > image
> > 
> > 
> > i meant that option **should** become obsolete after this plugin, but the SpotifyControls plugin would need to be updated to use hyperlinks instead of `openExternal`
> 
> would you like to make this change?

I tried, I wasn't able to do it.

crwn1337:
https://github.com/Vendicated/Vencord/commit/088a8bd1b681b6983c19a8b062e7f755374a4335

Vendicated:
oh I'm really sorry, I entirely forgot this pr was a thing. Thanks anyway for your effort

PR #1062, opened 2023-05-05
silentMessageToggle: Add setting to disable auto disable

SupertigerDev:
Currently, the silent message toggle is automatically disabled after sending a message, so you have to click on the toggle every time you send another message.

I've added an option in settings to allow users to keep it enabled after sending a message.


src/plugins/silentMessageToggle.tsx
```diff
@@ -33,6 +33,11 @@ const settings = definePluginSettings({
         onChange(newValue: boolean) {
             if (newValue === false) lastState = false;
         }
+    },
+    autoDisable: {
+        type: OptionType.BOOLEAN,
+        description: "Automatically disable the state of the silent message after sending a message.",
+        default: true,
```
Nuckyz, line 40:
```suggestion
        default: true
```

src/plugins/silentMessageToggle.tsx
```diff
@@ -33,6 +33,11 @@ const settings = definePluginSettings({
         onChange(newValue: boolean) {
             if (newValue === false) lastState = false;
         }
+    },
+    autoDisable: {
+        type: OptionType.BOOLEAN,
+        description: "Automatically disable the state of the silent message after sending a message.",
```
Vendicated, line 39:
```suggestion
        description: "Automatically disable the silent message toggle again after sending one",
```

PR #1067, opened 2023-05-06
Create splitLargeMessages plugin

nmsturcke:
Create a "splitLargeMessages" that splits messages when they're over the character limit.

Would also like to potentially patch the character count left/over too? Can't find how to tho :P
icZSD


PierreV23:
#800 :p

nmsturcke:
> #800 :p

Oops! I searched, but I guess not very well :P

PR #1068, opened 2023-05-06
reviewDB: Fix avatar z-index

nmsturcke:
Set z-index to 0 for avatars in reviewDB component

## Old
hO1zy

## New
bSv9L

### DIFF
--- a/src/plugins/reviewDB/components/ReviewComponent.tsx
+++ b/src/plugins/reviewDB/components/ReviewComponent.tsx
@@ -94,7 +94,7 @@ export default LazyComponent(() => {
                         className={classes(avatar, clickable)}
                         onClick={openModal}
                         src={review.sender.profilePhoto || "/assets/1f0bfc0865d324c2587920a7d80c609b.png?size=128"}
-                        style={{ left: "0px" }}
+                        style={{ left: "0px", zIndex: 0 }}
                     />
                     <span
                         className={classes(clickable, username)}


Vendicated:
how do u even get this? this seems likr a theme issue

nmsturcke:
> how do u even get this? this seems likr a theme issue

Don't have any themes enabled - It's just that the z-index of the avatar class is 1 by default

Vendicated:
but how do u get here? the profile popout is supposed to close when u open the modal

nmsturcke:
> but how do u get here? the profile popout is supposed to close when u open the modal

click on an avatar of a reviewer
https://i.gyazo.com/b4429db1fccd2e9ac2297a0de75b4ed7.mp4

TheKodeToad:
Maybe it's a Chromium issue because it's not happening on Firefox

nmsturcke:
Also: 
6H7cv

Vendicated:
i cannot reproduce any of this at all. Please provide more info about how you're using Discord, what themes and css you're using, etc

This pr is a poor bandaid fix and doesn't actually address the root issue (of not properly closing the profile popout)

Vendicated:
Gonna close this because it's not the right way of fixing it as mentioned above. Please still provide more info / open a proper fix pr if you can figure it out yourself

nmsturcke:
> Gonna close this because it's not the right way of fixing it as mentioned above. Please still provide more info / open a proper fix pr if you can figure it out yourself

For some reason I can't reproduce it now.. Weird because I had been experiencing it for a while and I have not changed anything within Vencord. Although my whole local cache got restored (~~when I got banned~~) so maybe the issue had something to do with that? No clue tbh

AutumnVN:
lol i accidently reproduced this silly with
```css
[class*=animator] {
  transform: none !important;
}
```

PR #1071, opened 2023-05-06
fix: do not highlight new member leaf in red in message logger

axu5:
Problem: The leaf icon for new members appears red instead of green after the message was deleted.

Fix: CSS `:not` attribute.

Before:
image
After:
image

P.S. Edited and highlighted messages look fine
image
image


Vendicated:
this is not really a good fix, a more specific element list should be used instead, because it affects more than just the leaf

axu5:
So would something more specific be `:not(h3 > span > span > div.newMemberBadge-3PdStX)`?

Vendicated:
> Maybe a better solution would be to only select the message text with the following selector

That is what I meant! Essentially, instead of making a very broad filter, then :not'ing specific stuff like this pr does, I would prefer making the selector more specific like that instead

Vendicated:
what do you think?

axu5:
I looked into it, here is some of my testing with the new solution, I think it's ready for review.

I tried to be specific so it wouldn't highlight things that it shouldn't, however I may have overlooked something.

> What do you think?

I think that this is a better solution incase discord changes the class name for the new member leaf, and so that it doesn't interfere with other plugins. I think being specific in a way where we don't use `:not(.some-specific-class)` allows us to get the correct results in the future too 👍 

**New users, and old users**
image
image

**Embedded content**
image

**Videos and links**
image
image
image

src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -1,4 +1,17 @@
-.messagelogger-deleted :is(div, h1, h2, h3, p) {
+/* Message content highlighting */
+.messagelogger-deleted
+    > .message-2CShn3
```
Vendicated, line 3: this should be obsolete, just content should be sufficient

src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -1,4 +1,17 @@
...
+    > .contents-2MsGLg
```
Vendicated, line 4:
```suggestion
    [class*="contents-"]
```

src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -1,4 +1,17 @@
...
+    > :is(div, h1, h2, h3, p) {
+    color: #f04747 !important;
+}
+
+/* Embed highlighting */
+.messagelogger-deleted
+    > .message-2CShn3
+    > .container-2sjPya
+    > article
```
Vendicated, line 13: article is likely enough? could delete message and container

Vendicated (approved):

PR #1072, opened 2023-05-06
Add "Open Settings" to TextReplace quick actions

ItzOnlyAnimal:
context for the ts-ignore, dont know how to fix it tbh, works anyways
explode

### DIFF
--- a/src/plugins/textReplace.tsx
+++ b/src/plugins/textReplace.tsx
@@ -20,8 +20,10 @@ import { DataStore } from "@api/index";
 import { addPreSendListener, removePreSendListener } from "@api/MessageEvents";
 import { definePluginSettings } from "@api/Settings";
 import { Flex } from "@components/Flex";
+import PluginModal from "@components/PluginSettings/PluginModal";
 import { Devs } from "@utils/constants";
 import { Logger } from "@utils/Logger";
+import { openModalLazy } from "@utils/modal";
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
Is this really needed?
<kbd>ctrl</kbd> + <kbd>,</kbd> does the job

ItzOnlyAnimal:
> Is this really needed? ctrl + , does the job

did you really read it? this goes 
# directly
to the settings. not opening them, searching "text..", clicking cog, then doing whatever you were gonna do

ArjixWasTaken:
> > Is this really needed? ctrl + , does the job
> 
> did you really read it? this goes 
> # directly
> to the settings. not opening them, searching "text..", clicking cog, then doing whatever you were gonna do
> 
> anyways, i will eventually expand this so any plugin's settings can be opened, i guess. since ven brought that up

Yeah I now get it, but I'd have to read the code to know that it opens plugin settings instead of normal settings, so calm down

Vendicated:
Closing this due to what we discussed on Discord:

Adding this to a specific plugin is very opiniated and people will definitely want other plugins too. Instead, there should be an easy way to add a custom list of plugins. My suggestion for how this would look:

```
-- Plugin Settings --
Plugin 1
Plugin 2
Configure plugins > plugin 1 [checkbox]
				    plugin 2 [checkbox]
```

PR #1092, opened 2023-05-09
ShowConnections: Add verified & copy/link icons in tooltip

Vendicated:
Not sure if Copy/Link icon would look better in front of the text instead of after

https://github.com/Vendicated/Vencord/assets/45497981/9b0d72ba-2386-4c65-a6af-ab7a2bc9c9eb


TheKodeToad:
Maybe have the tick before the copy/link

TheKodeToad (approved):
I haven't looked too closely tbh... but the code doesn't seem to be smelly and it looks good

PR #1094, opened 2023-05-10
Add Better Blocked Users plugin

IThundxr:


IThundxr:
this could be merged now but ill be adding more features soon

ArjixWasTaken:
> this could be merged now but ill be adding more features soon

Then this shouldn't be merged, as it is not complete.

IThundxr:
Changelog:

- Added Settings page to select weather blocked messages can be opened or cant
- changed how css is loaded
- add tooltip
- fix button so it works :/
- cleanup css and remove uneeded css
- make the bar better positioned and fix the sizing

IThundxr:
`pnpm test` succeds.

IThundxr:
`testWeb` and `test` both run sucessfully and this is ready to be merged

IThundxr:
Folder casing doesn't seem to be right I'll fix this later when I'm on pc

IThundxr:
Should be as simple as making the name smt else and then changing it again

IThundxr:
switched to camelCasing this pr is fully ready now!

(unless vee ends up finding bad code D: )

Vendicated:
your plugin has no description, could you please describe what this does and maybe show a screenshot?

src/plugins/betterBlockedUsers/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import "./style.css";
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { React, Tooltip } from "@webpack/common";
+
+export const settings = definePluginSettings({
+    showingAllowed: {
+        description: "Allow Blocked Messages to be opened",
```
Vendicated, line 28: this setting seems silly. Why would one want unopenable blocked messages? At that point one would rather use the HideBlockedMessages plugin

IThundxr:
> your plugin has no description, could you please describe what this does and maybe show a screenshot?

Normally when you block a user and they send a message in a mutual server it shows a big giant “Blocked message - show message” banner which gets really distracting and annoying and basically this plug-in replaces it with a small bar that can still be opened but if the user chooses they can disable the function of being able to view the blocked message

src/plugins/betterBlockedUsers/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        type: OptionType.BOOLEAN,
+        default: true
+    }
+});
+
+export default definePlugin({
+    name: "Better Blocked Messages",
```
Vendicated, line 35:
```suggestion
    name: "BetterBlockedMessages",
```

src/plugins/betterBlockedUsers/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+                <span className="vc-betterblockedusers-divider">
+                    <svg className="vc-betterblockedusers-svg" aria-hidden="true" role="img" width="8" height="13" viewBox="0 0 8 13">
+                        <path className="vc-betterblockedusers-color" stroke="currentColor" fill="transparent" d="M8.16639 0.5H9C10.933 0.5 12.5 2.067 12.5 4V9C12.5 10.933 10.933 12.5 9 12.5H8.16639C7.23921 12.5 6.34992 12.1321 5.69373 11.4771L0.707739 6.5L5.69373 1.52292C6.34992 0.86789 7.23921 0.5 8.16639 0.5Z"></path>
+                    </svg>
+                    {
+                        showingAllowed ?
+                            // if showing is allowed show a button and tooltip
+                            <Tooltip text={props.expanded ? "Hide Blocked Message" : "Show Blocked Message"}>
+                                {(tooltipProps: any) => (
+                                    <div style={{ display: "flex" }}>
```
Vendicated, line 70: i think it would semantically make more sense if this outside div would instead be the button, and then just have a plain string "Blocked Message" as last child

IThundxr: that div is needed otherwise it shows errors cause a JSX expression needs a parent div

src/plugins/betterBlockedUsers/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onClick: () => void;
+        expanded: boolean;
+        count: number;
+        compact: boolean;
+        collapsedReason: unknown;
+    }) {
+        if (!props) return null;
+        const { onClick } = props;
+
+        const { showingAllowed } = settings.use(["showingAllowed"]);
```
Vendicated, line 57: you're violating the rules of react hooks here, hooks must never come after a possible early return, that will throw errors. Either remove the `if (!props) return null` as props should never be null or move this hook call above that line

src/plugins/betterBlockedUsers/index.tsx
Vendicated: i don't think you understood me correctly. What i meant is: change this div to be the button instead and then replace the current button with plain text

src/plugins/betterBlockedUsers/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+                        showingAllowed ?
+                            <Tooltip text={props.expanded ? "Hide Blocked Message" : "Show Blocked Message"}>
+                                {(tooltipProps: any) => (
+                                    <button {...tooltipProps} className="vc-betterblockedmessages-button" onClick={onClick}>
+                                        Blocked Message
+                                    </button>
+                                )}
+                            </Tooltip>
+                            :
+                            <span className="vc-betterblockedmessages-span">Blocked Message</span>
```
ArjixWasTaken, line 74: Just a suggestion

```suggestion
        expanded: number; // actually a boolean, but fuck TS
        count: number;
        compact: boolean;
        collapsedReason: unknown;
    }) {
        const { onClick, count: msgCount } = props;

        const { showingAllowed } = settings.use(["showingAllowed"]);
        const text = msgCount === 1 ? "Blocked Message" : `${msgCount} Blocked Messages`;

        return (
            <div className="vc-betterblockedmessages-main" id="---blocked-message-bar" role="separator">
                <span className="vc-betterblockedmessages-divider">
                    <svg className="vc-betterblockedmessages-svg" aria-hidden="true" role="img" width="8" height="13" viewBox="0 0 8 13">
                        <path className="vc-betterblockedmessages-color" stroke="currentColor" fill="transparent" d="M8.16639 0.5H9C10.933 0.5 12.5 2.067 12.5 4V9C12.5 10.933 10.933 12.5 9 12.5H8.16639C7.23921 12.5 6.34992 12.1321 5.69373 11.4771L0.707739 6.5L5.69373 1.52292C6.34992 0.86789 7.23921 0.5 8.16639 0.5Z"></path>
                    </svg>
                    {
                        showingAllowed ?
                            <Tooltip
                                text={
                                    ["Show", "Hide"][props.expanded + 0] +
                                    " Blocked Message" +
                                    (msgCount > 1 && "s" || "")
                                }
                            >
                                {(tooltipProps: any) => (
                                    <button {...tooltipProps} className="vc-betterblockedmessages-button" onClick={onClick}>
                                        {text}
                                    </button>
                                )}
                            </Tooltip>
                            :
                            <span className="vc-betterblockedmessages-span">{text}</span>
```

RobinRMC:
I want to see messages from people I blocked, without having to click the button to show it.

Could you add this as a feature?

(I block people to prevent them from sending me DMs and friend requests, not to prevent myself from seeing their messages.)

PR #1096, opened 2023-05-10
USRBG: Hide Nitro badge if banner's source is USRBG

TheKodeToad:
The reason for doing this is that having Nitro's badge doesn't make much sense. There could be another badge but i'm pretty sure it's only intended for a plug in the first place!

### DIFF
--- a/src/plugins/usrbg/index.tsx
+++ b/src/plugins/usrbg/index.tsx
@@ -47,8 +47,8 @@ const settings = definePluginSettings({
 
 export default definePlugin({
     name: "USRBG",
-    description: "USRBG is a community maintained database of Discord banners, allowing anyone to get a banner without requiring Nitro",
-    authors: [Devs.AutumnVN, Devs.pylix],
+    description: "Displays user banners from USRBG, allowing anyone to get a banner without Nitro",
+    authors: [Devs.AutumnVN, Devs.pylix, Devs.TheKodeToad],
     settings,
     patches: [
         {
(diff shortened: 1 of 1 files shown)


src/plugins/usrbg/index.tsx
```diff
@@ -61,6 +61,10 @@ export default definePlugin({
                 {
                     match: /(\i)\.bannerSrc,/,
                     replace: "$self.useBannerHook($1),"
+                },
+                {
+                    match: /\?\(0,\i\.jsx\)\(\i,{type:\i,shown/,
+                    replace: "&&$self.showBadge(arguments[0])$&"
```
Vendicated, line 67:
```suggestion
                    replace: "&&$self.shouldShowBadge(arguments[0])$&"
```

src/plugins/usrbg/index.tsx
```diff
@@ -78,7 +82,7 @@ export default definePlugin({
 
     settingsAboutComponent: () => {
         return (
-            <Link href="https://github.com/AutumnVN/usrbg#how-to-request-your-own-usrbg-banner">CLICK HERE TO GET YOUR OWN BANNER</Link>
+            <Link href="https://github.com/AutumnVN/usrbg#how-to-request-your-own-usrbg-banner">How to request a banner</Link>
```
Vendicated, line 85: why? need all caps because people are genuinely blind/dumb
```suggestion
            <Link href="https://github.com/AutumnVN/usrbg#how-to-request-your-own-usrbg-banner">CLICK HERE TO GET YOUR OWN BANNER</Link>
```

TheKodeToad: Oh, it looked like YT clickbait to me which is why I changed it

Vendicated: yes but the average USRBG user needs youtube clickbait to even find the button

PR #1106, opened 2023-05-11
Settings: Fix resetting scroll/search when getting a ping

Vendicated:
The issue was with using a TabBar. I don't know exactly why and couldn't figure out how to fix it, so I just got rid of the TabBar and it's fixed

but yeah, tabbar gone
image


MeguminSama (approved):
:( goodbye tab bar

PR #1107, opened 2023-05-12
NoReplyMention: add option to only exclude specific users from pings

outfoxxed:
Adds a new behavior option to NoReplyMention that lets you avoid pinging only specific users instead of only not pinging specific users.

Note: Not sure if I was supposed to add myself to the author list or not, looking through commit history it seems like I am.

### DIFF
--- a/src/plugins/noReplyMention.tsx
+++ b/src/plugins/noReplyMention.tsx
@@ -22,11 +22,26 @@ import definePlugin, { OptionType } from "@utils/types";
 import type { Message } from "discord-types/general";
 
 const settings = definePluginSettings({
-    exemptList: {
+    userList: {
         description:
-            "List of users to exempt from this plugin (separated by commas or spaces)",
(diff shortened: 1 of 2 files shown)


Vendicated:
it's a bit hard to follow for me what this is trying to accomplish. isn't this already a feature if you enable the invert shift behaviour option?

outfoxxed:
Looks like it sorry. Haven't updated in a while so its not in my main browser, and misunderstood what it was doing while writing the pr.

Vendicated:
ooo no need to apologise, thanks for your work anyway!

outfoxxed:
Actually thats not quite right. Inverting shift just makes shit clicking the reply button not work. I'll check how the options interact and see if I need to change anything (I assume it works the way it does for a reason but idk)

outfoxxed:
This pr seems to work fine in its current state with inverseShiftReply as mentioned on the discord, and does not overlap feature wise.

PR #1110, opened 2023-05-12
feat(plugin): FavoriteEmojiFirst

Syncxv:
simple plugin that sorts emoji autocomplete results to show favorite emojis at the top.


Vendicated:
this is when u type in chat right?

Syncxv:
yep

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,73 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Emoji } from "@api/MessageEvents";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+
+const DisambiguatedEmojiContext = findByPropsLazy("getDisambiguatedEmojiContext");
```
Vendicated, line 24:
```suggestion
const EmojiStore = findStoreLazy("EmojiStore");
```

Also this is now used by 3 plugins (EmoteCloner, FakeNitro, Yours, maybe more?), so perhaps it could be moved to webpack commons and properly typed if you would like doing that

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,73 @@
...
+                match: /\(\(function\(.{1,15}\){.{1,400}(?<stateVar>\i)=\i\[0\].{1,200}selectedIndex\);/,
+                replace: "$&$self.sortEmojis($<stateVar>);"
+            }
+        }
+    ],
+
+    sortEmojis(state: EmojiAutocompleteState) {
+        if (
+            state.query == null
+            || state.query.type !== "EMOJIS_AND_STICKERS"
```
Vendicated, line 55: Using optional chaining here makes the null check obsolete
```suggestion
            || state.query?.type !== "EMOJIS_AND_STICKERS"
```

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,73 @@
...
+            || state.query.typeInfo?.sentinel !== ":"
+            || state.query.results?.emojis?.length === 0
```
Vendicated, line 57: You only check for 0 here and then later assume results.emojis is non nullish when it could be undefined. Just dropping the === 0 and instead negating solves that
```suggestion
            || !state.query.results?.emojis?.length
```

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,73 @@
...
+        ) return;
+
+        const emojiContext = DisambiguatedEmojiContext.getDisambiguatedEmojiContext();
+
+        state.query.results.emojis = state.query.results.emojis.sort((a: Emoji, b: Emoji) => {
```
Vendicated, line 62: sort sorts in place so there is no need to reassign
```suggestion
        state.query.results.emojis.sort((a: Emoji, b: Emoji) => {
```

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,73 @@
...
+            replacement: {
+                match: /\(\(function\(.{1,15}\){.{1,400}(?<stateVar>\i)=\i\[0\].{1,200}selectedIndex\);/,
+                replace: "$&$self.sortEmojis($<stateVar>);"
+            }
+        }
+    ],
+
+    sortEmojis(state: EmojiAutocompleteState) {
+        if (
+            state.query == null
```
Vendicated, line 54:
```suggestion
```

Vendicated:
doesn't seem to work unless im missing something, i have an emote called umuCool favourited
image

Vendicated:
could it be that the filter state only contains the top 6 results (aka what it's showing) and not actually all matches?

Syncxv:
> could it be that the filter state only contains the top 6 results (aka what it's showing) and not actually all matches?

yeah thats right. 
i tried fixing this in my most recent commit. try it and let me know if it works

Vendicated:
still doesn't seem to work for me

Vendicated:
changed it from -1 to Infinity, finally it seems to work for me 
image

src/plugins/favEmojiFirst.ts
```diff
@@ -0,0 +1,80 @@
...
+        query.results.emojis = query.results.emojis.sort((a: Emoji, b: Emoji) => {
+            const aIsFavorite = emojiContext.isFavoriteEmojiWithoutFetchingLatest(a);
+            const bIsFavorite = emojiContext.isFavoriteEmojiWithoutFetchingLatest(b);
+
+            if (aIsFavorite && !bIsFavorite) return -1;
+
+            if (!aIsFavorite && bIsFavorite) return 1;
+
+            return 0;
+        }).slice(0, 6);
```
Vendicated, line 78: Discord's limit is actually 10, what's up with that?
image

Vendicated: oh i see now, it does to 10 emojis but reduces the max number of emojis down to 6 if there are also gif matches

PR #1114, opened 2023-05-13
feat(rolecoloreverywhere): chat messages

lewisakura:
Applies the role color to chat messages.

Current state of PR does not update the chat messages when members are cached or roles change/update color, so you need to hover over the message for them to update correctly. Need to find an efficient way of triggering an update only for the users' messages.


Vendicated:
> Need to find an efficient way of triggering an update only for the users' messages.

Well Discord already has some state for the user's colour (as seen by the username colour updating in real time) so you could likely just use that

lewisakura:
I'm unsure of how the messages actually get updated... I can't see any React hooks or anything in the source from my read, so I don't know how to trigger a successful rerender. But, for now, this is workable and is the same behaviour as the original.

src/plugins/roleColorEverywhere.tsx
```diff
@@ -87,10 +93,21 @@ export default definePlugin({
...
                 }
             ],
             predicate: () => settings.store.voiceUsers,
+        },
+        // Chat messages
+        {
+            find: "Messages.SOURCE_MESSAGE_DELETED",
+            replacement: [
+                {
+                    match: /function \i\((\i)\).{5,20}className.{5,10}message.{40,80}isEdited.{250,300}jsxs.{20,35}ref:\i/,
```
Vendicated, line 106: this match is needlessly complicated. It would be much cleaner to either use arguments[0] to access the first argument to the surrounding function, or to use a lookbehind to obtain the parameter name

Vendicated: for example something like
```js
match: /ref:\i,className:.{0,20}markup/,
replace: "...$self.getInlineStyle({user:arguments[0].message.author,channelId:arguments[0].message.channel_id}),$&"
```

Vendicated:
it doesnt really work well
because the member isnt cached at the time the message is rendered

Nuckyz:
Superseded by https://github.com/Vendicated/Vencord/pull/2893

PR #1120, opened 2023-05-13
VcNarrator: Add 'Narrate Self' option

iAmSpace:
if anyone wishes to hear their own name on join this should work

### DIFF
--- a/src/plugins/vcNarrator.tsx
+++ b/src/plugins/vcNarrator.tsx
@@ -165,7 +165,7 @@ export default definePlugin({
                 if (!type) continue;
 
                 const template = Settings.plugins.VcNarrator[type + "Message"];
-                const user = Settings.plugins.VcNarrator["narrateSelf"] && isMe ? "" : UserStore.getUser(userId).username;
+                const user = isMe && !Settings.plugins.VcNarrator.sayOwnName ? "" : UserStore.getUser(userId).username;
                 const channel = ChannelStore.getChannel(id).name;
 
                 speak(formatText(template, user, channel));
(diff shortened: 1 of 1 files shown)


src/plugins/vcNarrator.tsx
```diff
@@ -165,7 +165,7 @@ export default definePlugin({
                 if (!type) continue;
 
                 const template = Settings.plugins.VcNarrator[type + "Message"];
-                const user = isMe ? "" : UserStore.getUser(userId).username;
+                const user = Settings.plugins.VcNarrator["narrateSelf"] && isMe ? "" : UserStore.getUser(userId).username;
```
Vendicated, line 168: narrateSelf is kind of a misleading name and this if should be inverted
```suggestion
                const user = isMe && !Settings.plugins.VcNarrator.sayOwnName ? "" : UserStore.getUser(userId).username;
```

iAmSpace:
> narrateSelf is kind of a misleading name and this if should be inverted
that is my mistake sorry :)

src/plugins/vcNarrator.tsx
```diff
@@ -230,6 +230,11 @@ export default definePlugin({
                 markers: [0.1, 0.5, 1, 2, 5, 10],
                 stickToMarkers: false
             },
+            narrateSelf: {
+                description: "Narrate Self",
```
Vendicated, line 234: did you forget to update the setting

iAmSpace: uhhh yes i am half asleep

Vendicated (approved):

PR #1125, opened 2023-05-14
DevPlugin: AssetExplorer

ArjixWasTaken:
## Warning
This plugin is for plugin devs, to use it you must build vencord in dev mode. (eg `pnpm watch`)

## About
This is a plugin to make finding icons an easier task for plugin devs.
Many icons are not exported, so you might have to either patch the icon to export it or copy the component and paste it in ur plugin.

## TODO
- [ ] Maybe load all the webpack chunks so all the icons are loaded?

image
image

### DIFF
--- a/src/plugins/assetExplorer.dev.tsx
+++ b/src/plugins/assetExplorer.dev.tsx
@@ -24,7 +24,7 @@ import { Text, Toasts } from "@webpack/common";
 
 const pluginName = "AssetExplorer" as const;
 export default definePlugin({
-    name: pluginName,
+    name: "AssetExplorer",
     description: "Dev util for exploring icons bundled with discord.",
     authors: [Devs.Arjix],
 
--- a/src/plugins/assetExplorer.dev.tsx
+++ b/src/plugins/assetExplorer.dev.tsx
@@ -1,0 +1,90 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


src/plugins/assetExplorer.dev.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <ModalRoot {...modalProps} size={ModalSize.LARGE}>
+            <ModalHeader>
+                <Text variant="heading-lg/semibold" style={{ flexGrow: 1 }}>Asset Explorer</Text>
+                <ModalCloseButton onClick={close} />
+            </ModalHeader>
+
+            <ModalContent>
+                <div style={{ marginBottom: "10px" }}></div>
+                {assets.map(Asset => {
+                    if (`${Asset}`.includes("\"svg\""))
```
Vendicated, line 72:
```suggestion
                    if (`${Asset}`.includes('"svg"'))
```

src/plugins/assetExplorer.dev.tsx
```diff
@@ -0,0 +1,90 @@
...
+        },
+        all: true
+    }],
+
+    toolboxActions: {
+        "Open Explorer"() {
+            const key = openModal(modalProps => (
+                <AssetExplorerModal
+                    modalProps={modalProps}
+                    close={() => closeModal(key)}
```
Vendicated, line 46: i noticed recently that you can just use modalProps.onClose to close the modal, so storing the key and using closeModal manually is actually not necessary!

src/plugins/assetExplorer.dev.tsx
```diff
@@ -0,0 +1,127 @@
...
+            const isWasm = res.includes(".module.wasm");
+
+
+            if (hasSVG && !isWasm) {
+                await wreq.e(id as any);
+                modulesLoaded++;
+                logger.debug(`Loaded module ${id} (${count}/${modulesCount})`);
+            }
+
+            await new Promise(r => setTimeout(r, 50));
```
Vendicated, line 61: from my experience this is too short and will possibly get u rate limited, i think i use 300ms or smth like that in the workflow and haven't gotten rate limited yet (although that will take a minute or two)

Vendicated: I'm also not a big fan of fetching this on startup. I would assume most devs use a devbuild 24/7 (i do!) so this would run every time and possibly make discord laggy (?)

What do you think about instead:
- Load all svgs the first time (if no saved version exists, see below)
- Save all the svgs in DataStore
- Use the saved ones from now on
- Add a "Rebuild Icon Cache" button in the modal

ArjixWasTaken: I did ask you if it should be a setting, and you 100% forgot that

PR #1138, opened 2023-05-15
PermViewer: Fix context menu for roleless users & muted channels

Vendicated:


src/plugins/permissionsViewer/index.tsx
```diff
@@ -137,10 +139,12 @@ function makeContextMenuPatch(childId: string, type?: MenuItemParentType): NavCo
                 case MenuItemParentType.Guild:
                     return MenuItem(props.guild.id);
                 default:
-                    throw new Error("Invalid type");
+                    return null;
             }
         })();
 
+        if (item == null) return;
```
Vendicated, line 146: remove this

src/plugins/permissionsViewer/index.tsx
```diff
@@ -137,10 +139,12 @@ function makeContextMenuPatch(childId: string, type?: MenuItemParentType): NavCo
                 case MenuItemParentType.Guild:
                     return MenuItem(props.guild.id);
                 default:
-                    throw new Error("Invalid type");
+                    return null;
```
Vendicated, line 142: why? this point should never be reached so throwing an error is more logical

src/api/ContextMenu.ts
```diff
@@ -89,15 +89,18 @@ export function removeGlobalContextMenuPatch(patch: GlobalContextMenuPatchCallba
 }
 
 /**
- * A helper function for finding the children array of a group nested inside a context menu based on the id of one of its childs
- * @param id The id of the child
+ * A helper function for finding the children array of a group nested inside a context menu based on the id(s) of its children
+ * @param id The id(s) of the child/children
```
Vendicated, line 93: this doc is kinda misleading, it makes it sound like all children must be present
```suggestion
 * @param id The id of the child. If an array is specified, all ids will be tried
```

PR #1160, opened 2023-05-16
PlatformIndicators: make mobile icon match Discord's

Nuckyz:


src/plugins/platformIndicators.tsx
```diff
@@ -29,15 +29,15 @@ import { User } from "discord-types/general";
 
 const SessionsStore = findStoreLazy("SessionsStore");
 
-function Icon(path: string, viewBox = "0 0 24 24") {
+function Icon(path: string, opts?: { viewBox?: string; size?: { width: number, height: number; }; }) {
```
Vendicated, line 32: why not have this flat? also considering it's always square, you could just change it from height & width to "size" or smth like that and use it for both
```suggestion
function Icon(path: string, opts?: { viewBox?: string; width?: number, height?: number; }; }) {
```

Vendicated (approved):
other than that looks good

PR #1161, opened 2023-05-16
New plugin : Party mode 🎉

UwUDev:
## Party mode 🎉

Enable the party mode + cool presets.

huge gif loading...

### Please note that's the first time of my life coding in TS so please don't trash me ;~;

I face 3 problems in my code :
- I set `GenericStore` in export from `stores.ts` to get the 'PoggermodeSettingsStore' store but i don't know if the best way to do it.
- I don't know if i am autorized to put spaces and emojis
- The way i imported `waitForStore` makes me feel bad


UwUDev:
Sorry sorry if my code was super ugly x)
And ty c:

Imbwbl:
That the best plugin I never see

src/plugins/partyMode/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+enum Intensity {
+    Normal,
+    Better,
+    ProjectX,
+}
+
+const settings = definePluginSettings({
+    superIntensePartyMode: {
+        description: "Party intensity (disable and re-enable plugin to apply changes)",
```
ArjixWasTaken, line 36: Since I added the `onChange` callback, this no longer applies.
```suggestion
        description: "Party intensity",
```

src/plugins/partyMode/index.ts
```diff
@@ -0,0 +1,107 @@
...
+*/
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findStoreLazy } from "@webpack";
+import { GenericStore } from "@webpack/common";
+
+
+const PoggerModeSettingsStore: GenericStore = findStoreLazy("PoggermodeSettingsStore");
```
ArjixWasTaken, line 26: Use the actual store from `src/webpack/common/stores.ts` (after you've defined it there ofc)

src/webpack/common/stores.ts
```diff
@@ -37,6 +37,7 @@ export let PermissionStore: GenericStore;
 export let GuildChannelStore: GenericStore;
 export let ReadStateStore: GenericStore;
 export let PresenceStore: GenericStore;
+export let PoggerModeSettingsStore: GenericStore;
```
ArjixWasTaken, line 40: Since you defined a type for it, you should export the actual store just like the other ones.

ArjixWasTaken (changes requested):

src/webpack/common/stores.ts
UwUDev: I had put the store here at the beginning but I preferred to put it in the plugin directly because this store is only used for the party mode and I will surely be the only one to use it

ArjixWasTaken: Then please remove the type from `@webpack/common/stores` as it is not really useful.

src/plugins/partyMode/index.ts
```diff
@@ -0,0 +1,107 @@
...
+function setPoggerState(state: boolean) {
+    Object.assign(PoggerModeSettingsStore.__getLocalVars().state, {
+        enabled: state,
+        settingsVisible: state
+    });
+}
+
+function setSettings(intensity: Intensity) {
+    const state = {
+        screenshakeEnabledLocations: { 0: true, 1: true, 2: true },
```
Vendicated, line 74:
```suggestion
        screenshakeEnabledLocations: [true, true, true],
```

ArjixWasTaken: I mean, it ain't wrong, it's the truth.

Vendicated:
Thank you!

Please don't pull request from the main branch in the future, use a feature branch

PR #1162, opened 2023-05-16
feat(InvisibleChat): use discords embed api

SammCheese:
Bin ballin

### DIFF
--- a/src/plugins/invisibleChat/index.tsx
+++ b/src/plugins/invisibleChat/index.tsx
@@ -22,7 +22,7 @@ import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants";
 import { getStegCloak } from "@utils/dependencies";
 import definePlugin, { OptionType } from "@utils/types";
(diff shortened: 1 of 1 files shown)


src/plugins/invisibleChat/index.tsx
```diff
@@ -178,25 +178,13 @@ export default definePlugin({
...
-
-        // AWS hosted url to discord embed object
-        const rawRes = await fetch(this.EMBED_API_URL, options);
-        clearTimeout(timeout);
-
-        return await rawRes.json();
+        const { body } = await RestAPI.post({
+            url: "/unfurler/embed-urls",
+            body: {
+                urls: [url]
```
Vendicated, line 184: considering this takes an array, i would change this to fetch all links in one batch instead of sending multiple reqeuests

SammCheese: If you look at Line 204, It always just fetches the first url. We dont want to spam the user with random embeds.

Vendicated (approved):

PR #1165, opened 2023-05-17
fix(TypingTweaks): use global displayName over username

0xallie:
This adds support for Discord's new global display name feature rolled out ahead of the new username system, and makes it prefer the global display name over the username like the stock client does.


0xallie:
It looks like `discord-types` needs to be updated, this definitely works though.

Vendicated:
yeah it's really outdated

0xallie:
I don't know if anything else is outdated in it, but this has been added now.  

Swishilicous/discord-types#2

PR #1166, opened 2023-05-18
Fix SHC broken patches; Sort PermViewer channel overwrites roles

Nuckyz:
Merge when broken SHC patches reach stable


Nuckyz:
It's on stable already

PR #1176, opened 2023-05-20
PluginModal: Anonymise authors

Vendicated:
Makes the user modal for plugin authors "anonymous" (just see the screenshot)

image

As to why: Many users think adding plugin authors or sending them unsolicited support dms is appropriate. This prevents people from interacting with the authors, hopefully promoting use of #support over dms

We probably want to hide the BOT tag


MeguminSama (approved):

TheKodeToad:
What about having a note not to contact for support?

Vendicated:
bold of u to assume the type of people to add/dm u for support will read that :P

PR #1187, opened 2023-05-23
ShowConnections: Remove redundant inferior Connections section, use `i18n`

TheKodeToad:
The roles most likely imply what service they have linked on the server


TheKodeToad:
image
in case it wasn't clear (I can't remember; it's been a while)
this is not pretty :c

Vendicated:
i see what this is now, the linked roles section

hiding this is definitely not a good idea... it may contain other unrelated things and hiding stock discord features is almost never a good idea

thank you anyway!

PR #1198, opened 2023-05-26
VencordToolbox: rework to add and remove plugin settings by using a checkbox

Redundakitties:
Added a few other quick actions such as reload discord, toggle css, and updater tab. plans to add more as well probably. 

Some things that are annoying  and I do not know how to fix: 
- contributors become bots 
- checkbox menu scrolls to the top every time you add a new plugin 
- no way to tell the user to restart if they change a plugin setting
- and probably others I forgot while writing this pull request


src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -249,4 +250,60 @@ function Updater() {
     );
 }
 
+function UpdaterModal ({ modalProps, close }: { modalProps: ModalProps; close(): void; }) {
```
Vendicated, line 253: this seems to be copy paste of a lot of code. Why not instead refactor the component to be more generic and used by both?

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,267 @@
...
+        description: "Plugin Settings",
+        default: true,
+        component: () =>
+            <Switch
+                value={settings.store.PluginSettings}
+                onChange={(v: boolean) => settings.store.PluginSettings = v}
+                note="Display plugin settings group for adding or removing plugin settings from toolbox"
+            >
+                Plugin Settings
+            </Switch>
```
Vendicated, line 146: 90% of the code is duplicated for each setting. Why not make it a function that takes parameters for the settings name and description?

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,267 @@
...
+            <Switch
+                value={settings.store.DevCompanion}
+                onChange={(v: boolean) => settings.store.DevCompanion = v}
+                note="Display DevCompanion Reconnect entry in toolbox"
+            >
+                DevCompanion
+            </Switch>
+    },
+    // For enabling and disabling individual plugin settings menus
+    PluginSettings: {
```
Vendicated, line 135: camelCase is preferred
```suggestion
    pluginSettings: {
```

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,267 @@
...
+                note="Display plugin settings group for adding or removing plugin settings from toolbox"
+            >
+                Plugin Settings
+            </Switch>
+
+    },
+});
+
+function VencordPopout({ onClose }: { onClose: () => void; }) {
+    const ps = settings.use(["includedPlugins"] as any) as unknown as { includedPlugins: string[]; }; // keeps track of added plugin settings entries
```
Vendicated, line 152: what in the world is this mess

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,267 @@
...
+                            );
+                        })}
+                    </Menu.MenuGroup>
+                ))
+            }
+
+            {settings.store.PluginSettings && // Main plugin setttings dropdown w/ checkboxes
+                <Menu.MenuGroup label="Plugin Settings">
+                    {settings.store.PluginSettings &&
+                        Object.values(Vencord.Plugins.plugins)
```
Vendicated, line 233: you reuse Object.values(plugins) multiple times. Do it at the top and store the result in a variable for better performance

Vendicated (changes requested):

src/plugins/vencordToolbox/index.tsx
Redundakitties: okay this one is explainable, before adding the checkbox dropdown menu I had set the settings name to be the name of the plugin to distinguish which plugin to add to the toolbox. but that's all been rendered obsolete ✅

src/plugins/vencordToolbox/index.tsx
Redundakitties: oops ignore that

Redundakitties: wait now i remember, this is correct. it's how I'm adding plugins to the dropdown. same method as welcomestickerpicker. is there a better way?

PR #1201, opened 2023-05-26
DailyStatus: New plugin to autogenerate ChatGPT status messages

darktohka:
dAIlystatus is a new Vencord plugin that serves to automatically generate brand new status messages for you on Discord every single day.

It uses the ChatGPT API to generate status messages. This generation is done every single day, on midnight (or whenever the user happens to start Vencord). The status message that is generated is then shown to the user as a modal, for manual verification:

image

If the user likes the status message, they can press the Apply button to automatically apply the new status message. If not, they can press the Regenerate button to generate a new status message using ChatGPT. If the user does not want to change their status message, they can simply dismiss the modal or press the Cancel button.

By default, this plugin uses the free ChatGPT API provided by free.churchless.tech. The actual ChatGPT API can also be used. The ChatGPT API endpoints, API keys and model selection is available through the plugin settings. If required, the user can also set the plugin up to automatically set the status messages, without manual user intervention and status message verification.


darktohka:
Resolved merge conflicts.

ArjixWasTaken:
Would be cool if it supported more values than every day.

Like every hour, every 2nd hour and so on

Vendicated:
Hi! I don't think this plugin is very mergeable for various reasons:
- It uses a suspicious seeming api that will likely die sooner or later
- Especially with arjix suggestion (and other people will likely also request it), this comes dangerously close to AnimatedStatus which is against our rules
- It doesn't seem very useful to me. Who would want an ai generated status? 

Thanks for your effort anyway! If you want to publish this in the #third-party-plugins channel on discord, feel free to ping me for access

PR #1206, opened 2023-05-28
Add option to show/hide magnifier circle.

mostafa-abdelbrr:
Hacky.
It just sets the magnifier circle size to a very large value (this is how I got the same effect on the original BD plugin anyway).

### DIFF
--- a/src/plugins/imageZoom/index.tsx
+++ b/src/plugins/imageZoom/index.tsx
@@ -198,11 +198,11 @@ export default definePlugin({
         if (instance.props.id === ELEMENT_ID) {
             if (!this.currentMagnifierElement) {
                 if(settings.store.showMagnifierCircle) {
-                this.currentMagnifierElement = <Magnifier size={settings.store.size} zoom={settings.store.zoom} instance={instance} />;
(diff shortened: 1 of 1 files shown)


Vendicated:
so instead of changing the code, why not just set `settings.store.size` to 1000?

mostafa-abdelbrr:
Already tried that when the plugin came out, 1000 turned out to be too small and just got me a kinda small/medium magnifier circle.

Vendicated:
> Already tried that when the plugin came out, 1000 turned out to be too small and just got me a kinda small/medium magnifier circle.

I'm confused. All this PR does is hardcode the setting to 1000, so why can't you just do that via settings?

mostafa-abdelbrr:
10K, not 1K.

Vendicated:
you cxan still just do it via settings!

mostafa-abdelbrr:
> you cxan still just do it via settings!

I see no such option to set it to 10K.

Vendicated:
```js
Vencord.Settings.plugins.ImageZoom.size = 10_000
```
in devtools

mostafa-abdelbrr:
Well, this means it is not easily available, but alright I guess. Thanks for taking the time to review it!

PR #1208, opened 2023-05-29
feat(plugin): Translate Messages

shrihanDev:
Translates messages using Google Translate.
(More or less a port of https://github.com/aeongdesu/vdplugins/tree/main/plugins/Dislate)


somerandomcloud:
I might be remembering wrong but I think this already exists

ItzOnlyAnimal:
😕 right, this already is a plugin, so i dont see a need for a duplicate unless this one does something different 
https://github.com/Vendicated/Vencord/tree/main/src/plugins/translate

Vendicated:
Yeah, unfortunately this plugin already exists so this is a duplicate. Thanks anyway!

PR #1216, opened 2023-05-30
Plugin: NoPendingCount

aamiaa:
This plugin removes the ping count of incoming friend requests, message requests, and nitro offers.
Ping indicators are removed in the following places:

- Home button (count reduced)
- Friends tab button
- Pending tab button
- Message Requests button
- Taskbar icon (count reduced)

The visibility logic of the Message Requests tab button is preserved (i.e. still appears when there are message requests, and disappears otherwise).

### DIFF
--- a/src/plugins/noPendingCount.ts
+++ b/src/plugins/noPendingCount.ts
@@ -1,6 +1,6 @@
 /*
  * Vencord, a modification for Discord's desktop app
- * Copyright (c) 2022 Vendicated and contributors
+ * Copyright (c) 2023 Vendicated and contributors
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/plugins/noPendingCount.ts
+++ b/src/plugins/noPendingCount.ts
@@ -17,10 +17,10 @@
 */
 
 
+import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy } from "@webpack";
-import { definePluginSettings } from "@api/Settings";
 
 const MessageRequestStore = findByPropsLazy("getMessageRequestsCount");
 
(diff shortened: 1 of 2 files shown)


src/plugins/noPendingCount.ts
```diff
@@ -0,0 +1,97 @@
...
+    settings: settings,
+
+    // Functions used to determine the top left count indicator can be found in the single module that calls getUnacknowledgedOffers(...)
+    // or by searching for "showProgressBadge:"
+    patches: [
+        {
+            find: ".getPendingCount=",
+            predicate: () => settings.store.hideFriendRequestsCount,
+            replacement: {
+                match: /(\i)\.getPendingCount=function\(\){return\s\i};/,
```
Vendicated, line 62:
```suggestion
                match: /\.getPendingCount=function\(\)\{/,
```

you can make this patch simpler by not unnecessarily matching \i and the function body and also using a lookbehind. Now your replacement only has to be `.getPendingCount=function(){return 0;`

You could make it even simpler by using a lookbehind:
```suggestion
                match: /(?<=\.getPendingCount=function\(\)\{)/,
```
Now your replacement becomes `return 0;`

src/plugins/noPendingCount.ts
```diff
@@ -0,0 +1,97 @@
...
+                replace: "$1.getMessageRequestsCount=function(){return 0};"
+            }
+        },
+        // This prevents the Message Requests tab from always hiding due to the previous patch (and is compatible with spam requests)
+        // In short, only the red badge is hidden. Button visibility behavior isn't changed.
+        {
+            find: ".getSpamChannelsCount(),",
+            predicate: () => settings.store.hideMessageRequestsCount,
+            replacement: {
+                match: /(,\i=\i\.getSpamChannelsCount\(\)),(\i)=\i\.getMessageRequestsCount\(\),(\i=\i>0\|\|\i>0;return\s*?\i}})/,
```
Vendicated, line 80:
```suggestion
                match: /(?<=getSpamChannelsCount\(\),\i=)\i\.getMessageRequestsCount\(\)/,
```

Now your replacement is just `$self.getRealMessageRequestCount()`

src/plugins/noPendingCount.ts
aamiaa: ah okay, for some reason I thought that causing the function to have double `return` would be bad

Vendicated: no it's completely fine, we do it a lot as it makes many things a lot simpler

PR #1222, opened 2023-06-01
feat(plugin): BiggerStreamPreview

philhk:
# BiggerStreamPreview

This pull request adds the BiggerStreamPreview plugin, which implements the idea proposed in #392. With this plugin, users can enlarge any Stream Preview.

image

The context menu mentioned above appears whenever you right-click on a user or directly on a stream.

image


src/components/MaskedLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { LazyComponent } from "@utils/react";
+import { find } from "@webpack";
+
+
+export type MaskedLink = React.ComponentType<{
+    onClick(): void;
+    trusted: boolean;
+}>;
+
+export const MaskedLink = LazyComponent(() => find(m => m?.type?.toString()?.includes("MASKED_LINK")) as MaskedLink);
```
Vendicated, line 28: this should be in webpack commons, not here

src/webpack/common/types/classes.d.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+export interface ImageModalClasses {
+    image: string,
+    modal: string,
+    responsiveWidthMobile: string;
+}
```
Vendicated, line 23: there are some more class commons that are in components, could you please move them here as well?

philhk: Sure. Is the `ButtonWrapperClasses` the only one or did I miss anything?

src/components/MaskedLink.tsx
philhk: [Okay](https://github.com/Vendicated/Vencord/pull/1222/files#diff-b579282bc8a47a3af7dcf9d446f2188a5bef1f493fbd14338dfeec369c5a3723R51) I have done that now.

src/plugins/biggerStreamPreview/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                />
+            </ModalRoot>
+        );
+    }
+    );
+};
+
+export const addViewStreamContext = (children, { userId }: { userId: string | bigint; }) => {
+    const streamPreviewItemIdentifier = "view-stream-preview";
+    if (children.some(child => child?.props?.id === streamPreviewItemIdentifier)) return;
```
Vendicated, line 80: instead of this check, you can return a function from your patch. Vencord will only call that function once

src/plugins/biggerStreamPreview/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    shouldAnimate={true}
+                    renderLinkComponent={props => <MaskedLink {...props} />}
+                />
+            </ModalRoot>
+        );
+    }
+    );
+};
+
+export const addViewStreamContext = (children, { userId }: { userId: string | bigint; }) => {
```
Vendicated, line 78:
```suggestion
export const addViewStreamContext = (children, { userId }: { userId: string | bigint; }) => () => {
```

for the suggestion below

src/plugins/biggerStreamPreview/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    target: HTMLElement,
+    stream: Stream,
+    theme: string,
+}
+
+export const handleViewPreview = async ({ guildId, channelId, ownerId }: ApplicationStream | Stream) => {
+    const previewUrl = await ApplicationStreamPreviewStore.getPreviewURL(guildId, channelId, ownerId);
+    if (!previewUrl) return;
+
+    openModalLazy(async () => {
```
Vendicated, line 58: ViewIcons also makes use of this exact modal. Thus it would be beneficial to move this into utils/discord so both plugins can use it!

src/webpack/common/stores.ts
```diff
@@ -90,3 +92,5 @@ waitForStore("GuildChannelStore", m => GuildChannelStore = m);
 waitForStore("MessageStore", m => MessageStore = m);
 waitForStore("WindowStore", m => WindowStore = m);
 waitForStore("EmojiStore", m => EmojiStore = m);
+waitForStore("ApplicationStreamPreviewStore", m => ApplicationStreamPreviewStore = m);
+waitForStore("ApplicationStreamingStore", m => ApplicationStreamingStore = m);
```
Vendicated, line 96: webpack commons like the name implies are for commonly used modules. These stores are very niche, i can't think of any other plugin that would want these, so they should probably be in your plugin instead

src/webpack/common/index.ts
```diff
@@ -16,12 +16,13 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+export * from "./classes";
 export * from "./components";
+export * from "./internal";
```
Vendicated, line 21: this is called internal for a reason, you shouldn't use it in ur plugin

src/utils/discord.tsx
```diff
@@ -77,3 +79,23 @@ export function sendMessage(
 
     return MessageActions.sendMessage(channelId, messageData, waitForChannelReady, extra);
 }
+
+export function openImageModal(url: string, props?: Partial<React.ComponentProps<ImageModal>>): Promise<string> {
+    return openModalLazy(async () => _props => (
```
Vendicated, line 84:
```suggestion
    return openModal(modalProps=> (
```

src/plugins/viewIcons.tsx
```diff
@@ -60,26 +57,45 @@ const settings = definePluginSettings({
                 value: "jpg",
             }
         ]
+    },
+    imgsize: {
+        type: OptionType.SELECT,
+        description: "Choose the default image size to use, reducing this value will decrease load time and bandwidth.",
```
Vendicated, line 63:
```suggestion
        description: "The image size to use",
```

src/plugins/viewIcons.tsx
```diff
@@ -60,26 +57,45 @@ const settings = definePluginSettings({
...
+        options: [
```
Vendicated, line 64: you could make this slightly prettier like so:
```js
options: [128, 256, 512, 1024, 2048, 4096].map(n => ({ label: n, value: n }))
```

src/plugins/viewIcons.tsx
```diff
@@ -60,26 +57,45 @@ const settings = definePluginSettings({
                 value: "jpg",
             }
         ]
+    },
+    imgsize: {
```
Vendicated, line 61:
```suggestion
    imgSize: {
```

src/plugins/biggerStreamPreview/webpack/types/stores.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export interface StreamingStoreState {
+    activeStreams: [string, Stream][];
+    rtcStreams: { [key: string]: RTCStream; };
+    streamerActiveStreamMetadatas: { [key: string]: StreamMetadata | null; };
+    streamsByUserAndGuild: { [key: string]: { [key: string]: ApplicationStream; }; };
+}
+
+/**
+ * Here is a example how an stream key could look like `call(type of connection):1116549917987192913(channelId):305238513941667851(ownerId)`
```
Vendicated, line 58:
```suggestion
 * example how a stream key could look like: `call(type of connection):1116549917987192913(channelId):305238513941667851(ownerId)`
```

src/plugins/biggerStreamPreview/webpack/stores.ts
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+import { waitForStore } from "@webpack/common";
+
+import * as t from "./types/stores";
+
+export let ApplicationStreamPreviewStore: t.ApplicationStreamPreviewStore;
+export let ApplicationStreamingStore: t.ApplicationStreamingStore;
+
+waitForStore("ApplicationStreamPreviewStore", m => ApplicationStreamPreviewStore = m);
```
Vendicated, line 26: as mentioned below, plugins shouldn't use this. Use findStoreLazy instead

Vendicated (changes requested):

src/plugins/viewIcons.tsx
```diff
@@ -60,26 +57,27 @@ const settings = definePluginSettings({
                 value: "jpg",
             }
         ]
+    },
+    imgSize: {
+        type: OptionType.SELECT,
+        description: "The image size to use",
+        options: [128, 256, 512, 1024, 2048, 4096].map(n => ({ label: `${n}`, value: n, default: n === 1024 }))
```
Vendicated, line 64:
```suggestion
        options: ["128", "256", "512", "1024", "2048", "4096"].map(n => ({ label: n, value: n, default: n === "1024" }))
```

Vendicated (approved):

Vendicated:
thanks!!

PR #1226, opened 2023-06-03
Fix cloud auth [object%20Object]

AutumnVN:


rivalryxd:
I need this ngl, I have multiple devices with vencord

Nuckyz (approved):

PR #1241, opened 2023-06-07
chore: update pnpm

ryanccn:
because yes


Vendicated:
this is kinda pointless

PR #1243, opened 2023-06-07
USRBG: Show USRBG banners in members list

AutumnVN:
# closed cuz dead and there is luna's pr


src/plugins/usrbg/index.tsx
```diff
@@ -77,6 +83,16 @@ export default definePlugin({
                     replace: "$self.voiceBackgroundHook($1),"
                 }
             ]
+        },
+        {
+            find: "[\"aria-selected\"])",
```
Vendicated, line 88: this fine seems very generic and could appear in many places. Is there a reason it's this way and doesn't target something more specific?

src/plugins/usrbg/index.tsx
```diff
@@ -86,17 +102,22 @@ export default definePlugin({
         );
     },
 
+    memberListBannerHook(props: any) {
+        const userId = props.avatar._owner.pendingProps.user.id;
```
Vendicated, line 106: this access seems dangerous. I would wrap this in a try catch for safety

src/plugins/usrbg/index.tsx
AutumnVN: i dont remember
also parent took away the laptop :(

PR #1245, opened 2023-06-07
feat(PronounDB): Add local pronoun overrides

nea89o:
Newer version of https://github.com/Vendicated/Vencord/pull/424

### DIFF
--- a/src/plugins/pronoundb/contextmenu.tsx
+++ b/src/plugins/pronoundb/contextmenu.tsx
@@ -1,0 +1,62 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { NavContextMenuPatchCallback } from "@api/ContextMenu";
+import { FluxDispatcher, Menu } from "@webpack/common";
+import { User } from "discord-types/general";
+
+import { getLocalPronounOverrideNow, setLocalPronounOverride } from "./pronoundbUtils";
+import { PronounCode, PronounCodes, PronounMapping } from "./types";
+
+export const PronounOverrideContextMenu: NavContextMenuPatchCallback = (children, { user }: { user: User; }) => () => {
+    const userId = user.id;
+    // We cannot useAwaiter here for some reason (probably because this is not a React component, and we sadly can also not return a custom component and use useAwaiter in there, since we are in a context menu), so we have to use the cached version, which should be fine since whenever we right click someone we probably have loaded their profile.
+    const selected = getLocalPronounOverrideNow(userId);
+    const action = (newOverride: PronounCode | null) => () => {
+        setLocalPronounOverride(userId, newOverride);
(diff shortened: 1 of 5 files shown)


src/plugins/pronoundb/contextmenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { NavContextMenuPatchCallback } from "@api/ContextMenu";
+import { FluxDispatcher, Menu } from "@webpack/common";
+import { User } from "discord-types/general";
+
+import { getLocalPronounOverrideNow, setLocalPronounOverride } from "./pronoundbUtils";
+import { PronounCode, PronounCodes, PronounMapping } from "./types";
+
+export const PronounOverrideContextMenu: NavContextMenuPatchCallback = (children, { user }: { user: User; }) => () => {
+    const userId = user.id;
+    // We cannot useAwaiter here for some reason (probably because this is not a React component, and we sadly can also not return a custom component and use useAwaiter in there, since we are in a context menu), so we have to use the cached version, which should be fine since whenever we right click someone we probably have loaded their profile.
```
ActuallyTheSun, line 28: React hooks can be added at the top level of a context menu patch
```ts
const ExampleContextMenu: NavContextMenuPatchCallback = (children, props) => {
    const [value, error, pending] = useAwaiter();
    return () => { 
        // ...
```

src/plugins/pronoundb/contextmenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const selected = getLocalPronounOverrideNow(userId);
+    const action = (newOverride: PronounCode | null) => () => {
+        setLocalPronounOverride(userId, newOverride);
+        FluxDispatcher.dispatch({ type: "CONTEXT_MENU_CLOSE" });
+    };
+    console.log(selected);
```
ActuallyTheSun, line 34:
```suggestion
```

src/plugins/pronoundb/contextmenu.tsx
romangraef: Ahhh, oki

romangraef: While i can use react hooks in there, those react hooks do not seem to work? The actual components i push do not seem to be updated by the hooks i use in the outer function.
image

ArjixWasTaken: > and we sadly can also not return a custom component and use useAwaiter in there, since we are in a context menu)

I kinda doubt that, what prevents you from adding ur custom component as a child?

romangraef: Discord has an explicit check for the type of the children array.
image

Vendicated:
Look at CustomGreetSticker for an example for a stateful radio menu

src/plugins/pronoundb/contextmenu.tsx
ArjixWasTaken: well, can't you pass ur custom component as a child to a menu item?

romangraef:
> Look at CustomGreetSticker for an example for a stateful radio menu
I don't think thats applicable here since the problem is that i cannot use react hooks in the inner components, since i cannot use a custom component.

src/plugins/pronoundb/contextmenu.tsx
Vendicated: there actually is:
```jsx
<Menu.MenuItem
	render={MyCustomComponent}
	id="stuff"
/>
```
but there's no reason to use it, just use discord's

romangraef: > well, can't you pass ur custom component as a child to a menu item?

i dont think so

> but there's no reason to use it, just use discord's

using discords components does not allow me to use react hooks

romangraef:
Hey @Vendicated can you look over this again? I'll need to rebase this, but I don't think the issue with the underlying react hook issues was addressed.

Vendicated:
a looot of time has passed (my apologies) and you can use hooks in context menu patches now, so technically you could do that now

anyway, i'm not sure if we even need this feature in pronoundb? we just really need an EditUsers plugin, that could also cover editing anyone's pronouns even without the pronoundb plugin

Vendicated:
this plugin has been removed a long time ago. I added pronoun editing to #2657, still gotta finish that pr one day...

PR #1249, opened 2023-06-08
strikethrough silentTyping's icon when it's disabled instead of when enabled

thymue:
silent type is enabled and silent message is enabled as well in the picture below right? at least that's what my head tells me but instead silent type is disabled and silent message is enabled. wouldn't it be better if both were enabled when they are not striked-through?
image of the reason why my 2 brain cells are being confused asf as to what is happening
picuter of my 2 braincells being confused
bonus: enabled game activity toggle
listen, I'm already feeling like shit but this just does it. at this point I'm reaching never before seen levels of suffering


Vendicated:
not really because the icon used by the plugin is exactly that of silent messages

image


but i do get what you mean

thymue:
dunno I got my friend to install Vencord and one of the first things that came up was him saying 
image
so I was thinking it's not just me and maybe it is a good idea to have consistency on when a button should and should not be striked-through.
the [silentMessageToggle plugin](https://github.com/Vendicated/Vencord/blob/main/src/plugins/silentMessageToggle.tsx#L91), [gameActivityToggle plugin](https://github.com/Vendicated/Vencord/blob/main/src/plugins/gameActivityToggle/index.tsx#L40) both use a `!enabled` and I think this should as well since the messaging when you hover over it states "Enable/Disable Silent Typing" therefore the button should indicate whether the plugin is enabled or not which would mean, **not** striked-through when enabled and striked-through when disabled.

lewisakura:
This is fair, it confused me a lot too when I first used these. It makes sense from a developer standpoint but is counterintuitive in terms of UX.

PR #1263, opened 2023-06-11
ViewRaw: Add setting for swapping left/right click

RadNotRed:


src/plugins/viewRaw.tsx
```diff
@@ -116,26 +117,58 @@ function openViewRawModal(msg: Message) {
     ));
 }
 
+const settings = definePluginSettings({
+    clickMethod: {
+        description: "Click method",
```
Vendicated, line 122: this is very meaningless. You should give it a better name

src/plugins/viewRaw.tsx
```diff
@@ -116,26 +117,58 @@ function openViewRawModal(msg: Message) {
...
+        type: OptionType.SELECT,
+        options: [
+            { label: "Right Click", value: "Right", default: true },
```
Vendicated, line 125: Same as above. 
```suggestion
            { label: "Right Click to copy", value: "Right", default: true },
```

src/plugins/viewRaw.tsx
```diff
@@ -116,26 +117,58 @@ function openViewRawModal(msg: Message) {
...
         addButton("ViewRaw", msg => {
-            return {
-                label: "View Raw (Left Click) / Copy Raw (Right Click)",
-                icon: CopyIcon,
-                message: msg,
-                channel: ChannelStore.getChannel(msg.channel_id),
-                onClick: () => openViewRawModal(msg),
-                onContextMenu: e => {
+            const handleClick = () => {
+                if (Settings.plugins.ViewRaw.clickMethod === "Left") {
```
Vendicated, line 141:
```suggestion
                if (settings.store.clickMethod === "Left") {
```

src/plugins/viewRaw.tsx
```diff
@@ -116,26 +117,58 @@ function openViewRawModal(msg: Message) {
...
+                    copyWithToast(msg.content);
+                } else {
+                    openViewRawModal(msg);
+                }
+            };
+
+            const handleContextMenu = e => {
+                if (Settings.plugins.ViewRaw.clickMethod === "Right") {
```
Vendicated, line 149:
```suggestion
                if (settings.store.clickMethod === "Right") {
```

src/plugins/viewRaw.tsx
```diff
@@ -116,26 +117,58 @@ function openViewRawModal(msg: Message) {
...
                     e.stopPropagation();
                     copyWithToast(msg.content);
+                } else {
+                    e.preventDefault();
+                    e.stopPropagation();
+                    openViewRawModal(msg);
                 }
             };
+
+            const label = Settings.plugins.ViewRaw.clickMethod === "Left"
```
Vendicated, line 160:
```suggestion
            const label = settings.store.clickMethod === "Left"
```

Vendicated (changes requested):

src/plugins/viewRaw.tsx
RadNotRed: I realized that I was doing it backwards, so I just switched the buttons around because it's called viewRaw and not copyRaw

Vendicated (approved):

Vendicated:
Please don't pr from the main branch in the future

Vendicated:
thank you!!

PR #1272, opened 2023-06-11
Custom Slash Commands Plugin

stasiekkk:
Preview:

https://github.com/Vendicated/Vencord/assets/53151845/c1473699-c257-4f38-a239-1f9757420ae5


TheKodeToad:
:c I think you can still try and integrate some of this into the existing plugin

stasiekkk:
> :c I think you can still try and integrate some of this into the existing plugin

it made sense as a whole plugin made by me i guess ima do something different

PR #1303, opened 2023-06-17
Add HideWebhooks Plugin

9021007:
Here's my plugin, HideWebhooks. It automatically hides webhook messages, unless the webhook ID is added to the whitelist in the plugin preference pane. Enjoy!


9021007:
I have found a bug. oops.

9021007:
Fixed bug

Vendicated:
Hi! I don't think this should be merged because it's really niche and will just cause stupid people to enable it then be confused why some messages are gone. I can still give you feedback if you would like though, also if you would like to post it in #third-party-plugins just let me know!

PR #1315, opened 2023-06-20
[Plugin] Discord Custom Server Logos

kckarnige:
If any of the code sucks, let me know so I could fix it
I didn't originally code this for TS, but I've tested it and I know it functions as intended

### DIFF
--- a/src/plugins/discordCSL.ts
+++ b/src/plugins/discordCSL.ts
@@ -1,0 +1,128 @@
+import definePlugin from "@utils/types";
+
+const serverHeader = "container-1NXEtd";
+const serverName = "name-3Uvkvr";
+const loadingSplash = "container-2RRFHK";
+const manifestVersion = "5.0.0";
+
+// CDSL Console Log
+function log(input: string, type: string | null): any {
+  let color;
+  if (type == "info") {
+    color = "#43b581";
+  } else if (type == "warn") {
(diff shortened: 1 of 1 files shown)


Vendicated:
Please explain a bit what this plugin does

Also based on a very quick skim of the code, this would require a full rewrite to be anywhere near mergeable.
- You should not use raw dom manipulation, especially not via polling. Use patches instead
- This plugin seems to allow remote code execution which needs to be fixed
- You should use Vencord's css bundling instead of adding it manually

kckarnige:
Don't tell me that 😭
I'll see what I could do later on

kckarnige:
> This is not true. You fetch logoURL as text then set an element's innerHTML to this response. If logoURL returns vulnerable html like ``, it leads to remote code execution. image
> 
> You're also using svgs which are inherently vulnerable to xss [research.securitum.com/do-you-allow-to-load-svg-files-you-have-xss](https://research.securitum.com/do-you-allow-to-load-svg-files-you-have-xss/)
> 
> For proper security, you need to stop using svg and use something safe like png or webp

Oh damn, thanks for letting me know 😵‍💫

Vendicated:
Closing for the aforementioned reasons. Feel free to reopen if you're interested in fixing these issues!

PR #1319, opened 2023-06-21
Add News Ticker plugin

9021007:
This plugin adds a news ticker to the bottom of the screen with current headlines. You can click them to open their associated articles. 

HideWebhooks removed from branch


lewisakura (changes requested):
Please please please please please use React and don't write the HTML like this. Vencord provides the facilities for you to create React components explicitly to avoid this.

lewisakura (changes requested):
Use the new settings system. You can find an example [here](https://github.com/Vendicated/Vencord/blob/bf34b2ae43ed8c31665be35d3d6293f19ed0e80e/src/plugins/roleColorEverywhere.tsx#L24-L43).

Vendicated:
Hello!

I don't find this very useful. Why would someone want this instead of just opening reddit directly? The code is also very messy, using raw dom manipulation instead of proper patches

As such, I will unfortunately have to reject this. Thanks for your effort anyway!

PR #1320, opened 2023-06-21
VencordToolbox: rework to add and remove plugin settings (2: electric boogaloo)

Redundakitties:
Adds the ability to quick reload discord, toggle css, and check updater tab. Best part is you can add plugin settings to menu using a checkbox dropdown menu. 

You get all of the features of the old PR with all of the same things I do not know to fix! 
- contributors still become bots
- and checkbox menu scrolls to the top every time you add a new plugin


src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -182,9 +183,7 @@ function Newer(props: CommonProps) {
     );
 }
 
-function Updater() {
-    const settings = useSettings(["notifyAboutUpdates", "autoUpdate", "autoUpdateNotification"]);
-
+function updaterBody() {
```
Redundakitties, line 186: created something more generic that is used by both updater tab and the updater modal

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,189 @@
...
+    // for enabling and disabling Vencord-wide quick actions
+    relaunchDiscord: settingsSwitch("Relaunch Discord", "relaunchDiscord", "Quit and restart discord from toolbox", IS_WEB),
+    notifs: settingsSwitch("Open Notification Log", "notifs", "View notifications log from toolbox"),
+    quickCss: settingsSwitch("Edit QuickCss", "quickCss", "Edit QuickCss from toolbox"),
+    toggleQuickCss: settingsSwitch("Toggle QuickCss", "toggleQuickCss", "Enable/Disable QuickCss from toolbox"),
+    updater: settingsSwitch("UpdaterTab", "updater", "Open UpdaterTab from toolbox", IS_WEB),
+
+    // for enabling and disabling misc plugin quick actions
+    BadgeAPI: settingsSwitch("BadgeAPI", "BadgeAPI", "Refetch Badges from toolbox"),
+    DevCompanion: settingsSwitch("DevCompanion", "DevCompanion", "Reconnect Dev Companion from toolbox"),
```
Redundakitties, line 66: I'm unable to use camelCase here because later (in line 81) I'm taking the settingsName of Object.entries(settings.store) and pushing them to a list to later be added to the toolbox modal. Thus their name here has to match their plugin name. I could slap a .toLowerCase() maybe but this is simpler.

src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -197,6 +196,33 @@ function Updater() {
...
+            <Forms.FormTitle tag="h5">Updates</Forms.FormTitle>
+
+            {isNewer ? <Newer {...commonProps} /> : <Updatable {...commonProps} />}
+        </>
+    );
+}
+
+function Updater() {
+    const settings = useSettings(["notifyAboutUpdates", "autoUpdate", "autoUpdateNotification"]);
+    const updaterContent = updaterBody();
```
Vendicated, line 224: React components should always be created via jsx, not by calling it as a function:
```suggestion
    const updaterContent = <UpdaterBody />;
```

src/components/VencordSettings/UpdaterTab.tsx
Vendicated, line 186: React components should be named in PascalCase

```suggestion
function UpdaterBody() {
```

src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -224,29 +250,35 @@ function Updater() {
...
-                        ? "Failed to retrieve - check console"
-                        : (
-                            <Link href={repo}>
-                                {repo.split("/").slice(-2).join("/")}
-                            </Link>
-                        )
-                }
-                {" "}(<HashLink hash={gitHash} repo={repo} disabled={repoPending} />)
-            </Forms.FormText>
+            {updaterContent}
```
Vendicated, line 253: There doesn't seem to be any reason for this to require a variable, is there?
```suggestion
            <UpdaterContent />
```

src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -224,29 +250,35 @@ function Updater() {
...
+    return (
+        <ModalRoot {...modalProps} size={ModalSize.LARGE}>
+            <ModalHeader>
+                <Text variant="heading-lg/semibold" style={{ flexGrow: 1 }}>Vencord Updater</Text>
+                <ModalCloseButton onClick={close} />
+            </ModalHeader>
 
-            <Forms.FormTitle tag="h5">Updates</Forms.FormTitle>
+            <ModalContent>
+                {updaterContent}
```
Vendicated, line 268: same as above

src/components/VencordSettings/UpdaterTab.tsx
```diff
@@ -224,29 +250,35 @@ function Updater() {
...
-        </SettingsTab>
+        </ModalRoot>
     );
 }
 
+export function openUpdaterModal() {
+    const key = openModal(modalProps => (
+        <UpdaterModal
+            modalProps={modalProps}
+            close={() => closeModal(key)}
```
Vendicated, line 279: there is no need for manually closing. You should instead just use modalProps.onClose()

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,189 @@
...
+            >
+                {description}
+            </Switch>
+        )
+    };
+}
+
+const settings = definePluginSettings({
+    // for enabling and disabling Vencord-wide quick actions
+    relaunchDiscord: settingsSwitch("Relaunch Discord", "relaunchDiscord", "Quit and restart discord from toolbox", IS_WEB),
```
Vendicated, line 58: BOOLEAN settings have recently been changed to switches so using a custom component should now be obsolete

src/plugins/vencordToolbox/index.tsx
```diff
@@ -19,59 +19,189 @@
...
 import { LazyComponent } from "@utils/react";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType, PluginSettingDef } from "@utils/types";
 import { findByCode } from "@webpack";
-import { Menu, Popout, useState } from "@webpack/common";
+import { Alerts, Menu, Popout, Switch, useState } from "@webpack/common";
 import type { ReactNode } from "react";
 
 const HeaderBarIcon = LazyComponent(() => findByCode(".HEADER_BAR_BADGE,", ".tooltip"));
+const getAllPlugins = makeLazy(() => Object.values(Vencord.Plugins.plugins));
```
Vendicated, line 37: What's this for? It would be better to just call this at the top of the component

Vendicated (changes requested):

src/plugins/vencordToolbox/index.tsx
Redundakitties, line 1: Writing reminder to myself to fix needing to hardcode misc plugin actions because it's dumb

Redundakitties (commented):
- Made changes to UpdaterBody creating jsx components and removing custom close from modal
- changed the custom switch to default boolean function
- created a singular for loop to sort all the plugins to their correct display list
- now plugin actions can be added and removed using a checkbox system instead of settings

There are three types of toolbox items:
Plugin Custom Actions (toolbox actions custom set by plugins) 
Plugin Settings (pinning plugin settings for easy access)
App Tools (relaunching discord, checking for updates, Opening QuickCSS)

Issue #1299, opened 2023-06-16
[Feature Request] ClearURLs (Add function for fx and vx)


SkyRider13:
### Discord Account

skyrider1570

### What is it that you'd like to see?

The plugin is great but at the moment it seems to be unable to remove thesame tracking stuff from any twitter URL I post that uses "vx" and "fx" for example 
https://twittervx.com/Dexerto/status/1669345035639492609?t=2c86_jcBxl9gK7EhefNT6Q&s=33, 
https://fxtwitter.com/NoContextHumans/status/1669279210609729536?t=PDA9-QyuwOY03_IfwPy9pA&s=19 and
https://vxtwitter.com/NoContextHumans/status/1669279210609729536?t=PDA9-QyuwOY03_IfwPy9pA&s=19 
don't get cleared but a normal twitter link of each of these still does.

### Request Agreement

- [X] This is not a plugin request

Vendicated:
This feature request doesn't make a lot of sense. These tracking params do nothing on twitter alternatives, so stripping them is kinda senseless. Plus if you use the Vencord TextReplace plugin to convert twitter -> fxtwitter automagically, the stripping works correctly

PR #1322, opened 2023-06-21
Adding Additional Rules for #1299

KannaDev:
Basically added a few different URL's for the issue. No major changes.
https://github.com/Vendicated/Vencord/issues/1299

Side Note: Quite funny how I added myself to dev's when i only added a couple lines 😅


Vendicated:
see https://github.com/Vendicated/Vencord/issues/1299#issuecomment-1606150104

Vendicated:
ClearURLs should just allow you to add custom rules. The only reason it doesn't is because it's so old, plugin settings didn't exist back then

PR #1324, opened 2023-06-22
New Plugin: BetterSessions

aamiaa:
This plugin improves the sessions (devices) tab in settings:

- Each session gets an os-specific color and a platform-specific icon
- Sessions can be given custom names (indicated with a trailing asterisk)
- Session timestamp labels are replaced with the markdown timestamp component, which displays the exact date on hover
- New sessions get the "NEW" badge, which goes away when settings are closed

Platform icons were obtained from https://fontawesomeicons.com/svg/icons. All icons listed there are under a free open source license, so they should be fine to use.
Icon colors should work well with both dark and light theme.


src/plugins/betterSessions/components/RenameModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return (
+        <ModalRoot {...props}>
+            <ModalHeader>
+                <Forms.FormTitle tag="h4">Rename</Forms.FormTitle>
+            </ModalHeader>
+
+            <ModalContent>
+                <Forms.FormTitle tag="h5" style={{ marginTop: "10px" }}>New device name</Forms.FormTitle>
+                <TextInput
+                    inputRef={ref}
```
Vendicated, line 53: this ref + onChange usage setting a plain variable is really weird. you should use state instead of a ref and make use of the `value` prop

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { openModal } from "@utils/modal";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+import { Button, React, RestAPI, Tooltip } from "@webpack/common";
+
+import { RenameModal } from "./components/RenameModal";
+import { SessionInfo } from "./types";
+import { fetchNamesFromDataStore, getDefaultName, GetOsColor, GetPlatformIcon, savedSessionsCache, saveSessionsToDataStore } from "./utils";
+
+const UserSettingsAccountActionCreators = findByPropsLazy("saveAccountChanges", "open");
```
Vendicated, line 31:
```suggestion
const UserSettingsModal = findByPropsLazy("saveAccountChanges", "open");
```

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+        const newSessions = data.body.user_sessions.filter((session: SessionInfo["session"]) => !savedSessionsCache.has(session.id_hash));
+        for (const session of newSessions) {
+            savedSessionsCache.set(session.id_hash, { name: "", isNew: true });
+
+            showNotification({
+                title: "BetterSessions",
+                body: `New session:\n${session.client_info.os} · ${session.client_info.platform} · ${session.client_info.location}`,
+                permanent: true,
+                onClick: () => UserSettingsAccountActionCreators.open(Constants.UserSettingsSections.SESSIONS)
```
Vendicated, line 176: i would just hardcode these constants since they're never gonna change. It's more likely for this find to break, than for the constant value to change. Plus, it costs like 2-20ms to search for it (which adds up)
```suggestion
                onClick: () => UserSettingsAccountActionCreators.open("Sessions")
```

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+                },
+                {
+                    match: /({variant:"text-sm\/medium",className:\i\.sessionInfoRow,children:.{70,110}{children:"\\xb7"}\),\(0,\i\.\i\)\("span",{children:)(\i\[\d+\])}/,
+                    replace: "$1$self.renderTimestamp(arguments[0], $2)}"
+                },
+                // Replace the icon
+                {
+                    match: /(currentSession:null\),children:\[)\(0,\i\.\i\)\("div",{className:\w+\.sessionIcon,children:\(0,\i\.\i\)\(\i,{width:\i,height:\i}\)}\),/,
+                    replace: "$1$self.renderIcon(arguments[0]),"
+                }
```
Vendicated, line 78: these patches aren't that good because they are way more precise / complex than they need to be, but since this has gone so long without breakage, it's fine

just for future references, try to keep matches as short as possible (while still being unique and not overly generic) and don't match things you don't actually need to match. and try to avoid matching brackets since they are very likely to change after a bundler change or if they add anyhting new

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+                        backgroundColor: "#ED4245",
+                        marginLeft: "2px"
+                    }}
+                >
+                    NEW
+                </div>
+            );
+        }
+
+        return children;
```
Vendicated, line 109: this is not really how you're supposed to use react. Instead of plain functions, you should prefer Components. Instead of pushing to an array in if statements, you should just conditionally render jsx. For example, this function would be better written as
```jsx
return (
	<>
		<span>{name}</span>
		<RenameButton session={session} state={state} />

		{(savedSession == null || savedSession.isNew) && (
			<div ...>NEW</div>
		)}
	</>
);
```

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+        );
+    },
+
+    async checkNewSessions() {
+        const data = await RestAPI.get({
+            url: "/auth/sessions"
+        });
+
+        const newSessions = data.body.user_sessions.filter((session: SessionInfo["session"]) => !savedSessionsCache.has(session.id_hash));
+        for (const session of newSessions) {
```
Vendicated, line 169: this is mainly a nitpick so you can just ignore this, but first filtering then immediately looping over the array is a bit bad style imo. It's cleaner and more performant (it doesn't matter here but if you have very large arrays it could matter) to just add an if statement inside the loop, like
```js
for (const session of data.body.user_sessions) {
	if (savedSessionsCache.has(session.id_hash)) continue;
	...
}
```
or even
```js
for (const session of data.body.user_sessions) if (!savedSessionsCache.has(session.id_hash)) {
	...
}
```

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+            });
+        }
+
+        saveSessionsToDataStore();
+    },
+
+    flux: {
+        // Note: for some reason this is dispatched with a blank array when settings are closed, hence the length check later on
+        FETCH_AUTH_SESSIONS_SUCCESS({ sessions }: { sessions: SessionInfo["session"][]; }) {
+            lastFetchedHashes = sessions.map(session => session.id_hash);
```
Vendicated, line 186: any reason you're keeping track of this yourself instead of using `AuthSessionsStore.getSessions()`?

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+            });
+
+            // Remove names of sessions that were removed
+            if (lastFetchedHashes.length > 0) {
+                savedSessionsCache.forEach((_, idHash) => {
+                    if (!lastFetchedHashes.includes(idHash)) savedSessionsCache.delete(idHash);
+                });
+                lastFetchedHashes = [];
+            }
+            saveSessionsToDataStore();
```
Vendicated, line 205: this function is a bit hard to understand and seems to be overcomplicated

if im understanding correctly..

first you add every lastFetchedHash to savedSessionsCache if it doesnt already exist
then you change all savedSessions to not new
finally you remove all saved sessions that werent in lastFetched. so this now leaves just lastFetched
why not just clear the map and then add lastFetchedHashes to it? the outcome seems the same

src/plugins/betterSessions/README.md
```diff
@@ -0,0 +1,5 @@
+# BetterSessions
+
+Enhances the sessions (devices) menu. Allows you to view exact timestamps, give each session a custom name, and receive notifications about new sessions.
+
+![the plugin in action](https://github.com/aamiaa/Vencord/assets/9750071/811ad58a-78f3-4d2c-9623-7004c88f9c6a)
```
Vendicated, line 5: this alt text is kinda meaningless, so i'd just remove it
```suggestion
![](https://github.com/aamiaa/Vencord/assets/9750071/811ad58a-78f3-4d2c-9623-7004c88f9c6a)
```

src/plugins/betterSessions/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+            // (it allows changing width so we can mirror that logic)
+            find: "this.getBadgePositionInterpolation(",
+            replacement: {
+                match: /(\i\.animated\.rect,{id:\i,x:48-\(\i\+8\)\+4,y:)28(,width:\i\+8,height:)24,/,
+                replace: (_, leftPart, rightPart) => `${leftPart} 48 - ((this.props.lowerBadgeHeight ?? 16) + 8) + 4 ${rightPart} (this.props.lowerBadgeHeight ?? 16) + 8,`
+            }
+        }
+    ],
+
+    renderName({ session }: SessionInfo) {
```
Vendicated, line 92: if any of your renderX() functions error, it will likely crash discord

to prevent that from happening, you should wrap any components that have the potential to error (anything that relies on some discord module that might change / break. e.g. if your component just renders a plain div, thats safe) in an ErrorBoundary:
```jsx
renderName: ErorrBoundary.wrap(({ session }: SessionInfo) => {
}),
```
this will catch any erorrs in your component and by default instead render a red error card as you might have seen before. it would look very out of place so you can instead make it render nothing via `ErrorBoundary.wrap(..., { noop: true })`

Vendicated (commented):
changes look great, thank you!! just one more small thing

PR #1326, opened 2023-06-22
fix removing DMs from BetterFolders sidebar (fixes #1323)

### LINKED ISSUE #1323: [Bug] BetterFolders duplicate the DM section
### Discord Account

rinlovesyou

BetterFolders seems to copy over the top section of the server list, which is for DMs
image

It shouldn't do that

1. Enable BetterFolders
2. See the unexpected behavior

### Errors

_No response_

### Request Agreement

Juby210:


Vendicated:
Pushed in https://github.com/Vendicated/Vencord/commit/5a0e5018297e83877a8db7b323d3c265b7663393, thank you!

PR #1333, opened 2023-06-23
MessageLogger: Add "Clear Message Log History"

nmsturcke:
Clear the Message Log History for a whole channel/user/gdm with one click.

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -18,7 +18,7 @@
 
 import "./messageLogger.css";
 
-import { addContextMenuPatch, NavContextMenuPatchCallback, removeContextMenuPatch } from "@api/ContextMenu";
+import { addContextMenuPatch, findGroupChildrenByChildId, NavContextMenuPatchCallback, removeContextMenuPatch } from "@api/ContextMenu";
 import { Settings } from "@api/Settings";
 import { disableStyle, enableStyle } from "@api/Styles";
 import ErrorBoundary from "@components/ErrorBoundary";
@@ -26,7 +26,7 @@ import { Devs } from "@utils/constants";
 import { Logger } from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy } from "@webpack";
-import { FluxDispatcher, i18n, Menu, moment, Parser, Timestamp, UserStore } from "@webpack/common";
+import { FluxDispatcher, i18n, Menu, MessageStore, moment, Parser, Timestamp, UserStore } from "@webpack/common";
 
 import overlayStyle from "./deleteStyleOverlay.css?managed";
 import textStyle from "./deleteStyleText.css?managed";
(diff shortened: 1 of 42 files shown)


src/plugins/messageLogger/index.tsx
```diff
@@ -89,18 +89,63 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+                    toDelete.forEach(message => {
+                        FluxDispatcher.dispatch({
+                            type: "MESSAGE_DELETE",
+                            channelId: channel.id,
+                            id: message.id,
+                            mlDeleted: true
+                        });
+                    });
+                    toClearEditHistory.forEach(message => {
+                        message.editHistory = [];
```
Vendicated, line 124: this doesn't rerender messages so edits will not properly be removed, no?

src/plugins/messageLogger/index.tsx
```diff
@@ -89,18 +89,63 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+    const messages: Array<any> = MessageStore.getMessages(channel.id)?._array;
+
+    const toDelete: Array<any> = [];
+    const toClearEditHistory: Array<any> = [];
+    for (const message of messages) {
+        if (message.deleted) toDelete.push(message);
+        else if (message.editHistory?.length) toClearEditHistory.push(message);
+    }
+
+    if (!toDelete.length) return;
```
Vendicated, line 104: shouldn't this be
```suggestion
    if (!toDelete.length && !toClearEditHistory.length) return;
```

src/plugins/messageLogger/index.tsx
```diff
@@ -89,18 +89,63 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
     ));
 };
 
+const patchChannelContextMenu: NavContextMenuPatchCallback = (children, props) => () => {
```
Vendicated, line 92:
```suggestion
const patchChannelContextMenu: NavContextMenuPatchCallback = (children, { channel }) => () => {
```

Vendicated (changes requested):

src/plugins/messageLogger/index.tsx
nmsturcke: https://github.com/Vendicated/Vencord/assets/30734036/30e66d44-bea0-47a3-a567-d32dd0e28d4d

Vendicated: yes, you have to hover it to rerender

nmsturcke: Ah.. I think it updates when the message is hovered, but I basically just copied what the message CTX Menu does
7G3cV

Vendicated: yes, but that one is fine because it will always rerender the message after you close the context menu

not the same for the normal chat

src/plugins/messageLogger/index.tsx
nmsturcke: c35523ac343208947842b52a23ff087b60e859bc

src/plugins/messageLogger/index.tsx
nmsturcke: Do I just leave this PR sitting here for now?

PR #1336, opened 2023-06-24
New Plugin Called IP

madonchik123:
Plugin Name - IP
Plugin Description - Get details about IPv4/IPv6 address or a domain name.
What it does? Gets info about the address using slash command 
Example: /ip query:google.com


Vendicated:
this is kinda useless, why would anyone ever need this info as a discord command?

madonchik123:
To check suspicious links you know the people that install discord right? the nitro scams and etc

Vendicated:
That makes no sense. You should just check such link with something like virustotal / google safe browsing

madonchik123:
Not a lot of people want to do that steps and I made this plugin because yes

madonchik123:
And that's good idea to make it work with virustotal api

aamiaa:
How would checking the domain's ip help identify scams? They use the same hosting providers as normal sites, and the ips are often proxied too.

A virustotal integration seems pointless as well since you can just use a normal antivirus at this point.

Vendicated:
closing as this isn't useful. If someone really needs this for some reason, you can still just add a text replace rule that replaces smth like `ip:` with `http://ip-api.com/line/`:
`ip:130.61.130.176` -> http://ip-api.com/line/130.61.130.176

PR #1339, opened 2023-06-25
New Plugin: XSOverlay

PennyBunny:
This plugin adds the function to send message notifications to XSOverylay's Notification API over UDP


Vendicated:
why is udp needed here? You should just use plain fetch or a normal websocket

PennyBunny:
XSOverlay's notification system uses UDP, I don't really have a choice other than distributing a whole new executable to do IPC over fetch/ws to XSO's UDP server

PennyBunny:
Ok, that is everything this this needs I think. Should be ready unless there is any more changes you guys need.

src/main/ipcPlugins.ts
```diff
@@ -0,0 +1,33 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { IpcEvents } from "@utils/IpcEvents";
+import dgram from "dgram";
```
Vendicated, line 20: nit
```suggestion
import { createSocket } from "dgram";
```

src/main/ipcPlugins.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ipcMain } from "electron";
+
+ipcMain.handle(IpcEvents.DGRAM_SEND, (_, data) => dgramSend(data));
```
Vendicated, line 23: why not inline this?
```suggestion
ipcMain.handle(IpcEvents.DGRAM_SEND, (_, data) => {
	...
});
```

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,176 @@
...
+    timeout: {
+        type: OptionType.NUMBER,
+        description: "Time in seconds the notifcation will be displayed",
+        default: 5
+    },
+    opacity: {
+        type: OptionType.SLIDER,
+        description: "Opacity of the notifcation displayed",
+        default: 1,
+        markers: [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]
```
Vendicated, line 42:
```suggestion
        markers: makeRange(0, 1, 0.1)
```

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,176 @@
...
+                    icon: result,
+                    sourceApp: "Vencord"
+                };
+                VencordNative.pluginHelpers.dgramSend(data);
+            });
+        }
+    }
+});
+
+function supposedToNotify(message, channel) {
```
Vendicated, line 156:
```suggestion
function shouldNotify(message, channel) {
```

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (message.author.id === UserStore.getCurrentUser().id) return false;
+    if (MuteStore.allowAllMessages(channel)) return true;
+
+    return message.mentioned;
+}
+
+function calculateHeight(content) {
+    if (content.length <= 100) {
```
Vendicated, line 164: nit: These could just be plain ifs
```js
if (content.length <= 100) return 100;
if (content.length <= 200) return 150;
if (content.length <= 300) return 200;
return 250;
```

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,176 @@
...
+    } else if (content.length <= 200) {
+        return 150;
+    } else if (content.length <= 300) {
+        return 200;
+    }
+    return 250;
+}
+
+function clearMessage(content) {
+    return content.replace(new RegExp("<[^>]*>", "g"), "");
```
Vendicated, line 175: this seems to completely remove stuff like emojis, mentions, etc

Why not instead transform these to human readable ones? For mentions, the username, for emotes the emoji name, etc

Otherwise `@jeff told me that @john did a very bad thing` suddenly becomes `told me that did a very bad thing` and suddenly it makes no sense anymore

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,176 @@
...
+export default definePlugin({
+    name: "XSOverlay",
+    description: "Sends notifications to XSOverlay to be easier to see in VR",
+    authors: [Devs.Penny],
+    settings,
+    flux: {
+        MESSAGE_CREATE({ message }) {
+            var finalMsg = message.content;
+
+            const author = UserStore.getUser(message.author.id);
```
Vendicated, line 55: what's this for? message.author is already a full user

Vendicated (changes requested):

src/plugins/xsoverlay.discordDesktop.ts
PennyBunny: This is only used to calculate the height.

src/main/ipcPlugins.ts
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+import { IpcEvents } from "@utils/IpcEvents";
+import { createSocket } from "dgram";
+import { ipcMain } from "electron";
+
+ipcMain.handle(IpcEvents.DGRAM_SEND, (_, data) => {
+    data.icon = Buffer.from(data.icon).toString("base64");
+    data = JSON.stringify(data);
+    const client = createSocket("udp4");
```
Vendicated, line 26: wouldn't it be better to keep the socket alive instead of constantly recreating and closing it?

src/plugins/xsoverlay.discordDesktop.ts
```diff
@@ -0,0 +1,172 @@
...
+
+            if (message.embeds.length !== 0) {
+                finalMsg += " [embed] ";
+                if (message.content === "") {
+                    finalMsg = "[embed]";
+                }
+            }
+
+            if (message.stickers) {
+                finalMsg += " [sticker] ";
```
Vendicated, line 95: why not include the sticker name?

src/main/ipcPlugins.ts
PennyBunny: I did this so people who didn't have the plugin enabled would not have a socket sitting around, but I don't know much about ts so if its not a big deal to just make it at the beginning then I can do that.

src/plugins/xsoverlay.discordDesktop.ts
PennyBunny: it would just be message.stickers.name? right?

src/main/ipcPlugins.ts
Vendicated: you can just lazy initialise it
```js
let xsoSocket: SomeTypeIdk;

ipcMain.handle(... => {
	xsoSocket ??= createSocket("udp4");
})

Vendicated:
Is this even ready or still a draft? Considering there are several TODO comments

PennyBunny:
Sorry this is a draft, should I convert it to a draft pr?

Vendicated:
Yes please! That makes it easier for me to keep track. Let me know once it's ready :3

PR #1344, opened 2023-06-26
PronounDB: fix caching not respecting user preference

0xallie:
We don't need to cache Discord pronouns in the `cache` object, because they're already cached in the user object.

### DIFF
--- a/src/plugins/pronoundb/index.ts
+++ b/src/plugins/pronoundb/index.ts
@@ -52,7 +52,7 @@ export default definePlugin({
             find: ".userTagNoNickname",
             replacement: {
                 match: /=(\i)\.pronouns/,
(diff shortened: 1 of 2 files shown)


Vendicated:
What exactly does this achieve? Unless I'm mistaken, previously discord pronouns were already not added to our own cache

0xallie:
They are, which causes them to get cached and not be overridden with PronounDB pronouns even if you have it set to prefer PronounDB if you open their profile before the client encounters a message from them.

Vendicated:
I don't exactly remember why I did it but I'm pretty sure there's a reason I made it obtain pronouns from discord as opposed to just fetching them myself, so that might or might not be problematic

Vendicated:
Seems to work fine, thanks

PR #1348, opened 2023-06-26
feat(ShowMeYourName): compare username and nickname case-insensitively

0xallie:
Pomelo usernames are lowercase-only and case-insensitive.


TheKodeToad:
I wouldn't be surprised if by the time this is merged pomelo is extinct

Vendicated:
hi! why's this a draft? does anything else need to be done?

0xallie:
IIRC I was told on Discord that I should add an option to turn this behavior off.

PR #1353, opened 2023-06-27
feat(plugin): unlimitedAccounts

controlado:
This PR adds the unlimitedAccounts plugin.
It allows you to add more than 5 accounts to your client.

With the introduction of pomelos, this function has become very interesting for **collectors**, like me!
I really like Vencord and I don't want to use another mod because of this plugin, so I decided to do it.
(and I've also seen users asking for something to do this using Vencord)

I already ran the `pnpm test`

<div align="center">
  <table>
    <tr>
      <td>
        <div align="center">Accounts</div>
        Accounts
      </td>
      <td>
        <div align="center">Options</div>
        Options
      </td>
    </tr>
  </table>
</div>

### DIFF
--- a/src/plugins/unlimitedAccounts.ts
+++ b/src/plugins/unlimitedAccounts.ts
@@ -22,7 +22,7 @@ import definePlugin, { OptionType } from "@utils/types";
 
 const settings = definePluginSettings({
     maxAccounts: {
-        description: "Number of accounts that can be added or 0 for infinite",
+        description: "Number of accounts that can be added, or 0 for no limit",
         default: Number.POSITIVE_INFINITY,
         type: OptionType.NUMBER,
         restartNeeded: true,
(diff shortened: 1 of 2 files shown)


lewisakura:
I do not see the need for anyone to have >5 accounts in a single client, let alone own >5 accounts in general (except for plurality), and encouraging collectors with this sort of plugin means less usernames that people would like to use; genuinely good pomelo usernames have already been sniped and they will never get used or they will be resold for extortionate prices. I'm not sure if this is a good idea to put in the mod.

Personal opinions aside, if it's a cap remover, why does it have the choice to configure the cap? Seems odd to me due to the name.

Vendicated:
Like lewi said, there is no reason for anyone to need more than 5 accounts. I can only see this being used for malicious purposes, like name squatting and logging into stolen accounts

somerandomcloud:
Collecting Discord usernames is scummy in general...

PR #1357, opened 2023-06-27
feat(plugin): add importquickcss plugin

castdrian:
I've been adding a bunch of css snippets this past week and got the point of being incredibly annoyed by the process, so here;s a plugin that adds a context menu entry to messages that contain a css codeblock, and imports them according to one of three import strategies, to please all the ocd poatients out there, video attached

https://github.com/Vendicated/Vencord/assets/22133246/200bfbe1-6c41-413c-bb15-1ac79b55f456


castdrian:
now also supports multiple snippets per message (thanks ven)
Discord_psDbXvYaMI

src/plugins/importQuickCss.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { Menu, Toasts } from "@webpack/common";
+
+const enum AddStrategy {
+	Replace,
+	Append,
+	Prepend
+}
+
+function importCssSnippet(snippet: string, strategy: AddStrategy) {
+    VencordNative.quickCss.get().then(quickCss => {
```
Vendicated, line 33: using async await would avoid some nesting here

castdrian: done

castdrian:
some improvements, now the multiple import consists of nested elements, and instead of plain ol' numbers it shows a couple characters of the css as context
Discord_CW5XLHkBxc

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,306 @@
...
+*/
+
+import { addContextMenuPatch, NavContextMenuPatchCallback, removeContextMenuPatch } from "@api/ContextMenu";
+import * as DataStore from "@api/DataStore";
+import { definePluginSettings, Settings } from "@api/Settings";
+import { CSSFileIcon } from "@components/Icons";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { Button, Menu, Toasts } from "@webpack/common";
+import React from "react";
```
Vendicated, line 26:
```suggestion
```

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,306 @@
...
+    for (const snippetId of snippetIds) {
+        await removeCssSnippet(snippetId);
+    }
+};
+
+async function removeCssSnippet(snippetId: string) {
+    let quickCss = await VencordNative.quickCss.get();
+
+    const regex = new RegExp(
+        `\\/\\*\\nsnippet ${snippetId}[^]*?\\/\\* end snippet ${snippetId} \\*\\/`,
```
Vendicated, line 91: Escaping / is pointless, no? And to avoid having to double escape, you can just use String.raw
```suggestion
        String.raw`/\*\nsnippet ${snippetId}[^]*?/\* end snippet ${snippetId} \*/`,
```

Vendicated (changes requested):

src/VencordNative.ts
```diff
@@ -48,6 +48,16 @@ export default {
 
         addChangeListener(cb: (newCss: string) => void) {
             ipcRenderer.on(IpcEvents.QUICK_CSS_UPDATE, (_, css) => cb(css));
+
+            return {
+                remove: () => {
+                    ipcRenderer.removeListener(IpcEvents.QUICK_CSS_UPDATE, cb);
```
Vendicated, line 54: hmm. I don't think this will work like you expect it to. The .on creates a new array function wrapping cb, which differs from cb. You'll have to store the arrow function in a variable and use the same one in the remove

src/VencordNative.ts
```diff
@@ -48,6 +48,16 @@ export default {
...
+                },
+            };
+        },
+
+        removeChangeListener(listener: { remove: () => void }) {
+            listener.remove();
```
Vendicated, line 60: What's the point of this function? Can't you just call .remove() directly?

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+};
+
+
+const fetchSnippetIds = async () => {
+    cachedSnippetIds = await DataStore.get(STORE_KEY) || [];
+};
+
+const saveSnippetIds = async () => {
+    const storedSnippetIds = await DataStore.get(STORE_KEY) || [];
+    const newSnippetIds = cachedSnippetIds.filter(id => !storedSnippetIds.includes(id));
```
Vendicated, line 57: You're likely better off just storing a Set in the DataStore. This guarantees unique snippets and has way nicer methods for removing single items

This function is also not atomic, meaning it can lead to race conditions. DataStore.update should be used instead for an atomic update

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+    }
+};
+
+const removeAllSnippets = async (snippetIds: string[]) => {
+    for (const snippetId of snippetIds) {
+        await removeCssSnippet(snippetId);
+    }
+};
+
+const syncSnippetIds = async () => {
```
Vendicated, line 85: same as above, this function should use DataStore.update for an atomic update

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+        await DataStore.set(STORE_KEY, snippetIds.filter((id: string) => !snippetIdsToRemove.includes(id)));
+        await fetchSnippetIds();
+    }
+};
+
+const removeCssSnippet = async (snippetId: string) => {
+    let quickCss = await VencordNative.quickCss.get();
+
+    const regex = new RegExp(
+        String.raw`/\*\nsnippet ${snippetId}[^]*?/\* end snippet ${snippetId} \*/`,
```
Vendicated, line 101: what does this [^] do?

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+        type: Toasts.Type.SUCCESS,
+        id: Toasts.genId(),
+        options: {
+            duration: 2000,
+            position: Toasts.Position.BOTTOM,
+        },
+    });
+};
+
+const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
```
Vendicated, line 160: Unless you have a reason not to, why not destructure props right here? And type it to avoid bugs

```suggestion
const messageContextMenuPatch: NavContextMenuPatchCallback = (children, { message }: { message: Message }) => () => {
```

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+    const { message } = props;
+    const { content, timestamp } = message;
+
+    const hasCSSCodeblock = content.includes("```css\n") && content.includes("\n```");
```
Vendicated, line 164: codeblocks don't require a newline before the closing brackets so this is not fully correct

```suggestion
    const hasCSSCodeblock = /```css\n.+?```/s.test(content);
```
(You should likely store this regex in a global variable for better performance

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+    if (!hasCSSCodeblock) return;
+
+    const strategy = Settings.plugins.ManageQuickCSS.addStrategy;
+
+    const items: any[] = [];
```
Vendicated, line 169:
```suggestion
    const items: ReactNode[] = [];
```

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+    const snippets: { snippetId: string; snippet: string; }[] = [];
+
+    const re = /```css\n(.+?)```/gs;
+    let match: string[] | null;
+
+    const snippetIds = cachedSnippetIds;
+    const importableSnippets: { snippetId: string; snippet: string; }[] = [];
```
Vendicated, line 176: this inline type is used in at least 3 places. It would be better to make it an interface

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+        ]
+    }
+});
+
+export default definePlugin({
+    name: "ManageQuickCSS",
+    authors: [Devs.castdrian, Devs.Ven],
+    description: "Allows you to import and remove QuickCSS snippets contained within messages",
+    settings,
+    cssListener: undefined as ReturnType<typeof VencordNative.quickCss.addChangeListener> | undefined,
```
Vendicated, line 306:
```suggestion
    cssListener: null as ReturnType<typeof VencordNative.quickCss.addChangeListener> | null,
```

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+        this.cssListener = VencordNative.quickCss.addChangeListener(async () => {
+            await syncSnippetIds();
+        });
+    },
+
+    async stop() {
+        await saveSnippetIds();
+        removeContextMenuPatch("message", messageContextMenuPatch);
+
+        if (this.cssListener) {
```
Vendicated, line 321:
```suggestion
        this.cssListener?.remove();
```

src/plugins/manageQuickCss.tsx
```diff
@@ -0,0 +1,325 @@
...
+};
+
+const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
+    const { message } = props;
+    const { content, timestamp } = message;
+
+    const hasCSSCodeblock = content.includes("```css\n") && content.includes("\n```");
+    if (!hasCSSCodeblock) return;
+
+    const strategy = Settings.plugins.ManageQuickCSS.addStrategy;
```
Vendicated, line 167: this way of accessing plugin settings is deprecated since it is very error-prone. Instead, you should use the strongly typed settings.store where settings is your own settings object created via definePluginSettings

Vendicated (changes requested):

src/plugins/manageQuickCss.tsx
castdrian: shorthand for [\s\S]

Vendicated: just use `.` with the s flag

castdrian: oki doki, done

Vendicated:
hmm, i tried this and it just doesn't do anything for me 
image

castdrian:
well that's odd, uh, maybe the regex for the css codeblocks broke, will check in a bit

castdrian:
uh, hm, cannot replicate, latest commit off of the branch works for me
image

kingofcube:
doesnt work...

castdrian:
This branch hasn’t been updated in like half a year, not surprised, main problem is neither me nor Ven ever figured out what the issue was, so it’s been kinda stuck in limbo since

PR #1359, opened 2023-06-28
More consistent username formatting

mdashlw:
- In PermissionsViewer, don't use guild nickname because Discord itself doesn't in channel settings
- Fixed TypingIndicator not using global names

### DIFF
--- a/src/utils/discord.tsx
+++ b/src/utils/discord.tsx
@@ -18,13 +18,14 @@
 
 import { MessageObject } from "@api/MessageEvents";
 import { findByCodeLazy, findByPropsLazy, findLazy } from "@webpack";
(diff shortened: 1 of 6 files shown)


src/utils/discord.tsx
```diff
@@ -122,5 +123,20 @@ export async function openUserProfile(id: string) {
...
+
+/**
+ * Returns user's display name applicable for all contexts:
+ * - guild nickname (if guildId is provided)
+ * - friend nickname (only if guildId is not provided)
+ * - globalName
+ * - username (pomelo)
+ * - username#1234
+ */
+export function getDisplayName(guildId: string | null, userId: string): string {
```
Vendicated, line 137: wouldn't it make more sense to swap these two parameters? we're getting the user's displayName afterall, not the guild's. this way, the second parameter could also be optional, not requiring you to pass null

furthermore, maybe instead of the userId, it should take a user object? seems more consistent with discord's functions and also the getUniqueUsername function above

Vendicated (commented):
thank you, looks good!

could you please resolve conflicts? my apologies for only getting to this now!

PR #1382, opened 2023-07-05
feat(cloud): support multiple user accounts

lewisakura:
This modifies the key that is used to store secrets to include the current user ID. This may have issues on first migration if the current user, for instance, is on the wrong account. In this case, running this in DevTools will wipe out the existing secrets and allow a user to reconfigure without erasing their data:

```javascript
Vencord.Api.DataStore.del("Vencord_cloudSecret");
```

### DIFF
--- a/src/utils/cloud.tsx
+++ b/src/utils/cloud.tsx
@@ -22,17 +22,13 @@ import { Settings } from "@api/Settings";
 import { findByProps } from "@webpack";
 import { UserStore } from "@webpack/common";
 
-import { proxyLazy } from "./lazy";
 import { Logger } from "./Logger";
 import { openModal } from "./modal";
 
 export const cloudLogger = new Logger("Cloud", "#39b7e0");
-export const cloudUrl = proxyLazy(() => new URL(Settings.cloud.url));
+export const getCloudUrl = () => new URL(Settings.cloud.url);
 
-const cloudUrlOrigin = proxyLazy(() => cloudUrl.origin);
-
-// can't make this lazy as this may change at runtime (user switches accounts), therefore using a cached value is
-// incorrect
+const cloudUrlOrigin = () => getCloudUrl().origin;
 const getUserId = () => {
     const { id } = UserStore.getCurrentUser();
     if (!id) throw new Error("User not yet logged in");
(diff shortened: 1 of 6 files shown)


src/utils/cloud.tsx
```diff
@@ -28,23 +28,40 @@ import { openModal } from "./modal";
...
 export const getCloudUrl = () => new URL(Settings.cloud.url);
 
+const getUserId = () => UserStore.getCurrentUser().id;
+
 export async function getAuthorization() {
     const secrets = await DataStore.get<Record<string, string>>("Vencord_cloudSecret") ?? {};
-    return secrets[getCloudUrl().origin];
+
+    // we need to migrate from the old format here
+    if (secrets[getCloudUrl().origin]) {
```
Vendicated, line 37: you have 6 instances of `getCloudUrl().origin`. I think you should extract it into a variable

src/utils/cloud.tsx
```diff
@@ -28,23 +28,40 @@ import { openModal } from "./modal";
 export const cloudLogger = new Logger("Cloud", "#39b7e0");
 export const getCloudUrl = () => new URL(Settings.cloud.url);
 
+const getUserId = () => UserStore.getCurrentUser().id;
```
Vendicated, line 31: this might technically return undefined in case the user is logged out. Some handling would probably be good. I think just a throw in getAuthorization() should be good enough?

src/utils/cloud.tsx
lewisakura: Admittedly the alternative of this isn't much cleaner (`const cloudUrlOrigin = () => getCloudUrl().origin;`) but I suppose it is better in readability.

Vendicated: did you forget variables exist? :P 

```js
const { origin } = getCloudUrl();
```

lewisakura: shhhhhhhhhh

src/utils/cloud.tsx
```diff
@@ -28,23 +28,45 @@ import { openModal } from "./modal";
 export const cloudLogger = new Logger("Cloud", "#39b7e0");
 export const getCloudUrl = () => new URL(Settings.cloud.url);
 
+const cloudUrlOrigin = () => getCloudUrl().origin;
+const getUserId = () => {
+    const { id } = UserStore.getCurrentUser();
```
Vendicated, line 33:
```suggestion
    const id = UserStore.getCurrentUser()?.id;
```

Vendicated:
thank you!

PR #1384, opened 2023-07-06
feat(ClearURLs): add Threads share link tracking param

ryanccn:


Vendicated:
at this point i feel like just adding settings to add custom rules to the plugin is the better option, otherwise we could add thousands of params

PR #1386, opened 2023-07-06
feat(plugin): Favorite Gif Search

Syncxv:
lets you search your favorite gifs

idea: https://github.com/Vendicated/Vencord/discussions/1385

https://github.com/Vendicated/Vencord/assets/47534062/e4905548-0800-4b10-ba09-ff15da6f16da


Vendicated:
holy shit

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,111 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findByPropsLazy, findLazy } from "@webpack";
+import { useCallback, useState } from "@webpack/common";
+
+
+interface Sizes {
```
Vendicated, line 25: nit: this could be better written as
```suggestion
type Sizes = Record<"SMALL" | "MEDIUM" | "LARGE", string>;
```

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,111 @@
...
+            find: "renderCategoryExtras",
+            replacement: {
+                match: /(renderHeader=function.{1,500}return)(.{1,100}renderHeaderContent.{1,50})}/,
+                replace: "$1[$self.renderSearchBar(this), $2]}"
+            }
+        }
+    ],
+
+
+    renderSearchBar: (instance: Instance) => instance.state?.resultType === "Favorites" ? <SearchBar instance={instance} /> : null
```
Vendicated, line 79:
```suggestion
    renderSearchBar: ErrorBoundary.wrap((instance: Instance) => instance?.state?.resultType === "Favorites" ? <SearchBar instance={instance} /> : null, { noop: true })
```

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,111 @@
...
+});
+
+
+function SearchBar({ instance }: { instance: Instance; }) {
+    const [query, setQuery] = useState("");
+
+    const onChange = useCallback((quwery: string) => {
+        setQuery(quwery);
+        const { props } = instance;
+        if (!props.originalFav) props.originalFav = props.favorites;
```
Vendicated, line 89:
```suggestion
        props.originalFav ||= props.favorites;
```

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,111 @@
...
+    name: "FavoriteGifSearch",
+    authors: [Devs.Aria],
+    description: "Adds a search bar for favorite gifs",
+
+    patches: [
+        {
+            find: "renderCategoryExtras",
+            replacement: {
+                match: /(renderHeader=function.{1,500}return)(.{1,100}renderHeaderContent.{1,50})}/,
+                replace: "$1[$self.renderSearchBar(this), $2]}"
```
Vendicated, line 73: i don't like the way this patch works. Wouldn't it be way easier to patch renderHeaderContent?

the FAVORITES case (frame 1) would only need to be changed to return the SearchBar similar to in frame 2. Then you could even capture the SearchBar component instead of having to webpack find it

image

Vendicated (changes requested):

src/plugins/favGifSearch.tsx
Syncxv: alright i tried doing that. what do you think?

MaddyUnderStars:
holy shit that was fast. thanks

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+        props.favorites = props.originalFav.filter(gif => {
+            const url = new URL(gif.url ?? gif.src);
+            switch (settings.store.searchOption) {
+                case "both":
+                    return url.href.toLowerCase().includes(quwery.toLowerCase());
+                case "path":
+                    return url.pathname.toLowerCase().includes(quwery.toLowerCase());
+                case "origin":
+                    return url.origin.toLowerCase().includes(quwery.toLowerCase());
+            }
```
MaddyUnderStars, line 125: while using `.includes` works, wouldn't it be better to do, say, fuzzy find instead? or similar? perhaps a lil sneaky matching of individual words? idk.
otherwise searching for gifs you don't remember the URL of will be slightly annoying, you'll need to search for 1 word only since you might be wrong on the order of words perhaps. also it prevents queries like `tenor cat` or `discord dog`.
also also shouldn't you replaceAll `-` with spaces for convenience?

regardless: thank you very much. this has bothered me for so long lol

Syncxv: thanks

Syncxv:
alright i think adding the dropdown in the gif picker is a bad idea.

The problem is that the Select component uses a popout to render its items, and having a popout within a popout makes it difficult to select any of the items in the dropdown

I have a fix but implementing it in this plugin may not be the best idea

so basically its because of `target === g.current` the target doenst match the current active element, which makes discord think we clicked outside the gif picker and causes it to be closed.

```js
// WebpackModule656823 line 7440
handleDocumentMouse = o.useCallback(function (e) {
    if (!hasAnyModalOpen()) {
        var target = e.target;
        if (/* ... */) {
            for (; (0, ko.k)(target); ) {
                // HERE is the issue. when we click the options in the select dropdown these 2 dont match and closes the gif picker
                if (target === g.current) return;
                // we need to add data-menu-item="true" to the options but we can only render teh labels not the actual option container :pensive:
                if ("true" === target.getAttribute("data-menu-item")) return;
                // ...
                target = target.parentNode;
            }
            (0, Do._Q)(); // Close gif picker
            // ...
        }
    }
}, [/* ... */]);

o.useLayoutEffect(function () {
    // ...
    P.addEventListener("mousedown", handleDocumentMouse);
    P.addEventListener("contextmenu", handleDocumentMouse);
    // ...
    return function () {
        P.removeEventListener("mousedown", handleDocumentMouse);
        P.removeEventListener("contextmenu", handleDocumentMouse);
        // ...
    };
}, [/* ... */]);


```

so the solution is either to patch the select component and add `data-menu-item="true"` to each option container or patch this function
i dont like either of these options and i think we shouldnt add the dropdown

I tried to fix it by using the renderOptionLabel prop of the Select component, but this did not work as expected. The first element does not work when other items are selected. also if we do `pointer-events: all` then the hover stuff wont work for the options 😔 

Here is the code I tried
```tsx
<Select
    // ...
    renderOptionLabel={option => {
        return (
            <>
                <div data-menu-item={true} style={{ position: "absolute", top: "0", left: "0", width: "100%", height: "100%", zIndex: "9999", pointerEvents: "none" }} />
                <div data-menu-item={true}>
                    {option.label}
                </div>
            </>
        );
    }}
    // ...
/>
```

but if there is another way to render the dropdown in a popout let me know

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,218 @@
...
+            {
+                label: "Entire Url",
+                value: "url"
+            },
+            {
+                label: "Path Only (/somegif.gif)",
+                value: "path"
+            },
+            {
+                label: "Host Only (tenor.com)",
```
Vendicated, line 78: this honestly seems useless

src/plugins/favGifSearch.tsx
```diff
@@ -0,0 +1,286 @@
...
+const bitFlagsCache = new Map<string, number>(); // (str, flags)
+
+const containerClasses: { searchBar: string; } = findByPropsLazy("searchBar", "searchHeader");
+
+export const settings = definePluginSettings({
+    searchOption: {
+        type: OptionType.SELECT,
+        description: "The part of the url you want to search",
+        default: "hostandpath",
+        options: [
```
Vendicated, line 70: shouldn't this specify a default option?

Vendicated:
thank you! i am too lazy to review further so i will just merge :trollface:

PR #1387, opened 2023-07-06
New Plugin: UserProfile

Duckulus:


Vendicated:
Hi! This is not really substantial enough to be a plugin. Perhaps add it to moreCommands instead?

PR #1389, opened 2023-07-06
Update quickCss.ts

LaptopCat:
fixes really really tiny issue that would update the style when editing quickcss even though quickcss is disabled in settings

### DIFF
--- a/src/utils/quickCss.ts
+++ b/src/utils/quickCss.ts
@@ -28,7 +28,11 @@ export async function toggle(isEnabled: boolean) {
             style = document.createElement("style");
             style.id = "vencord-custom-css";
             document.head.appendChild(style);
-            VencordNative.quickCss.addChangeListener(css => style.textContent = css);
+            VencordNative.quickCss.addChangeListener(css => {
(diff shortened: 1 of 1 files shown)


Vendicated:
hello!

how exactly is this an issue? does style.disabled not prevent the css from having any effect?

LaptopCat:
https://github.com/Vendicated/Vencord/assets/98688203/0a755041-875f-4a51-aee2-d141a5ac3af7

LaptopCat:
steps to reproduce issue:

1. disable custom css (quickcss) in settings
if you didnt have it enabled, enable it and then disable it

2. click on the "open quickcss file" button

3. edit something

4. changes get applied, though custom css is disabled

Vendicated:
Thank you! This seems to be a bug / quirky oversight in the html spec
For some reason setting style.textContent resets style.disabled

In any case, this fix is not optimal because it will potentially lead to stale content. I have pushed a better fix in https://github.com/Vendicated/Vencord/commit/2a318e390e77f8269fcba45ab9a0fd7008ca8d0a

PR #1408, opened 2023-07-11
Add plugin: richerAppleMusic.desktop.tsx

Arch-Storm:
This adds better rpc integration for Apple Music (Preview) and the unofficial rpc client [AMWin-RP](https://github.com/PKBeam/AMWin-RP/). This is mostly a direct copy from the old richerCider plugin. I made this after hearing about their opinions and uninstalling that app for good.


Vendicated:
Hi! The original plugin was in part removed because it was an overly specific plugin and only covers one single niche use case. This plugin suffers from the same problem. It would be much better to make a more general purpose plugin that allows you to modify all sorts of Activities, not just this specific one

PR #1416, opened 2023-07-13
Add "Show New" option in plugin settings

MrDiamondDog:
First contribution, resolves #323 

Adds a very simple "Show New" option in plugin settings that will show plugins with the "New" badge.

### DIFF
--- a/src/components/PluginSettings/index.tsx
+++ b/src/components/PluginSettings/index.tsx
@@ -230,7 +230,7 @@ export default function PluginSettings() {
         const enabled = settings.plugins[plugin.name]?.enabled;
         if (enabled && searchValue.status === SearchStatus.DISABLED) return false;
         if (!enabled && searchValue.status === SearchStatus.ENABLED) return false;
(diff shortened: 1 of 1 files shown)


src/components/PluginSettings/index.tsx
```diff
@@ -229,6 +230,7 @@ export default function PluginSettings() {
         const enabled = settings.plugins[plugin.name]?.enabled;
         if (enabled && searchValue.status === SearchStatus.DISABLED) return false;
         if (!enabled && searchValue.status === SearchStatus.ENABLED) return false;
+        if (!newPlugins?.includes(plugin.name) && searchValue.status === SearchStatus.NEW) return false;
```
Vendicated, line 233:
```suggestion
        if (searchValue.status === SearchStatus.NEW && !newPlugins?.includes(plugin.name)) return false;
```

Vendicated (approved):

Vendicated:
Thank you

PR #1455, opened 2023-07-19
fix(customRPC): replace empty string with undefined

Ju1-js:
### Changes
- Fix the blank tooltip showing over the large image when hovered.
image -> image
*The cursor was hovering over the images when the screenshots were taken, it is just not in the images.*

- The small image's tooltip doesn't exhibit this behaviour, so no extra code was needed.

#### To reproduce the bug:
1. Set large image's tooltip to anything (other than undefined/null), and save.
2. Toggle plugin.
3. Set tooltip to nothing, and save.
4. Toggle plugin.


Vendicated:
superseded by #1481

PR #1469, opened 2023-07-23
VoiceMessages UI

Vap0r1ze:
Adds waveforms and VoiceRecording preview component

### DIFF
--- a/src/plugins/voiceMessages/VoicePreview.tsx
+++ b/src/plugins/voiceMessages/VoicePreview.tsx
@@ -1,0 +1,60 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { findByCodeLazy } from "@webpack";
+import { useEffect, useState } from "@webpack/common";
+import { ComponentType } from "react";
+
+import { cl } from "./utils";
+
+type VoiceMessage = ComponentType<{
+    src: string;
+    waveform: string;
+}>;
+const VoiceMessage: VoiceMessage = findByCodeLazy('["onVolumeChange","volume","onMute"]');
+
+export type VoicePreviewOptions = {
+    src?: string;
+    waveform: string;
+    recording?: boolean;
+};
+export const VoicePreview = ({
+    src,
+    waveform,
+    recording,
+}: VoicePreviewOptions) => {
+    const [seconds, setSeconds] = useState(0);
+
+    useEffect(() => {
+        if (!recording) return;
+
+        const interval = setInterval(() => {
+            setSeconds(s => s + 1);
+        }, 1000);
+        return () => clearInterval(interval);
+    }, [recording]);
+
+    const duration = Math.floor(seconds / 60) + ":" + (seconds % 60).toString().padStart(2, "0");
+
+    if (src && !recording) return <VoiceMessage key={src} src={src} waveform={waveform} />;
+    return <div className={cl("preview", recording ? "preview-recording" : [])}>
+        <div className={cl("preview-indicator")}></div>
+        <div className={cl("preview-time")}>{duration}</div>
+        <div className={cl("preview-label")}>{recording ? "RECORDING" : "----"}</div>
+    </div>;
+};
--- a/src/plugins/voiceMessages/VoicePreview.tsx
+++ b/src/plugins/voiceMessages/VoicePreview.tsx
@@ -16,9 +16,8 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { LazyComponent } from "@utils/react";
+import { LazyComponent, useTimer } from "@utils/react";
 import { findByCode } from "@webpack";
-import { useEffect, useMemo, useState } from "@webpack/common";
 
 import { cl } from "./utils";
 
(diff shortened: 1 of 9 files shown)


Vap0r1ze:
this is ready now

src/plugins/voiceMessages/index.tsx
```diff
@@ -99,28 +114,59 @@ function useObjectUrl() {
...
+            let squares = 0;
+            for (let sampleOffset = 0; sampleOffset < samplesPerBin; sampleOffset++) {
+                const sampleIdx = binIdx * samplesPerBin + sampleOffset;
+                squares += channelData[sampleIdx] ** 2;
+            }
+            bins[binIdx] = ~~(Math.sqrt(squares / samplesPerBin) * 0xFF);
+        }
+
+        return {
+            waveform: btoa(String.fromCharCode(...bins)),
```
Vendicated, line 148:
```suggestion
            waveform: window.btoa(String.fromCharCode(...bins)),
```

src/plugins/voiceMessages/VoicePreview.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useEffect, useState } from "@webpack/common";
+import { ComponentType } from "react";
+
+import { cl } from "./utils";
+
+type VoiceMessage = ComponentType<{
+    src: string;
+    waveform: string;
+}>;
+const VoiceMessage: VoiceMessage = findByCodeLazy('["onVolumeChange","volume","onMute"]');
```
Vendicated, line 29: you should use LazyComponent for components

src/plugins/voiceMessages/WebRecorder.tsx
```diff
@@ -16,16 +16,21 @@
...
 import type { VoiceRecorder } from ".";
 
-export const VoiceRecorderWeb: VoiceRecorder = ({ setAudioBlob }) => {
+export const VoiceRecorderWeb: VoiceRecorder = ({ setAudioBlob, setIsRecording }) => {
     const [recording, setRecording] = useState(false);
     const [paused, setPaused] = useState(false);
     const [recorder, setRecorder] = useState<MediaRecorder>();
     const [chunks, setChunks] = useState<Blob[]>([]);
 
+    useEffect(() => {
```
Vendicated, line 29: this is kinda bad, i would just move the recording state into the parent and pass both value and setter to child

Vap0r1ze: the reason i didnt do that is cus theres the case that the recorder cant record, so the recorder should be in control of the recording state. there definitely is a better way to go about it still though

Vap0r1ze: i pushed a slight change to the use effect idea, not much better but yea idk what else to do without too much thinking

src/plugins/voiceMessages/index.tsx
```diff
@@ -114,7 +116,7 @@ function useObjectUrl() {
 }
 
 function Modal({ modalProps }: { modalProps: ModalProps; }) {
-    const [isRecording, setIsRecording] = useState(false);
+    const [isRecording, setRecording] = useState(false);
```
Vendicated, line 119: why the rename?

src/plugins/voiceMessages/settings.ts
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+import { definePluginSettings } from "@api/Settings";
+import { OptionType } from "@utils/types";
+
+export const settings = definePluginSettings({
+    noiseSuppression: {
+        type: OptionType.BOOLEAN,
+        description: "Noise Suppression",
+        defaultValue: true,
```
Vendicated, line 26: 🤨 isn't it default

Vap0r1ze: isnt what default

Vendicated:
```suggestion
        default: true,
```

Vap0r1ze: oh lmaoo

Vendicated:
🔥 🔥 👯‍♀️ 🚒 💣 👍🏻 😸

src/plugins/voiceMessages/VoicePreview.tsx
```diff
@@ -0,0 +1,60 @@
...
+    waveform,
+    recording,
+}: VoicePreviewOptions) => {
+    const [seconds, setSeconds] = useState(0);
+
+    useEffect(() => {
+        if (!recording) return setSeconds(0);
+
+        const interval = setInterval(() => {
+            setSeconds(s => s + 1);
```
Vendicated, line 47: nit: this is technically an anti pattern since intervals will slowly drift (aka not be consistently invoked every 1000ms). To avoid drift you should generally store the start timestamp and always recalculate this. But it also probably doesnt matter in this case

src/plugins/voiceMessages/VoicePreview.tsx
```diff
@@ -0,0 +1,60 @@
...
+        }, 1000);
+        return () => clearInterval(interval);
+    }, [recording]);
+
+    const duration = Math.floor(seconds / 60) + ":" + (seconds % 60).toString().padStart(2, "0");
+
+    if (src && !recording) return <VoiceMessage key={src} src={src} waveform={waveform} />;
+    return <div className={cl("preview", recording ? "preview-recording" : [])}>
```
Vendicated, line 55: CODE NEEDS ROOM TO BREATHE ITS SUFFOCATING :(((((((((( 💔 😦 

```suggestion
    return (
    	<div className={cl("preview", recording ? "preview-recording" : [])}>
```

src/plugins/voiceMessages/index.tsx
```diff
@@ -23,25 +23,33 @@ import { Flex } from "@components/Flex";
...
+    setAudioBlob(blob: Blob): void;
+    onRecordingChange?(recording: boolean): void;
+}>;
 
 const VoiceRecorder = IS_DISCORD_DESKTOP ? VoiceRecorderDesktop : VoiceRecorderWeb;
 
 export default definePlugin({
     name: "VoiceMessages",
     description: "Allows you to send voice messages like on mobile. To do so, right click the upload button and click Send Voice Message",
     authors: [Devs.Ven],
```
Vendicated, line 51:
```suggestion
    authors: [Devs.Ven, Devs.Vap],
```

src/plugins/voiceMessages/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { classNameFactory } from "@api/Styles";
+
+export const cl = classNameFactory("vc-vmsg-");
+export const clamp = (n: number, min: number, max: number) => Math.min(Math.max(n, min), max);
```
Vendicated, line 22: i feel like this could be @utils/misc instead

Vendicated (approved):
looks really good other than the aforementioned nits thank you a lot

src/plugins/voiceMessages/VoicePreview.tsx
Vap0r1ze: based

Vap0r1ze:
can u test that commit

Vendicated:
no :heart:

PR #1470, opened 2023-07-23
New Plugin: DownloadVoiceMessage

steveruu:


Vendicated:
you can just do this via ViewRaw so this plugins is not very useful. Thank you anyway!

PR #1473, opened 2023-07-24
feat(plugin): UrlShortener

stanislawkuriata:
Simple and fast url shortener


Vendicated:
why would anyone need this?

Thanks anyway

PR #1487, opened 2023-07-26
fix: open links externally in Quick CSS editor

ryanccn:


src/main/utils/externalLinks.ts
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { type BrowserWindow, shell } from "electron";
+
+export function makeLinksOpenExternally(win: BrowserWindow) {
+    win.webContents.setWindowOpenHandler(({ url }) => {
+        switch (url) {
+            case "about:blank":
+            case "https://discord.com/popout":
```
Vendicated, line 25: canary & ptb

Vendicated: well actually this isn't really needed cause the quick css editor will never open this but its whatever

Vendicated:
uhoh conflict juice

PR #1497, opened 2023-07-26
feat(plugin): showMoreLines

lumap:
This plugin removes the 100 lines limit of Discord's file preview

Before:

After:


lumap:
this shit makes discord even more laggy, might wait a few years for electron to be good in term of
performance aka never to make this possible.

Vap0r1ze:
i should add file previews as a shiki feature to make discord's performance better

PR #1505, opened 2023-07-26
Wikisearch plugin: add a send article link setting

lumap:
This technically reopens #1454 because I sorted my fork and deleted/renamed the wrong branches.

This opt-in option makes Clyde send the article URL and embed instead of the fancy embed Clyde sends. This makes it far easier to copy an article link and send it in chat.


Vendicated:
this PR has a lot of unnecessary whitespace changes. please restructure the code to avoid these changes by using an early return instead of an if else:
```js
if (setting) {
    doStuff();
    return;
}

old code goes here
```

src/plugins/wikisearch.ts
```diff
@@ -18,13 +18,22 @@
 
 import { ApplicationCommandInputType, ApplicationCommandOptionType, findOption, sendBotMessage } from "@api/Commands";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
+import { RestAPI } from "@webpack/common";
+import { Settings } from "Vencord";
```
Vendicated, line 23:
```suggestion
import { Settings } from "@api/settings";
```

src/plugins/wikisearch.ts
```diff
@@ -47,63 +56,94 @@ export default definePlugin({
...
-                if (!data) return;
+                    if (!data) {
+                        return;
+                    }
 
-                if (!data.query?.search?.length) {
-                    console.log(data);
-                    return sendBotMessage(ctx.channel.id, { content: "No results given" });
-                }
+                    const url = data[3][0];
```
Vendicated, line 71: this seems kind of unsafe. Are you sure it will always be exactly like this? What if there are no results? A .find() would likely be more appropriate

lumap: Their API hasn't changed in the slightest in the past few years but I guess I could implement a .find just to be sure. Will do in a bit

lumap:
> this PR has a lot of unnecessary whitespace changes. please restructure the code to avoid these changes by using an early return instead of an if else:
> 
> ```js
> if (setting) {
>     doStuff();
>     return;
> }
> 
> old code goes here
> ```

Committed a fix, tell me what you think about it

lumap:
no one;
GitHub: let's make a review that is marked as resolved not count towards the reviews needed for merging

lumap:
cba to rewrite since rspack + useless

PR #1516, opened 2023-07-27
feat(loadingQuotes) more Quotes for the Plugin

Clone69420:
- [x] Only show famous quotes
- [x] more text


Clone69420:
- [ ] Loading Text Color Options?
- [ ] Loading Text Size Options?
- [ ] when loading play a custom sound?
- [ ] more text

PR #1544, opened 2023-07-28
VoiceMessages: Improve permission check

lumap:
The upload button is not supposed to appear in channels in which you can't send messages. However, misconfigured channel permissions (or intentional idk) can make it appear thanks to one perm: create threads. In this case, the "Send voice message" button would still appear, but voice messages can't be sent in that channel. This fixes it.

Note: The third commit undoes the first commit since I suck at git so you can ignore those two

### DIFF
--- a/src/plugins/voiceMessages/index.tsx
+++ b/src/plugins/voiceMessages/index.tsx
@@ -215,7 +215,7 @@ function Modal({ modalProps }: { modalProps: ModalProps; }) {
 }
 
 const ctxMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
-    if (props.channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel)) return;
+    if (props.channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel) || props.channel.guild.id && !PermissionStore.can(PermissionsBits.SEND_MESSAGES, props.channel)) return;
 
     children.push(
         <Menu.MenuItem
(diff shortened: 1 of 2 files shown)


src/plugins/voiceMessages/index.tsx
```diff
@@ -215,7 +215,7 @@ function Modal({ modalProps }: { modalProps: ModalProps; }) {
 }
 
 const ctxMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
-    if (props.channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel)) return;
+    if (props.channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel) || props.channel.guild.id && !PermissionStore.can(PermissionsBits.SEND_MESSAGES, props.channel)) return;
```
Vendicated, line 218:
```suggestion
    if (props.channel.guild_id && !(PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel) && PermissionStore.can(PermissionsBits.SEND_MESSAGES, props.channel)) return;
```

src/plugins/voiceMessages/index.tsx
```diff
@@ -215,7 +215,8 @@ function Modal({ modalProps }: { modalProps: ModalProps; }) {
 }
 
 const ctxMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
-    if (props.channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_VOICE_MESSAGES, props.channel)) return;
+    console.log(!(PermissionStore.can(PermissionsBits.SEND_MESSAGES, props.channel)));
```
Vendicated, line 218: combust
```suggestion
```

lumap: anytime

PR #1547, opened 2023-07-28
New Plugin: Fact Of The Day

MrDiamondDog:
Shows a completely useless fact every time you open Discord, almost exact copy of the BetterDiscord one

Resolves #1534


Vendicated:
this is a pretty bad implementation. i don't think anyone wants a modal to open every time they open Discord

the only thing that would make sense is showing it as a loading quote. you should edit the customloadingquotes plugin to add this as a feature if anything

MrDiamondDog:
I basically just copied what the bd plugin did cause that's what the discussion was referring to, I can see how it is kinda bad though

PR #1548, opened 2023-07-29
New Plugin: Platform Emulator

Rawiros:
Trick Discord into thinking that you are using a different operating system.

https://github.com/Vendicated/Vencord/discussions/763

### DIFF
--- a/src/plugins/PlatformEmulator.ts
+++ b/src/plugins/PlatformEmulator.ts
@@ -30,9 +30,9 @@ const settings = definePluginSettings({
             win32: { label: "Windows (Desktop)", default: IS_DISCORD_DESKTOP },
             darwin: { label: "Mac OS (Desktop)" },
             linux: { label: "Linux (Desktop)" },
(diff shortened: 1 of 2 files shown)


Vendicated:
I'm worried about the safety of people's accounts when using this. There's also not much reason to want to spoof your platform

I don't think this is a good plugin to have

Thanks anyway!

TomH5634142b7:
> I'm worried about the safety of people's accounts when using this. There's also not much reason to want to spoof your platform
> 
> I don't think this is a good plugin to have
> 
> Thanks anyway!

Please reconsider this, I've been using the more recent PlatformSpoofer pull request for a week and my account is still good.
Some people may wanna appear online but on mobile instead of desktop to avoid the "I see you online, play [so and so]" messages, or when using custom clients that use the browser version under the hood to make it actually show that you're on desktop.
The only PlatformEmulator plugin on BetterDiscord has also been broken for a while so there's no alternatives other than to make your own as far as I'm aware.
<sup>Also I could see this having more use than the "Moyai" plugin, but that's just my opinion.</sup>

PR #1552, opened 2023-07-29
feat(plugins/moreCommands): add skill issue command

isabelroses:
I think people would love this.
A simple and small command.

Example:
image


h4rldev:
This would be really funny

TheKodeToad:
honestly I think this gets old really quickly lol

isabelroses:
> honestly I think this gets old really quickly lol

Yeah it probably could but better that it there for anyone that still finds it useful etc.

h4rldev:
Ngl a figlet command would be better

Vendicated:
pretty funny but you can just achieve this with text replace so it's not that useful

PR #1563, opened 2023-07-29
feat(plugin): MessageTimeAgo

aelew:
Adds relative time ("time from now") to chat message timestamps.
Created a new PR because I accidentally closed the old one by renaming my fork branch :p

### Example
Example

### DIFF
--- a/src/plugins/messageTimeAgo.ts
+++ b/src/plugins/messageTimeAgo.ts
@@ -31,7 +31,7 @@ export default definePlugin({
             find: "Messages.MESSAGE_EDITED_TIMESTAMP_A11Y_LABEL",
             replacement: {
                 match: /.{13}"LT"\):\(.{8}(.{1})\)/,
-                replace: (match, moment) => `(${match})+' • '+$self.getRelativeTime(${moment})`
+                replace: (match, moment) => `${match}+' • '+$self.getRelativeTime(${moment})`
             }
         }
     ],
--- a/src/plugins/messageTimeAgo.ts
+++ b/src/plugins/messageTimeAgo.ts
@@ -1,0 +1,48 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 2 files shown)


AutumnVN:
i think it shouldn't change the small timestamp here, it takes 3-4 lines
also new message somehow always come from future "in 5 seconds" instead of "now"
image

aelew:
> i think it shouldn't change the small timestamp here, it takes 3-4 lines also new message somehow always come from future "in 5 seconds" instead of "now" image

I've gone ahead and removed relative time from compact timestamps.
I'm not sure why new messages are displaying in the future for you, but I pushed a commit that should fix that as well.

src/plugins/messageTimeAgo.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default definePlugin({
+    name: "MessageTimeAgo",
+    description: "Adds relative time (\"time from now\") to chat message timestamps.",
+    authors: [Devs.aelew],
+    patches: [
+        {
+            find: "Messages.MESSAGE_EDITED_TIMESTAMP_A11Y_LABEL",
+            replacement: {
+                match: /.{13}"LT"\):\(.{8}(.{1})\)/,
```
Vendicated, line 33: what exactly is this matching? Matching exactly n chars is not robust and matching 13 of any character at the beginning is pretty meaningless. Please rewrite this regex to be more precise

src/plugins/messageTimeAgo.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+const units: Intl.RelativeTimeFormatUnit[] = ["second", "minute", "hour", "day", "week", "month", "year"];
+const cutoffs = [60, 3600, 86400, 86400 * 7, 86400 * 30, 86400 * 365, Infinity];
+
+export default definePlugin({
+    name: "MessageTimeAgo",
+    description: "Adds relative time (\"time from now\") to chat message timestamps.",
```
Vendicated, line 27:
```suggestion
    description: 'Adds relative time ("time from now") to chat message timestamps.',
```

Vendicated:
closing this due to being very stale with unresolved feedback

if you'd like to continue work on this, feel free to reply and we can reopen this

thanks for your contribution anyway!

nutellaman111:
had the same idea, can i contribute to this plugin?

PR #1565, opened 2023-07-29
VoiceMessages: Support replies

nmsturcke:
Add a reply function to the voice message plugin.

### DIFF
--- a/src/plugins/voiceMessages/index.tsx
+++ b/src/plugins/voiceMessages/index.tsx
@@ -100,7 +100,7 @@ function sendAudio(blob: Blob, meta: AudioMetadata) {
                     waveform: meta.waveform,
                     duration_secs: meta.duration,
                 }],
-                message_reference: reply ? MessageCreator.getSendMessageOptionsForReply(reply).messageReference : null,
+                message_reference: reply ? MessageCreator.getSendMessageOptionsForReply(reply)?.messageReference : null,
             }
         });
     });
--- a/src/plugins/voiceMessages/index.tsx
+++ b/src/plugins/voiceMessages/index.tsx
@@ -74,7 +74,7 @@ const EMPTY_META: AudioMetadata = {
 function sendAudio(blob: Blob, meta: AudioMetadata) {
     const channelId = SelectedChannelStore.getChannelId();
     const reply = PendingReplyStore.getPendingReply(channelId);
-    (reply && FluxDispatcher.dispatch({ type: "DELETE_PENDING_REPLY", channelId }));
+    if (reply) FluxDispatcher.dispatch({ type: "DELETE_PENDING_REPLY", channelId });
 
     const upload = new CloudUpload({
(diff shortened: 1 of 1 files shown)


src/plugins/voiceMessages/index.tsx
```diff
@@ -37,6 +37,8 @@ import { VoicePreview } from "./VoicePreview";
 import { VoiceRecorderWeb } from "./WebRecorder";
 
 const CloudUpload = findLazy(m => m.prototype?.uploadFileToCloud);
+const MessageCreator = findByPropsLazy("getSendMessageOptionsForReply", "sendMessage");
+const PendingReplyStore = findByPropsLazy("getPendingReply");
```
TheKodeToad, line 41:
```suggestion
const PendingReplyStore = Vencord.Webpack.findStoreLazy("PendingReplyStore");
```

Vap0r1ze: This but you should just `import { findStoreLazy } from "@webpack";`

TheKodeToad: I copied it from DevTools :woman_facepalming:

src/plugins/voiceMessages/index.tsx
```diff
@@ -71,6 +73,8 @@ const EMPTY_META: AudioMetadata = {
 
 function sendAudio(blob: Blob, meta: AudioMetadata) {
     const channelId = SelectedChannelStore.getChannelId();
+    const reply = PendingReplyStore.getPendingReply(channelId);
+    (reply && FluxDispatcher.dispatch({ type: "DELETE_PENDING_REPLY", channelId }));
```
Vap0r1ze, line 77:
```suggestion
    if (reply) FluxDispatcher.dispatch({ type: "DELETE_PENDING_REPLY", channelId });
```
just a little more readable

TheKodeToad: why not
```suggestion
    if (reply)
       FluxDispatcher.dispatch({ type: "DELETE_PENDING_REPLY", channelId });
```

src/plugins/voiceMessages/index.tsx
nmsturcke: I just copied the spotify share plugin for pending reply lol

src/plugins/voiceMessages/index.tsx
```diff
@@ -95,7 +99,8 @@ function sendAudio(blob: Blob, meta: AudioMetadata) {
                     uploaded_filename: upload.uploadedFilename,
                     waveform: meta.waveform,
                     duration_secs: meta.duration,
-                }]
+                }],
+                message_reference: reply ? MessageCreator.getSendMessageOptionsForReply(reply).messageReference : null,
```
Vendicated, line 103:
```suggestion
                message_reference: reply ? MessageCreator.getSendMessageOptionsForReply(reply)?.messageReference : null,
```

PR #1605, opened 2023-07-31
feat(plugin): zipServerEmojis

lumap:
Adds a `/zip-server-emojis` slash command that downloads all of the emojis in a server as a zip file.


src/plugins/zipServerEmojis.ts
```diff
@@ -0,0 +1,74 @@
...
+                    return sendBotMessage(cmdCtx.channel.id, {
+                        content: "Server ID is invalid",
+                    });
+                }
+                sendBotMessage(cmdCtx.channel.id, {
+                    content: "This shouldn't take long...",
+                });
+
+                const zipContents: { file: Uint8Array, filename: string; }[] = [];
+                const waitForFetches = new Promise((resolve, _) => {
```
Vendicated, line 55: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

lumap: pushed a commit to use that, looks good?

lumap: wait nvm i broke something

lumap: unbroke it

src/plugins/zipServerEmojis.ts
```diff
@@ -0,0 +1,76 @@
...
+                        content: "Server ID is invalid",
+                    });
+                }
+                sendBotMessage(cmdCtx.channel.id, {
+                    content: "This shouldn't take long...",
+                });
+
+                const fetchEmojis = async e => {
+                    const emoji = await fetch(`https://cdn.discordapp.com/emojis/${e.id}${e.animated ? ".gif" : ".png"}?size=96&quality=lossless`).then(res => res.blob());
+                    return { file: new Uint8Array(await emoji.arrayBuffer()), filename: `${e.id}${e.animated ? ".gif" : ".png"}` };
```
TheKodeToad, line 56:
```suggestion
                    const filename = e.id + (e.animated ? ".gif" : ".png");
                    const emoji = await fetch(`https://cdn.discordapp.com/emojis/${filename}?size=96&quality=lossless`).then(res => res.blob());
                    return { file: new Uint8Array(await emoji.arrayBuffer()), filename };
```
maybe? i hope this works (typescript :sob)?

Vendicated:
I don't find this useful/versatile enough to include it as a plugin since this is something you'd use maybe once then never again, this seems like something better suited for the #js-snippets channel

src/plugins/zipServerEmojis.ts
```diff
@@ -0,0 +1,80 @@
...
+                const emojis = EmojiStore.getGuilds()[opts[0].value]?.emojis;
+                if (!emojis) {
+                    return sendBotMessage(cmdCtx.channel.id, {
+                        content: "Server ID is invalid",
+                    });
+                }
+
+                const fetchEmojis = async e => {
+                    const filename = e.id + (e.animated ? ".gif" : ".png");
+                    const emoji = await fetch(`https://cdn.discordapp.com/emojis/${filename}?size=96&quality=lossless`).then(res => res.blob());
```
Vendicated, line 53: using a bigger size seems better to ensure full quality
```suggestion
                    const emoji = await fetch(`https://cdn.discordapp.com/emojis/${filename}?size=512&quality=lossless`).then(res => res.blob());
```

Vendicated:
> I don't find this useful/versatile enough to include it as a plugin since this is something you'd use maybe once then never again, this seems like something better suited for the #js-snippets channel

I'll be closing this PR for that reason, feel free to post it in the snippet channel though! Maybe we could add some sort of official snippet section on https://vencord.dev for other stuff like this? Seems like a cool idea

thanks for your efforts anyway!!

PR #1607, opened 2023-08-02
feat(plugin): PreviewMessage

Syncxv:
lets you preview your message before you send it
basically this plugin: https://betterdiscord.app/plugin/PreviewMessage

image


src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export default definePlugin({
+    name: "PreviewMessage",
+    description: "Lets you preview your message before sending it.",
+    authors: [Devs.Aria],
+    patches: [
+        {
+            find: ".activeCommandOption",
+            replacement: {
+                match: /(.)\.push.{1,30}disabled:(\i),.{1,20}\},"gift"\)\)/,
```
Vendicated, line 76: omg we really need to make an api for this hahaha this is getting ridiculous how many plugins have this patch copy pasted

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+                replace: "$&;try{$2||$1.unshift($self.previewIcon(arguments[0]))}catch{}",
+            }
+        },
+    ],
+
+    previewIcon: ErrorBoundary.wrap(PreviewButton, { noop: true }),
+
```
Vendicated, line 83:
```suggestion
```

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+        analyticsName: string;
+    };
+}
+
+export function PreviewButton(chatBoxProps: Props) {
+    if (chatBoxProps.type.analyticsName !== "normal") return null;
+
+    return (
+        <Tooltip text="Preview Message">
+            {(tooltipProps: any) => (
```
Vendicated, line 41: why the any type here? this is already strongly typed so this shouldn't be necessary. If ts is complaining about Button not taking some of these props, you could just pick the specific ones you want, usually onMouseEnter and onMouseLeave

Vendicated: this button also needs an aria-label, but I'm pretty sure the tooltip props have aria label that defaults to the text value

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    {...tooltipProps}
+                    onClick={() => {
+                        const channelId = SelectedChannelStore.getChannelId();
+                        const draft = DraftStore.getDraft(channelId, DRAFT_TYPE);
+                        if (!draft) return;
+                        sendBotMessage(channelId, {
+                            content: draft,
+                        });
+                    }}
+                    disabled={DraftStore.getDraft(SelectedChannelStore.getChannelId(), DRAFT_TYPE).length === 0}
```
Vendicated, line 52: maybe return null would be better? to keep the bar less bloated

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+*/
+
+import { sendBotMessage } from "@api/Commands";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findStoreLazy } from "@webpack";
+import { Button, ButtonLooks, ButtonWrapperClasses, SelectedChannelStore, Tooltip } from "@webpack/common";
+
+const DraftStore = findStoreLazy("DraftStore");
```
Vendicated, line 26: might be worthwhile to move this to commons since I think this is used by multiple plugins?

Vendicated (changes requested):
looks good other than these mostly nitpicks

src/plugins/previewMessage.tsx
Vendicated: (not saying you should do it in this pr (unless you would like to), just a general comment)

Vendicated:
maybe make the botmessage have your pfp and name? For the most complete ✨ immersion ✨

Syncxv:
thanks for reviewing

> why the any type here? 

oops i was just copy pasting.

> maybe make the botmessage have your pfp and name? For the most complete ✨ immersion ✨

ill try add that tomorrow

> might be worthwhile to move this to commons since I think this is used by multiple plugins?

sure, i tried doing that in the latest commit 6c25c08. maybe check if its okay?

ill continue tomorrow

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface Props {
+    type: {
+        analyticsName: string;
+    };
+}
+
+function getDraft(channelId: string) {
+    const draft = DraftStore.getDraft(channelId, 0);
+    return draft;
```
Vendicated, line 33: please remove this pointless variable

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Button, ButtonLooks, ButtonWrapperClasses, DraftStore, SelectedChannelStore, Tooltip, UserStore } from "@webpack/common";
+
+interface Props {
+    type: {
+        analyticsName: string;
+    };
+}
+
+function getDraft(channelId: string) {
+    const draft = DraftStore.getDraft(channelId, 0);
```
Vendicated, line 32: it would benefit readability to move this 0 into a constant, something like

```js
const DRAFT_TYPE_CHANNEL_MESSAGE = 0;
```

src/webpack/common/types/stores.d.ts
```diff
@@ -148,3 +148,34 @@ export class EmojiStore extends FluxStore {
         get favoriteEmojisWithoutFetchingLatest(): Emoji[];
     };
 }
+
+type DraftType = 0 | 1 | 2 | 3;
+// ChannelMessage = 0,
+// ThreadSettings = 1,
+// FirstThreadMessage = 2,
+// ApplicationLauncherCommand = 3
```
Vendicated, line 156: how about making this an enum and exporting it from stores.ts? (kinda similar as above)

src/webpack/common/types/stores.d.ts
```diff
@@ -148,3 +148,34 @@ export class EmojiStore extends FluxStore {
...
+    timestamp: number;
+    draft: string;
+}
+
+interface DraftState {
+    [userId: string]: {
+        [channelId: string]: {
+            [key in DraftType]?: Omit<DraftObject, "channelId">;
+        } | undefined;
+    } | undefined;
```
Vendicated, line 169: instead of `| undefined` you should use `[userId: string]?: {`

src/webpack/common/types/stores.d.ts
```diff
@@ -148,3 +148,34 @@ export class EmojiStore extends FluxStore {
...
+}
+
+export class DraftStore {
```
Vendicated, line 172:
```suggestion
export class DraftStore extends Store {
```

src/plugins/previewMessage.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </Button>
+            )}
+        </Tooltip>
+    );
+
+}
+
+export default definePlugin({
+    name: "PreviewMessage",
+    description: "Lets you preview your message before sending it.",
```
Vendicated, line 73:
```suggestion
    description: "Lets you preview your message before sending it",
```

Vendicated:
looks good other than the suggestions

src/webpack/common/types/stores.d.ts
Syncxv: i tried  that but since its a .d.ts file it cant export enums. i get a webpack/types not resolved
here is the commit that tries to import DraftType enum from the  store.d.ts
[f2d3625](https://github.com/Vendicated/Vencord/pull/1607/commits/f2d3625f743e93e1f359dc429d2eafc3f1e45c6c)

here is the error
https://github.com/Vendicated/Vencord/actions/runs/5844460147/job/15847602652#step:6:34

src/webpack/common/types/stores.d.ts
Syncxv: yeah i tried that but theres an error 😔 and i have no idea how to solve it. no clear error message
error

Vendicated: try using a type instead of an interface

Syncxv: i tried making it a type too
apparently you cant do `[userId: string]?: {` in ts
https://stackoverflow.com/questions/57337815/why-cant-this-index-signature-be-optional

 `[key in DraftType]?: ` works probably because its a mapped type and not an index signature ¯\_(ツ)_/¯

Vendicated: thats fine then

Vendicated (approved):

Vendicated:
thank you

PR #1608, opened 2023-08-02
messageLogger: apply deleted style to ephemeral messages

VMGuy23:
The bot "thinking" text was previously not shown red when deleted.
I noticed this bug when using YAGPDB's poll feature. When the poll is created, a "YAGPDB.xyz is thinking..." during the creation of the poll, and deleted after. However, it is not shown in red once deleted. Image below.
image
This PR should fix this bug.

### DIFF
--- a/src/plugins/messageLogger/deleteStyleText.css
+++ b/src/plugins/messageLogger/deleteStyleText.css
@@ -2,7 +2,10 @@
 .messagelogger-deleted [class*="contents-"] > :is(div, h1, h2, h3, p) {
     color: #f04747 !important;
 }
-
+/* Bot "thinking" text highlighting */
+.messagelogger-deleted [class*="colorStandard-"] :is(div) {
+    color: #f04747 !important;
+}
 /* Embed highlighting */
 .messagelogger-deleted article :is(div, span, h1, h2, h3, p) {
     color: #f04747 !important;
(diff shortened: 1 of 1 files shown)


MrDiamondDog:
That's because it's not an actual message. It is a placeholder for when the bot is loading it's response. It can not be changed, and will always say `(Bot name) is thinking...`.

VMGuy23:
> That's because it's not an actual message. It is a placeholder for when the bot is loading it's response. It can not be changed, and will always say `(Bot name) is thinking...`.

That doesn't stop it from being styled red to show that it is deleted. The right-click menu still correctly shows the options for a deleted message

VMGuy23:
why is the check only running occasionally

Vendicated:
since you are a first time contributor, i have to manually approve workflow runs https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -3,6 +3,11 @@
     color: #f04747 !important;
 }
 
+/* Bot "thinking" text highlighting */
+.messagelogger-deleted [class*="colorStandard-"] :is(div) {
```
Vendicated, line 7: this :is() is obsolete, since there is only one selector. :is is used to group multiple selectors

VMGuy23: Should  be fixed now

Vendicated:
thank you

PR #1609, opened 2023-08-03
feat(plugin): MessageScheduler

WorriedArrow:
The **MessageScheduler** plugin adds a simple command to schedule a message to send in the channel the command was run in after a delay, even the channel is not currently selected!


Vendicated:
I can't imagine a non-malicious use case for such a feature. As such, I dont think having this plugin is a good idea. Thank you anyway!

PR #1619, opened 2023-08-04
feat(plugin): ColorHighlight

AutumnVN:
close because it's not useful anyway
-
Highlight color code in chat (like a role mention) and click to copy them
image

currently has this silly bug but idk how to fix it yet 😭
image


src/plugins/colorHighlight.tsx
```diff
@@ -0,0 +1,144 @@
...
...]\d|25[0-5])\s*(?:,|\s)\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*(?:,|\s)\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*\)|rgb\(\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:,|\s)\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:,|\s)\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*\)|rgba\(\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*,\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*,\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*,\s*(?:(?:1|0?\.\d+)|(?:[0-9]?\d(?:\.\d+)?|100)%)\s*\)|rgb\(\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:\/\s*(?:(?:1|0?\.\d+)|(?:[0-9]?\d(?:\.\d+)?|100)%)\s*)?\)|rgb\(\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*(?:[01]?\d\d?|2[0-4]\d|25[0-5])\s*\/\s*(?:(?:1|0?\.\d+)|(?:[0-9]?\d(?:\.\d+)?|100)%)\s*\)|hsl\(\s*(?:[012]?\d\d?|3[0-5]\d|360)(?:deg)?\s*(?:,|\s)\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:,|\s)\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*\)|hsla\(\s*(?:[012]?\d\d?|3[0-5]\d|360)(?:deg)?\s*,\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*,\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*,\s*(?:(?:1|0?\.\d+)|(?:[0-9]?\d(?:\.\d+)?|100)%)\s*\)|hsl\(\s*(?:[012]?\d\d?|3[0-5]\d|360)(?:deg)?\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*(?:[0-9]?\d(?:\.\d+)?|100)%\s*\/\s*(?:(?:1|0?\.\d+)|(?:[0-9]?\d(?:\.\d+)?|100)%)\s*\))/ig;
```
Vendicated, line 103: my brain went "wtf is this regex" then i read the variable name and hahahah

src/plugins/colorHighlight.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export default definePlugin({
+    name: "ColorHighlight",
+    authors: [Devs.AutumnVN],
+    description: "Highlight color code in chat and click to copy them.",
+    patches: [{
+        find: ".Messages.MESSAGE_EDITED,",
+        replacement: {
+            match: /var .,.,.=(.)\.className,.=.\.message,.=.\.children,.=.\.content,.=.\.onUpdate/gm,
```
Vendicated, line 112: replace these `.` with `\i`

Vendicated: why are you even matching so much only to $& it? you can likely remove most of this match, and I also recommend using a lookaround to avoid matching stuff you don't need matched

src/plugins/colorHighlight.tsx
```diff
@@ -0,0 +1,144 @@
...
+            replace: "try {if ($1) $1.content = $self.replaceMessageContent($1); } catch {};$&"
+        }
+    }],
+    replaceMessageContent(comp: { content: any[]; }) {
+        const final: any[] = [];
+        for (let i = 0; i < comp.content.length; i++) {
+            const component = comp.content[i];
+            if (typeof component.props.children !== "string") {
```
Vendicated, line 120: I'm not sure what content could be, but react nodes could be null, strings, etc, so checking .props.children might be unsafe

PR #1623, opened 2023-08-06
feat(plugin): NoSentryConsole

WorriedArrow:
**NoSentryConsole**: Disables the sentry console log hook to return breadcrumbs to console. Very useful for development and reporting issues alike.

### The problem
Discord uses Sentry and it will install a hook into the console object.
Doesn't seem that bad, right?
Well, it makes the breadcrumbs/origin in the console from the file that installs that hook,
which ends up making it very hard to tell where non-error logs are coming from!

### The solution
So this plugin just deletes that hook.
The hook doesn't even do anything anyway, thanks to NoTrack.
This just removes that unnecessary code and allows for a much better experience when debugging,
developing plugins, or reporting issues!
What it does

### DIFF
--- a/src/plugins/_core/noTrack.ts
+++ b/src/plugins/_core/noTrack.ts
@@ -55,7 +55,7 @@ export default definePlugin({
         {
             find: ".installedLogHooks)",
             replacement: {
-                match: /if\(\w\.getDebugLogging\(\)&&!\w.installedLogHooks\)/,
+                match: /if\(\i\.getDebugLogging\(\)&&!\i.installedLogHooks\)/,
                 replace: "if(false)"
             }
         },
--- a/src/plugins/_core/noTrack.ts
+++ b/src/plugins/_core/noTrack.ts
@@ -22,7 +22,7 @@ import definePlugin from "@utils/types";
 export default definePlugin({
     name: "NoTrack",
     description: "Disable Discord's tracking ('science'), metrics and Sentry crash reporting",
-    authors: [Devs.Cyn, Devs.Ven, Devs.Nuckyz],
+    authors: [Devs.Cyn, Devs.Ven, Devs.Nuckyz, Devs.Arrow],
     required: true,
     patches: [
         {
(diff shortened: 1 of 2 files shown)


src/plugins/noSentryConsole.tsx
```diff
@@ -0,0 +1,57 @@
...
+            // Disables Sentry console instrument and returns breadcrumbs to console
+            find: ".installedLogHooks)",
+            replacement: {
+                match: /if\(\w\.getDebugLogging\(\)&&!\w.installedLogHooks\)/,
+                replace: "if(false)"
+            }
+        },
+    ],
+
+    settingsAboutComponent() {
```
Vendicated, line 39: instead of using this, you should give your plugin a folder and add a README.md file

Vendicated:
why not just add this to the NoTrack plugin instead?

src/plugins/noSentryConsole.tsx
WorriedArrow: Will add to NoTrack.

src/plugins/_core/noTrack.ts
```diff
@@ -51,6 +51,13 @@ export default definePlugin({
                     replace: "return;"
                 }
             ]
-        }
+        },
+        {
+            find: ".installedLogHooks)",
+            replacement: {
+                match: /if\(\w\.getDebugLogging\(\)&&!\w.installedLogHooks\)/,
```
Vendicated, line 58:
```suggestion
                match: /if\(\i\.getDebugLogging\(\)&&!\i.installedLogHooks\)/,
```

src/plugins/_core/noTrack.ts
```diff
@@ -51,6 +51,13 @@ export default definePlugin({
                     replace: "return;"
                 }
             ]
-        }
+        },
+        {
+            find: ".installedLogHooks)",
+            replacement: {
+                match: /if\(\i\.getDebugLogging\(\)&&!\i.installedLogHooks\)/,
```
Nuckyz, line 58:
```suggestion
                match: /if\(\i\.getDebugLogging\(\)&&!\i\.installedLogHooks\)/,
```

Nuckyz (dismissed):

Vendicated:
looks good, thank you!

PR #1625, opened 2023-08-07
feat(plugin) HiddenLinks

VilariStorms:
A simple plugin to disguise links with a slash command.
Example:


Vendicated:
no + use textreplace

PR #1627, opened 2023-08-08
Quote Plugin (Bring back old Quote system that many people loves)

UDIA-GOD:


src/plugins/quotes.tsx
```diff
@@ -0,0 +1,66 @@
...
+    name: "Quote",
+    description: "This plugin brings back old Discord quote option.",
+    authors: [Devs.udiagod],
+    dependencies: ["MessagePopoverAPI"],
+
+    QuoteIcon: () => <QuoteIcon />,
+
+    async start() {
+        addButton("Quote", message => {
+            return message?.content
```
TheKodeToad, line 48:
```suggestion
            return message?.type == 0
```
I think this would work better (..?)

src/plugins/quotes.tsx
```diff
@@ -0,0 +1,66 @@
...
+                ? {
+                    label: "Quote",
+                    icon: this.QuoteIcon,
```
TheKodeToad, line 51:
```suggestion
                    icon: QuoteIcon,
```
pretty sure this without putting the lambda in the plugin would work just fine (I don't know if this is good practice, though)

TheKodeToad (commented):
Does this even work properly with multiline messages?

src/plugins/quotes.tsx
UDIA-GOD: It works but I've noticed that more plugins do it like that

Vendicated: that's for calling it outside of a react context (aka inside patches)

In your case this is obsolete

UDIA-GOD:
So ya changed 1 thing, but mostly i've seen using a lambda. Also ya it works with multiline :D

UDIA-GOD:
Tested it and it works fine rn

TheKodeToad:
See https://github.com/Vendicated/Vencord/pull/1627#discussion_r1287702475; you probably shouldn't use a lambda

UDIA-GOD:
k changed that, ur the boss.

TheKodeToad:
Yes, I am totally the boss B)

UDIA-GOD:
Please can someone approve this workflow thing

Vendicated:
I did, please fix lint!

UDIA-GOD:
fixed

src/plugins/quotes.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { addButton, removeButton } from "@api/MessagePopover";
+import { Devs } from "@utils/constants";
+import { insertTextIntoChatInputBox } from "@utils/discord";
+import definePlugin from "@utils/types";
+import { ChannelStore } from "@webpack/common";
+
+function QuoteIcon() {
+    return (
+
```
Vendicated, line 27:
```suggestion
```

src/plugins/quotes.tsx
```diff
@@ -0,0 +1,64 @@
...
+export default definePlugin({
+    name: "Quote",
+    description: "This plugin brings back old Discord quote option.",
+    authors: [Devs.udiagod],
+    dependencies: ["MessagePopoverAPI"],
+
+    async start() {
+        addButton("Quote", message => {
+            return message?.type === 0 && message?.content?.length > 0
```
Vendicated, line 46:
```suggestion
            return message?.content?.length
```

Vendicated: or maybe keep type check but there are also other types you can reply to, i think you can actually reply to all types (?)

Vendicated:
this will add the button in channels you don't have send message perms in, please add a permission check

UDIA-GOD:
I'll add perm check soon. Been out for some time.

src/plugins/quotes.tsx
TheKodeToad: I forgot replies have a different message type. oops :woozy_face:
I assumed type 0 was the only type you'd want to quote

Vendicated:
Hi, I think this is too niche (and abandoned :p), but thank you nevertheless for making this pr!

PR #1630, opened 2023-08-11
Add more features to oneko addon

korbough:
Changes:

- Added built-in skinning functionality
- Added speed slider
- No more eval 🎉


Vendicated:
it would probably be better to edit the script to have this logic instead of running it 10 times

src/plugins/oneko.ts
```diff
@@ -17,24 +17,43 @@
...
+    settings,
 
     start() {
-        fetch("https://raw.githubusercontent.com/adryd325/oneko.js/5977144dce83e4d71af1de005d16e38eebeb7b72/oneko.js")
-            .then(x => x.text())
-            .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif"))
-            .then(eval);
+
+        for(var i = 0; i < settings.store.quantity; i++) {
+            fetch("https://raw.githubusercontent.com/Korbaux/oneko.js/main/oneko.js")
```
Vendicated, line 43: this is a backdoor since you could edit the script at any time to add malicious code. Pin it to specific long commit hash

somerandomcloud: What would be better is making a folder for the plugin and including the script in the repo itself instead of fetching it every time

Korbaux: oops i will fix this immediately

Korbaux:
> it would probably be better to edit the script to have this logic instead of running it 10 times

from what i see, i would have to rewrite a lot of the script to do this

Korbaux:
> it would probably be better to edit the script to have this logic instead of running it 10 times

i instead made it only fetch once, then replace and eval 10 times

Korbaux:
ok so im doing a different feature instead of that now

Korbaux:
whoops

Korbaux:
Added built-in skinning functionality

stevejobless2020:
> why the oneko no work? i just added it on vencord doesnt work

korbough:
ven i swear to shiggy

korbough:
let the water hold me down

korbough:
letting the days go by

korbough:
sad to see this pr die, smh my head

korbough:
code was quite shit though

PR #1633, opened 2023-08-11
youtify (placeholder name)

exhq:
shows you the corresponding youtube link of a spotify song 

TODO:

- add settings for invidious instances (done)
- make it look nicer


Vendicated:
please avoid profanity in source code & github discussions

MaGames53812:
Alternate Name Ideas That Don't Include Profanity:
SpoTubefy
I don't use Spotify
Spotifyn't
SpotifYT Music

I dunno, not that good at names

Vendicated:
i think it would be a lot better to use song.link instead of your api, since it supports a lot more platforms both ways (source and destination)

for spotify urls it's `https://song.link/s/:track-id`, eg https://song.link/s/1wdK548KKmN40wxH051bSy

exhq:
the CF worker does use song.link, I think I made it because song.link had CORS problems
either way, the code is garbage. I'll look into using the song.link API when I rewrite the entire plugin

Vendicated:
ah i see! well you can bypass CORS via `native.ts` now

Vendicated:
very stale and too niche, probably better done via a user app, e.g. via context menu or slash command

thanks for your efforts :3

PR #1639, opened 2023-08-12
feat(PermissionsViewer): separated list scrolling, hide irrelevant perms

bb010g:
The role listing in the permissions viewer can now be scrolled separately from the permissions listing. Additionally, the permission list scroll position is preserved when changing which role is displayed.

The permissions viewer permissions listing now has a toggle to hide permissions which are irrelevant to the computed permission set. By default, all permissions are shown, but this default can be changed. You can also hide or show irrelevant permissions for just the current permissions viewer instance. This hiding is performed via CSS, for efficiency.

## Screenshots

DiscordCanary_14wtHuIeR7

DiscordCanary_abGgfrd2z1

### DIFF
--- a/src/plugins/permissionsViewer/components/icons.tsx
+++ b/src/plugins/permissionsViewer/components/icons.tsx
@@ -18,9 +18,9 @@
 
 import { PermissionValue } from "./RolesAndUsersPermissions";
 
-export function PermissionDeniedIcon() {
+export function PermissionDeniedIcon({ ...props }: {} & React.SVGProps<SVGSVGElement>) {
     return (
-        <svg
+        <svg {...props}
             height="24"
             width="24"
             viewBox="0 0 24 24"
@@ -31,9 +31,9 @@ export function PermissionDeniedIcon() {
     );
 }
 
-export function PermissionPassthroughIcon() {
+export function PermissionPassthroughIcon({ ...props }: {} & React.SVGProps<SVGSVGElement>) {
     return (
-        <svg
+        <svg {...props}
             height="24"
             width="24"
             viewBox="0 0 16 16"
@@ -46,9 +46,9 @@ export function PermissionPassthroughIcon() {
     );
 }
 
-export function PermissionAllowedIcon() {
+export function PermissionAllowedIcon({ ...props }: {} & React.SVGProps<SVGSVGElement>) {
     return (
-        <svg
+        <svg {...props}
             height="24"
             width="24"
             viewBox="0 0 24 24"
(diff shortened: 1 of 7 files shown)


src/plugins/permissionsViewer/components/icons.tsx
```diff
@@ -29,30 +35,38 @@ export function PermissionDeniedIcon() {
     );
 }
 
-export function PermissionAllowedIcon() {
+export function PermissionDefaultIcon(): JSX.Element {
```
Vendicated, line 38:
```suggestion
export function PermissionDefaultIcon() {
```
don't type the return type unless necessary (eg overloads or recursive funcs)

src/plugins/permissionsViewer/styles.css
```diff
@@ -112,37 +114,49 @@
...
     align-items: center;
-    padding: 10px;
-    border-bottom: 2px solid var(--background-modifier-active);
+    padding: var(--vc-permviewer-perms-perms-item-outer-margin) var(--vc-permviewer-perms-perms-inner-margin);
+    border-bottom: var(--vc-permviewer-perms-perms-item-border-width) solid var(--background-modifier-active);
 }
 
-.vc-permviewer-perms-perms-item:last-child {
-    border: 0;
+.vc-permviewer-perms-perms-hide-irrelevant-permissions > .vc-permviewer-perms-perms-item:where(:not([data-vc-permviewer-permission-state]), [data-vc-permviewer-permission-state="default"], [data-vc-permviewer-permission-state="deny"]):where(:not([data-vc-permviewer-permission-overwrite-state]), [data-vc-permviewer-permission-overwrite-state="default"]) {
```
Vendicated, line 142: this selector is veeery complex, it would likely benefit readability to split it into multiple selectors or split it across multiple lines

Nuckyz (changes requested):
Instead of messing with new classes, attributes and css the code should simply skip rendering the irrelevant permissions

src/plugins/permissionsViewer/styles.css
bb010g: I initially split this over multiple lines, but Stylelint's indentation rule required 0 leading spaces for each subsequent line of the same rule, which reads the same as multiple rules (separated by `,`) would, so I avoided that for clarity.

bb010g:
> the code should simply skip rendering the irrelevant permissions

Done. While handling this I realized that the list rendering wasn't keyed, which gets more relevant when adding/removing classes from a list, so that's done for both listings now.

The nesting got severe after introducing `<Fragment>` to handle permissions that aren't rendered, so permission item display has been factored out to a standalone `PermissionItem` component.

bb010g:
Refactored and cleaned up the patch stack some. Should be much easier to review commit-by-commit now.

Vendicated:
This has since been implemented

PR #1640, opened 2023-08-12
feat(.vscode): local workspace config overrides

bb010g:
Currently accomplished via the "Workspace Config+" extension. This should really be a built-in VS Code feature, but here we are.

This does not break VS Code instances with Workspace Config+ disabled.


bb010g:
This now uses the `workspaceConfigPlus.replaceMergedSymlinks` option to ensure a smooth migration, implemented in <https://github.com/swellaby/vscode-workspace-config-plus/pull/144>.

Vendicated:
a lot of time has passed and your linked pr has not been merged. i also personally don't really see the use case for this change, and we haven't had anyone else need this, so it seems very niche

thus, i'll be closing this, sorry. thank you a lot for your contribution regardless!

PR #1641, opened 2023-08-12
plugins(NormalizeResourceLinks): create plugin

bb010g:
Finally, I can stop manually removing the `canary.` from my copied links

This patches the copy link button action (which directly reads ambient `location.host`), because any deeper patches would risk breaking functionality.

### DIFF
--- a/src/plugins/normalizeResourceLinks/index.ts
+++ b/src/plugins/normalizeResourceLinks/index.ts
@@ -15,8 +15,8 @@ export default definePlugin({
         {
             find: ".Messages.COPY_MESSAGE_LINK,",
             replacement: {
-                match: /(\.concat\()(location\.host)(\))/,
-                replace: "$1$self.normalizeResourceHost($2)$3",
+                match: /\.concat\(location\.host\)/,
+                replace: ".concat(self.normalizeResourceHost(location.host))",
             },
         },
     ],
--- a/src/plugins/normalizeResourceLinks/index.ts
+++ b/src/plugins/normalizeResourceLinks/index.ts
@@ -1,0 +1,27 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NormalizeResourceLinks",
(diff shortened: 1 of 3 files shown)


TheKodeToad:
Can't you use TextReplace?

TheKodeToad:
Never mind. I'm stupid.

src/plugins/normalizeResourceLinks/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default definePlugin({
+    name: "NormalizeResourceLinks",
+    description: "Normalize resource links to match stable Discord",
+    authors: [Devs.bb010g],
+    patches: [
+        {
+            find: ".Messages.COPY_MESSAGE_LINK,",
+            replacement: {
+                match: /(""\.concat\(location\.protocol,"\/\/"\)\.concat\()(location\.host)(\)\.concat\()/,
```
Vendicated, line 18: this matcher seems overly complex. You just need to replace location.host, right? Something like this should work fine

```js
match: /\.concat\(location\.host\)/,
replace: ".concat(location.host.replace(/^(canary|ptb)\\./, ''))"
```

bb010g: Is this safe to do? I don't want to mess up other code that depends on the actual `location.host` value.

Vendicated: the find limits it to only the module you want to match, which in this case is literally just the element and nothing else
image

src/plugins/normalizeResourceLinks/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    name: "NormalizeResourceLinks",
+    description: "Normalize resource links to match stable Discord",
+    authors: [Devs.bb010g],
+    patches: [
+        {
+            find: ".Messages.COPY_MESSAGE_LINK,",
+            replacement: {
+                match: /(\.concat\()(location\.host)(\))/,
+                replace:
+                    "$1Vencord.Plugins.plugins.NormalizeResourceLinks.normalizeResourceHost($2)$3",
```
Vendicated, line 20:
```suggestion
                    "$1$self.normalizeResourceHost($2)$3",
```

Vendicated:
try to avoid redundant groups, it makes patches way harder to read. just repeat short constant strings or use lookarounds

Vendicated:
thank you

PR #1658, opened 2023-08-16
plugins(LightshotRoulette): create

Deltara3:
This plugin is a fun little plugin that generates random `prnt.sc` links for you to look at. It's a gamble between interesting, not interesting, or nothing. This might not be enough to warrant a whole new plugin and migration to another plugin like MoreCommands is perfectly fine.


Vendicated:
besides not being that useful, this plugin also has much deeper issues in that it might post illegal or just nsfw content

thank you anyway!

Deltara3:
No problem!

PR #1659, opened 2023-08-16
feat(plugin): TwitterIcon

ActuallyTheSun:
twitter icon in profile


Vendicated:
why not just use css

ActuallyTheSun:
fair enough

PR #1660, opened 2023-08-16
feat(PronounDB): update to v2 API

0xallie:
The new version of PronounDB supports pronoun sets like "he/any" and "he/she/they".

For multiple pronouns, it will show up like "he/any" or "He/Any", but for single pronouns it will still show up like "he/him" or "He/Him" or "Any pronouns".

Multi-language pronouns are also supported now, but I've chosen to hardcode English for now. This can be reevaluated later.

### DIFF
--- a/src/plugins/pronoundb/pronoundbUtils.ts
+++ b/src/plugins/pronoundb/pronoundbUtils.ts
@@ -128,7 +128,7 @@ async function bulkFetchPronouns(ids: string[]): Promise<PronounsResponse> {
     params.append("ids", ids.join(","));
 
     try {
-        const req = await fetch("https://pronoundb.org/api/v1/lookup-bulk?" + params.toString(), {
+        const req = await fetch("https://pronoundb.org/api/v2/lookup?" + params.toString(), {
             method: "GET",
             headers: {
                 "Accept": "application/json",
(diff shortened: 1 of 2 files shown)


src/plugins/pronoundb/pronoundbUtils.ts
```diff
@@ -150,14 +151,21 @@ async function bulkFetchPronouns(ids: string[]): Promise<PronounsResponse> {
 }
 
 export function formatPronouns(pronouns: string): string {
+    let formattedPronouns;
```
Vendicated, line 154:
```suggestion
    let formattedPronouns: string;
```
or better yet, use a ternary

0xallie: I applied the suggested change. Ternary seems like it would make the line too long, or split into multiple lines which is generally discouraged for a ternary, so idk.

Vendicated:
hmm, something seems wrong here, there's a lot of type errors

0xallie:
Yeah I'll take a look tomorrow, I didn't have a local TS linting environment set up but the code seemed to work.

Vendicated:
tomorrow has been quite a while ago now. are you still interesting in fixing the last bits? 

if not, i might just fix it

0xallie:
Yeah oops, I thought about it recently but I've been lazy... if you want to fix it yourself go ahead, it will probably be faster.

PR #1661, opened 2023-08-17
build(scripts): allow linting user code & lint ESM source files

bb010g:


bb010g:
@Vendicated Requesting review for a pretty minor change set. The `.m{j,t}s{,x}` linting added here is useful when modifying build scripts, which is why #2446 also adds that.

Vendicated:
i don't think this should be added. it makes the build scripts more convoluted for no real gain and thus worsens maintainabality. these scripts are mainly useful for ci, for development you should really just set up linting inside your editor. this use case has also never come up among those who developed user plugins

PR #1662, opened 2023-08-17
style: use `node:` module protocol

bb010g:
Depends on #1661 for `pnpm run lint:fix` to fix all files, including ESM source files.


Vendicated:
please avoid redundant style changes / unnecessary refactors, this contributes nothing relevant and just pollutes history and wastes both of our time

PR #1663, opened 2023-08-17
build(scripts): type `runInstaller`

bb010g:
Depends on #1662 (due to use of the `node:` module prefix). Hard depends on #1661 (due to use of the `.mts` file extension).


bb010g:
`runInstaller` now avoids a possible race condition around the ETag cache file.

Vendicated:
please avoid redundant style changes / unnecessary refactors, this contributes nothing relevant and just pollutes history and wastes both of our time

bb010g:
> please avoid redundant style changes / unnecessary refactors, this contributes nothing relevant and just pollutes history and wastes both of our time

Could this be reopened? It actually fixes a handful of bugs, or at least brings a few things to type compliance. It's also been refactored to no longer land on top of on #1662.

Ideally, this would also fix hang I'm running into with this script on Windows, but I'm still debugging that.

PR #1664, opened 2023-08-18
style: recommend not using `node` module scheme

bb010g:
Replacement for #1662 after discussion on Discord.


bb010g:
Project style is intentionally undefined & allowed to be inconsistent here.

PR #1666, opened 2023-08-18
customrpc: fixes

AutumnVN:
- refresh rpc when hit save, dont need to restart whole discord anymore
- make isValid not prevent user input
- fix some of my silly mistake :p


src/components/PluginSettings/components/SettingTextComponent.tsx
```diff
@@ -33,11 +33,9 @@ export function SettingTextComponent({ option, pluginSettings, definedSettings,
...
         if (typeof isValid === "string") setError(isValid);
         else if (!isValid) setError("Invalid input provided.");
-        else {
-            setError(null);
-            setState(newValue);
-            onChange(newValue);
-        }
+        else setError(null);
+        setState(newValue);
+        onChange(newValue);
```
Vendicated, line 38: does this not let you save a setting that is invalid? that would be undesired

AutumnVN: as i remember i tested yesterday it still prevent save button, just let you change the input while it invalid.

PR #1667, opened 2023-08-18
Fix patching Win32 updater with OpenAsar

CanadaHonk:
This is more generic rewrite allowing for more paths to be added in the future for whatever reason (like a rename in future Discord versions).

(The "OpenAsar" code previously was completely wrong lol!)


Vendicated:
thank you!

PR #1671, opened 2023-08-21
feat(plugin): BetterJoinedDate

AutumnVN:
image


src/plugins/betterJoinedDate.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { Tooltip } from "@webpack/common";
+
+export default definePlugin({
+    name: "BetterJoinedDate",
+    authors: [Devs.AutumnVN],
+    description: "Add a tooltip to the joined date showing the exact time and how many days ago it was",
+    patches: [{
```
Vendicated, line 15: both of these matches are pretty bad. `children:(\(.+stuff` means find the first children:( in the file, then match all characters up to the first instance of stuff. the odds of this matching the wrong thing are incredibly high

AutumnVN: ok i changed it

UlyssesZh:
I didn't see this pr, but I implemented this again in #1784.

Vendicated:
hmm i think it'd be better to format the date exactly like Discord's own timestamps
grafik

i think they use moment for that

PR #1674, opened 2023-08-21
messageLinkEmbeds: fix AutoModEmbed on compact mode

lumap:
Tried to find a solution with the Embed type but I didn't find anything

### DIFF
--- a/src/plugins/messageLinkEmbeds.tsx
+++ b/src/plugins/messageLinkEmbeds.tsx
@@ -319,7 +319,7 @@ function ChannelMessageEmbedAccessory({ message, channel, guildID }: MessageEmbe
     />;
 }
 
-const compactModeEnabled = getSettingStoreLazy<boolean>("textAndImages", "messageDisplayCompact");
+const compactModeEnabled = getSettingStoreLazy<boolean>("textAndImages", "messageDisplayCompact")!;
 
 function AutomodEmbedAccessory(props: MessageEmbedProps): JSX.Element | null {
     const { message, channel, guildID } = props;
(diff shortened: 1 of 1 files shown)


src/plugins/messageLinkEmbeds.tsx
```diff
@@ -336,7 +337,7 @@ function AutomodEmbedAccessory(props: MessageEmbedProps): JSX.Element | null {
                 <span>{isDM ? " - Direct Message" : " - " + GuildStore.getGuild(channel.guild_id)?.name}</span>
             </Text>
         }
-        compact={false}
+        compact={getSettingStoreLazy<boolean>("textAndImages", "messageDisplayCompact")?.getSetting() || false}
```
Vendicated, line 340: this should be stored inside a variable on the top level

lumap: did i do it right

Vendicated: no, that is not top level. you should store the settings store globally

src/plugins/messageLinkEmbeds.tsx
```diff
@@ -336,7 +338,7 @@ function AutomodEmbedAccessory(props: MessageEmbedProps): JSX.Element | null {
                 <span>{isDM ? " - Direct Message" : " - " + GuildStore.getGuild(channel.guild_id)?.name}</span>
             </Text>
         }
-        compact={false}
+        compact={compactModeEnabled?.getSetting() || false}
```
Vendicated, line 341:
```suggestion
        compact={compactModeEnabled.getSetting()}
```

PR #1675, opened 2023-08-22
Use rolePillBorder conditionally

TheKodeToad:
image
This fixes this problem


src/plugins/permissionsViewer/index.tsx
```diff
@@ -163,13 +163,13 @@ export default definePlugin({
         {
             find: ".Messages.BOT_PROFILE_SLASH_COMMANDS",
             replacement: {
-                match: /showBorder:.{0,60}}\),(?<=guild:(\i),guildMember:(\i),.+?)/,
-                replace: (m, guild, guildMember) => `${m}$self.UserPermissions(${guild},${guildMember}),`
+                match: /showBorder:(.{0,60})}\),(?<=guild:(\i),guildMember:(\i),.+?)/,
```
Vendicated, line 166: this seems kinda dangerous. if they add an additional prop after showBorder, it will also match that and produce broken code

TheKodeToad: Yeah, I think that occurred to me previously but i'm lazy

Vendicated:
seems broken now

Nuckyz:
Superseded by https://github.com/Vendicated/Vencord/commit/18d63af241938501c64c88877c5942b39733bbf8

PR #1679, opened 2023-08-24
feat(memberCount): format count according to user locale

ryanccn:


src/plugins/memberCount.tsx
```diff
@@ -30,6 +30,8 @@ const ChannelMemberStore = findStoreLazy("ChannelMemberStore") as FluxStore & {
     getProps(guildId: string, channelId: string): { groups: { count: number; id: string; }[]; };
 };
 
+const numberFormat = (value: number) => new Intl.NumberFormat().format(value);
```
Vendicated, line 33: creating number formats is pretty expensive, you should cache and reuse the same instance. just make a global instance

PR #1682, opened 2023-08-24
BetterDiscord theme store plugin

CREAsTIVE:
um
just a plugin

### DIFF
--- a/src/VencordNative.ts
+++ b/src/VencordNative.ts
@@ -48,7 +48,7 @@ export default {
         },
 
         addThemeChangeListener(cb: () => void) {
-            ipcRenderer.on(IpcEvents.THEME_UPDATE, cb);
+            ipcRenderer.on(IpcEvents.THEME_UPDATE, () => cb());
         },
 
         openFile: () => invoke<void>(IpcEvents.OPEN_QUICKCSS),


CREAsTIVE:
sry for commitins settings for vsc btw

Vendicated:
hello

this code is very messy, there are many issues:
- use const/let over var
- use SCREAMING_SNAKE_CASE for constants
- use a css file instead of inline styles
- storing react state in global variables is almost always, most definitely in your case, an anti pattern
- avoid using resources from random websites
- lines should be broken if they get too long
- follow the project's style and lint
- push to the customSections of settings instead of copy pasting the patch

Regardless of this, I'm still not very convinced this is a good addition. just visit the BD site 🤷🏻‍♂️
in any case, this plugin should most definitely not be enabled by default

src/plugins/betterDiscordThemeLoader/index.tsx
```diff
@@ -0,0 +1,44 @@
+import definePlugin from "@utils/types";
+import "./BetterDiscordThemesTab";
+import BetterDiscordThemesTab from "./BetterDiscordThemesTab";
+
+export default definePlugin({
+    name: "BetterDiscord Theme Loader",
```
Vendicated, line 6: this name is misleading
```suggestion
    name: "BDThemeRepo",
```

CREAsTIVE:
About bad code is expected - I wrote for the first time on TS while learning the syntax on the go) (usually I work with C #, C ++, sometimes java, rarely rust and js / css / html)
Therefore, I also planned to completely rewrite the code from scratch, refactor it. And about the need for a plugin - when I first downloaded Vencord, I wanted to look through the plugins, but it takes about 10-15 seconds to insert them, which does not allow them to be viewed normally. Therefore, for beginners (perhaps) it will be much easier to choose the right topic for themselves in this way.

CREAsTIVE:
Wait for the refactor)

CREAsTIVE:
А как мне использовать customSections?

CREAsTIVE:
I m done with refactor. What about now?

CREAsTIVE:
If you still don't like the code in some places, then tell me - I'll try to improve it

CREAsTIVE:
Oh wait, now this plugin can't be turned off...

CREAsTIVE:
fix

CREAsTIVE:
I checked everything, now everything works exactly without errors

PR #1683, opened 2023-08-24
feat: add copyUserURLs plugin

castdrian:
this adds a button to copy user urls, if this is useless just close, I mainly made it for myself since I use user urls a lot

### DIFF
--- a/src/plugins/copyUserURLs.tsx
+++ b/src/plugins/copyUserURLs.tsx
@@ -20,7 +20,7 @@ import { addContextMenuPatch, NavContextMenuPatchCallback, removeContextMenuPatc
 import { LinkIcon } from "@components/Icons";
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
-import { Menu, Clipboard, Toasts } from "@webpack/common";
+import { Clipboard, Menu, Toasts } from "@webpack/common";
 import type { Channel, User } from "discord-types/general";
 
 interface UserContextProps {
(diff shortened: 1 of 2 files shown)


CREAsTIVE:
why not a plugin?

src/plugins/copyUserURLs.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface UserContextProps {
+    channel: Channel;
+    guildId?: string;
+    user: User;
+}
+
+const UserContextMenuPatch: NavContextMenuPatchCallback = (children, { user }: UserContextProps) => () => {
+    function copyUserURL() {
+        Clipboard.copy(`<https://discord.com/users/${user.id}>`);
+        Toasts.show({
```
Vendicated, line 35: just use copyWithToast instead

castdrian: just removed the toast and cleaned it up, copy user ID doesn't have any confirmation either, dunno what my 2 AM brain was thinking

src/plugins/copyUserURLs.tsx
```diff
@@ -0,0 +1,55 @@
...
+interface UserContextProps {
+    channel: Channel;
+    guildId?: string;
+    user: User;
+}
+
+const UserContextMenuPatch: NavContextMenuPatchCallback = (children, { user }: UserContextProps) => () => {
+    children.push(
+        <Menu.MenuItem
+            id="copy-user-url"
```
Vendicated, line 35:
```suggestion
            id="vc-copy-user-url"
```

PR #1684, opened 2023-08-25
Betterdiscord themes store plugin 

CREAsTIVE:
refactored plugin


Vendicated:
like i mentioned before on Discord, i don't think this is a good addition. in its current state, this also has a multitude of issues

as such i'm closing this. thank you for your effort regardless!

PR #1688, opened 2023-08-25
Add favorite plugins

CREAsTIVE:
Added favorite button inside plugin info/settings
Favorite plugins are now placed before of others
ezgif com-video-to-gif

### DIFF
--- a/src/api/Settings.ts
+++ b/src/api/Settings.ts
@@ -44,6 +44,7 @@ export interface Settings {
     plugins: {
         [plugin: string]: {
             enabled: boolean;
+            favorite: boolean;
             [setting: string]: any;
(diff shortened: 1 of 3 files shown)


CREAsTIVE:
Updated color:
image

TheKodeToad:
wouldn't it be good to show the stars on the cards too?

CREAsTIVE:
> wouldn't it be good to show the stars on the cards too?

Только если на картах со включенной звездой. Иначе это выглядит плохо

CREAsTIVE:
> wouldn't it be good to show the stars on the cards too?

Only if on maps with the star turned on, otherwise it looks bad

PR #1689, opened 2023-08-25
feat(moreCommands): Add 'ghost ping' command

RobertsSpaceIndustries:
😈


lewisakura:
There is no purpose except to be annoying with this.

RobertsSpaceIndustries:
> There is no purpose except to be annoying with this.

Yeah, that's why I made it

PR #1691, opened 2023-08-27
messageLinkEmbeds: fix AutoModEmbed on compact mode

lumap:
I should probably stop doing commits on main


Vendicated:
thank you!

PR #1692, opened 2023-08-27
previewMessage: Hide the button once a message is sent

lumap:


src/plugins/previewMessage.tsx
```diff
@@ -33,7 +33,7 @@ const getDraft = (channelId: string) => DraftStore.getDraft(channelId, DraftType
 export function PreviewButton(chatBoxProps: Props) {
     if (chatBoxProps.type.analyticsName !== "normal") return null;
     const channelId = SelectedChannelStore.getChannelId();
-    const draft = getDraft(channelId);
+    const draft = useStateFromStores([DraftStore], () => getDraft(channelId));
```
Vendicated, line 36: hooks cannot be conditional. this must be above the early return

PR #1695, opened 2023-08-30
feat(plugin): commandHistory

HexoKnight:
Use Shift+Up/Down to cycle through previously sent messages like in a terminal.
A setting allows Ctrl to be used instead
A setting invert this (to use just Up/Down instead)

~~This does present issues with multi-line messages in that use can't use up/down arrow as usual, if anyone has a good workaround, I'd appreciate it but it isn't a main concern.~~
_**toggling the setting resolves this**_

~~Only main issue currently is that using `ComponentDispatch.dispatchToLastSubscribed("CLEAR_TEXT")` as in discord's own code seems to mess with `ComponentDispatch.dispatchToLastSubscribed("INSERT_TEXT", ...)`, causing the latter to only work exactly every other usage and I have no clue why. So if anyone has a solution, this plugin would be complete (other than any minor issues with the code that reviewers might have) and ready for merge
EDIT: on closer inspection, it seems `.dispatchToLastSubscribed("CLEAR_TEXT")` doesn't really function correctly on discord's own side (without vencord) either as it is only used in the 'Upload your message as a file' functionality (as far as I can tell ) seems to just fail to clear the text box resulting in weird behaviour~~
_**Fixed by adding a delay**_


TheKodeToad:
What if Ctrl+Up and Ctrl+Down went through history..?
Or maybe Ctrl+Up could be used for the usual behaviour

HexoKnight:
Ctrl+Up is actually currently used to return to normal behaviour (including editing the previous message) but it means that you can't go up one line at a time as the default Ctrl-Up behaviour (of going to the top of the paragraph I think) takes precedent. But I do think the idea of making it configurable it a good one regardless.

HexoKnight:
ah, that's an issue :(

TheKodeToad:
No I mean generally. not with this plugin.

HexoKnight:
oh, I see, sorry, but yeah that's why I chose Ctrl-Up, it seems to be one of the few '*-Up's that doesn't conflict with any other default behaviour (like switching between channels, etc)

HexoKnight:
Okay I think everything is fine now

Vendicated:
Seems like this has been a stale draft for very long. 

I also think this feature isn't very useful for anything besides spamming... Why would you need to resend similar messages? Not to mention that the code is not really up to standard for various reasons (extremely overcomplicated patch, `var` usage, ternary sauce, not following project formatting (e.g. using snake_case instead of camelCase)

Thank you for your work nevertheless!

HexoKnight:
The motivating usecase was commands in bot DMs, rather than in servers. Though I'm not planning on working on this as that usecase is gone for me, I haven't used this branch in months. I'm fine with this being closed.
Thanks for all your work on Vencord!!

PR #1700, opened 2023-09-02
feat: usercss

lewisakura:
Full UserCSS support for Vencord, featuring a settings dialogue, compilers for stylesheets (LESS and Stylus supported), and hot-reload (both for variable changes and when the stylesheet code is updated).


lewisakura:
TODO: 
- settings UI
- storing usercss settings somewhere

WONTFIX:
- conflicting variable names (this will usually result from using two large themes at once which is a very silly idea, and snippets should really be using unique variable names unless they're doing the same thing)

lewisakura:
I have written [some tests](https://github.com/Vendicated/Vencord/files/12561088/Testing.UserCSS.zip) that demo all of the compilers and some error resiliency in case you want to try it out. If your text goes red, it did the thing.

src/main/ipcMain.ts
```diff
@@ -47,18 +49,30 @@ function readCss() {
...
 
         const data = await getThemeData(fileName).then(stripBOM).catch(() => null);
         if (data == null) continue;
 
-        themeInfo.push(getThemeInfo(data, fileName));
+        if (fileName.endsWith(".user.css")) {
+            // handle it as usercss
+            themeInfo.push({
+                type: "usercss",
+                header: usercssParse(data, fileName)
```
Vendicated, line 67: why do you parse the header here? it doesn't seem necessary and also means that the usercss dependency is duplicated across both the renderer and the main build

src/main/ipcMain.ts
```diff
@@ -47,18 +49,30 @@ function readCss() {
...
+            });
+        } else {
+            // presumably BD but could also be plain css
+            themeInfo.push({
+                type: "bd",
```
Vendicated, line 72: perhaps?
```suggestion
                type: "other",
```

lewisakura (commented):
i'm so tired

src/main/ipcMain.ts
lewisakura: unsure, possibly? haven't actually tried to see if the BD parser / themes feature will break without properly formatted BD themes

lewisakura:
As far as I'm concerned, the actual feature implementation is now complete, so it's just refactoring and cleanup. I'll open it for review for suggestions on the aforementioned.

lewisakura:
> conflicting variable names (this will usually result from using two large themes at once which is a very silly idea, and snippets should really be using unique variable names unless they're doing the same thing)

This is not really the case and I sort of "fixed" it by isolating UserCSS variable configuration into their own settings subsets. In the case of vanilla CSS though (where it's injected using a `:root` selector) this can still cause conflicts, but in other compiled themes (i.e., Less, Stylus, and USO) it will work without conflict.

src/main/ipcMain.ts
lewisakura: shrug, should be ok

lewisakura:
Updated PR description to better reflect status.

lewisakura:
@Vendicated can this PR be prioritised? every time we make a change to themes/whatever has been moved, it requires a manual merge because the changes are apparently too complex for Git and GitHub to resolve (and for some reason the incoming changes become blank, so I have to manually go through and work out what the changes are to fix it).

coolesding:
me want usercss

effctcodes:
wow you did a lot of work on this, nice job

lewisakura:
image

coolesding:
I would love the same functionality for online usercss themes as with local. I think that would make adopting usercss for every user easier, since most people want to import the most recent version of a theme instead of having to download a snapshot of it

lewisakura (commented):
usercss sweep, 100 morbillion dollars

lewisakura:
calling agent @Nuckyz for support on the colorpicker when they're free

.vscode/settings.json
lewisakura, line 1: what i didnt edit this???

lewisakura:
@coolesding fixed, thanks for flag

re delay: one of the features is instant hotreloading - parse errors should generally be silent anyway so this shouldn't be a user facing issue

coolesding:
i thought more of like a two second delay, but yeah the notification is pretty small and doesnt matter much

coolesding:
is this still not merged?

wolfieboy09:
This has not been merged yet @coolesding

lewisakura:
this closure/deletion was accidental (just doing spring cleaning), however the PR itself is severely out of date, and will need to be rewritten anyway when usercss is standardized, which iirc was dependent on BD merging its usercss implementation

PR #1701, opened 2023-09-03
Add ShareX plugin

Rainnny7:


Vendicated:
similar plugins were already rejected before. this plugin also has a lot of issues

thank you anyway

PR #1702, opened 2023-09-03
Ignore user ID for display pronoun in messages

User9684:
This PR is useful for ignoring specific users that have abnoxious pronouns, such as a spam or outright slurs.


ImLvna:
Could you add a context menu entry to user-context to toggle the block without going into settings?

ImLvna:
like copyUserUrls.tsx

src/plugins/pronoundb/settings.ts
```diff
@@ -52,6 +52,11 @@ export const settings = definePluginSettings({
             }
         ]
     },
+    ignoreUsers: {
+        type: OptionType.STRING,
+        description: "Comma-separated list of user IDs to hide pronounds in messages",
```
lewisakura, line 57:
```suggestion
         description: "Comma-separated list of user IDs to hide pronouns in messages",
```

ImLvna: pronounds😋😋😋😋😋😋

User9684: oh man I'm so good at typing

User9684:
> Could you add a context menu entry to user-context to toggle the block without going into settings?

Sure!

TheKodeToad:
Personally I'd just block or ban those users in a lot of cases (depending on severity). I thought about opening a PR for an option to just ignore Discord pronouns!

User9684:
Not really an option if the pronoun set is just a bunch of dashes by one of your close friends who just wanna mess with people, or if its in a server you aren't a moderator in

> Personally I'd just block or ban those users in a lot of cases (depending on severity). I thought about opening a PR for an option to just ignore Discord pronouns!

TheKodeToad:
yes - I understand. I feel like there are a *lot* of people abusing this feature though, which is what gave me the other idea. So I'm not necessarily against this idea (and I'm obviously not a maintainer either) - but I can imagine the list getting quite long.

User9684:
> yes - I understand. I feel like there are a _lot_ of people abusing this feature though, which is what gave me the other idea. So I'm not necessarily against this idea (and I'm obviously not a maintainer either) - but I can imagine the list getting quite long.

Haha yeah the list would probably get very long, also, isn't there already an option to block discord pronouns in the plugin?

User9684:
Nevermind, there is not.

PR #1707, opened 2023-09-06
HideAttachments: fix broken css

AutumnVN:
no syntax highlight moment


Vendicated:
arguably this shouldnt use css in the first place but we can still fix that later

PR #1708, opened 2023-09-06
viewRaw: dont add context menu to guild folders

AutumnVN:
cuz it isnt a guild and jumpscare you with red error banner when click it


ImLvna:
i forgot theres more on guild-context than just guilds whoops

PR #1710, opened 2023-09-06
LoadingQuotes: add a new quote

lovenginx:
i consider this a significant improvement to Vencord


Vendicated:
banger pr and banger account

PR #1711, opened 2023-09-06
add oneko setting for custom spritesheet URL

fmleo:
This commit adds a setting to the oneko plugin for custom oneko.gif URL support.


Korbaux:
Can't this just be done with CSS?

Vendicated:
yes it can. and this pr is dangerous because this could be abused to execute arbitrary code for example by getting someone to use your settings backup

PR #1713, opened 2023-09-07
LastFM: Add setting for using name + artist as activity name

Mushrrom:
This just adds a setting to the last.fm plugin that allows you to set the status name to the current song and artist

image

image


Vendicated:
the order seems unconventional, isn't it usually `Artist - Song`? maybe the order should be configurable

Mushrrom:
Added settings for custom text, `artist - song name`, and `song name - artist`
image
Happy to change those settings if you want

Vendicated:
please use feature branches instead of pring from the main branch

Vendicated:
thank you!

boobthecat:
Hi I really don't know how to mess with the pull requests shit but is it possible to make the Last.FM plugin turn off when a track not available on Spotify is playing? Here's what I mean:

image I'm not listening to this song anymore and the status stay as this because I haven't scrobbled yet.

image When using Spotify, we can see the name and length of the song, which isn't ideal but is better than nothing at all. Thanks in advance.

PR #1716, opened 2023-09-08
fix(serverProfile): styling tweaks

ryanccn:
looks a bit better :3


Vendicated:
1) please show a comparison
2) the tab bar styles are directly copied from the user profile tabs so im not sure changing them is a good idea

ryanccn:
Increases paddings and margins and also makes the profile scrollable

PR #1720, opened 2023-09-08
InvisibleChat: fixup decryption modal

lovenginx:
`secret` is used for both the message with encryption and the message obtained after decrypting -- clarified difference between them in code and modal itself.
also added spacing between the two fields as they were colliding before and didn't look great
(attached: before and after changes)
image
image


lovenginx:
this is important and business

Vendicated:
thank

PR #1733, opened 2023-09-15
ClearURLs: Add si parameter on youtube.com

XAlboX:
It removes ?si= from youtube.com/shorts/ links, TESTED, it works.

### DIFF
--- a/src/plugins/clearURLs/defaultRules.ts
+++ b/src/plugins/clearURLs/defaultRules.ts
@@ -127,6 +127,7 @@ export const defaultRules = [
     "redircnt@yandex.*",
     "feature@youtube.com",
     "kw@youtube.com",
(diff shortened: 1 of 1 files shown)


Cynosphere (changes requested):
Can you add `pp@youtube.com` while you're at it, since it leaks search query (base64 encoded)

src/plugins/clearURLs/defaultRules.ts
```diff
@@ -127,6 +127,7 @@ export const defaultRules = [
     "redircnt@yandex.*",
     "feature@youtube.com",
     "kw@youtube.com",
+    "si@youtube.com",
```
Vendicated, line 130:
```suggestion
    "si@youtube.com",
    "pp@youtube.com",
```

PR #1735, opened 2023-09-17
feat(startup timings): add Time-To-Interactive info

MeguminSama:
Adds TTI information & some payload stats to the startup timings tab
image
image

### DIFF
--- a/src/plugins/startupTimings/StartupTimingPage.tsx
+++ b/src/plugins/startupTimings/StartupTimingPage.tsx
@@ -16,12 +16,18 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import "./StartupTimingPage.css";
+
+import { classNameFactory } from "@api/Styles";
 import ErrorBoundary from "@components/ErrorBoundary";
 import { Flex } from "@components/Flex";
 import { Margins } from "@utils/margins";
 import { classes } from "@utils/misc";
 import { findByPropsLazy } from "@webpack";
 import { Forms, React } from "@webpack/common";
+
+
+export const cl = classNameFactory("vc-startuptimings-");
 
 interface ITTITrackerEvent {
     emoji: string;
@@ -109,11 +115,11 @@ function TimingSection({ title, logs, traceEnd }: TimingSectionProps) {
         <Forms.FormSection title={title} tag="h1">
             <code>
                 {traceEnd && (
-                    <div style={{ color: "var(--header-primary)", marginBottom: 5, userSelect: "text" }}>
+                    <div className={cl("server-trace")} style={{ marginBottom: 5 }}>
                         Trace ended at: {(new Date(traceEnd)).toTimeString()}
                     </div>
                 )}
-                <div style={{ color: "var(--header-primary)", display: "grid", gridTemplateColumns: "repeat(3, auto) 1fr", gap: "2px 10px", userSelect: "text" }}>
+                <div className={classes(cl("grid"), cl("4-cols"))}>
                     <span>Start</span>
                     <span>Interval</span>
                     <span>Delta</span>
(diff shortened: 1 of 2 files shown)


src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -129,6 +147,78 @@ function ServerTrace({ trace }: ServerTraceProps) {
     );
 }
 
+function TTIAnalytics() {
+    const analytics = TTITracker.serializeTTITracker();
+    const filteredAnalytics = Object.entries(analytics).filter(([key, value]) => !/_start$|_end$/.test(key) && value !== null && value !== undefined);
```
Vendicated, line 152:
```suggestion
    const filteredAnalytics = Object.entries(analytics).filter(([key, value]) => !/_start$|_end$/.test(key) && value != null);
```
these two conditions should also be flipped

src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -129,6 +147,78 @@ function ServerTrace({ trace }: ServerTraceProps) {
...
+                        ))}
+                    </div>
+                </code>
+            </Forms.FormSection>
+        </ErrorBoundary>
+    );
+}
+
+function TTITimings() {
+    const records: [string, ITTITrackerEvent][] = (Object.entries(TTITracker) as [string, ITTITrackerEvent][])
```
Vendicated, line 173:
```suggestion
    const records = (Object.entries(TTITracker) as [string, ITTITrackerEvent][])
```
avoid redundant type annotations

src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -129,6 +147,78 @@ function ServerTrace({ trace }: ServerTraceProps) {
...
+        .filter(([, value]) => value instanceof Object && value.hasData?.()) as any;
```
Vendicated, line 174:
```suggestion
        .filter(([, value]) => value instanceof object && value.hasData?.()) as any;
```

src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -129,6 +147,78 @@ function ServerTrace({ trace }: ServerTraceProps) {
...
 
+function TTIAnalytics() {
+    const analytics = TTITracker.serializeTTITracker();
+    const filteredAnalytics = Object.entries(analytics).filter(([key, value]) => !/_start$|_end$/.test(key) && value !== null && value !== undefined);
+
+    return (
+        <ErrorBoundary>
+            <Forms.FormSection title="TTI Analytics" tag="h2">
+                <code>
+                    <div style={{ color: "var(--header-primary)", display: "grid", gridTemplateColumns: "auto 1fr", gap: "2px 10px", userSelect: "text" }}>
```
Vendicated, line 158: consider using a css file instead of inline styles

src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -129,6 +147,78 @@ function ServerTrace({ trace }: ServerTraceProps) {
...
+                            </React.Fragment>
+                        ))}
+                    </div>
+                </code>
+            </Forms.FormSection>
+        </ErrorBoundary>
+    );
+}
+
+function UnregisteredTimings() {
```
Vendicated, line 198: these components are almost identical. you could likely merge them into one single component that takes props

src/plugins/startupTimings/StartupTimingPage.tsx
```diff
@@ -119,13 +143,66 @@ function ServerTrace({ trace }: ServerTraceProps) {
...
+    records: [string, ITTITrackerEvent][];
+    title: string;
+    type: "registered" | "unregistered";
+}
+
+function TTITimings({ records, title, type }: TTITimingsProps) {
+    const isRegistered = type === "registered";
+
+    return (
+        <ErrorBoundary>
```
Vendicated, line 188: instead of using the errorboundaries here, it'd be much better to wrap the entire component, as otherwise it will only catch errors in the child components, but not the component itself. In any case, wrapping just the outmost component of yours should be enough

Vendicated: reminder that there is still unresolved feedback blocking this from being merged

PR #1740, opened 2023-09-18
Bundle dependencies with extensions for webstore rule compliance

Vendicated:
Since we need to move away from remote resources for extensions

This PR:
- embeds monaco & rnnoise into the extension zip
- bundles gifenc and apng-canvas with normal vencord as they are only ~30kb combined
- makes shiki & invisChat desktop only for simplicity


Vendicated:
need to test this in firefox

Vendicated:
other than that it's ready now

spikespaz:
Can we please have a separate release that includes the "desktop only" extensions so that ArmCord and WebCord can use it?

Maybe a better approach would be to have a release specifically for AMO.

ImLvna:
not possible with how armcord (and I'm assuming webcore) loads Vencord

Vendicated:
what they mean is bringing back the few plugins i made desktop only for simplicity (so that we do not have to bundle their deps)

that's definitely possible but kinda tricky to implement, since we do not currently have a way to target every platform except one, we only have a way to target one specific platform (hence desktop)

ImLvna:
ahh, okay. would it be possible to modify globPlugins to give the plugin object something like a `supported_clients` array? would allow for better control than file names and be a good starter

Vendicated:
no, unless you want to manually read the file and try to parse js with basic string parsing (not very robust)

ImLvna:
maybe something like `export const supported_clients = ...` and use a simple regex to parse that?

spikespaz:
Why do you not just let the caller decide these things rather than depending on such leaky abstractions? If you architect this like a library, instead of working from the inside out, the binary host can decide what features to use/not use. This means less work for you also.

For example, what settings pages should be visible. For clients that use their own theme loading, the one from Vencord is redundant. I'd like to see CSS handling go away because I think ArmCord handles it better (I can generate the files declaratively with Nix).

Vendicated:
I don't really understand what you're trying to say, sorry

Vencord has multiple different build targets (desktop, vesktop, generic web, extension, userscript) that include/exclude different parts of the code

what things are visible is entirely determined by the build target, as the code in question is removed completely. there is no runtime support for toggling modules and that is also not planned / possible. if you want to hide specific settings tabs like themes yourself, you are free to do so with css

PR #1742, opened 2023-09-20
Added Plugin: WatchTogetherActivityAdblock

kumpz:
Simple plugin that injects a modified version AdGuard's BlockYouTubeAdsShortcut code into youtube frames.
Basically done, just needs some testing.


ImLvna:
This plugin looks like it would only work on desktop/vesktop, as it uses electronside code. The plugin file should be suffixed with .desktop.ts

kumpz:
@ImLvna Ah, I see. Thanks for the input!

n1ckoates:
This works very well, thank you for making it! Is there anything else that needs to be done for this to be ready? I noticed there's a lot of CSS styles specified in the script, are all of those necessary for Discord's embed?

kumpz:
Replaced by #2021

PR #1743, opened 2023-09-22
BARP Plugin

Wuemeli:
Hey there,

BARP stands for "Burn After Read Paste." You can click the "Fire" button in your text input field, and then you can input a message that's then converted into a link that burns after being read.

Here's how it works:

When you send a message via the modal, it first gets sent to my proxy (the proxy is open source, so you users can self-host it). The proxy then sends it to the paste service (which can also be changed to another paste service; by default, it's paste.wuemeli.com, but for self-hosting, you will need to set up a Lenpaste instance). The paste service creates the paste and sends it back to the proxy, and the proxy, in turn, sends it back to the client. Then, the URL is in the message box, and you only need to hit "Enter."

Now, you may wonder why this proxy is necessary:

Lenpaste has a CORS policy that doesn't allow sending from the original discord.com domain. The proxy service acts as a CORS bypass.


Wuemeli:
@Vendicated Any Updates on this?

Wuemeli:
The Proxy can be found here: https://github.com/Wuemeli/BARP-Proxy

Vendicated:
be patient.

Wuemeli:
Okay No Problem

PR #1748, opened 2023-09-23
VcNarrator: Ignore multiple underscores

example-git:
Changing this plugin to include underscores in the ignored characters when you enable the latin character only option. The reason I am adding this is because the extreme usecase of a user who's username is literal @______________ joining and leaving a channel is as horrible to listen to as you'd imagine. If there is a more eloquent way to achieve this where it only does this if someone has multiple underscores in a row that would most likely be a potentially better way to achieve this.

### DIFF
--- a/src/plugins/vcNarrator.tsx
+++ b/src/plugins/vcNarrator.tsx
@@ -66,6 +66,7 @@ function clean(str: string) {
 
     return str.normalize("NFKC")
         .replace(replacer, "")
+        .replace(/_+/g, " ")
         .trim();
 }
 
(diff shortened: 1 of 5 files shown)


src/plugins/vcNarrator.tsx
```diff
@@ -66,6 +66,7 @@ function clean(str: string) {
 
     return str.normalize("NFKC")
         .replace(replacer, "")
+        .replace(/_+/g, " ")
```
Vendicated, line 69:
```suggestion
        .replace(/_{2,}/g, "_")
```

Vendicated: `____` -> `_`

example-git (commented):
Thank you, that should make this much better for this extreme use case

PR #1756, opened 2023-09-26
web: fix themes tab

TheKodeToad:
This fixes the themes tab on web. The reviewdb icons should look almost the same, in fact maybe slightly better since the viewBox seemed to be different to Discord's (and incorrect?).

### DIFF
--- a/src/plugins/reviewDB/components/MessageButton.tsx
+++ b/src/plugins/reviewDB/components/MessageButton.tsx
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import { DeleteIcon, ReportIcon } from "@components/Icons";
 import { classes } from "@utils/misc";
 import { findByPropsLazy } from "@webpack";
 import { Tooltip } from "@webpack/common";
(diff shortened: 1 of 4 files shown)


TheKodeToad:
Oh, looks like I "accidentally" moved out ReportIcon too. Not sure if that makes a lot of sense since I don't think any other plugins use the icon

src/plugins/reviewDB/components/MessageButton.tsx
```diff
@@ -31,10 +32,7 @@ export function DeleteButton({ onClick }: { onClick(): void; }) {
                     className={classes(iconClasses.button, iconClasses.dangerous)}
                     onClick={onClick}
                 >
-                    <svg width="16" height="16" viewBox="0 0 20 20">
-                        <path fill="currentColor" d="M15 3.999V2H9V3.999H3V5.999H21V3.999H15Z" />
-                        <path fill="currentColor" d="M5 6.99902V18.999C5 20.101 5.897 20.999 7 20.999H17C18.103 20.999 19 20.101 19 18.999V6.99902H5ZM11 17H9V11H11V17ZM15 17H13V11H15V17Z" />
-                    </svg>
+                    <DeleteIcon width={20} height={20} />
```
Vendicated, line 35: nit
```suggestion
                    <DeleteIcon width="20" height="20" />
```

PR #1758, opened 2023-09-27
ClearURLs: Add Discord Copy Link Tags

brandonator365:
Copying links in the Discord app now appends a ton of extra tags to the end. This seeks to fix that.
This was added for a day or two, removed for a day or two, and has since been added back again.
This implementation uses wildcards in case Discord changes this for old media.discordapp.net links.
That being said, I did a bit of research, and it doesn't seem copying media.discordapp.net links appends anything extra to the end, so this may be unnecessary. If this causes problems, let me know.

### DIFF
--- a/src/plugins/clearURLs/defaultRules.ts
+++ b/src/plugins/clearURLs/defaultRules.ts
@@ -139,4 +139,7 @@ export const defaultRules = [
     "si@open.spotify.com",
     "igshid",
     "share_id@reddit.com",
+    "ex@*.discordapp.*",
+    "is@*.discordapp.*",
+    "hm@*.discordapp.*",
 ];


Cynosphere:
My personal opinion is to wait until the feature is 1. announced properly to stop the misconception of why these are even added **(its not tracking)** and 2. the client has [the proper link renewal mechanisms in place](https://toaster.sh/i/kdrdrqeu.png).

Cynosphere:
Also to note that the parameters are not finalized as `hm` used to be `sg`, they could very well be changed again.

brandonator365:
Noted. Where was the screenshot in the second part of the first comment taken from? It's obvious I'm missing something here.
Is this an attempt to keep Discord share links on Discord?

Cynosphere:
> Where was the screenshot in the second part of the first comment taken from?

Discord API

> Is this an attempt to keep Discord share links on Discord?

Yes. It was originally added for server shops but they've decided to eventually roll it out for all attachments (other CDN assets are unaffected)

brandonator365:
I appreciate the insight. I'll just keep an eye on this for now then.

dolfies:
Even when this feature is finalized, stripping these parameters will eventually result in an unusable link.

PR #1763, opened 2023-09-29
feat(plugin): PermissionFreeWill

lewisakura:
aka FreiheitlicheBerechtigungsKonfiguration

### DIFF
--- a/src/plugins/permissionFreeWill/index.ts
+++ b/src/plugins/permissionFreeWill/index.ts
@@ -34,7 +34,7 @@ export default definePlugin({
             find: "Messages.SELF_DENY_PERMISSION_BODY",
             replacement: [
                 {
-                    match: /case "DENY":.{0,50}if\(\i\.\i\.can/,
+                    match: /case"DENY":.{0,50}if\((?=\i\.\i\.can)/,
                     replace: "$&true||"
                 }
             ],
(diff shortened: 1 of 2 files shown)


src/plugins/permissionFreeWill/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    description: "Disables the client-side restrictions for channel permission management.",
+    authors: [Devs.lewisakura],
+
+    patches: [
+        // Permission lockout, just set the check to true
+        {
+            find: "Messages.SELF_DENY_PERMISSION_BODY",
+            replacement: [
+                {
+                    match: /"Unexpected boolean action".{5,20}allow.{5,20}.{80,100}if\(/,
```
Vendicated, line 37: this does not seem very robust.
It's supposed to match this if (C.Z.can), right? 
grafik

much more robust would be smth like
```suggestion
                    match: /case "DENY":.{0,50}if\(\i\.\i\.can/,
```

src/plugins/permissionFreeWill/index.ts
```diff
@@ -0,0 +1,56 @@
...
+                }
+            ],
+            predicate: () => settings.store.lockout
+        },
+        // Onboarding, same thing but we need to prevent the check
+        {
+            find: "Messages.ONBOARDING_CHANNEL_THRESHOLD_WARNING",
+            replacement: [
+                {
+                    match: /case 1:if\((?=!\i\.sent)/,
```
Vendicated, line 48: like above, this is not robust. you're patching regeneratorRuntime, their async await polyfill. what you're patching here is literally the equivalent of `if (!await...)`

but you can just use the CANNOT_CHANGE_CHANNEL_PERMS right after to reliably match it
grafik

PR #1767, opened 2023-09-30
TextReplace: Use settings.json for storing rules instead of DataStore API

HeLau1337:
The changes of this PR change the way the existing TextReplace plugin stores the rules (find, replace). It uses the settings.store (aka. the settings.json file in the AppData directory) instead of Vencord's DataStore API (aka. the IndexedDB of Chromium).

Using the settings.json would have a few benefits imo:
- rules could be synced with Vencord Cloud Sync
- users can edit their rules in the settings.json directly using any text editor (which is a blessing if you want to edit a lot of rules or copy/paste them)
- sharing an entire set of rules with other users would be easier, too

I'm not aware of potential drawbacks. Let me know what you think about it.

image

### DIFF
--- a/src/plugins/textReplace/index.tsx
+++ b/src/plugins/textReplace/index.tsx
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import { DataStore } from "@api/index";
 import { addPreSendListener, removePreSendListener } from "@api/MessageEvents";
 import { definePluginSettings } from "@api/Settings";
 import { Flex } from "@components/Flex";
@@ -28,7 +29,6 @@ import { Button, Forms, React, TextInput, useState } from "@webpack/common";
 
 const STRING_RULES_KEY = "TextReplace_rulesString";
 const REGEX_RULES_KEY = "TextReplace_rulesRegex";
-const PLUGIN_NAME = "TextReplace";
 
 type Rule = Record<"find" | "replace" | "onlyIfIncludes", string>;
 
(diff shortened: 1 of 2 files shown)


ImLvna:
use settings.store instead of vencord.settings.plugins

src/plugins/textReplace/index.tsx
```diff
@@ -242,15 +265,18 @@ const TEXT_REPLACE_RULES_CHANNEL_ID = "1102784112584040479";
...
     dependencies: ["MessageEventsAPI"],
 
     settings,
 
     async start() {
-        stringRules = await DataStore.get(STRING_RULES_KEY) ?? makeEmptyRuleArray();
-        regexRules = await DataStore.get(REGEX_RULES_KEY) ?? makeEmptyRuleArray();
+        getRulesFromSettingsJson();
+        if (!("rules" in Vencord.Settings.plugins[PLUGIN_NAME])) {
+            // If the "rules" key doesn't exist in the settings.json yet, create it at startup so users can directly start defining their rules in the settings.json using a blank template.
```
ImLvna, line 277: this would have to be changed relatively soon

HeLau1337:
> use settings.store instead of vencord.settings.plugins

Oops, will fix that, thanks for letting me know!

src/plugins/textReplace/index.tsx
```diff
@@ -16,7 +16,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { DataStore } from "@api/index";
```
ImLvna, line 19: allow user to keep settings from datastore

ImLvna:
also
> new rule if you update a plugin that has no readme you are legally obligated to write one
- vee

src/plugins/textReplace/index.tsx
HeLau1337: Good point, didn't think of that. My bad.

ImLvna:
optimally this rewrite should wait on #696

ImLvna:
also this is already implimented in #696 by the looks of it. 
> the example impl is a TextReplace plugin

src/plugins/textReplace/index.tsx
HeLau1337: I added data migration in my latest commit.

src/plugins/textReplace/index.tsx
HeLau1337: Sorry, what exactly are you referring to? And does it still apply to the current state of the code (I pushed changes)?

HeLau1337:
I was told that my idea was suggested before and got declined: https://github.com/Vendicated/Vencord/pull/540#issuecomment-1450623262

The Discord user vending.machine also made me aware of the drawbacks:

> settings.json is a very inefficient storage 
it's blocking & json is inherently kinda shit
it's also stored in localStorage in browsers which is potentially limited to a few kb
DataStore/indexeddb is much more efficient, not blocking (async) and has no storage limit
settings should only be used for basic booleans and small data. text replace rules can potentially become plenty and large storing them in settings is bad

PR #1773, opened 2023-10-03
feat(permissionfreewill): onboarding deletion

lewisakura:
onboarding has client-side restrictions to prevent deleting channels used in it, although in reality this is actually useless since you can still use onboarding without them


Vendicated:
i don't really understand your draft reason... is it just that you're unsure if all channel types can actually be deleted server side? i think it's fine to merge this way and if some can't be, surely people will complain sooner or later

if it's really just that, let's just merge it (if the patch even still works)

lewisakura:
It was drafted because the patch half-worked and half-didn't. I just haven't had time to go through it and fix it up.

PR #1776, opened 2023-10-05
Fix command API

wntiv-main:
Fix the command API, broken by discord removing the /giphy command, which was used to locate the injection point for the CommandAPI _init() function. This fix is only a sketchy temporary fix and i am uncertain whether it should be used over a better, more stable solution.. but it works for now for me ig? I am mostly worried that i have seen some users who still have /giphy.. so this would break  the CommandAPI for them..

### DIFF
--- a/src/plugins/_api/commands.ts
+++ b/src/plugins/_api/commands.ts
@@ -26,7 +26,7 @@ export default definePlugin({
     patches: [
         // obtain BUILT_IN_COMMANDS instance
         {
-            find: '"giphy","tenor"',
+            find: '"gif","tenor"',
             replacement: [
                 {
                     // Matches BUILT_IN_COMMANDS. This is not exported so this is
@@ -34,7 +34,7 @@ export default definePlugin({
                     // patch simpler
 
                     // textCommands = builtInCommands.filter(...)
-                    match: /(?<=\w=)(\w)(\.filter\(.{0,30}giphy)/,
+                    match: /(?<=\w=)(\w)(\.filter\(.{0,30}gif)/,
                     replace: "Vencord.Api.Commands._init($1)$2",
                 }
             ],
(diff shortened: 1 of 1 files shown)


JDJGInc:
Would an if check work here?

wntiv-main:
ahh sorry i got no idea im not very experienced with vencord plugining yet..

src/plugins/_api/commands.ts
```diff
@@ -26,15 +26,15 @@ export default definePlugin({
     patches: [
         // obtain BUILT_IN_COMMANDS instance
         {
-            find: '"giphy","tenor"',
+            find: '"gif","tenor"',
```
Vendicated, line 29:
```suggestion
            find: ',"tenor"',
```

src/plugins/_api/commands.ts
```diff
@@ -26,15 +26,15 @@ export default definePlugin({
...
             replacement: [
                 {
                     // Matches BUILT_IN_COMMANDS. This is not exported so this is
                     // the only way. _init() just returns the same object to make the
                     // patch simpler
 
                     // textCommands = builtInCommands.filter(...)
-                    match: /(?<=\w=)(\w)(\.filter\(.{0,30}giphy)/,
+                    match: /(?<=\w=)(\w)(\.filter\(.{0,30}gif)/,
```
Vendicated, line 37:
```suggestion
                    match: /(?<=\w=)(\w)(\.filter\(.{0,60}tenor)/,
```

Vendicated (changes requested):
this should make it work with both versions. i haven't tested it but it should work

wntiv-main:
yeah that looks like it should work for both, ive tested on mine (which has new /gif) with no issue, i can only assume it would work on old (/giphy) also

wntiv-main:
also, most users will have updated discord by the time that vencord update also, no?

Vendicated (approved):

PR #1779, opened 2023-10-06
feat(Plugin): PackManager

arHSM:
Simple utility to (un)collect packs with just the guild id, and some patches for fixing packs from joined servers.


arHSM:
A little demo
pxnla4ISF8

arHSM:
killed by pisscord

PR #1780, opened 2023-10-06
fix moreUserTags

sunnniee:


src/plugins/moreUserTags/index.tsx
```diff
@@ -230,10 +228,10 @@ export default definePlugin({
         },
         // in the member list
         {
-            find: ".renderBot=function(){",
+            find: ".Messages.GUILD_OWNER,",
             replacement: {
-                match: /\.BOT;return null!=(\i)&&.{0,10}\?(.{0,50})\.botTag,type:\i/,
-                replace: ".BOT;var type=$self.getTag({...this.props,origType:$1.bot?0:null,location:'not-chat'});return type!==null?$2.botTag,type"
+                match: /(.USE_EXTERNAL_EMOJIS,.{0,50}(?<channel>\i)\),\(0,.{0,350})(?<type>\i)=\(null==(?<user>\i).{0,50}\.BOT,(.{0,10})&&\i.bot/,
```
Vendicated, line 233: this has way too many wildcards. why do you match this much code? that's most likely not needed

sunnniee: it was indeed not needed 👍

sunnniee:
so is this getting merged or

Vendicated:
thank you

PR #1642, opened 2023-08-12
feat(VcNarrator): add `{{DISPLAY_NAME}}` as placeholder

Marocco2:
`{{DISPLAY_NAME}}` is placeholder for user's `global_name` object if it exists, otherwise it defaults to `username`

PR #1781, opened 2023-10-07
VcNarrator: Add 'Use Display Name' option + update discord-types from 1.3.26 -> 1.3.3

iAmSpace:
Allows users to hear the display name of a user instead of the username, which is useful as some usernames dont sound right when narrated as they don't have any spaces and many people are now using periods, making `username.` sound like "username..... has joined your channel"


Vendicated:
dupe of #1642

PR #1787, opened 2023-10-10
new plugin NoServerEmoji ~ hides server emojis from autocomplete

UlyssesZh:
Don't show server emojis in autocomplete menu (optionally, show the current server's emojis but not others).


src/plugins/noServerEmojis/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    description: "Do not show server emojis in the autocomplete menu.",
+    settings: definePluginSettings({
+        includeGuilds: {
+            description: "If disabled, server emojis will not be shown in the autocomplete menu at all.",
+            type: OptionType.BOOLEAN,
+            default: false
+        },
+        includeExternalGuilds: {
+            description: "If disabled, emojis from external servers will not be shown in the autocomplete menu. " +
+                "This setting has no effect if \"Include guild emojis\" is disabled. " +
```
Vendicated, line 23: remove this and instead use isDisabled or whatever the property is called to automatically disable this option fi the other is disabled

src/plugins/noServerEmojis/index.ts
```diff
@@ -0,0 +1,47 @@
...
+            default: false
+        }
+    }),
+    patches: [
+        {
+            find: "queryEmojiResults:",
+            replacement: {
+                match: /searchWithoutFetchingLatest\(\{/,
+                replace: "$&" +
+                    "includeExternalGuilds:$self.settings.store.includeExternalGuilds," +
```
Vendicated, line 35: slightly prettier would be:
```js
replace: "$&...$self.getExtraProps()"
```
and getExtraProps returns object

src/plugins/noServerEmojis/index.ts
UlyssesZh: I switched to a select instead of two checkboxes in [133c7d8](https://github.com/Vendicated/Vencord/pull/1787/commits/133c7d8596f47e835fd508412f67567ea9d15f82).

src/plugins/noServerEmojis/index.ts
UlyssesZh: Improved in [133c7d8](https://github.com/Vendicated/Vencord/pull/1787/commits/133c7d8596f47e835fd508412f67567ea9d15f82).

src/plugins/noServerEmojis/index.ts
```diff
@@ -0,0 +1,51 @@
...
+            find: "}searchWithoutFetchingLatest(",
+            replacement: {
+                match: /(searchWithoutFetchingLatest.+?=(\i);)(.+?reduce\(\((\i),(\i)\)=>\{)/,
+                replace: "$1 let includeGuilds = $2.includeGuilds ?? true; $3 if ($5.type === 'GUILD_EMOJI' && !includeGuilds) { return $4; }"
+            }
+        }
+    ],
+    getExtraProps() {
+        return {
+            includeExternalGuilds: this.settings.store.shownEmojis === "all",
```
Vendicated, line 47: you don't reference this at all, seems like a mistake?

the currentServer option shows all guilds so this is likely the culprit

Vendicated (commented):
somehow the patches still work after all this time O_O, sorry for the long wait!

UlyssesZh:
It turns out that, if FakeNitro is enabled, and this plugin is set to show only emojis in current server, it would not work. Seems like FakeNitro is forcing all emoji to be shown. I'll see what I can do.

Vendicated:
oh i see, that explains! i indeed have it enabled

UlyssesZh:
OK, I saw the problem. The user needs to turn off "Enable Emoji Bypass" in the settings of FakeNitro. Then my plugin works. I guess I don't have to specially modify my plugin to make it have higher precedence over that FakeNitro setting because they are just in conflict in nature. @Vendicated Please check if there is still a problem when you turn off "Enable Emoji Bypass". Do I need to add a notice about this conflict somewhere?

Vendicated:
please attempt to fix the problem in a way that makes the two plugins compatible

if you could explain the exact problem a bit better, maybe i could advise how to do that

UlyssesZh:
I seem to have fixed the problem now.

src/plugins/noServerEmojis/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const settings = definePluginSettings({
+    shownEmojis: {
+        description: "The types of emojis to show in the autocomplete menu.",
+        type: OptionType.SELECT,
+        default: "onlyUnicode",
+        options: [
+            { label: "Only unicode emojis", value: "onlyUnicode" },
+            { label: "Unicode emojis and server emojis from current server", value: "currentServer" },
+            { label: "Unicode emojis and all server emojis (Discord default)", value: "all" }
```
Vendicated, line 19: what's the point of this option? isn't this the same as just disabling the plugin?

Vendicated:
can confirm it works as expected now in that regard! 

however there seems to be a new issue: when using the currentServer option, the emoji picker (ctrl + e) will only show emojis of the current guild. the other two options will show all servers

i think the purpose of this plugin is to just hide it from the autocomplete menu when you type `:a`, right? so i think this is a bug

UlyssesZh:
I used a cleaner patching approach. The <kbd>Ctrl</kbd>+</kbd>E</kbd> picker is also fixed.

src/plugins/noServerEmojis/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    name: "NoServerEmojis",
+    authors: [Devs.UlyssesZhan],
+    description: "Do not show server emojis in the autocomplete menu.",
+    settings,
+    patches: [
+        {
+            find: "}searchWithoutFetchingLatest(",
+            replacement: {
+                match: /searchWithoutFetchingLatest.{20,300}get\((\i).{10,40}?reduce\(\((\i),(\i)\)=>\{/,
+                replace: `
```
Vendicated, line 34: i feel bad asking for changes so often but having this much code in your replacement is error prone (e.g. super easy to miss a bracket). always prefer declaring a method on your plugin and using that


```suggestion
                replace: "$& if($self.someFuncYouPickTheName($1,$3))return $2;" 
```

Vendicated:
looks good now! but i still don't understand the purpose of `Unicode emojis and all server emojis (Discord default)`. how does it differ from just disabling the plugin?

UlyssesZh:
No difference, but you can switch to that option without refreshing your client. This is useful when one occasionally wants to send an external emoji.

Vendicated:
cursed, but sure

Vendicated:
thank you!

PR #1788, opened 2023-10-10
Plugin: Fix Twitter Embeds

csh:
Simple plugin that automatically replaces twitter.com and x.com URLs with ones that have been "fixed" so embeds continue to work.


Vendicated:
yes there are already textreplace rules for this. thank you anyway!

PR #1790, opened 2023-10-12
feat(Plugin): SoundTriggers

battlesquid:
Plugin to allow users to define rules for playing audio when text/messages are sent.


justinfernald:
I like this plugin, is good 👍

Vendicated:
hi, thank you for your contribution! and apologies for the late comment

so this seems to just be a customisable version of the Moyai plugin? what exactly is the use case for this, why would someone need this?

k26pl:
Probably similar to the Moyai, and from what i understand it can allow for custom notification sounds for certain messages, which may be useful.

Battlesquid:
> it can allow for custom notification sounds for certain messages
Basically this. The original goal was to make the moyai plugin but for the skull emoji with the bad to the bone riff. Figured it would be duplicate work so I opted to make a generalized plugin that allowed you to make your own custom sounds.

It's been a hot minute since this was updated so I'll need to do that if you're okay with actually adding it.

Battlesquid:
Fixes ended up being relatively straightforward, should be working as of now.

Vendicated:
closing as this is not useful

PR #1791, opened 2023-10-12
feat(plugin): NoMosaic

AutumnVN:


src/plugins/noMosaic/index.ts
```diff
@@ -0,0 +1,42 @@
...
+            {
+                match: /\i===\i\.\i\.MOSAIC/,
+                replace: "true",
+            },
+            {
+                match: /null!==\(\i=\i\.get\(\i\)\)&&void 0!==\i\?\i:"INVALID"/,
+                replace: '"INVALID"',
+            },
+            {
+                match: /\i\.length>1/,
```
Vendicated, line 34: this looks kinda dangerous, if it ever matches the wrong thing a crash is very likely. maybe make it slightly more specific

src/plugins/noMosaic/index.ts
```diff
@@ -0,0 +1,42 @@
...
+                replace: "false",
+            },
+        ],
+    }],
+    start() {
+        enableStyle(style);
```
Vendicated, line 40: stop disableStyle in stop (i know stop currently isnt used for plugins with patches but still)

PR #1794, opened 2023-10-13
new plugin CtrlEnterSend

UlyssesZh:
Use <kbd>Ctrl</kbd>+<kbd>Enter</kbd> to send messages. The actual behavior can be customized.


src/plugins/ctrlEnterSend/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+
+let compiledSubmitRule: ((event: KeyboardEvent, codeblock: boolean) => boolean) | null = null;
+let compiledNewlineRule: ((event: KeyboardEvent, codeblock: boolean) => boolean) | null = null;
+
+const plugin = definePlugin({
```
Vendicated, line 14: storing the plugin in a variable is ugly. just make the functions you need global functions so you don't need to use the plugin object

src/plugins/ctrlEnterSend/index.ts
```diff
@@ -0,0 +1,99 @@
...
+            textArea.value += "\r\n";
+            textArea.dispatchEvent(new Event("input", { bubbles: true, cancelable: true }));
+        }
+    },
+    start() {
+        this.compileSubmitRule(this.settings.store.submitRule);
+    },
+    compileSubmitRule(rule: string): boolean {
+        try {
+            compiledSubmitRule = new Function("event", "codeblock", `
```
Vendicated, line 69: do not eval user input, that's a terrible terrible terrible idea

src/plugins/ctrlEnterSend/index.ts
UlyssesZh: Avoided storing the plugin in a variable in [95f8513](https://github.com/Vendicated/Vencord/pull/1794/commits/95f85130eb3568ed26044c6f24bb009f53f60afd).

src/plugins/ctrlEnterSend/index.ts
UlyssesZh: OK. I don't want to implement an AST parser, so I just sacrificed some customizability. This should still be useful enough, however.

Vendicated: yeah that's way better, thanks

Vendicated:
seems like this broke, unless i'm missing something :( the patch still applies but it doesn't do anything and the function is never called

looks good other than that, so it could be merged if you fixed it

UlyssesZh:
The code that I patch here is now unused by Discord. I cannot find where the logic is defined now.

Nuckyz:
> The code that I patch here is now unused by Discord. I cannot find where the logic is defined now.

Module 780748 in stable, check it

UlyssesZh:
I already found there.
```js
case l.KeyboardKeys.ENTER:
    // `E()` here does nothing (calls a function that calls a function short-circuited by a guard clause).
    if ((0,
    o.checkModifiers)(e, {}) && (null == E ? void 0 : E(e))) {
        e.preventDefault(),
        e.stopPropagation();
        return
    }
}
// `i()` contains a switch-case, none of whose cases is entered.
if ((null === (i = t.onKeyDown) || void 0 === i ? void 0 : i.call(t, e)) === !0) {
    e.preventDefault(),
    e.stopPropagation();
    return
}
e.which === l.KeyboardKeys.ENTER && (!e.altKey && !e.shiftKey && (!u || e.ctrlKey) && !(0,
s.hasOpenPlainTextCodeBlock)(t) || !I) && (e.preventDefault(),
e.stopPropagation(),
// `T()` handles bot command suggestions.
T()),
// `d()` has some switch-cases, none of whose cases is entered, and then calls a function that handles tab highlighting.
null == d || d(e)
```
Nothing higher in the call stack of these seems related to the things I want. I also looked at other places.

Nuckyz:
T calls a method which calls onSubmit which handles submitting the message

UlyssesZh:
Thanks to @Nuckyz, I fixed this.

src/plugins/ctrlEnterSend/index.ts
```diff
@@ -0,0 +1,68 @@
...
+            description: "Whether to send a message in the middle of a code block",
+            type: OptionType.BOOLEAN,
+            default: true,
+        }
+    }),
+    patches: [
+        {
+            find: "KeyboardKeys.ENTER&&(!",
+            replacement: {
+                match: /((\i)\.which.+?KeyboardKeys.ENTER)&&.+?(\(0,\i\.hasOpenPlainTextCodeBlock\)\(\i\)).+?&&(\(\i\.preventDefault)/,
```
Vendicated, line 45: try to match as little as possible and avoid using `.+`
```suggestion
                match: /(?<=(\i)\.which===\i\.KeyboardKeys.ENTER&&).{0,100}(\(0,\i\.hasOpenPlainTextCodeBlock\)\(\i\)).{0,100}&&(?=\(\i\.preventDefault)/,
```

src/plugins/ctrlEnterSend/index.ts
```diff
@@ -0,0 +1,68 @@
...
+                replace: "$1 && $self.shouldSubmit($2, $3) && $4"
```
Vendicated, line 46:
```suggestion
                replace: "$self.shouldSubmit($1, $2) &&"
```

Vendicated (commented):
thanks for the fix!

untested but it should probably work

Vendicated:
thank you

let's hope discord doesn't rewrite their entire chatbox key handler logic again

PR #1795, opened 2023-10-13
LoadingQuotes: more customization & custom quotes support

### LINKED ISSUE #1654: [Feature Request] LoadingQuotes Add custom Quotes
### Discord Account

totallynotmwa

Hi i think it would be nice to add custom Quotes 

### Request Agreement

UlyssesZh:
Resolve #1654. Resolve Vencord/plugin-requests#185.

Fix quotes not replaced during Halloween.


src/plugins/loadingQuotes/index.ts
```diff
@@ -63,13 +64,36 @@ const quotes = [
...
+            default: "|",
+        },
+    }),
     patches: [
         {
             find: ".LOADING_DID_YOU_KNOW",
             replacement: {
-                match: /\._loadingText=.+?random\(.+?;/s,
-                replace: "._loadingText=$self.quote;",
+                match: /(\._loadingText=\(\i=)(\[.{0,4000}\])(\)\[.{0,10}random\(.+?;)/s,
```
Vendicated, line 95: using such large ranges is weird. either use .+? if it's safe, or find a better way to do this

src/plugins/loadingQuotes/index.ts
```diff
@@ -80,7 +104,15 @@ export default definePlugin({
         return String.fromCharCode(...codes);
     },
 
-    get quote() {
-        return this.xor(quotes[Math.floor(Math.random() * quotes.length)]);
+    quotes(preset: string[]) {
+        let result: string[] = this.settings.store.enableDiscordPresetQuotes ? preset : [];
```
Vendicated, line 108: make a global variable for the settings and use that instead of using it via this

src/plugins/loadingQuotes/index.ts
```diff
@@ -80,7 +104,15 @@ export default definePlugin({
...
+        if (this.settings.store.enablePluginPresetQuotes) {
+            result = result.concat(quotes.map(this.xor));
```
Vendicated, line 110:
```suggestion
           result.push(...quotes.map(this.xor));
```

src/plugins/loadingQuotes/index.ts
```diff
@@ -80,7 +104,15 @@ export default definePlugin({
...
+        }
+        result = result.concat(this.settings.store.additionalQuotes.split(this.settings.store.additionalQuotesDelimiter));
```
Vendicated, line 112: same as above. you should likely also .trim() the elements

src/plugins/loadingQuotes/index.ts
```diff
@@ -80,7 +104,15 @@ export default definePlugin({
...
+        if (result.length === 0) {
```
Vendicated, line 113: bad style. `return result || "Loading..."`

Vendicated (changes requested):

src/plugins/loadingQuotes/index.ts
UlyssesZh: The range is large because there are many quotes in the middle. Setting the upper bound to 4000 is to prevent it from searching the rest of the module.

src/plugins/loadingQuotes/index.ts
UlyssesZh: `result` is an array, so `[] || ["something"]` will return `[]`, which is undesired.

src/plugins/loadingQuotes/index.ts
UlyssesZh: `trim()` may be undesired if users want to have their quotes with leading or trailing spaces.

src/plugins/loadingQuotes/index.ts
Vendicated: i realised that too now. but you should still use a ternary

src/plugins/loadingQuotes/index.ts
```diff
@@ -99,7 +121,12 @@ export default definePlugin({
...
     },
 
-    get quote() {
-        return this.xor(quotes[Math.floor(Math.random() * quotes.length)]);
+    quotes(preset: string[]) {
+        const result: string[] = settings.store.enableDiscordPresetQuotes ? preset : [];
+        if (settings.store.enablePluginPresetQuotes) {
+            result.push(...quotes.map(this.xor));
+        }
+        result.push(...settings.store.additionalQuotes.split(settings.store.additionalQuotesDelimiter));
```
Vendicated, line 129: split on an empty string will yield `[""]` so if you don't use the custom quote feature there's a small chance to get an empty quote
```suggestion
        result.push(...settings.store.additionalQuotes.split(settings.store.additionalQuotesDelimiter).filter(Boolean));
```

src/plugins/loadingQuotes/index.ts
```diff
@@ -61,36 +61,58 @@ const quotes = [
     "Wdn`khc'|f*eghl{%"
 ];
 
+
 const settings = definePluginSettings({
     replaceEvents: {
```
Vendicated, line 66: with the new option of toggling default discord badges, maybe this setting should be changed to "enableEventQuotes" that's independent of normal discord quotes? not sure how feasible that would be

UlyssesZh: What do you mean by being independent of normal Discord quotes? Discord disables the normal quotes during events, so they are exclusive to each other. If you mean enabling the normal quotes on top of the event quotes during events, I may be able to do that.

What new option of toggling default Discord badges?

Vendicated: > What new option of toggling default Discord badges?

quotes, not badges, brainfart! i was mainly referring to this setting being quite confusing now that you can toggle default discord quotes. it'd probably be better to reword it a bit

UlyssesZh: Ah I see your point. What about this: `replaceEvents` says "This plugin effects events (e.g. Halloween)", and `enableDiscordPresetQuotes` says "Enable Discord's preset quotes (including event quotes, during events)".

UlyssesZh:
Converting to a draft because I suddenly cannot get my installation of Vencord working. The last commit is untested.

UlyssesZh:
Back on. It should be good now.

lewisakura:
Can you please outright remove the Vaporeon copypasta quote (see #2474)

UlyssesZh:
Already removed. See the last commit message.

lewisakura:
Outright, not commented out, I don't think we need to keep the line if it's never seen.

src/plugins/loadingQuotes/quotes.js
```diff
@@ -0,0 +1,61 @@
+/*
```
Vendicated, line 1: why's this a source file? just name it `quotes.txt` and make it newline separated quotes

you can import it in ur plugin via 
```js
import QuotesText from "~fileContent/quotes.txt";

const Quotes = QuotesText.trim().split("\n")
```

UlyssesZh: Hmm, I didn't know one can do this in JS/TS. I use this approach now.

UlyssesZh:
> Outright, not commented out, I don't think we need to keep the line if it's never seen.
Done. The commenting syntax is retained, though, in case people want to leave comments there.

lewisakura:
thank you!

Vendicated:
the patches were kinda bad for the following reasons, thus I changed them:
- reliant on the current easter egg that might get removed at any time
- reassigning the variable which might be made a `const` in the future. the plugin now mutates the array in place
- just generally using fixed length anything matchers is best avoided

but the rest is good now so I'll go ahead and merge this, thank you for your work!!

PR #1799, opened 2023-10-14
Prevent videos playing in modal with NoMosaic

TheKodeToad:


TheKodeToad:
Huh looks like this also disables the carousel. Maybe I should make it more specific and split into multiple options.

Cynosphere:
```js
{
  find: /poster:\i\.toString\(\),naturalWidth/,
  replacement: {
    match: /\i\[(\i)\.src\]=function\(\i\){return \i\(\i,\i,\i,\i\)}/,
    replace: (orig, state) => `if(!${state}?.component?.props?.poster)${orig}`
  },
}
```

TheKodeToad:
what?

Cynosphere:
Patch that allows playing videos inline without disabling carousel entirely

TheKodeToad:
thinking about it carousel doesn't make as much sense if clicking the video doesn't open it

PR #1800, opened 2023-10-14
[Plugin Request] UserPFP

Yeetov:
I have created a plugin for UserPFP which is for having an animated profile picture without nitro.


TheKodeToad:
..why would you have a plugin to import CSS?
wouldn't it be better to patch the code which determines the src as this appears to be pretty inefficient

lewisakura:
this is literally just a CSS snippet, add it to your quickcss

PR #1804, opened 2023-10-15
betterNotes, userVoiceShow: fix padding issue

AutumnVN:
- make betterNotes adds padding when note is hidden and note is last section (so self profile wont missing padding bottom)

because of above change and userVoiceShow is between note box and message box:
- add padding-bottom and remove padding-top of userVoiceShow when in self profile
  - remove padding-top because the above change added space above userVoiceShow in self profile even if note is hidden or not
  - add padding-bottom because it's now last section, self profile dont have message box below


Vendicated:
thanks

Issue #1556, opened 2023-07-29
[Feature Request] Option: Alt + F4 minimize to tray instead of closing.


ChickenIQ:
### What is it that you'd like to see?

I would like an option that makes Discord minimize itself to the system tray when using Alt+F4, instead of closing completely, similar to how it works on Linux.

### Request Agreement

- [X] This is not a plugin request

PR #1808, opened 2023-10-16
minimize on alt+f4 (#1556)

gorgulenkozxc:
resolve https://github.com/Vendicated/Vencord/issues/1556

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.6.6",
+    "version": "1.6.7",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
     "bugs": {


Vendicated:
hi! thanks for the contribution but this is a bad way to solve this problem for various reasons
- linux and mac use a different shortcut than alt f4
- the current code reads and parses the settings on every single key stroke. this is very inefficient. it would be much better to instead disable the electron keybinds responsible for it

gorgulenkozxc:
> hi! thanks for the contribution but this is a bad way to solve this problem for various reasons
> 
>     * linux and mac use a different shortcut than alt f4
> 
>     * the current code reads and parses the settings on every single key stroke. this is very inefficient. it would be much better to instead disable the electron keybinds responsible for it

hello, thank you for taking the time to write a complete response!

indeed, I should've been more aware of the existing code and register the key combination according to the OS

as you suggested, I tried to use the registration of the shortcut via the electron.js native ways: `globalShortcut` and `submenu`. here are the pieces of code I tried to solve the problem with:
`index.ts in main`
```ts
globalShortcut.register("Alt+F4", () => {
    app.hide();
});
```

`pathcher.ts, analogy with winCtrlQ`
```ts
const originalBuild = Menu.buildFromTemplate;
Menu.buildFromTemplate = function (template) {
    if (template[0]?.label === "&File") {
        const { submenu } = template[0];
        if (Array.isArray(submenu)) {
            submenu.push({
                label: "Alt+F4's supress",
                visible: false,
                acceleratorWorksWhenHidden: true,
                accelerator: "Alt+F4",
                click: () => { }
            });
        }
    }
    return originalBuild.call(this, template);
};
```

none of the methods was able to prevent the application from being completely closed by pressing alt+f4

I also don't quite get the statement «the current code reads and parses the settings on every single key stroke» and consider that the settings will be read only if alt and F4 keys are pressed at the same time

PR #1816, opened 2023-10-18
bugfix: fix csp on non-chromium browsers

junepark678:
Although I am aware of official support dropping on Firefox(-based) browsers, this feels necessary for many plugins that relies on external requests to work reliably.


Vendicated:
what is this supposed to do? this just duplicates already existing code

PR #1818, opened 2023-10-18
meta: Add .gitkeep file to mark userplugins folder

nea89o:
This makes it easier for users to find where to put userplugins, as that folder already exists

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,8 @@ lerna-debug.log*
 .pnpm-debug.log*
 *.tsbuildinfo
 
-src/userplugins
+src/userplugins/*
+!src/userplugins/.gitkeep
 
 ExtensionCache/
 settings/


Vendicated:
.gitkeep is weird, i'd just create a .gitignore inside the userplugins folder that excludes everything except itself

nea89o:
.gitkeep is standard for this kind of stuff. i can also change it to use a .gitignore, but i feel like this would be more normal

Vendicated:
> .gitkeep is standard for this kind of stuff
that's not true. .gitignore is the only standardised way to do this. in any case, as discussed on the Discord server, having an empty directory in the source code is meh, so i'll reject this one. thanks anyway!

nea89o:
> > .gitkeep is standard for this kind of stuff
> 
> that's not true. .gitignore is the only standardised way to do this. in any case, as discussed on the Discord server, having an empty directory in the source code is meh, so i'll reject this one. thanks anyway!

Not standardized by git itself, but by the community. .gitkeep files are so common there are literally hundreds of thousands of them in git repos everywhere. even the git source code itself uses a .gitkeep file in some of it's tests. If you still want to use .gitignore we can do that as well, but this is how it is done pretty much everywhere.

I also think that not having the userplugins folder is just massively backwards for user comfort. I get that you as the main dev for vencord probably dont feel much need for user plugins, but for people who *do* want to use userplugins this would make it a lot clearer where to put the files.

PR #1821, opened 2023-10-19
Plugin: BannersEverywhere

lillithrosepup:
# BannersEverywhere

Displays nitro and USRBG banners as a background in the member list

Adds banners to member list


src/plugins/bannersEverywhere/index.css
```diff
@@ -0,0 +1,14 @@
+/* stylelint-disable property-no-vendor-prefix */
+.vc-banners-everywhere-memberlist {
+    opacity: 0.8;
+    -webkit-mask-image: linear-gradient(to right, transparent 20%, #fff);
```
TheKodeToad, line 4: seems like you stopped needing this some time in 2022?

ImLvna: chrome (or atleast the chrome electron uses) doesnt have mask-image without the vendor prefix

TheKodeToad: yeah it's weird but seems to be true.
https://caniuse.com/css-masks
confusing docs :woozy_face:

src/plugins/bannersEverywhere/index.css
```diff
@@ -0,0 +1,14 @@
...
+    mask-image: linear-gradient(to right, transparent 20%, #fff);
+    width: 100%;
+    height: 100%;
+    content: "";
```
Vendicated, line 8: don't think this is needed, no?

ImLvna: it seems so, iirc that was one of the snippets from autumn's pr that i took

src/plugins/bannersEverywhere/index.css
```diff
@@ -0,0 +1,11 @@
+/* stylelint-disable property-no-vendor-prefix */
+.vc-banners-everywhere-memberlist {
+    opacity: 0.8;
+    -webkit-mask-image: linear-gradient(to right, transparent 20%, #fff);
```
Vendicated, line 4: you should also add a non prefix variant. likely necessary on firefox, and possibly latest chromium which removed the prefix
```suggestion
    -webkit-mask-image: linear-gradient(to right, transparent 20%, #fff);
    mask-image: linear-gradient(to right, transparent 20%, #fff);
```

src/plugins/bannersEverywhere/README.md
```diff
@@ -0,0 +1,5 @@
+# BannersEverywhere
+
+Displays nitro and USRBG banners as a background in the member list
+
+![Adds banners to member list](https://github.com/Vendicated/Vencord/assets/44179559/d8a3a2f2-8491-4a4f-b43e-22fdf46622e5)
```
Vendicated, line 5:
```suggestion
![](https://github.com/Vendicated/Vencord/assets/44179559/d8a3a2f2-8491-4a4f-b43e-22fdf46622e5)
```

src/plugins/bannersEverywhere/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { disableStyle, enableStyle } from "@api/Styles";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { RestAPI } from "@webpack/common";
+import { User } from "discord-types/general";
+
+import style from "./index.css?managed";
```
Vendicated, line 14: your style does not need to be managed

src/plugins/bannersEverywhere/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+const settings = definePluginSettings({
+    animate: {
+        description: "Animate banners",
+        type: OptionType.BOOLEAN,
+        default: false
+    },
+});
+const data: { [key: string]: string | null; } = {};
```
Vendicated, line 23:
```suggestion
const data: Record<string, string | null> = {};
```

src/plugins/bannersEverywhere/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        const { banner } = res.body;
+
+        if (!banner) {
+            data[memberId] = null;
+            continue;
+        }
+
+        data[
```
Vendicated, line 76: why keep your own data collection? get it from and subscribe to the UserProfileStore

you also don't seem to be handling guild profiles at all

src/plugins/bannersEverywhere/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        description: "Animate banners",
+        type: OptionType.BOOLEAN,
+        default: false
+    },
+});
+const data: { [key: string]: string | null; } = {};
+
+let enabled = false;
+
+async function refreshVisibleMembers() {
```
Vendicated, line 27: this entire function is incredibly awful. you should patch the component of a member row and `useStateFromStores([UserProfileStore], ...)` in there

Vendicated (changes requested):

Nonstopgaming75:
pls fix it

src/plugins/bannersEverywhere/index.tsx
ImLvna: im sorry, i dont understand? this uses dom to see what userlist elements are in the viewport and fetches them, i dont think thats possible from just react side

Vendicated: i was suggesting that instead of doing this, you just patch the memberlist row component to insert your own hook call that subscribes to the user profile store and fetches profile if missing

this will then automatically rerender the component when the user profile has eventually been fetched, and Discord's lazy scroller already makes it so that only visible users will be rendered

like
```js
"$self.useFetchMemberProfile($1)"

useFetchMemberProfile(userId: string) {
	const profile = useStateFromStore([UserProfileStore], () => UserProfileStore.getProfile(userId));
	
	useEffect(() => {
		let cancel = false;

		queue.push(() => {
			if (cancel) return;
			return fetchMember(userId).finally(() => sleep(1000));
		})

		return () => cancel = true;
	}, []);
}
```

Vendicated: hope that makes sense!

src/plugins/bannersEverywhere/index.tsx
ImLvna: > you also don't seem to be handling guild profiles at all, so probably should try to add that

i don't see a utility function to fetch member profiles, only users. how would i go about that?

Nonstopgaming75:
It works but I saw you removed Userbg support

ImLvna:
added back usrbg support! only fetches the user profile if usrbg.getImageUrl returns null

ImLvna:
not using usrbg banners if the user does not want animated banners, as there is no way of knowing if a usrbg banner is animated or not

ImLvna:
confirmed working after doomsday!
image

sadan4:
how would this work with nameplates

Issue #1417, opened 2023-07-14
[Bug] Userscript not working on Safari


itz-coffee:
### Discord Account

itz_coffee

### What happens when the bug or crash occurs?

The userscript isn't working with either Tampermonkey or Userscripts on safari

### What is the expected behaviour?

Vencord should load on the web page

Safari: Version 17.0 (19616.1.20.11.3)
Discord: Stable 212637 (6da172b)
Mac: Sonoma 14.0

### How do you recreate this bug or crash?

1. Download Tampermonkey or Userscripts from the app store
2. Add vencord userscript
3. Open the discord web app

### Errors

Tampermonkey
```
Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy.
```
Userscripts
```
[Warning] Vencord.user.js @inject-into value set to 'content' due to @grant values: GM_xmlhttpRequest - https://github.com/quoid/userscripts/issues/265#issuecomment-1213462394 (content.js, line 384)
[Info] Injecting Vencord (js) (content.js, line 278)
[Error] Vencord.user.js error
ReferenceError: Can't find variable: unsafeWindow
(anonymous function)
(anonymous function)
(anonymous function)
(anonymous function)
(anonymous function)
(anonymous function)
(anonymous function)
anonymous
injectJS
processJS
(anonymous function)
	injectJS (content.js:291)
	processJS (content.js:229)
	(anonymous function) (content.js:419)
```

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well

ilikecereal1:
I tried it on my iPad and it doesn’t work.

Vendicated:
safari is not supported, use chromium/firefox

itz-coffee:
You should update the script then as it currently says its compatible 
https://github.com/Vencord/builds/tree/main/Vencord.user.js#L16

joulev:
Bumping this, it doesn't work on Safari 17.3.1 and Tampermonkey v4.20.6188. The error is the same as above.

> [Error] Refused to execute a script because its hash or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy. (about:blank, line 1)
> [Error] Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy. (about:blank, line 0)
[...]

PR #1823, opened 2023-10-19
Remove safari compatibility tag

itz-coffee:
Removing it out of the userscript because of https://github.com/Vendicated/Vencord/issues/1417#issuecomment-1646371117

### DIFF
--- a/browser/userscript.meta.js
+++ b/browser/userscript.meta.js
@@ -13,7 +13,6 @@
 // @compatible      firefox Firefox Tampermonkey
 // @compatible      opera Opera + Tampermonkey or Violentmonkey
 // @compatible      edge Edge + Tampermonkey or Violentmonkey
-// @compatible      safari Safari + Tampermonkey or Violentmonkey
 // ==/UserScript==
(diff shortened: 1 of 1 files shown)


Vendicated:
no
grafik

itz-coffee:
I don't know what that is a screenshot of but I am on safari 17.0 (like your picture has it green as) and it doesn't work, you said it yourself that it doesn't work on safari so why keep it?

Vendicated:
it should work on more recent safari versions. it only doesn't work on the old versions

itz-coffee:
Still not working for me but I'll just move to vencord desktop

Trying to get it to work with "userscripts" (safari clone of tampermonkey) but won't work, I did buy tampermonkey couple months ago but refunded it since same issue. Didn't know it worked for others so I won't keep bugging

Vendicated:
it probably works with tampermonkey or violentmonkey if those exist for safari. i don't own a mac, and never used safari in my life so i can't test it but there's no reason it shouldn't

PR #1825, opened 2023-10-20
feat(plugin): Invidious

twnlink:
This plugin makes YouTube embeds use Invidious instead of the default YouTube embed.


twnlink:
nevermind im retarded

PR #1831, opened 2023-10-22
[RelationshipNotifier] add notifications when friend is added/accepts your request

nin0-dev:
This pr improves the relationshipnotifier plugin by adding notifications for accepted friend requests or when someone uses your friend invite.
This feature can be helpful as discord web/desktop does not have notification for friend request acceptances/friend invite clicked..

However, enabling "Friends Add" and the "Offline Removals/Adds" settings will trigger a bunch of notifications "___ added you as a friend" due to the plugin syncing with the friend list. That's why the Friends Add option is toggled off by default


nin0-dev:
> However, enabling "Friends Add" and the "Offline Removals/Adds" settings will trigger a bunch of notifications "___ added you as a friend" due to the plugin syncing with the friend list. That's why the Friends Add option is toggled off by default
seems like this is pretty hit or miss, but it only happened once, so marked it as ready for review. we can always improve it later on

PR #1848, opened 2023-10-25
fix: sortFriendRequests patches

MeguminSama:
Fixes the patches for sortFriendRequests

### DIFF
--- a/src/plugins/sortFriendRequests/index.tsx
+++ b/src/plugins/sortFriendRequests/index.tsx
@@ -31,12 +31,8 @@ export default definePlugin({
     patches: [{
         find: "getRelationshipCounts(){",
         replacement: {
-            match: /\.sortBy\((\i)=>\i\.comparator\)/,
-            replace: (_, row) => `.sortBy((${row}) => {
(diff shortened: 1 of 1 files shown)


src/plugins/sortFriendRequests/index.tsx
```diff
@@ -29,24 +29,24 @@ export default definePlugin({
...
-        replacement: [{
-            match: /\.sortBy\(\(function\((\w)\){return \w{1,3}\.comparator}\)\)/,
-            // If the row type is 3 or 4 (pendinng incoming or outgoing), sort by date of receipt
-            // Otherwise, use the default comparator
-            replace: (_, row) => `.sortBy((function(${row}) {
+        find: "getRelationshipCounts(){",
+        replacement: {
+            match: /\.sortBy\((\i)=>\i\.comparator\)/,
+            replace: (_, row) => `.sortBy((${row}) => {
                 return ${row}.type === 3 || ${row}.type === 4
```
Vendicated, line 36: i am not a big fan of having so much code in the patch. stuff like that is always best made a method on ur plugin (yes this used to be the same way before but might as well fix it while we're at it

PR #1858, opened 2023-10-25
forceOwnerCrown: fix

D3SOX:
Also resolves #1830 as it returns the original `isOwner` by default now

### DIFF
--- a/src/plugins/forceOwnerCrown/index.ts
+++ b/src/plugins/forceOwnerCrown/index.ts
@@ -27,23 +27,24 @@ export default definePlugin({
     authors: [Devs.D3SOX, Devs.Nickyux],
     patches: [
         {
-            // This is the logic where it decides whether to render the owner crown or not
-            find: ".MULTIPLE_AVATAR",
+            find: "AVATAR_DECORATION_PADDING:",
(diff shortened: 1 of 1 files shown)


Vendicated:
use \i instead of \w+, it matches js identifiers

D3SOX:
done, still used to using `\w+` because that's how I did it when I wrote my first plugin ^^

src/plugins/forceOwnerCrown/index.ts
```diff
@@ -27,23 +27,24 @@ export default definePlugin({
     authors: [Devs.D3SOX, Devs.Nickyux],
     patches: [
         {
-            // This is the logic where it decides whether to render the owner crown or not
-            find: ".MULTIPLE_AVATAR",
+            find: "AVATAR_DECORATION_PADDING:",
             replacement: {
-                match: /(\i)=(\i)\.isOwner,/,
-                replace: "$1=$self.isGuildOwner($2),"
+                match: /var (\i)=function\((\i)\){/g,
```
Vendicated, line 32: this does not seem very robust. adding some more unique anchor like the previous isOwner would be much better

PR #1863, opened 2023-10-25
fix: NoticesAPI

redstonekasi:


src/plugins/_api/notices.ts
```diff
@@ -29,12 +29,12 @@ export default definePlugin({
             find: 'displayName="NoticeStore"',
             replacement: [
                 {
-                    match: /(?=;\i=null;.{0,70}getPremiumSubscription)/g,
-                    replace: ";if(Vencord.Api.Notices.currentNotice)return false"
+                    match: /(?<=return!1;)\i=null;/g,
```
Vendicated, line 32: that looks not very robust, adding some more unique anchor like the old getPremiumSubscription would be a lot better

PR #1894, opened 2023-10-25
fix: ImageZoom + PiP

lumap:
- ImageZoom: removed CSS that caused media mosaic to warp all the way to the top
- PiP: fixed a bug that replaced the remove attachment button with the PiP button (even on non-video attachments)

### DIFF
--- a/src/plugins/imageZoom/styles.css
+++ b/src/plugins/imageZoom/styles.css
@@ -25,12 +25,6 @@
     box-shadow: none;
 }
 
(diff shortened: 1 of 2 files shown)


Docn1ght:
sorry im new here, is there any way to force this to apply to my imagezoom plugin? just wondering since the mosaic thing is annoying

Docn1ght:
awesome, thanks. all working by the way!

PR #1917, opened 2023-10-27
feat(plugin): CustomAppIcons

nakoyasha:
Adds a plugin that lets you add your own icons to the in-app icon list.
<p align="center">

</p>

Non-nitro users have to use FakeNitro alongside this plugin as the list is completely disabled for them; re-enabling it here is pointless and out of the scope for this PR anyway.
(thanks @RealSimplyData for fixing my garbage code)


camila314:
good plugin idk why they are taking so long to approve it

PR #1921, opened 2023-10-28
feat: anonymous telemetry

lewisakura:
Depends on https://github.com/Vencord/Vencloud/pull/20

Collecting information about:
- Vencord versions in use
- Plugins in use
- Operating systems in use

This will not be traced back to a user.

**ONLY MERGE IF COMMUNITY APPROVAL IS GAINED. HEAVILY WORK IN PROGRESS.**

### DIFF
--- a/src/Vencord.ts
+++ b/src/Vencord.ts
@@ -34,6 +34,7 @@ import { patches, PMLogger, startAllPlugins } from "./plugins";
 import { localStorage } from "./utils/localStorage";
 import { relaunch } from "./utils/native";
 import { getCloudSettings, putCloudSettings } from "./utils/settingsSync";
+import { sendTelemetry } from "./utils/telemetry";
 import { checkForUpdates, update, UpdateLogger } from "./utils/updater";
 import { onceReady } from "./webpack";
 import { SettingsRouter } from "./webpack/common";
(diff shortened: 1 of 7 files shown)


FieryFlames:
🤓☝️ you need to remove "no telemetry" from the readme

Zippybonzo (approved):
My quick lookthrough shows it's pretty good. - I would say don't send it on devbuilds so uncomment that but it seems good beyond that.

lewisakura:
> My quick lookthrough shows it's pretty good. - I would say don't send it on devbuilds so uncomment that but it seems good beyond that.

We aren't going to send on devbuilds, it's to be added back before this gets merged (if it does) but it needs to be commented during testing.

thororen1234 (approved):
looks cool and could be used for debugging n stuff even tho the user will complain but it could help listing errors as in if a certain error keeps popping up its high priority ect.

IThundxr:
Telemetry 👍 👍

IThundxr (approved):
Yummy

taskylizard (approved):

lewisakura:
No motivation to continue doing this from our side at the minute, so it's on hold until further notice.

PR #1930, opened 2023-10-30
feat(plugin): FakeProfileThemesAndEffects

ryan-0324:
This plugin is meant to replace FakeProfileThemes.
It allows profile theming and the usage of profile effects by hiding the colors and effect ID in your About Me using invisible, zero-width characters.
- It includes a UI (called the FPTE Builder) for generating the profile theme/effect (FPTE) strings in the User Profile and Server Profiles settings pages.
- The FPTE strings can be made backwards compatible so that users of older plugins will still see your profile theme.
- When the "Build backwards compatible FPTE" setting is off, the generated strings will use at most 22 characters (12 codepoints) of your About Me.
- The FPTE Builder can be hidden through the plugin settings.


src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+        description: "Hide the 3y3 Builder in the profiles settings page",
+        type: OptionType.BOOLEAN,
+        default: false
+    }
+});
+
+export default definePlugin({
+    name: "FakeProfileThemesAndEffects",
+    description: "Allows profile theming and the usage of profile effects by hiding the colors and effect ID in your About Me using invisible 3y3 encoded characters",
+    authors: [
```
MrDiamondDog, line 209: Add yourself to `Devs`, see other plugins

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+const settings = definePluginSettings({
+    prioritizeNitro: {
+        description: "Source to use if profile theme colors / effects are set by both Nitro and About Me",
+        type: OptionType.SELECT,
+        options: [
+            { label: "Nitro", value: true },
+            { label: "About Me", value: false, default: true },
+        ]
+    },
+    "Hide 3y3 Builder": {
```
MrDiamondDog, line 199:
```suggestion
    hide3y3Builder: {
```
and change things accessing this

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+                        <li>{'Click the "Copy 3y3" button'}</li>
+                        <li>{"Paste the invisible text anywhere in your About Me"}</li>
+                    </ol>
+                </Forms.FormText>
+            </Forms.FormSection>
+        );
+    },
+    settings,
+    decodeUserBio3y3Hook(user: UserProfile | undefined): UserProfile | undefined {
+        if (user !== undefined) {
```
MrDiamondDog, line 255:
```suggestion
        if (user === undefined) return;
```
Do this for a bunch of the other `if` statements, prevents nesting as much and improves readability

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+                updateUserThemeColors(user, primaryColor, base125StrToBase10CSSColor(decoded3y3[1]));
+                updateUserEffectID(user, base125StrToBase10ItemID(decoded3y3[2]));
+            }
+        }
+        return user;
+    },
+    add3y3Builder(): JSX.Element {
+        if (settings.store["Hide 3y3 Builder"] === true) return <></>;
+
+        const [primaryColor, setPrimaryColor]: [number, (v: number) => void] = useState(-1);
```
MrDiamondDog, line 291: You don't need to type these, as they will get their types from `useState`

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+                    <Button
+                        innerClassName={this.name + "TextOverflow"}
+                        color={Button.Colors.PRIMARY}
+                        style={{
+                            borderTopLeftRadius: "0",
+                            borderBottomLeftRadius: "0",
+                            paddingLeft: "0",
+                            paddingRight: "0"
+                        }}
+                        onClick={(): void => {
```
MrDiamondDog, line 425: This whole function is completely unreadable, move to a different function and use the `if` statement strategy from before, space out some blocks, and maybe some comments

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+                            let stringToCopy: string = "";
+                            if (buildLegacy3y3 === true) {
```
MrDiamondDog, line 427:
```suggestion
                            if (buildLegacy3y3) {
```

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+                    size={Button.Sizes.TINY}
+                    style={{
+                        display: primaryColor === -1 && accentColor === -1 && effectID === "" ? "none" : "inline",
+                        height: "14px",
+                        marginLeft: "4px",
+                        minHeight: "0",
+                        paddingBottom: "0",
+                        paddingTop: "0"
+                    }}
+                    onClick={(): void => {
```
MrDiamondDog, line 318: Not everything needs types

```suggestion
                    onClick={() => {
```

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+import { Margins } from "@utils/margins";
+import { copyWithToast } from "@utils/misc";
+import { Button, Forms, showToast, Slider, Switch, Text, Toasts, useState } from "@webpack/common";
+import { openColorPickerModal } from "./components/ColorPickerModal";
+import { openProfileEffectModal } from "./components/ProfileEffectModal";
+import { User } from "discord-types/general";
+
+interface UserProfile extends User {
+    themeColors: [number, number] | undefined;
+    profileEffectID: string | undefined;
```
MrDiamondDog, line 30:
```suggestion
    profileEffectID?: string;
```

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,496 @@
...
+}
+
+function base10NumToBase125Str(base10: number): string {
```
MrDiamondDog, line 33: Add some jsdocs to these functions

MrDiamondDog (changes requested):

src/plugins/fakeProfileThemesAndEffects/index.tsx
ryan-0324: This must be done otherwise the setting name will show as "Hide3y3 Builder" in the plugin settings modal.

src/plugins/fakeProfileThemesAndEffects/index.tsx
ryan-0324: `profileEffectID?: string` is not the same as `profileEffectID: string | undefined`. `string | undefined` requires the property `profileEffectID` to exist and have a type of string or undefined, whereas the optional marker does not require `profileEffectID` to exist at all.

MrDiamondDog: It's a lot easier to let `profileEffectID` to not exist rather than just making it undefined

ryan-0324: I extended the type `User` from `discord-types/general` as it is not up to date. The user object returned by the `getUserProfile` function always contains `profileEffectID` even if it is undefined. That is why `string | undefined` is correct and the optional marker is not.

MrDiamondDog: Makes sense then

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,567 @@
...
+                >
+                    {"Build backwards compatible 3y3"}
+                </Switch>
+            </>
+        );
+    },
+    start(): void {
+        document.documentElement.appendChild(Object.assign(
+            document.createElement("style"),
+            { textContent: "." + this.name + "TextOverflow{white-space:normal!important;text-overflow:clip!important}" }
```
MrDiamondDog, line 564: You shouldn't have to do this, inline styles exist

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,567 @@
...
+            updateUserEffectID(user, base125StrToBase10ItemID(decoded3y3[1]));
+        } else {
+            updateUserThemeColors(user, primaryColor, base125StrToBase10Color(decoded3y3[1]));
+            updateUserEffectID(user, base125StrToBase10ItemID(decoded3y3[2]));
+        }
+
+        return user;
+    },
+    add3y3Builder(): JSX.Element {
+        if (settings.store["Hide 3y3 Builder"] === true) return <></>;
```
MrDiamondDog, line 419:
```suggestion
        if (settings.store["Hide 3y3 Builder"]) return <></>;
```

MrDiamondDog (changes requested):

src/plugins/fakeProfileThemesAndEffects/index.tsx
ryan-0324: I tried using inline styles at first. The problem is that the `Button` component from `@webpack/common` has its text contained in a separate `div` element. The only way I was able to style the text was by using the `innerClassName` property.

MrDiamondDog: Include the `<style>` tag inside that component then, don't insert it into the dom like this

src/plugins/fakeProfileThemesAndEffects/index.tsx
```diff
@@ -0,0 +1,664 @@
...
+        callback(suggestedColors);
+    }).catch(e => {
+        console.error(e);
+        showToast("Unable to retrieve suggested colors.", Toasts.Type.FAILURE);
+        callback([]);
+    });
+}
+
+function fetchProfileEffects(callback: (v: any) => void): void {
+    fetch("/api/v9/user-profile-effects", { mode: "same-origin", cache: "only-if-cached" })
```
Vendicated, line 310: this is terrible. use Discord's RestAPI module

Vendicated:
this is a crazy amount of code for such a small feature. why is all of this needed? it seems very overengineered

ImpishMoxxie:
LGTM 👍

aaronateataco2 (changes requested):
Shorten the plugin name as it goes off screen in the Plugin Browser:
image
Suggestions: fakeET 
(Effects and Themes)

tiagorangel2011:
Looks good

ImpishMoxxie (approved):

MrDiamondDog (approved):
Seems over engineered as ven said but not sure how you can improve it

beatmakerofficial:
Where do I download it?

ryan-0324:
@rniii Do you know what could be causing the `Invalid header` errors? All of my headers were automatically added on save.

omegion3:
get that thing **declined,** because this is _reupload from discord assets._

now if only we get custom profile effects, then just a reupload one, this would've be good.

MrDiamondDog:
> get that thing **declined,** because this is _reupload from discord assets._
> 
> 
> 
> now if only we get custom profile effects, rather then just a reupload one, this would've be prefect.

What are you talking about

MrDiamondDog:
> nah i'm just kidding, even though it's a reupload assets, but i hope this plugin will merge 

It's not reuploaded assets..?

omegion3:
> It doesn't have reuploaded assets..?
well it does sadly, but hopefully someone might plan to make custom profile effect or this plugin will finally merge.

CompeyDev:
What's the status on this? Anything blocking a merge on this?

Sqaaakoi:
> What's the status on this? Anything blocking a merge on this?

vee gets around to PRs "*eventually*"
this is also (relatively) a big plugin

CompeyDev:
Makes sense, I'm running a manually compiled build from ryan's repo and it seems to work about fine. Haven't really looked much at the code though.

ryan-0324:
> Some issues:
> 
> 1. The stock profile effect editor is patched by this plugin resulting in it displaying all effects as unlocked and are unable to apply any & it falsely showing the remove effect button. Should behave like Decor where it doesn't modify the stock feature.
> 2. The FPTE Builder should auto fill with the values from the bio.
> 3. The FPTE Builder should have buttons to save to bio without manual copy & pasting being required And likewise have a clear button.

1. This seems to be a recent breakage. Thanks for letting me know.
2. I've though about adding something like this before, and it seems reasonable.
3. This has been suggested before, and I don't think it's worth the added complexity:
    1. Should I just prepend the FPTE string to the user's bio, displaying an error if their bio doesn't have enough space or already starts with an FPTE character?
    2. Or should I try to automatically detect preexisting FPTE strings and strings from other plugins, and remove them from the user's bio, showing an error if there still isn't enough space?
    3. If not (ii), when the user's bio already starts with an FPTE character, should I add a non-FPTE character as a separator, or should I display an error?
    4. Should I send an API request to save the user's bio, or should I have the user do it manually?

alyavoidx:
can other can see it

Vendicated:
This is really overengineered. Also honestly not something I want to add, profile effects are so useless and annoying

Also if we add profile effects, next it's gonna be animated pfps, profile frames, nameplates, etc. There are so many things that could be spoofed but I think theme is already more than enough

PR #1936, opened 2023-10-31
textReplace: uh update ig

AutumnVN:
Added:
- bigger modal
- merge both type of rule into one with regex toggle
- convert old data to this new one
- 2000 char max length
- realtime update test rule as typing no need to unfocus
- create new rule button
- import rules with merge & dedupe
- export rules
- rule toggle
- preview aside after each rule
- move rule up/down
- onlyIfIncludes use regex test if it's regex rule

Removed:
- auto delete empty row
- auto create last row if not empty

image


Vendicated:
so much code in one single file. it'd likely help readability and maintanibility to split it across multiple files

src/components/InviteLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Link } from "@components/Link";
+import { findByPropsLazy } from "@webpack";
+import { FluxDispatcher, showToast } from "@webpack/common";
+
+interface Props extends React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement> {
+    target: string;
+    disabled?: boolean;
+}
+
+const InviteActions = findByPropsLazy("resolveInvite");
```
Vendicated, line 28: maybe could be wp commons candidate? i dont remember if its used in more places, but at the very least Vesktop uses it as well

FieryFlames: Decor uses this as well along with a utility function to open invite modal

src/components/InviteLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export function InviteLink(props: React.PropsWithChildren<Props>) {
+    if (props.disabled) {
+        props.style ??= {};
+        props.style.pointerEvents = "none";
+        props["aria-disabled"] = true;
```
Vendicated, line 34: you don't actually seem to be using these props unless i'm missing something? so this probably does nothing

you should pass the disabled prop to <Link> (and maybe do this with css instead)

src/components/InviteLink.tsx
Vendicated: The themes page also uses it for invites (thanks fiery)

Vendicated:
i'll look at the main textReplace code once it's split across files (or not) as suggested above

src/components/PluginSettings/PluginModal.tsx
```diff
@@ -182,7 +182,7 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
     }
 
     return (
-        <ModalRoot transitionState={transitionState} size={ModalSize.MEDIUM} className="vc-text-selectable">
+        <ModalRoot transitionState={transitionState} size={plugin.modalSize || ModalSize.MEDIUM} className="vc-text-selectable">
```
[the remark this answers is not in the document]
Vendicated: also, just in case Discord ever changes this to a numerical enum, you don't want to override enum value 0
```suggestion
        <ModalRoot transitionState={transitionState} size={plugin.modalSize ?? ModalSize.MEDIUM} className="vc-text-selectable">
```

src/components/InviteLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Link } from "@components/Link";
+import { openInviteModal } from "@utils/discord";
+
+interface Props extends React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement> {
```
Vendicated, line 22: you don't seem to be using any other props and even disabled isnt used. either fix the types or use them
```suggestion
interface Props {
```

src/utils/discord.tsx
```diff
@@ -18,14 +18,28 @@
...
 import { ImageModal, ModalRoot, ModalSize, openModal } from "./modal";
 
 const MessageActions = findByPropsLazy("editMessage", "sendMessage");
 const UserProfileActions = findByPropsLazy("openUserProfileModal", "closeUserProfileModal");
 
+export const InviteActions = findByPropsLazy("resolveInvite");
+
+export async function openInviteModal(code: string) {
+    const { invite } = await InviteActions.resolveInvite(code, "Desktop Modal");
+    if (!invite) return showToast("Invalid or expired invite", Toasts.Type.FAILURE);
```
Vendicated, line 33: to make this func more easily reusable, it'd be better to not have this toast here, and instead either don't handle the error, or return true/false to indicate success

FieryFlames: throw a error if it's invalid so it can be caught

PR #1937, opened 2023-10-31
Custom Snippets

Stefanuk12:
The purpose of this plugin is to stop people from repeating themselves. It's quite similar to a feature that Dyno has with their autoresponder but works on your client instead.

There are improvements that could be made that are outside of my reach but doable. The improvements are noted within the comments at the top.


Stefanuk12:
I realise that this plugin is quite similar to TextReplace, however custom snippets could be more flexible - if expanded upon with the suggested improvements.

Vendicated:
This already exists as MessageTags

PR #1938, opened 2023-10-31
EmoteCloner: make the error toasts useful

nakoyasha:
image
yea

### DIFF
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -143,6 +143,26 @@ async function fetchBlob(url: string) {
     return res.blob();
 }
 
+function tryParseJSON(data: string) {
+    let parsingError = undefined;
+
+    try {
+        var out = JSON.parse(data);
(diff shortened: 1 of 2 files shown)


src/plugins/emoteCloner/index.tsx
```diff
@@ -155,10 +155,14 @@ async function doClone(guildId: string, data: Sticker | Emoji) {
             type: Toasts.Type.SUCCESS,
             id: Toasts.genId()
         });
-    } catch (e) {
+    } catch (e: any) {
+        const exceptionText = e.text as string;
+        const jsonDecoded = JSON.parse(exceptionText);
+        const errorMessage = jsonDecoded.message;
```
Vendicated, line 161: this seems very not robust. what if the error is very different? you should add null safety / error handling and a sensible default message

nakoyasha: fixed with 3b4a5a9

Vendicated:
thanks!

PR #1940, opened 2023-11-01
feat(plugin): UserPFP

thororen1234:
This plugin is usrbg but for profile pictures in simple terms.
It allows for non-nitro users to have a gif and such as their profile picture.


Vendicated:
all this plugin does is import a css file. you should just turn it into a theme / post it in the snippets channel

thororen1234:
fair

PR #1941, opened 2023-11-01
feat: add dropdown to choose vibrancy value on macOS

eramdam:
Preview: 
image

I was working on a custom theme that [takes advantage of the vibrancy option](https://github.com/eramdam/slacker-discord) when I noticed the vibrancy option was "only" set to `sidebar` which isn't as see-through as `fullscreen-ui` so this PR adds a dropdown to choose a specific [vibrancy type](https://www.electronjs.org/docs/latest/api/browser-window#winsetvibrancytype-macos) value. Those values in turn map to native [macOS materials](https://developer.apple.com/documentation/appkit/nsvisualeffectview/material)

I tried to order options from the most opaque to the most transparent. I'm not a fan of the item descriptions but I don't know if there is a better way to name them 🤷 Also not all values are there because a few of them have the same effect as others so I removed "duplicates".

Here's a preview of what each option does:

| Material     | Light                                                                                                             | Dark                                                                                                             |
| ------------ | ----------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
| Under Page   | A_underpage-light   | A_underpage-dark   |
| Content      | B_content-light     | B_content-dark     |
| Window       | C_window-light      | C_window-dark      |
| Selection    | Q_selection-light   | Q_selection-dark   |
| Titlebar     | R_titlebar-light    | R_titlebar-dark    |
| Header       | S_header-light      | S_header-dark      |
| Sidebar      | T_sidebar-light     | T_sidebar-dark     |
| Under Window | U_underwindow-light | U_underwindow-dark |
| Tooltip      | V_tooltip-light     | V_tooltip-dark     |
| Menu         | W_menu-light        | W_menu-dark        |
| Popover      | X_popover-light     | X_popover-dark     |
| Fullscreen   | Y_fullscreen-light  | Y_fullscreen-dark  |
| HUD          | Z_hud-light         | Z_hud-dark         |


src/main/patcher.ts
```diff
@@ -85,9 +85,11 @@ if (!IS_VANILLA) {
                     options.backgroundColor = "#00000000";
                 }
 
-                if (settings.macosTranslucency && process.platform === "darwin") {
+                const needsVibrancy = process.platform === "darwin" || (settings.macOsVibrancyStyle || settings.macosTranslucency);
```
ryanccn, line 88: Maybe merge the two settings into one and migrate the old option to `"sidebar"`

ryanccn (changes requested):

src/main/patcher.ts
eramdam: I'm not sure I get what you mean 😅 Do you meant change the type of `macosTranslucency`? Wouldn't that require a migration to support the old settings properly? (I couldn't find where to do that)

ryanccn: Keep the old setting in the types but don't make it editable by the user, and migrate the old option to the new option if the old option is set and the new option isn't

eramdam: Done! https://github.com/Vendicated/Vencord/pull/1941/commits/81dd0fd11e73bb4b46bc4336e8a76cf05473a582

eramdam:
@ryanccn any chance of getting a re-review on this one? 😅

PR #1945, opened 2023-11-02
Ipc Plugins

ImLvna:
Allows plugins to inject into discord frames. This is done by creating ipcPlugin.ts inside the plugin folder. This must have the same name as the plugin it belongs to. The plugin will be injected into frames specified by the `matcher` param. I have migrated fixSpotifyEmbeds, and included #1742 as a proof of concept

### DIFF
--- a/scripts/build/build.mjs
+++ b/scripts/build/build.mjs
@@ -1,25 +1,22 @@
 #!/usr/bin/node
 /*
- * Vencord, a modification for Discord's desktop app
- * Copyright (c) 2022 Vendicated and contributors
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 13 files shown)


AAGaming00:
ipcPlugin is kind of a misleading name for this IMO, it doesn't seem to inject into the ipc (what?) and doesn't inject into main, but into subframes

(at least if I'm understanding correctly)

Vendicated:
i find this too oddly specific

i think it would be much more logical to instead allow plugins to have a native.ts file. this file would be bundled into the nodejs code. any exports of this file would automatically be turned into ipc methods. we could then export some apis from src/main that you would be able to export from there for doing things like this

### Example:

native.ts
```ts
export function doThing(foo: string, bar: number) {}
```
now we would make the bundler import this file and iterate over its exports to register this method as ipc handler for `PluginName_MethodName`, and generate an object like this:
```js
const pluginIpc = {
     MyPlugin: ["doThing"]
}
```
this object will then be requested from preload to auto generate bindings on VencordNative.ts

i hope this makes sense

Vendicated:
superseded by https://github.com/Vendicated/Vencord/pull/1965 - thanks a lot aynway for your effort!

#1965 doesn't actually have a convenient way to inject into frames, but rather just the ability for any plugin to have native code

However, it would be fairly trivial to export a basic frame patching api from ipcMain, like
```js
addFrameCallback("MyPlugin", frame => {
	if (frame.url.startsWith("https://open.spotify.com/embed")) {
		frame.executeJavascript("...")
	}
})
```

PR #1959, opened 2023-11-04
fix(dearrow): don't replace thumbnail if only original available

NurMarvin:
Currently the plugin tries to replace thumbnails even when only the original thumbnail is available which fails because the `timestamp` for original thumbnails is `null`. This PR simply adds a check if the timestamp is a non-null value.

Example video where this issue currently occurs for me: https://www.youtube.com/watch?v=I6xwMIUPHss

image


Vendicated:
thanks!

PR #1963, opened 2023-11-06
Add NoSuperReactions plugin

throwaway-123456:
This plugin treats super reactions as regular reactions by immediately normalizing reaction objects upon receiving applicable store events. I've tested it, and it seems to work just fine. Thanks.


camila314:
perhaps the code could be spaced out more and formatted better but other than that its good

rniii:
I really don't think you can just PR from a throwaway account and make ven the plugin author

PR #1972, opened 2023-11-10
MessageClickActions: Ignore Ephemeral & Deleted Messages

nmsturcke:
Prevent replying to/editing ephemeral/deleted messages when double clicking.


src/plugins/messageClickActions/index.ts
```diff
@@ -81,6 +81,9 @@ export default definePlugin({
                 } else {
                     if (!settings.store.enableDoubleClickToReply) return;
 
+                    const EPHEMERAL = 64;
+                    if ((msg.flags & EPHEMERAL) === EPHEMERAL) return;
```
Vendicated, line 85:
```suggestion
                    if (msg.hasFlag(EPHEMERAL) return;
```

src/plugins/messageClickActions/index.ts
```diff
@@ -72,6 +72,7 @@ export default definePlugin({
                 if (event.detail < 2) return;
                 if (settings.store.requireModifier && !event.ctrlKey && !event.shiftKey) return;
                 if (channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_MESSAGES, channel)) return;
+                if (msg.deleted === true) return;
```
Vendicated, line 75: no. we deliberately added MessageLogger support, where message click actions will clear deleted message history

nmsturcke: Hmm... When double clicking a deleted message it prompts me to edit it:

oMsFJ

PR #1973, opened 2023-11-11
feat(plugin): NoReplyPings

MrDiamondDog:
Doesn't mention you when someone who left mentions on replies to you

### DIFF
--- a/src/plugins/noReplyPings/index.ts
+++ b/src/plugins/noReplyPings/index.ts
@@ -7,6 +7,16 @@
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
 import { MessageStore, UserStore } from "@webpack/common";
+import { Message } from "discord-types/general";
+
+// for some reason discord has two different types for messages, isn't that just amazing design
+type ReceivedMessage = {
+    message_reference: {
+        guild_id?: string;
+        channel_id: string;
+        message_id: string;
+    } | undefined;
+} & Omit<Message, "messageReference">;
 
 export default definePlugin({
     name: "NoReplyPings",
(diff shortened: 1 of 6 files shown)


src/plugins/noReplyPings/index.ts
```diff
@@ -0,0 +1,43 @@
...
+export default definePlugin({
+    name: "NoReplyPings",
+    description: "Disables getting pinged by replies, even if the sender didn't turn it off",
+    authors: [Devs.MrDiamond],
+    patches: [
+        {
+            find: "_channelMessages",
+            replacement: {
+                match: /receiveMessage\((\i)\)\{/,
+                replace: "$&if($self.isReplyMention($1))$1.mentions=[];"
```
Vendicated, line 30: you should not remove all mentions. you should only remove your own mention

src/plugins/noReplyPings/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { MessageStore, UserStore } from "@webpack/common";
+import { Message } from "discord-types/general";
+
+// for some reason discord has two different types for messages, isn't that just amazing design
```
Vendicated, line 12: - api message (raw json from gateway)
- local message (serialised & proper class)

MrDiamondDog: Changed to `MessageJSON` type which is the api message

ant0n-0x0000:
This has been updated in PR #1975

PR #1974, opened 2023-11-11
ci: automatically enforce contributor requirement on issues

lewisakura:
haven't tested yet


Vendicated:
i don't think this is a good idea, since some issues might still be okay

or sometimes i ask people to open an issue when i confirmed their issue and want to track it

PR #1976, opened 2023-11-11
 [Plugin] LargeProfileAvatars

BoardTM:
Makes avatars take up the full width of the profile modal and renders them in higher resolution.
before
after


src/plugins/largeProfileAvatars/index.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and Board
```
ant0n-0x0000, line 3: I think this needs to include contributors.

```suggestion
 * Copyright (c) 2023 Vendicated, Board, and contributors
```

ant0n-0x0000 (commented):
Could be a good idea to add fallbacks incase discord update UI but that's up to you.

ant0n-0x0000 (changes requested):
Oh also, add a README.md for your plugin!

MrDiamondDog:
This can be done with css

BoardTM:
> This can be done with css

Correct me if I’m wrong, but css will not be able to change the resolution of the avatar which is why this is a plugin and not just a theme.

MrDiamondDog:
I don't think you can change the resolution but most pfps are high enough resolution already, either way this doesn't feel like it should be a whole plugin

BoardTM:
Discord only grabs the picture in 80x80px by default this plugin makes it user selectable.

BoardTM:
Turns out the patch is kinda flawed so I’m closing this so I can rework it.

BoardTM:
Completely rewrote the patch such that it no longer touches chat avatars

Vendicated:
honestly the way this looks is kinda ugly. might be better to have it in larger size here instead and push the rest down?
image

alternatively, and especially considering this plugin mostly consists of css, you could instead just define a larger pfp link as a data attribute on the pfp node and leave the rest of the work to themes / quick css? this way a theme could do something like the following and then adjust the ui accordingly
```css
.avatar {
	content: attr(large-avatar-url);
}
```
this could be part of the ThemeAttributes plugin

let me know your thoughts! (and sorry for taking so long to get to this)

BoardTM:
> alternatively, and especially considering this plugin mostly consists of css, you could instead just define a larger pfp link as a data attribute on the pfp node and leave the rest of the work to themes / quick css? this way a theme could do something like the following and then adjust the ui accordingly
> 
> ```css
> .avatar {
> 	content: attr(large-avatar-url);
> }
> ```
> 
> this could be part of the ThemeAttributes plugin
> 
> let me know your thoughts! (and sorry for taking so long to get to this)

I like this idea, would probably need a completely new patch since it would probably be better to try to add it to all avatarstack elements so that it could be used in more places. I'll keep this PR open for the time being (if that's alright) but I assume it would be better to open a new one for the addition to ThemeAttributes.

Vendicated:
assuming this can be closed now as superseded by #2249?

PR #1979, opened 2023-11-12
fix(dearrow): support DeArrow thumbnail submissions at 0 seconds

ajayyy:
0 is falsey and would trigger the check before. The original boolean can be used for this instead.


Vendicated:
thanks!

PR #1982, opened 2023-11-13
feat(PatchHelper): Paste Full Patch

MrDiamondDog:
image

Pasting your patch into the `Full Patch` field will auto-fill all the other fields


src/components/CheckedTextInput.tsx
```diff
@@ -55,6 +55,10 @@ export function CheckedTextInput({ value: initialValue, onChange, validate }: Te
         }
     }
 
+    useEffect(() => {
```
Vendicated, line 58: ?

MrDiamondDog: I forgot why I did this, I think something didn't work without it

MrDiamondDog: This is necessary so that when I use setMatch it will update the checked text input, otherwise it doesn't

Vendicated:
thanks

PR #1987, opened 2023-11-14
Clear URLs Plugin: Update defaultRules.ts

hir0xygen:
Add Discord's new tracking parameters that are added when copying a link to an image or video from a chat

### DIFF
--- a/src/plugins/clearURLs/defaultRules.ts
+++ b/src/plugins/clearURLs/defaultRules.ts
@@ -139,4 +139,7 @@ export const defaultRules = [
     "si@open.spotify.com",
     "igshid",
     "share_id@reddit.com",
+    "ex@*.discordapp.com",
+    "is@*.discordapp.com",
+    "hm@*.discordapp.com",
 ];


FieryFlames:
Why? they will be needed to view them at all eventually.

hir0xygen:
> Why? they will be needed to view them at all eventually.

Sorry, why is that? Removing the tracking parameters currently doesn't have any effect on the media linked.

FieryFlames:
> > Why? they will be needed to view them at all eventually.
> 
> Sorry, why is that? Removing the tracking parameters currently doesn't have any effect on the media linked.

In the future they will be required. Not sure if this matters in client and the client will just refresh all cdn links, but I don't see the point of removing them at this point

hir0xygen:
> > > Why? they will be needed to view them at all eventually.
> > 
> > 
> > Sorry, why is that? Removing the tracking parameters currently doesn't have any effect on the media linked.
> 
> In the future they will be required. Not sure if this matters in client and the client will just refresh all cdn links, but I don't see the point of removing them at this point

Just asking out of interest, how are you certain they will be required?

FieryFlames:
> > > > Why? they will be needed to view them at all eventually.
> > > 
> > > 
> > > Sorry, why is that? Removing the tracking parameters currently doesn't have any effect on the media linked.
> > 
> > 
> > In the future they will be required. Not sure if this matters in client and the client will just refresh all cdn links, but I don't see the point of removing them at this point
> 
> Just asking out of interest, how are you certain they will be required?

They will be used to authenticate the image is being used from within discord, they're cracking down on using discord as a file host. I'm not sure if they will be required when sending in-app (maybe the api will re-add them) but I don't see the point in removing them to begin with if the api is just going to re-add them

Cynosphere:
Duplicate of #1758 as well

PR #1990, opened 2023-11-15
shikiCodeblocks: added toLowerCase to avoid failed detections due to case

xero-lib:
Discovered bug with `Rust` not being a valid language tag, despite Discords builtin code block markdown accepting it. Most likely affects other language identifiers as well.


Nuckyz:
Looks good but avoid PRing from the main branch

Nuckyz (approved):

PR #1996, opened 2023-11-16
[Plugin] XFix

Twin1dev:
fixes twitter/X embeds not showing up.

literally just makes the presend event replace twitter.com with fxtwitter.com or x.com with fixupx.com


rniii:
Use TextReplace

Twin1dev:
wdym by this? i am using replace, is there something im not aware of?

nyakowint:
The TextReplace plugin with a rule can do exactly what this does so use it instead

Twin1dev:
ohh im so sorry, i look so stupid rn, thanks!

Issue #1749, opened 2023-09-24
Plugin README images thread


Vendicated:
yes

TheKodeToad:
no

Vendicated:
FavoriteEmojiFirst
![image](https://github.com/Vendicated/Vencord/assets/45497981/419c8c16-1afc-46e0-9cc2-20b9c3489711)
![image](https://github.com/Vendicated/Vencord/assets/45497981/4f57626d-cfc6-4155-a47c-2eac191231bb)

Vendicated:
FavoriteGifSearch
![image](https://github.com/Vendicated/Vencord/assets/45497981/19552adc-d921-4153-976e-e9361dc8fdaf)

Vendicated:
ImageZoom
![the plugin in action](https://github.com/Vendicated/Vencord/assets/45497981/408cd77d-c5f4-40bc-8de2-f977a31b3e5f)
![the context menu options offered by the plugin](https://github.com/Vendicated/Vencord/assets/45497981/3bede636-f1ce-493f-af46-788b920cb81c)

Vendicated:
PreviewMessage
![the plugin in action](https://github.com/Vendicated/Vencord/assets/45497981/3ce32860-e5cd-4ea2-bdab-e121f1703579)

Vendicated:
SearchReply
![the plugin in action](https://github.com/Vendicated/Vencord/assets/45497981/07e741d3-0f97-4e5c-82b0-80712ecf2cbb)

lumap:
MessageClickActions
![270138524-55422202-8804-4adf-95db-1062d5995adc](https://github.com/Vendicated/Vencord/assets/55940580/6885aca2-4021-4910-b636-bb40f877a816)

lumap:
QuickMention
![270138537-27b716f8-2c32-45ba-96d1-1fdb17090583](https://github.com/Vendicated/Vencord/assets/55940580/82d3fec7-4196-4917-b3c2-6e652b2aff9e)

lumap:
QuickReply
![270138545-c2ecb003-6d83-4a73-8990-4061e121754e](https://github.com/Vendicated/Vencord/assets/55940580/df79a27a-6529-4c70-8870-3c17d3637e4f)

lumap:
PictureInPicture
[...]

PR #1998, opened 2023-11-17
feat(plugin): AllowedMentions

arHSM:
# `AllowedMentions`

Fine grained control over whom to ping when sending or editing a message.

PS: Editing `allowed_mentions` in a message requires changing the content as the API persists the mentions if content is not changed


lewisakura:
I do wonder if this is actually safe to do for users since this is mainly a feature for bots, no? Just curious if this has the potential to flag.

lewisakura (changes requested):
Plugin needs a README.md

src/plugins/allowedMentions/README.md
```diff
@@ -0,0 +1,6 @@
+# AllowedMentions
+
+Fine grained control over whom to ping when sending a message.\
+Legacy chat input not supported for a11y reasons.
+
+![demo](./demo.gif)
```
lewisakura, line 6: use this for images, the site won't load them otherwise: https://github.com/Vendicated/Vencord/issues/1749

lewisakura (changes requested):

src/plugins/allowedMentions/README.md
arHSM: fixed

src/plugins/allowedMentions/AllowedMentions.tsx
```diff
@@ -0,0 +1,382 @@
...
+    // Guild/Friend Nickname > Global Name > Username
+
+    // User id if not cached in any stores
+    return nickname ?? globalName ?? username ?? userId;
+}
+
+function getDisplayableRoleName(roleId: string, guildId: string | null) {
+    // You *can* mention roles in DMs but is it really worth adding the UI for
+    // it in DMs
+    const role = guildId ? Object.values(GuildStore.getGuild(guildId).roles).find(r => r.id === roleId)?.name : undefined;
```
jamesbt365, line 84: You need to do this now.
```suggestion
    const role = guildId ? GuildStore.getRole(guildId, roleId)?.name : undefined;
```

jamesbt365 (commented):
I can't seem to figure out why you can't toggle everyone mentions when you don't have the permission to do so in the guild but you do in the channel, but this change is requires as the roles field is not directly present when you use the getGuild method anymore.

jamesbt365:
Seems that some recent changes broke about half the patches.

CatLocky:
@lewisakura May I continue working on this PR by opening a new draft PR?

PR #1999, opened 2023-11-17
fix(dearrow): remove > from DeArrow titles

ajayyy:
These are used by the extension to keep their case. Since this plugin does no autoformatting, it just needs to delete them. Regex only finds words that start with > (`>word`)


Vendicated:
thanks!

PR #2000, opened 2023-11-18
feat(plugin): CharCounter

ryan-0324:
Adds a character counter to text boxes.
The character counter text is fully customizable.


src/plugins/charCounter/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    get charCounter() {
+        return (
+            <span
+                style={{
+                    fontFamily: settings.store.useMonospacedFont ? "var(--font-code, revert)" : "revert",
+                    position: "absolute",
+                    right: "0",
+                    bottom: "-32px",
+                    whiteSpace: "nowrap"
+                }}
```
AutumnVN, line 78: give it a class name and use css

src/plugins/charCounter/index.ts
```diff
@@ -0,0 +1,110 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import "./styles.css";
```
AutumnVN, line 7:
```suggestion
import style from "./styles.css?managed";

```
and use enableStyle/disableStyle from api/Styles in start() stop() so css only applied when plugin is on (since you use discord class name now)

D3SOX:
Using this impacts the client's performance noticeably, even on beefy machines

src/plugins/charCounter/styles.css
```diff
@@ -0,0 +1,19 @@
...
+
+[class*="messagesWrapper_"]:has([class*="jumpToPresentBar__"]) + * [class*="characterCount_"] {
+    top: -48px !important;
+}
+
+[class*="messagesWrapper_"]:has([class*="jumpToPresentBar__"]) + * :has([class*="attachedBars_"]) [class*="characterCount_"] {
+    top: -45px !important;
+}
+
+[class*="messagesWrapper_"]:has(+ * [class*="characterCount_"]) [class*="container__"]:has(+ [class*="emptyForum__"]) {
```
Vendicated, line 17: these styles likely are why. this is some insane css. combining :has with * is awful for performance and definitely not needed

src/plugins/charCounter/index.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { disableStyle, enableStyle } from "@api/Styles";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+
+import style from "./styles.css?managed";
+
+class ValuePointer {
```
Vendicated, line 14: ?

Vendicated:
sorry but this code makes no sense to me and seems very overengineered. it's hardly maintainable. the complex css rules make no sense, the complex shouldShow makes no sense (who needs this? what's the use case?), the NumberPointer class makes no sense

as such i will have to reject this. thanks for your efforts regardless

src/plugins/charCounter/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    toString() {
+        return `${this.value}`;
+    }
+}
+
+const maxChars = new ValuePointer();
+const currChars = new ValuePointer();
+const remChars = new ValuePointer();
```
lewisakura, line 28: Just use `let` and numbers, you're already using `let` just below this

ryan-0324:
I agree that, in its current state, the plugin is overengineered and likely hard to maintain. However, this is a very outdated version of the plugin. I was waiting for #2149, which adds `FormattedMessage` to `webpack/common`, to be merged before committing the updated and largely simplified version of the plugin. 
Regarding the lag, I can assure you that it is not due to the usage of the `:has()` pseudo-class, which was the only way to ensure the character counter did not overlap with other elements in various situations. I have found that the lag is due to the absolute positioning of the character counter. More specifically, the lag is due to the fact that the CSS moves the counter outside and above the chat box, causing Slate, the library Discord uses for the chat box, to do expensive layout calculations involving the messages in the current channel. The lag caused by these calculations scales with the number of messages in the channel. You can see this for yourself by using the profiler in Chrome DevTools. Unfortunately, it seems the only practical way around this lag would be to implement a character counter from scratch, which I believe would be even more complex and harder to maintain.

PR #2003, opened 2023-11-20
Added AutoScroll plugin

PavelDobCZ23:
This is a plugin that was made for BetterDiscord to allow auto scrolling on non-Windows OSs (e.g. GNU/Linux), modified by me to work with Vencord. It was originally licensed with MIT license, which is compatible with GNU. The original source code can be found [here](https://github.com/hackermare/BD-AutoScroll). I have updated the code to use APIs provided by Vencord, transformed it from JavaScript to TypeScript and made sure it's type-safe and follows the project's code style rules.


Vendicated:
this is a native chromium feature that can be enabled in electron with a single command line flag

PavelDobCZ23:
> this is a native chromium feature that can be enabled in electron with a single command line flag

I know, but the problem is that the feature is quite incomplete, as it doesn't stop event propagation and therefore still pastes the clipboard into the message bar as GNU/Linux does by default, which is extremely annoying and makes the flag kinda pointless. I did some real work on this and so did the original creators of course, I've made it to work well with Vencord as quality is my top priority. For the reasons I've mentioned, there might be someone who finds it useful. It is ultimately your decision to include it in the base Vencord or it can always stay as self-compiled user plugin, but I would be happy to see it in base Vencord and will keep it up-to-date, if anything would need a fix.

Vendicated:
you can really easily stop the middle click pasting behaviour, either by adding a global keydown listener or by disabling it on a system level

PavelDobCZ23:
> you can really easily stop the middle click pasting behaviour, either by adding a global keydown listener or by disabling it on a system level

Alright you got me, close this PR if you wish. You know more than me about this, I just found this plugin and wanted it to work with Vencord, but sure if it doesn't meet your standards I understand that what I did was a waste of time probably.

Vendicated:
thank you anyway!

npc-glitch:
> I would've found such a plugin useful. I also wouldn't want to disable the middle-click pasting behavior on a System-Level. I wouldn't call it a waste of time.

I second this, only reason keeping me from using vencord

RandyGraham:
This would be useful. I have been down the rabbit hole of trying to disable middle click paste for Vesktop without disabling middle click entirely. it is not as easy as "just adding an event listener", and the Chromium flag which "natively enables this feature" does enable auto-scroll, but it does not disable pasting which results in the clipboard being pasted into the discord text box everytime I autoscroll.

PR #2004, opened 2023-11-20
fix: strikethrough in SilentMessageToggle's disabled SVG

korbough:


korbough:
image
This is what it looks like now, don't really know how to describe it, but there's now a little space/border between the strikethrough and the icon itself.

korbough:
everything should be good, is there anything else we are waiting on

PR #2008, opened 2023-11-21
MessageLogger: add context menu option to clear channel history

nmsturcke:
Clear the Message Log History for a whole channel/user/gdm with one click.

Updated version of #1333, with message edits now rendering - Not sure why I didn't do this in the first place lol


src/plugins/messageLogger/index.tsx
```diff
@@ -89,18 +89,66 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+    group?.push(
+        (<Menu.MenuItem
+            id="ml-clear-channel"
+            key="ml-clear-channel"
+            label="Clear Message Log History"
+            color="danger"
+            action={
+                () => {
+                    toDelete.forEach(message => {
+                        FluxDispatcher.dispatch({
```
CatGirlDShadow, line 114: Multiple messages should be handled through bulk delete dispatch, providing list of ids to delete

src/plugins/messageLogger/index.tsx
```diff
@@ -89,13 +89,56 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
     name: "MessageLogger",
     description: "Temporarily logs deleted and edited messages.",
-    authors: [Devs.rushii, Devs.Ven, Devs.AutumnVN],
+    authors: [Devs.rushii, Devs.Ven, Devs.AutumnVN, Devs.Nickyux],
 
     contextMenus: {
-        "message": patchMessageContextMenu
+        "message": patchMessageContextMenu,
+        "channel-context": patchChannelContextMenu,
+        "user-context": patchChannelContextMenu,
```
Vendicated, line 140: why user-context?

src/plugins/messageLogger/index.tsx
```diff
@@ -89,13 +89,56 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
     ));
 };
 
+const patchChannelContextMenu: NavContextMenuPatchCallback = (children, { channel }) => {
+    const messages: Array<any> = MessageStore.getMessages(channel.id)?._array;
```
Vendicated, line 93:
```suggestion
const patchChannelContextMenu: NavContextMenuPatchCallback = (children, { channel }) => {
	if (!channel) return;
    const messages: Array<any> = MessageStore.getMessages(channel.id)?._array;
    if (!messages) return;
```

src/plugins/messageLogger/index.tsx
```diff
@@ -89,13 +89,56 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+        else if (message.editHistory?.length) toClearEditHistory.push(message);
+    }
+
+    if (!toDelete.length && !toClearEditHistory.length) return;
+
+    const group = findGroupChildrenByChildId("mark-channel-read", children);
+    group?.push(
+        (<Menu.MenuItem
+            id="ml-clear-channel"
+            key="ml-clear-channel"
```
Vendicated, line 108:
```suggestion
            id="vc-ml-clear-channel"
```

src/plugins/messageLogger/index.tsx
```diff
@@ -89,13 +89,56 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+            label="Clear Message Log History"
```
Vendicated, line 109:
```suggestion
            label="Clear MessageLogger History"
```

src/plugins/messageLogger/index.tsx
```diff
@@ -89,13 +89,56 @@ const patchMessageContextMenu: NavContextMenuPatchCallback = (children, props) =
...
+                        channelId: channel.id,
+                        ids: toDelete.map(m => m.id),
+                        mlDeleted: true
+                    });
+                    toClearEditHistory.forEach(message => {
+                        message.editHistory = [];
+                        FluxDispatcher.dispatch({
+                            type: "MESSAGE_UPDATE",
+                            message: message,
+                        });
```
Vendicated, line 124:
```suggestion
                        updateMessage(channel.id, message.id)
```

Vendicated (commented):
we finally have a proper way to update messages without screwing over embeds, so we can revive this now!

src/plugins/messageLogger/index.tsx
nmsturcke: It's the only way I found to patch DMs as well... Not sure if there's a better way

We5Xi

Vendicated: oh alright, makes sense!

src/plugins/messageLogger/index.tsx
nmsturcke: There seems to be the same original issue, in which edited messages aren't updated unless hovered over. Just dispatching "MESSAGE_UPDATE" worked fine for me, but the `updateMessage()` doesn't for me

Not sure if anyone could double check just in case it's a me issue

Vendicated:
fixed it up now! thank you for the contirbution

PR #2009, opened 2023-11-22
feat(patcher): Grouped replacements

FieryFlames:
draft, would like input on if throwing and handling in the catch is the right thing to do

### DIFF
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -41,6 +41,8 @@ export interface Patch {
     all?: boolean;
     /** Do not warn if this patch did no changes */
     noWarn?: boolean;
+    /** Only patch the module if all replacements succeed */
+    group?: boolean;
     predicate?(): boolean;
 }
 
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -41,7 +41,7 @@ export interface Patch {
     all?: boolean;
     /** Do not warn if this patch did no changes */
     noWarn?: boolean;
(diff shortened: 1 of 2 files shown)


src/utils/types.ts
```diff
@@ -41,6 +41,8 @@ export interface Patch {
     all?: boolean;
     /** Do not warn if this patch did no changes */
     noWarn?: boolean;
+    /** Only patch the module if all replacements succeed */
```
Nuckyz, line 44:
```suggestion
    /** Only apply the patches inside this find if all replacements succeed */
```

Nuckyz (changes requested):

src/webpack/patchWebpack.ts
```diff
@@ -204,6 +204,9 @@ function patchFactories(factories: Record<string | number, (module: { exports: a
             if (code.includes(patch.find)) {
                 patchedBy.add(patch.plugin);
 
+                const preMod = mod;
```
Vendicated, line 207:
```suggestion
                const previousMod = mod;
```

src/utils/types.ts
```diff
@@ -41,6 +41,8 @@ export interface Patch {
     all?: boolean;
     /** Do not warn if this patch did no changes */
     noWarn?: boolean;
+    /** Only apply the patches inside this find if all replacements succeed */
```
Vendicated, line 44:
```suggestion
    /** Only apply this set of patches if all of them succeed. Use this, if your patches depend on each other */
```

PR #2015, opened 2023-11-24
Updater: Axios Update

thororen1234:
This update makes the client updater use axios. this makes a lot of things easier such as the requests in general as http is outdated.


thororen1234:
scratch the http is outdated was just told otherwise after they told me it was outdated 😭 
either way I feel axios is better over all!

MrDiamondDog:
What was wrong with it before? I see no reason to change this

nyakowint:
Why do this just to be added to devs? It works fine as is lol

Vendicated:
axios is extremely bloated

PR #2017, opened 2023-11-25
Utility function for loading Discord chunks

Nuckyz:


src/webpack/webpack.ts
```diff
@@ -69,7 +69,7 @@ export const filters = {
 export const subscriptions = new Map<FilterFn, CallbackFn>();
 export const listeners = new Set<CallbackFn>();
 
-export type CallbackFn = (mod: any, id: number) => void;
+export type CallbackFn = (mod: any, id: string) => void;
```
Vendicated, line 72: why this change?

src/webpack/webpack.ts
```diff
@@ -201,43 +201,53 @@ export const findBulk = traceFunction("findBulk", function findBulk(...filterFns
 
     if (found !== length) {
         const err = new Error(`Got ${length} filters, but only found ${found} modules!`);
-        if (IS_DEV) {
-            if (!devToolsOpen)
-                // Strict behaviour in DevBuilds to fail early and make sure the issue is found
-                throw err;
-        } else {
-            logger.warn(err);
-        }
```
Vendicated, line 210: why did you change this code? the warning is kinda annoying when you're testing filters in devtools, hence why this is here

src/webpack/webpack.ts
```diff
@@ -201,43 +201,53 @@ export const findBulk = traceFunction("findBulk", function findBulk(...filterFns
...
-export const lazyWebpackSearchHistory = [] as Array<["find" | "findByProps" | "findByCode" | "findStore" | "findComponent" | "findComponentByCode" | "findExportedComponent" | "waitFor" | "waitForComponent" | "waitForStore" | "proxyLazyWebpack" | "LazyComponentWebpack", any[]]>;
+/**
+ * Find the first module factory that includes all the given code
+ * @returns The module factory or null
+ */
+export function findModule(...code: string[]) {
+    const id = findModuleId(...code);
+    if (!id) return null;
+
+    return wreq.m[id].original ?? wreq.m[id];
```
Vendicated, line 247: why? we do not want people to use the original ones

src/webpack/webpack.ts
```diff
@@ -272,26 +282,28 @@ export function LazyComponentWebpack<T extends object = any>(factory: () => any,
...
-/**
- * Find the first module that has the specified properties
- */
-export function findByProps(...props: string[]) {
-    const res = find(filters.byProps(...props), { isIndirect: true });
+function indirectFind(name: string, filter: FilterFn) {
+    const res = find(filter, { isIndirect: true });
     if (!res)
-        handleModuleNotFound("findByProps", ...props);
+        handleModuleNotFound(name, filter);
```
Vendicated, line 296: this goes entirely against the point. you're now logging the `filters.byFoo()` return value which is identical for every filter and has 0 meaningful info unless you check the closure

src/webpack/webpack.ts
Nuckyz: oh that's true, the issue is it's returning our factory method instead of the mod, we isnt what we want

src/webpack/webpack.ts
Nuckyz: to not cast to number so much, since when we get the key it's always a string

src/webpack/webpack.ts
Nuckyz: sorry I was confused on why it didnt warn in dev mode

FieryFlames (approved):

PR #2019, opened 2023-11-29
feat(plugin): FixEmojiCrash

AutumnVN:
fix emoji has name `default` cause discord explod

step to reproduce:
1. create emoji with name `default`
2. search some emoji in chat by typing something like `:abcdef`
3. watch your discord explod :trollface:


lewisakura:
as comical as this would be to fix for them, we shouldn't really be writing patches for their own broken code in the first place (yes, we have other "fix" plugins, but in reality they're actually just features disguised as fixes), especially since what this targets will be fixed in one or two client versions

also, this doesn't seem to happen for everyone, so it would be confusing to introduce such a plugin to fix a non-issue.

Issue #2022, opened 2023-11-30
[Bug] SpotifyUtils: POST /next /previous running twice.


MeguminSama:
### Discord Account

megumin

### What happens when the bug or crash occurs?

When clicking next/previous when using a device via Spotify Connect, the request is getting wrongly rejected.

While caused by Discord's code, it would be good for us to fix this functionality.

`SpotifyUtils.SpotifyAPI` wrongly rejects the promise if the status is `202 (Request Accepted)`, which causes Discord to re-send the request after 5000ms. This means that the track is skipped twice. I'm unsure as to why Discord treats a 202 as a failure.

Possible solutions would be:
- Set the `retries` argument (the 4th function parameter) to `0` (it defaults to 1), however this could have unintended (albeit minimal) effects if a request genuinely fails.
- Write a regex patch to stop the request function from rejecting `202` statuses
- Reimplement the request function to not reject `202` statuses

### What is the expected behaviour?

On click, should skip once.

### How do you recreate this bug or crash?

1. Run spotify via Spotify Connect (smart home/spotifyd/etc)
2. Use SpotifyControls skip/prev on Vencord.
3. Wait 5s, it will run the command a second time.

### Errors

meow

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [X] I have read the requirements for opening an issue above

PR #2023, opened 2023-11-30
fix(SpotifyControls): Requests double-sending when using Spotify Connect

MeguminSama:
Fixes https://github.com/Vendicated/Vencord/issues/2022

Disables the check for status 202 which causes the promise to reject.

### DIFF
--- a/src/plugins/spotifyControls/index.tsx
+++ b/src/plugins/spotifyControls/index.tsx
@@ -55,13 +55,19 @@ export default definePlugin({
                 replace: "return [$self.renderPlayer(),$1]"
             }
         },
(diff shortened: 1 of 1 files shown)


src/plugins/spotifyControls/index.tsx
```diff
@@ -70,7 +76,7 @@ export default definePlugin({
                 match: /repeat:"off"!==(.{1,3}),/,
                 replace: "actual_repeat:$1,$&"
             }
-        }
```
Nuckyz, line 73: can you remove the trailing comma

Nuckyz (approved):

PR #1746, opened 2023-09-23
Improve permission checks on several plugins

lumap:
This PR makes it impossible to trigger the edit/reply in read-only channels. Additionally, I moved this plugin to it's own folder and added a README

PR #2028, opened 2023-12-05
Fix QuickMention not working in DMs

fres621:
Fix the check added in #1746 to make the button show up in direct messages.
One remaining issue is the official Discord DM where the button should not show up but I don't know what's the best way to deal with it. First PR, forgive me if something's wrong :p

### DIFF
--- a/src/plugins/quickMention/index.tsx
+++ b/src/plugins/quickMention/index.tsx
@@ -31,7 +31,7 @@ export default definePlugin({
     start() {
         addButton("QuickMention", msg => {
             const channel = ChannelStore.getChannel(msg.channel_id);
-            if (!PermissionStore.can(PermissionsBits.SEND_MESSAGES, channel)) return null;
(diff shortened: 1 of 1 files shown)


src/plugins/quickMention/index.tsx
```diff
@@ -31,7 +31,7 @@ export default definePlugin({
     start() {
         addButton("QuickMention", msg => {
             const channel = ChannelStore.getChannel(msg.channel_id);
-            if (!PermissionStore.can(PermissionsBits.SEND_MESSAGES, channel)) return null;
+            if (!channel.guild_id || !PermissionStore.can(PermissionsBits.SEND_MESSAGES, channel)) return null;
```
Vendicated, line 34: uhh, shouldn't this be
```suggestion
            if (channel.guild_id && !PermissionStore.can(PermissionsBits.SEND_MESSAGES, channel)) return null;
```

fres621: oh yeah damn I am sorry about that

Vendicated:
thanks

PR #2032, opened 2023-12-06
feat(MessageLogger): Add delete & edit toggle

hypergonial:
Add the ability to toggle the logging of edits or deletes seperately. Quick testing seems to show that it works but I probably missed things.

P.S. I have no idea what I'm doing, feedback is welcome!

P.P.S. Should I add myself to the authors for small changes like these?


hypergonial:
> perhaps it could be done as a runtime check in `shouldIgnore`? that way the user doesn't have to restart after toggling?

Good point, done!

src/plugins/messageLogger/index.tsx
```diff
@@ -197,7 +207,7 @@ export default definePlugin({
         return cache;
     },
 
-    shouldIgnore(message: any) {
+    shouldIgnore(message: any, is_edit: boolean = false) {
```
Vendicated, line 210:
```suggestion
    shouldIgnore(message: any, isEdit: boolean = false) {
```

keep casing consistent! in javascript we use camelCase

src/plugins/messageLogger/index.tsx
```diff
@@ -197,7 +207,7 @@ export default definePlugin({
...
         const { ignoreBots, ignoreSelf, ignoreUsers, ignoreChannels, ignoreGuilds } = Settings.plugins.MessageLogger;
```
Vendicated, line 211: add additional destructures here instead of calling them on the full path object below

hypergonial: I'm not entirely sure what you mean by this, sorry. Can you clarify?

Vendicated:
```suggestion
        const { ignoreBots, ignoreSelf, ignoreUsers, ignoreChannels, ignoreGuilds, logEdits, logDeletes } = Settings.plugins.MessageLogger;
```

hypergonial: Done, thanks!

Vendicated:
thank you!

PR #2036, opened 2023-12-08
AlwaysAnimate: Add guild banner

zImPatrick:
Basically makes the guild banner always animate, even when not hovered over

### DIFF
--- a/src/plugins/alwaysAnimate/index.ts
+++ b/src/plugins/alwaysAnimate/index.ts
@@ -46,6 +46,13 @@ export default definePlugin({
                 match: /(?<=\.activityEmoji,.+?animate:)\i/,
                 replace: "!0"
             }
+        },
+        {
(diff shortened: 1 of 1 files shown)


src/plugins/alwaysAnimate/index.ts
```diff
@@ -46,6 +46,13 @@ export default definePlugin({
                 match: /(?<=\.activityEmoji,.+?animate:)\i/,
                 replace: "!0"
             }
+        },
+        {
+            find: ".animatedBannerHoverLayer,onMouseEnter:",
+            replacement: {
+                match: /(?<=guildBanner:\i,animate:)\i(?=}\))/,
```
Nuckyz, line 53:
```suggestion
                match: /(?<=guildBanner:\i,animate:)\i)/,
```

Nuckyz:
Thank you!

PR #2039, opened 2023-12-09
feat(plugin): blurs all images and videos until hovered

HellBri8nger:
Plugins blurs all images and videos until hovered, you can set the blur strength in settings


AutumnVN:
it is just a css snippet, doesnt need to be a plugin

lewisakura:
>plugin
>look inside
>css

PR #2042, opened 2023-12-10
updater: Add top and bottom margins between commits

TheEvilSkeleton:
I haven't tested it locally, because the updater doesn't work with a custom Vencord location

image


Vendicated:
thanks :3

PR #2048, opened 2023-12-11
Add IrcColors plugin

jedenastka:
The plugin generates unique colors for usernames in chat, similar to how it works in many IRC clients.

Chat with Compact++ and IrcColors enabled


jedenastka:
This is a very good MR please review

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "IrcColors",
+    description: "Makes username colors in chat unique, like in IRC clients",
+    authors: [Devs.Grzesiek11],
+    patches: [
+        {
+            find: ".useCanSeeRemixBadge)",
```
Vendicated, line 28: this find will break when the remix experiment is removed

would be better to use something like
```suggestion
            find: '="SYSTEM_TAG',
```

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,60 @@
...
+            replacement: {
+                match: /style:"username"===\i&&null!=\i\?{color:\i}:void 0/,
```
Vendicated, line 30: this match is incompatible with BetterRoleDot
image

Vendicated:
```suggestion
                match: /(?<=className:\i\.username,style:.{0,50})\{color:\i\}/,
```

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,60 @@
...
+                replace: "style:{color:$self.calculateNameColor(BigInt(arguments[0].message.author.id))}"
```
Vendicated, line 31:
```suggestion
                replace: "{color:$self.calculateNameColor(arguments[0])}"
```

Vendicated (dismissed):

src/plugins/ircColors/index.ts
jedenastka: This is supposed to be an override, so is something like this good instead? `(?<=className:\i\.username,style:).{0,50}(?=,onClick)`

Vendicated: true, i missed that you also need to remove the boolean check. in that case, it's easiest to just duplicate the property:
```js
style: null != C ? {} : void 0,
style: yourStyle
```

jedenastka: I'm kinda surprised this is valid, but then again this is JavaScript. Gonna do that then.

jedenastka: Is `(?<=className:\i\.username,style:.{0,50}:void 0,)` a good match value for this?

Vendicated:
Please add a README.md file with a screenshot

Vendicated:
its a bit weird that this doesn't affect the member list

jedenastka:
I added a README, and a setting to control the lightness component of the color for people using themes different than the default dark.

> its a bit weird that this doesn't affect the member list

It's supposed to only be present in the chat. In the member list I think having the role color be present is more valuable (like the role dot).

craftxbox:
> It's supposed to only be present in the chat. In the member list I think having the role color be present is more valuable (like the role dot).

Maybe have an option for it? In my client (hexchat) the members list is colored the same as the chat names, and to the left of the name is a colored dot to indicate a user's status (blue for voice, cyan for halfop, green for op, and so on)

jedenastka:
Okay, added this as an opt-in feature.

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        description: "Lightness, in %. Change if the colors are too light or too dark.",
+        restartNeeded: true,
+        type: OptionType.NUMBER,
+        default: 70,
+    },
+    memberListColors: {
+        description: "Replace role colors in the member list",
+        restartNeeded: true,
+        type: OptionType.BOOLEAN,
+        default: false,
```
Vendicated, line 60:
```suggestion
        default: true,
```

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    },
+});
+
+export default definePlugin({
+    name: "IrcColors",
+    description: "Makes username colors in chat unique, like in IRC clients",
+    authors: [Devs.Grzesiek11],
+    patches: [
+        {
+            find: "=\"SYSTEM_TAG\"",
```
Vendicated, line 70:
```suggestion
            find: '="SYSTEM_TAG"',
```

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            predicate: () => settings.store.memberListColors,
+        },
+    ],
+    settings,
+    calculateNameColorForMessageContext(context: any) {
+        return calculateNameColorForUser(BigInt(context.message.author.id));
+    },
+    calculateNameColorForListContext(context: any) {
+        return calculateNameColorForUser(BigInt(context.user.id));
+    },
```
Vendicated, line 91: Since both of these will run for elements that are always rendered, if they break in the future it means crash loop. Thus, adding some safety like optional chaining is vital. Not sure if calculating colour for 0 is good, maybe it would be better to just return no colour or smth instead

```suggestion
    calculateNameColorForMessageContext(context: any) {
        return calculateNameColorForUser(BigInt(context?.message?.author?.id ?? 0));
    },
    calculateNameColorForListContext(context: any) {
        return calculateNameColorForUser(BigInt(context?.user?.id ?? 0));
    },
```

src/plugins/ircColors/README.md
```diff
@@ -0,0 +1,17 @@
...
+making distinguishing between different users easier. Inspired by the feature
+in many IRC clients, such as HexChat or WeeChat.
+
+Keep in mind this overrides role colors in chat, so if you wish to know
+someone's role color without checking their profile, enable the role dot: go to
+**User Settings**, **Accessibility** and switch **Role Colors** to **Show role
+colors next to names**.
+
+Created for use with the [Compact++](https://gitlab.com/Grzesiek11/compactplusplus-discord-theme)
+theme.
```
Vendicated, line 17: This kinda reads like an ad and isn't relevant to plugin users so it should be removed

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+    const idHash = hash(idBuffer);
+
+    return `hsl(${idHash % 360n}, 100%, ${settings.store.lightness}%)`;
+}
+
+const settings = definePluginSettings({
+    lightness: {
+        description: "Lightness, in %. Change if the colors are too light or too dark.",
+        restartNeeded: true,
```
Vendicated, line 52: i don't think this actually needs a restart, just to reopen chat, no?
```suggestion
```

src/plugins/fixCodeblockGap/README.md
```diff
@@ -0,0 +1,9 @@
+# FixCodeblockGap
```
Vendicated, line 1: is this meant to be here? I think that's okay, but the images seem to be missing their image links? xD

Vendicated (commented):
Looks good for the most part so this can definitely be merged once these minor things are fixed

src/plugins/fixCodeblockGap/README.md
jedenastka: LMAO it's not

src/plugins/ircColors/README.md
jedenastka: See [this conversation on the Discord server](https://discord.com/channels/1015060230222131221/1015063227299811479/1231143941743575153), you have personally approved this lol. That said, I can remove it or tweak it if you changed your mind.

src/plugins/ircColors/index.ts
jedenastka: Indeed. Not sure why I didn't notice that.

src/plugins/ircColors/index.ts
jedenastka: So you are saying this is better as an opt-out?

Vendicated: yes

jedenastka: Sure.

src/plugins/ircColors/index.ts
jedenastka: This should work.

src/plugins/ircColors/index.ts
```diff
@@ -0,0 +1,99 @@
...
+        result ^= BigInt(byte);
+        result = (result * fnvPrime) % 2n**32n;
+    }
+
+    return result;
+}
+
+// Calculate a CSS color string based on the user ID
+function calculateNameColorForUser(id: bigint) {
+    const idBuffer = new ArrayBuffer(16);
```
Vendicated, line 39: shouldn't this be 8? we're working with a u64 which is 8 bytes, so why 16bytes size? you're just gonna be reading 8 null bytes every time

image

jedenastka: Yes, it should be. Just a typo.

Vendicated:
there's actually one bug I noticed: for people who use the colour dot instead of role name (or use BetterRoleDot), the roledot has the wrong colour

image

also this doesn't integrate with RoleColorEverywhere which kinda sucks and should be added

image

In any case, I will just merge this now and we can make these fixes later

jedenastka:
Thanks for the merge!

> there's actually one bug I noticed: for people who use the colour dot instead of role name (or use BetterRoleDot), the roledot has the wrong colour

I think I already said this on Discord (a year ago, when I initially created the branch lol), but I don't really see value in having the role dot on while having the plugin enabled. Really, I considered patching out the dot completely, but decided against it because it's an additional patch that requires work (lazy) and it doesn't really hurt anybody (maybe someone wants to still see the role color as well, in which case they can use this combination (I don't see this being used by anyone but still)).

> also this doesn't integrate with RoleColorEverywhere which kinda sucks and should be added

Agree. This is pure laziness on my part, I would probably use RoleColorEverywhere if it was integrated.

PR #2049, opened 2023-12-12
feat: only show one QuickCSS editor window

ryanccn:
Focuses an existing QuickCSS editor if it already exists when opening QuickCSS editor.

### DIFF
--- a/src/main/ipcMain.ts
+++ b/src/main/ipcMain.ts
@@ -138,20 +138,26 @@ export function initIpc(mainWindow: BrowserWindow) {
     }));
 }
 
+let monacoEditorWin: BrowserWindow;
+
 ipcMain.handle(IpcEvents.OPEN_MONACO_EDITOR, async () => {
-    const win = new BrowserWindow({
-        title: "Vencord QuickCSS Editor",
-        autoHideMenuBar: true,
-        darkTheme: true,
-        webPreferences: {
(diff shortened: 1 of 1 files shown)


TheKodeToad (approved):

ryanccn:
@Vendicated

src/main/ipcMain.ts
```diff
@@ -138,20 +138,26 @@ export function initIpc(mainWindow: BrowserWindow) {
...
-        webPreferences: {
-            preload: join(__dirname, IS_DISCORD_DESKTOP ? "preload.js" : "vencordDesktopPreload.js"),
-            contextIsolation: true,
-            nodeIntegration: false,
-            sandbox: false
-        }
-    });
-
-    makeLinksOpenExternally(win);
+let monacoEditorWin: BrowserWindow;
```
Vendicated, line 141: i don't really like this way of storing the window. it will unnecessarily keep it around even after a close

i think instead of storing it at all you could use `BrowserWindow.getAllWindows().find(w => w.title === "Vencord QuickCSS Editor")`. or at least attach an `.on("close"` handler to set the variable to null when the window is closed

Vendicated:
please refrain from force pushing. it makes it very hard to understand what you changed after review and makes me have to re-review everything.

why did you not take the other approach of dynamically finding the browserwindow?

Vendicated:
superseded by https://github.com/Vendicated/Vencord/commit/2ab1c50c7355849e42ad81afcfef3d008b459247 which uses that approach instead, thanks though!

PR #2050, opened 2023-12-12
fix issue with non apng gif stickers being uploaded

sapphonie:
It seems that `getStickerLink` will occasionally give back a `.png` on a gif sticker, despite the `.png` not even being an APNG. I don't know if this is a discord bug or not, I haven't dug that deep, but it seems like this is the simplest solution.

Apologies if this is poor style, I don't write much TS/JS.

Thanks
Sappho

### DIFF
--- a/src/plugins/fakeNitro/index.ts
+++ b/src/plugins/fakeNitro/index.ts
@@ -787,7 +787,13 @@ export default definePlugin({
                 if (sticker.available !== false && (canUseStickers || sticker.guild_id === guildId))
                     break stickerBypass;
 
-                const link = this.getStickerLink(sticker.id);
+                // [12/12/2023] 
+                // Work around an annoying bug where discord API will return StickerType.GIF
+                // but give us a normal non animated png for no reason?
+                let link = this.getStickerLink(sticker.id);
+                if (sticker.format_type == StickerType.GIF && link.includes(".png")) {
+                    link = link.replaceAll(".png", ".gif");
+                }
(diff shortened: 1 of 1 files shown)


src/plugins/fakeNitro/index.ts
```diff
@@ -787,7 +787,13 @@ export default definePlugin({
                 if (sticker.available !== false && (canUseStickers || sticker.guild_id === guildId))
                     break stickerBypass;
 
-                const link = this.getStickerLink(sticker.id);
+                // [12/12/2023] 
+                // Work around an annoying bug where getStickerLink will return StickerType.GIF,
+                // but will give us a normal non animated png for no reason
```
Nuckyz, line 792:
```suggestion
                // [12/12/2023] 
                // Work around an annoying bug where getStickerLink will return StickerType.GIF,
                // but will give us a normal non animated png for no reason
                // TODO: Remove this workaround when it's not needed anymore
```

Nuckyz (changes requested):

src/plugins/fakeNitro/index.ts
sapphonie: 👍

Nuckyz (approved):

Nuckyz:
Lint is failing

sapphonie:
> Lint is failing

Resolved

Nuckyz:
Thank you!

PR #2052, opened 2023-12-13
feat(plugin): Added Plugin FixTwitterEmbed

jhonnysc:
Replaces x.com or twitter.com links with fxtwitter.com links to fix embeds
More details on fx links -> https://github.com/FixTweet/FixTweet

### DIFF
--- a/src/plugins/fixTwitterEmbed/index.ts
+++ b/src/plugins/fixTwitterEmbed/index.ts
@@ -1,0 +1,36 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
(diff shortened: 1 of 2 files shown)


Vendicated:
this can be done with textreplace

Vendicated:
TextReplace rule

To use a different service than FxTwitter, just change the replacement to whatever u want
```
Find: https://twitter.com
Replace: https://fxtwitter.com
Only if includes: /status/
```

thanks for your contribution regardless!

jhonnysc:
The purpose of this plugin was to make easy for users who doesn't know about TextReplacer (just like me) but if we feel like the similarity between them is too much, it's ok, I installed TextReplacer and already have it working, thanks!

PR #2054, opened 2023-12-14
Fix SilentTyping icon to be more consistent with the SilentMessageToggle plugin

MeowcaTheoRange:
As a user of both SilentTyping and SilentMessageToggle, seeing both icons next to each other is a little (very) confusing.

So, I'm proposing an alternate icon, adding the "Z" from the SilentMessageToggle bell icon onto the keyboard and inverting the state of the red slash on the icon.

## Old
image
image

## New
image
image


Vendicated:
this doesnt seem right

PR #2057, opened 2023-12-15
feat: translation

lewisakura:
This PR contains a draft translation system for Vencord.

Of course, this is purely a draft idea for now, but the actual groundwork for a translation system now exists:
- translations are bundled into the mod itself by esbuild, so no external file loading or downloads for now - in future we will use an asar file.
- it uses Discord's i18n system to work out what language the system should use (including when the language updates, since the system will clear the messages it has cached)

If we decide to go forward with this idea, we should:
- pick the languages we want to initially translate into first

- decide who leads what language since we need trusted localizers
- write a translation guide (i.e., the rules on how things get translated)

This is obviously a very complex project and will take a very long time most likely to finish, but I'm opening this now to prove it can be done and how it would work.

There is also plenty to discuss, but this also comes down to writing the translation guide, so let's talk about the viability and consequences of such a system, with possible resolutions, before we get any deeper.


lewisakura:
blitzing this PR to work on a different version with a different system

PR #2059, opened 2023-12-17
fix cursor style plugins filter

YOMAN1792:
## Summary

This PR fixes the cursor style for the "Show All" option in the plugin filter dropdown. Previously, the cursor would change to a text cursor when hovering over this option, which was misleading as this is a clickable element and not a text field. This PR changes the cursor to a pointer when hovering over this option to indicate that it is clickable.

## Test Plan

1. Navigate to the plugin management interface.
2. Hover over the "Show All" option in the plugin filter dropdown.
3. Observe that the cursor changes to a pointer, indicating that this option is clickable.


Vendicated:
seems fixed now, so this change is no longer necessary

thank you regardless!

PR #2046, opened 2023-12-11
feat(plugin): Fixed LastFMRichPresence albums reporting incorrectly

KontrollFreek:
For some reason LastFM decides to default to the last used album if it has no clue what it's supposed to be. Also did some grammar and capitalization stuff and accidentally committed by .gitignore settings. sorry.

PR #2060, opened 2023-12-17
new plugin DontRoundMyTimestamps: round 7.6y -> 7y instead of 8y

KontrollFreek:
Stops relative timestamps (<t:1000:**R**>) from using Math.round.


src/plugins/lastfm/index.tsx
```diff
@@ -101,81 +101,86 @@ function setActivity(activity: Activity | null) {
```
MrDiamondDog, line 101: This entire file's changes are completely unrelated and unnecessary

KontrollFreek: i have no clue how PRs work, these are from a previous PR (#2046)

MrDiamondDog: You have to make different branches on your fork so the changes aren't combined

src/plugins/dontRoundMyTimestamps/index.ts
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { moment } from "@webpack/common";
+
+export default definePlugin({
+    name: "DontRoundMyTimestamps",
+    authors: [Devs.Lexi],
+    description: "Stops relative timestamps from being rounded and therefore incorrect.",
```
Vendicated, line 26: this description and name are misleading. floor is still a round (down)

Vendicated: something like
```suggestion
    description: "Always rounds relative timestamps down, so 7.6y becomes 7y instead of 8y",
```

Vendicated:
thanks!!

PR #2061, opened 2023-12-17
Add TextEmoji plugin

jedenastka:
The plugin prevents Discord from replacing emoji with images, leaving it up to the browser to render them.

obraz


jedenastka:
Here's how the ZWJ issue looked like (pretty funny if you ask me):



Now it works correctly:

src/plugins/textEmoji/index.ts
```diff
@@ -0,0 +1,44 @@
...
+            replacement: {
+                match: /(if\(null!=.{1,2}\))return .{1,2}\.default\.getURL\(.{1,2}\)/,
+                replace: "$1return null",
+            },
+        },
+        {
+            find: ",findInlineEmojisFromSurrogates:",
+            replacement: {
+                match: /if\(!0!==.{1,2}&&!.{1,2}\.test\(.{1,2}\)\)/,
+                replace: "",
```
TheKodeToad, line 40:
```suggestion
                replace: "if (true)",
```
feels slightly better? /shrug

jedenastka: I like it the way it is, I can change it if it's preffered though. (There also might be a minimal performance difference, but that'd be a microoptimization so it's not like it really matters).

TheKodeToad: It's not about performance, IMO it's clearer - and it also won't break if an else is added.
I'm not a maintainer so it's up to you - or a maintainer once they see this.

src/plugins/textEmoji/index.ts
```diff
@@ -0,0 +1,44 @@
...
+            find: "getSrc(){",
+            replacement: {
+                match: /(if\(null!=.{1,2}\))return .{1,2}\.default\.getURL\(.{1,2}\)/,
+                replace: "$1return null",
+            },
+        },
+        {
+            find: ",findInlineEmojisFromSurrogates:",
+            replacement: {
+                match: /if\(!0!==.{1,2}&&!.{1,2}\.test\(.{1,2}\)\)/,
```
TheKodeToad, line 39:
```suggestion
                match: /if\(!0!==\i&&!\i\.test\(\i\)\)/,
```
Forgot this, oops. this is definitely something the maintainers would want

src/plugins/textEmoji/index.ts
jedenastka: The performance part was just a sidenote (as I said, I know it doesn't really matter), it's just that in my opinion removing the `if` entirely is cleaner... It's ultimately a matter of opinion so I'm gonna leave it up to the maintainers for now. Though  your point that it will break when an `else` is added is fair...

src/plugins/textEmoji/styles.css
```diff
@@ -0,0 +1,3 @@
+.reaction_fef95b .emoji {
```
TheKodeToad, line 1:
```suggestion
[class^="reaction_fef95b"] .emoji {
```
AA here we go again (I should've just looked at the whole thing in one go)

jedenastka: Is this really a good idea? After some discussion on [Discord](https://discord.com/channels/1015060230222131221/1134844326933954622/1188452136648978432) I am convinced it's not, and it'd be better to just update the plugin when it breaks. I reverted the change, though if the maintainers will want this I will re-introduce it.

TheKodeToad: partial matching is faster and more compatible. I don't understand why you wouldn't use it.

Vendicated: kode's suggestion is correct. do not hardcode class hashes

jedenastka: Re-introduced the change, as it is preferred by the project.

jedenastka:
Fixed this for the latest client (also bump).

jedenastka:
Added a README here as well, because why not.

jedenastka:
Fixed on latest client, also bump.

jedenastka:
This is partially broken on latest client, just saying that I am aware.

Vendicated:
I don't think there is much demand for this plugin

PR #2063, opened 2023-12-19
Added Commentor plugin and a Chat Icon

bloomsirenix:
:heavy_plus_sign:  Commentor plugin that allows you to comment on messages by clicking the chat icon it even has a style.css for the modal(it might still does need work it ain't the prettiest looking modal I did my best)
:heavy_plus_sign:  added ChatIcon to Icons.tsx (as it might be handy for other plugins in the future)
📜  Note: the modal is kinda bad looking im sorry about that it is however fully scrollable and functional
Comments button on messages:
image
Empty comment section:
image
Single to two comments:
image
With scrollbar enables after 3 or more comments!
image


bloomsirenix:
The UI has been updated  
image
it now looks better

MrDiamondDog:
Can other people see these comments? If not, what exactly is the point of this plugin?

bloomsirenix:
> Can other people see these comments? If not, what exactly is the point of this plugin?

They cant at the moment but I plan to implement it if I can find a secure host to do so
it is an way to comment on messages kinda like threats but less annoying and wont make a new channel

bloomsirenix:
it was inspired by that secret message plugin for better discord but it looks more native

Vendicated:
like Diamond said, this has little use in its current state

even if you were to add a backend for this, what exactly differentiates this from just creating a thread?

overall this is way too niche for being merged

thank you a lot regardless <3 (and my apologies for only looking at this now)

PR #2064, opened 2023-12-19
Add FixCodeblockGap plugin

jedenastka:
The plugin removes the gap between codeblocks and text by fixing Discord's regex to also match the optional following newline.

Vanilla Discord:

FixCodeblockGap:


src/plugins/fixCodeblockGap/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export default definePlugin({
+    name: "FixCodeblockGap",
+    description: "Removes the gap between codeblocks and text",
+    authors: [Devs.Grzesiek11],
+    patches: [
+        {
+            find: ".default.Messages.DELETED_ROLE_PLACEHOLDER",
+            replacement: {
+                match: /\/\^```\(\?:\(\[a-z0-9_\+\\-\.#\]\+\?\)\\n\)\?\\n\*\(\[\^\\n\]\[\^\]\*\?\)\\n\*```\//,
```
Vendicated, line 30: this is kinda hell. you could avoid all the escaping by doing something like
```js
{
	match: String.raw`/^\`\`\`...`,
	replace: whatever
}
```

jedenastka: Didn't know that JS has raw strings. However, you apparently aren't able to escape backticks in them, so I needed to do something like:

```js
String.raw`/^${"```"}...`
```

Looks weird but seems like there's no better solutions. Is this okay?

I changed the `replace` value as well, to get rid of the backslashes.

src/plugins/fixCodeblockGap/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export default definePlugin({
+    name: "FixCodeblockGap",
+    description: "Removes the gap between codeblocks and text",
+    authors: [Devs.Grzesiek11],
+    patches: [
+        {
+            find: ".default.Messages.DELETED_ROLE_PLACEHOLDER",
+            replacement: {
+                match: String.raw`/^${"```"}(?:([a-z0-9_+\-.#]+?)\n)?\n*([^\n][^]*?)\n*${"```"}/`,
```
Vendicated, line 30:
```suggestion
                match: String.raw`/^${"```"}(?:([a-z0-9_+\-.#]+?)\n)?\n*([^\n][^]*?)\n*${"```"}`,
```
if you avoid matching this `/`, you can just set replace to `replace: "$&\n?/"`

src/plugins/fixCodeblockGap/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "FixCodeblockGap",
+    description: "Removes the gap between codeblocks and text",
```
Vendicated, line 24:
```suggestion
    description: "Removes the gap between codeblocks and text below it",
```

PR #2070, opened 2023-12-24
petpet: fix /petpet image command

AutumnVN:


Vendicated:
i literally tested this the other day and it worked :face_with_spiral_eyes: 

anyway, it might be more future proof to add `export const { DraftType } = findByPropsLazy("DraftType")` to webpack commons and use that everywhere?

Vendicated:
i changed it to use the webpack common, but there's still the following issue:

it doesn't clear the slash command draft after you submit it. the fix is likely to just clearDraft(DraftType.SlashCommand), but i'm confused because Discord doesn't do that and it works fine for them :/

PR #2071, opened 2023-12-24
Decor: /config endpoint

ImLvna:
Allows the user to set a custom decor server via a hidden setting, fetching the client id and cdn url from the server, allowing easier development on a seperate server


Vendicated:
i don't think this is a good idea to have. dumb users may change this field then wonder why it doens't work

plus this is so niche, someone who needs this should probably just build from source. thank you anyway!

PR #2072, opened 2023-12-26
feat(plugin): Added fakeProfile

sampathgujarathi:
This plugin allows users to use profile effects, themes, animated avatar, banner, avatar decorations without nitro.
image


ryan-0324:
Duplicate of #1930

sampathgujarathi:
> Duplicate of #1930

In Effecto plugin users dont have to copy 3y3 id and paste it in their bio.

FieryFlames:
> > Duplicate of #1930
> 
> In Effecto plugin users dont have to copy 3y3 id and paste it in their bio.

There's no clear way for users to actually set their own effect though.

sampathgujarathi:
> > > Duplicate of #1930
> > 
> > 
> > In Effecto plugin users dont have to copy 3y3 id and paste it in their bio.
> 
> There's no clear way for users to actually set their own effect though.

Users can to join server and select their profile effects from there, then it will be automatically updated. It's same as USRBG

FieryFlames:
I missed the settingsAboutComponent.. I think the 3y3 version is better, userbg has a reason to use a bot (needs to store images) but this can clearly be done with 3y3.

sampathgujarathi:
> I missed the settingsAboutComponent.. I think the 3y3 version is better, userbg has a reason to use a bot (needs to store images) but this can clearly be done with 3y3.

3y3 is not better because it will take up to 20 characters from about me. I will be making this way more simple after few days

ItsCrusaderDude69:
well i like it

Vendicated:
as i said on discord, this mostly duplicates already existing plugins, and i don't trust you with being responsible for the API, because i don't know you

so i'll be rejecting this. thanks for your efforts regardless

PR #2076, opened 2023-12-28
SendTimestamps: add setting to disable message replacement

jedenastka:
SendTimestamps adds two features allowing to send timestamps: the button in the input area, and the replacement of timestamps in message contents. This allows disabling content replacement if the user only wants to have the input area button.

Something I'm not sure on is whether a corresponding setting for the input area button should be added as well. The description of the setting might also use some improvement.


src/plugins/sendTimestamps/index.tsx
```diff
@@ -77,7 +86,7 @@ function PickerModal({ rootProps, close }: { rootProps: ModalProps, close(): voi
                 />
 
                 <Forms.FormTitle>Timestamp Format</Forms.FormTitle>
-                <Select
+                <Selec
```
AutumnVN, line 89: you ate the letter t

Vendicated: yummy

jedenastka: My sleepy ass forgot to check `git diff`...

jedenastka:
Forgot to run `pnpm test` after renaming the variable as well.

Vendicated:
thanks!

PR #2080, opened 2023-12-29
Urban Dictionary: Chooses top rated definition & more results.

jewdev:
I've noticed that I get irrelevant definitions, so I made these changes:

Before, API gave only 10 results and the plugin chose the first result:
before

After, API gives the number of results you put in the settings and the plugin chooses the result with the most likes:
after

P.S:
These changes make a huge difference!

### DIFF
--- a/src/plugins/urbanDictionary/index.ts
+++ b/src/plugins/urbanDictionary/index.ts
@@ -18,14 +18,22 @@
 
 import { ApplicationCommandOptionType, sendBotMessage } from "@api/Commands";
 import { ApplicationCommandInputType } from "@api/Commands/types";
+import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
 
 export default definePlugin({
     name: "UrbanDictionary",
     description: "Search for a word on Urban Dictionary via /urban slash command",
     authors: [Devs.jewdev],
     dependencies: ["CommandsAPI"],
+    options: {
+        resultsAmount: {
+            type: OptionType.NUMBER,
+            description: "The amount of results you want to get (+ Better results, - Slower)",
+            default: 10
+        }
+    },
     commands: [
         {
             name: "urban",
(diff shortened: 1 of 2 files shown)


src/plugins/urbanDictionary/index.ts
```diff
@@ -41,12 +49,18 @@ export default definePlugin({
             ],
             execute: async (args, ctx) => {
                 try {
-                    const query = encodeURIComponent(args[0].value);
-                    const { list: [definition] } = await (await fetch(`https://api.urbandictionary.com/v0/define?term=${query}`)).json();
+                    const query: string = encodeURIComponent(args[0].value);
+                    const resultsAmount: number = Settings.plugins.UrbanDictionary.resultsAmount || 10;
+                    const response: Response = await fetch(`https://api.urbandictionary.com/v0/define?term=${query}&per_page=${resultsAmount}`);
+                    const { list } = await response.json();
```
lewisakura, line 55:
```suggestion
                    const { list } = await fetch(`https://api.urbandictionary.com/v0/define?term=${query}&per_page=${resultsAmount}`).then(response => response.json());
```
Feels slightly redundant to bother defining response if you're going to immediately destructure it after.

src/plugins/urbanDictionary/index.ts
```diff
@@ -18,14 +18,22 @@
...
 
 export default definePlugin({
     name: "UrbanDictionary",
     description: "Search for a word on Urban Dictionary via /urban slash command",
     authors: [Devs.jewdev],
     dependencies: ["CommandsAPI"],
+    options: {
+        resultsAmount: {
+            type: OptionType.NUMBER,
+            description: "The amount of results you want to get (+ Better results, - Slower)",
```
lewisakura, line 33: The bit after the description doesn't really make sense. Are you trying to say pros/cons? If so, it'd be better to reword it into a sentence in brackets rather than shorthanding it:
```suggestion
            description: "The amount of results you want to get (more gives better results, but is slower)",
```

src/plugins/urbanDictionary/index.ts
```diff
@@ -18,14 +18,22 @@
...
+            default: 10
+        }
+    },
```
lewisakura, line 36: Use `definePluginSettings`. The `options` API is deprecated.

src/plugins/urbanDictionary/index.ts
```diff
@@ -41,12 +49,18 @@ export default definePlugin({
             ],
             execute: async (args, ctx) => {
                 try {
-                    const query = encodeURIComponent(args[0].value);
-                    const { list: [definition] } = await (await fetch(`https://api.urbandictionary.com/v0/define?term=${query}`)).json();
+                    const query: string = encodeURIComponent(args[0].value);
+                    const resultsAmount: number = Settings.plugins.UrbanDictionary.resultsAmount || 10;
```
lewisakura, line 53: Switch to your local settings object after switching to `definePluginSettings`. Also, you shouldn't need to do `|| 10` if you have a default already set.

lewisakura (changes requested):
Since you're updating the plugin, can you also add a `README.md` to it as well?

lewisakura (approved):

PR #2081, opened 2023-12-30
MessageLogger: Add a whitelist add/remove option to the user context menu

nakoyasha:
PowerToys CropAndLock_apxzkrRQOZ PowerToys CropAndLock_cosb6zVVrz


nakoyasha:
oops just realized theres commits from an unrelated feature

PR #2083, opened 2023-12-31
Add HTTPHook plugin

lockieluke:
This plugin allows users to control their Discord client via HTTP endpoints, currently there are two endpoints:

`/start-vc`: used with `userId` query param to start a call with another user
`/end-vc`: used to end an ongoing voice call

No code is executed dynamically which means there's no way to hijack the client remotely, server only runs on localhost(port 1675 by default), security should be okay!

Use cases:
- [Raycast](https://www.raycast.com/) extension for controlling Discord(mute, unmute, joining a VC)
- Interoperability with Minecraft mods so they can control your Discord client too

### DIFF
--- a/src/plugins/HTTPHooks/index.ts
+++ b/src/plugins/HTTPHooks/index.ts
@@ -1,0 +1,71 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { getCurrentChannel, openPrivateChannel } from "@utils/discord";
+import definePlugin, { OptionType, PluginNative } from "@utils/types";
+
+const Native = VencordNative.pluginHelpers.HTTPHooks as PluginNative<typeof import("./native")>;
+
+const settings = definePluginSettings({
+    serverPort: {
+        type: OptionType.NUMBER,
+        description: "The port the HTTP server should listen on",
+        target: "DESKTOP",
+        default: 1675,
+        restartNeeded: true,
+        async onChange() {
+            await Native.stopServer();
+        }
+    }
+});
+
+export default definePlugin({
+    name: "HTTPHooks",
+    description: "Plugin that supports controlling Discord via HTTP endpoints",
+    authors: [{
+        id: 549866818269872138n,
+        name: "lockieluke3389"
+    }],
+    settings,
+    async start() {
+        if (!IS_DISCORD_DESKTOP) {
+            console.error("HTTPHooks is only available in the Desktop client");
+            return;
+        }
+        await Native.startServer(settings.store.serverPort);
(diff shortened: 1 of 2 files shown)


Vendicated:
[as i mentioned on discord previously](https://discord.com/channels/1015060230222131221/1015063227299811479/1190821261740355614), this plugin will not be merged because it's way too niche

i will still leave a few review comments because there are some serious issues

src/plugins/HTTPHooks/index.ts
```diff
@@ -0,0 +1,71 @@
...
+export default definePlugin({
+    name: "HTTPHooks",
+    description: "Plugin that supports controlling Discord via HTTP endpoints",
+    authors: [{
+        id: 549866818269872138n,
+        name: "lockieluke3389"
+    }],
+    settings,
+    async start() {
+        if (!IS_DISCORD_DESKTOP) {
```
Vendicated, line 35: instead of doing this, rename ur plugin folder to `*.desktop` and it will only be included on desktop

src/plugins/HTTPHooks/index.ts
```diff
@@ -0,0 +1,71 @@
...
+            console.error("HTTPHooks is only available in the Desktop client");
+            return;
+        }
+        await Native.startServer(settings.store.serverPort);
+        console.log("HTTPHooks enabled");
+
+        window.httphooks = {
```
Vendicated, line 42: don't pollute the window. you can access your plugin via window.Vencord.Plugins.plugins.PluginName

src/plugins/HTTPHooks/index.ts
```diff
@@ -0,0 +1,71 @@
...
+            startVC: (userId: string) => {
+                openPrivateChannel(userId);
+
+                const timer = setInterval(() => {
+                    const joinVC = document.querySelector("#app-mount > div.appAsidePanelWrapper__714a6 > div.notAppAsidePanel__9d124 > div.app_b1f720 > div > div.layers__1c917.layers_a23c37 > div > div > div > div > div.chat__52833 > section > div > div.toolbar__88c63 > div:nth-child(1)");
```
Vendicated, line 47: don't use dom. you can grab this func from webpack, for example via `findByProps("selectVoiceChannel")`

src/plugins/HTTPHooks/native.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        switch (pathname) {
+            case "/":
+                res.writeHead(200, { "Content-Type": "text/html" });
+                res.write("<h1>Vencord HTTP Hooks Plugin</h1>");
+                break;
+
+            case "/start-vc":
+                const userId = searchParams.get("userId");
+                await mainWindow.webContents.executeJavaScript(`window.httphooks.startVC("${userId}")`);
```
Vendicated, line 26: this is vulnerable to XSS. as i mentioned many times on discord, you must use JSON.stringify to ensure safety

PR #2087, opened 2024-01-02
feat(anonymisefilenames): add icon to toggle anonymising

rniii:
this PR supersedes #553

does basically the same thing:

image


Vendicated:
i hate how bad git is at determining renames, makes the diff sooo bad to look at

src/plugins/anonymiseFileNames/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    Random,
+    Consistent,
+    Timestamp,
+}
+
+const tarExtMatcher = /\.tar\.\w+$/;
+
+const settings = definePluginSettings({
+    anonymiseByDefault: {
+        description: "Whether to anonymise or keep original file name by default",
```
Vendicated, line 40: this is bad wording for a boolean option
```suggestion
        description: "Whether to anonymise file names by default",
```

Vendicated:
nitpicky (kinda): discord usually makes strike-throughs go from bottom left to top right. your icon's strike-through goes from bottom right to top left. it would be better to be consistent with discord

image
image

image

Vendicated:
pushed a somewhat dirty fix that just mirrors the strike-through svg with css. it looks fine to me, so this is probably okay? unless you have a better svg for it

rniii:
I'd just mirror the actual svg but honestly no idea how without throwing it in an editor and have it spit out an even bigger one
I got the crossed out one from the same source so yea

fawni:
> pushed a somewhat dirty fix that just mirrors the strike-through svg with css. it looks fine to me, so this is probably okay? unless you have a better svg for it

i think this is good enough :shipit:

PR #2088, opened 2024-01-02
feat(plugin): Fixed LastFMRichPresence albums reporting incorrectly

KontrollFreek:
For some reason LastFM decides to default to the last used album if it has no clue what it's supposed to be. Also did some grammar and capitalization stuff.

I had to make a new PR because I'm an idiot and deleted the main branch instead of just renaming it lmao


src/plugins/lastfm/index.tsx
```diff
@@ -234,7 +239,7 @@ export default definePlugin({
             // why does the json api have xml structure
             return {
                 name: trackData.name || "Unknown",
-                album: trackData.album["#text"],
+                album: trackData.album.mbid === "" ? (trackData.name || "Unknown") : trackData.album["#text"],
```
Vendicated, line 242: this doesn't seem right.. why use trackData.name here? isn't that the song name?

KontrollFreek: This is just in case there's no album, usually indicating the song is a single and is therefore its own album

KontrollFreek: i don't entirely remember but im fairly certain displaying nothing still displays something and i have no clue why honestly lmao

Vendicated:
yeah this still doesn't look right, displaying the track (instead of just nothing) makes little sense

furthermore you have introduced so many unrelated changes :(

thanks a lot for your contribution but i'll be rejecting this

PR #2090, opened 2024-01-03
New Plugin added: hideBlockedPeople

rhaym-tech:
This plugin completely hides the blocked people like they doesn't exist!
Before:
image
After:
image

PSA: this still an initial version, more improvements are coming soon

TODO:

- [x] Hide blocked members messages
- [ ] Hide blocked members completely from server's members list 
- [ ] Rename their mentions from `@username` to `@blocked`, and real tag can be viewed by hovering on it

### DIFF
--- a/src/plugins/hideBlockedPeople/index.ts
+++ b/src/plugins/hideBlockedPeople/index.ts
@@ -1,0 +1,34 @@
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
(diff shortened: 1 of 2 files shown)


Vendicated:
hi. this is already a plugin: https://vencord.dev/plugins/NoBlockedMessages

vencord also does not use raw dom manipulation, you should use patches

thanks a lot regardless

rhaym-tech:
ichigo sad

PR #2092, opened 2024-01-05
fix(toolbox): Check if plugin is enabled before adding it

### LINKED ISSUE #2079: [Bug] Toolbox actions are present even if plugin is disabled
### Discord Account

kyuuhachi

Toolbox actions are present in the Vencord Toolbox regardless of whether the owning plugin is enabled or not.

They should only be present if the plugin is enabled.

Define a tiny plugin with a toolbox action:
```js

export default definePlugin({
    name: "0_Example",
    description: "Showcases a bug with toolbox actions",
    authors: [{ id: 236588665420251137n, name: "Kyuuhachi" }],

    toolboxActions: {
        async "Example action"() { }
    },
});
```

Open the Vencord Toolbox in the top right corner. The action(s) from the plugin are still there, even if the plugin is disabled.

This happens regardless of whether the plugin is in `src/plugins` or `src/userplugins`.

### Errors

None.

### Request Agreement

EdVraz:
fixes #2079


Vendicated:
thanks but the way you check if the plugin is enabled is wrong. thus, superseded by https://github.com/Vendicated/Vencord/commit/3eada99ad623ae783a39eb2c8238ac66ed371132

PR #2094, opened 2024-01-06
Custom ScreenShare Resolution and FPS

KawaiianPizza:
[Discord_VP9.webm](https://github.com/Vendicated/Vencord/assets/22826626/26567f28-7330-4cf2-8fd9-e96111619ac3)

Didn't know if easter eggs were allowed so they're disabled by default

### DIFF
--- a/src/plugins/customScreenShare.desktop/goofs.tsx
+++ b/src/plugins/customScreenShare.desktop/goofs.tsx
@@ -1,0 +1,39 @@
+export function goofs() {
+    setTimeout(() => {
+        (document.querySelector('[class^="wordmarkWindows"]')?.firstElementChild as any).style.display = "none";
+        document.querySelector('[class^="wordmarkWindows"]')?.prepend(Object.assign(document.createElement("div"), {
+            style: `height: 16px;
+                        width: 51px;
+                        color: var(--text-muted);
+                        font-size: xx-small;
+                        text-align: center;
+                        margin-top: 5px;
+                        font-weight: 600;
+                        transform: scaleX(2.5);
+                        letter-spacing: -1px;`,
+            innerText: "Skype"
+        }
+        ));
+        document.onmousedown = e => {
+            const hit = Object.assign(document.createElement('div'), {
+                style: `left: ${e.clientX}px;
+                            top: ${e.clientY}px;
+                            color: white;
+                            text-align: center;
+                            position: absolute;
+                            z-index: 9999999;
+                            pointer-events: none;
+                            transform: translate(-50%, -50%);
+                            font-size: 2rem;`,
+                innerText: '×'
+            });
+            const hitSound = Object.assign(document.createElement('audio'), { src: 'https://cdn.discordapp.com/attachments/545600181744173068/1090841136458899516/hit.wav', autoplay: true, volume: 0.2 });
+            hitSound.onended = () => {
+                hit.removeChild(hitSound);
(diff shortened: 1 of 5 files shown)


Oi-jojo:
this is epic, thank you

Marocco2:
Is it possible to add a bitrate target slider? I wanna experimenting some video settings that won't make my stream too much blocky.

Thanks 👍

KawaiianPizza:
> Is it possible to add a bitrate target slider? I wanna experimenting some video settings that won't make my stream too much blocky.
> 
> Thanks 👍

I just messed around with the bitrate min, max, and target values. The target bitrate didn't seem to do anything even when streaming 4k footage. The stream was hovering around 4mbps regardless if the target was 5kbps or 20mbps. The min changed the stream to be crystal clear when set to >8mbps with the downside being it constantly using 8mbps even on a still frame. When changing the max bitrate too high, nothing would happen (because of the target bitrate staying around 4mbps regardless of changing it). Setting the max bitrate too low caused blockiness instead of blurriness. Setting the resolution to 480p, 720p, 1080p, 1440p, and Source all had the same bitrate results with the only difference being blurriness the lower in res I went (this is just from having to upscale 480p to 1080p on the viewers end, not bitrate related).

The trade off of quality comes in latency. When streaming at higher bitrates, the stream would get more and more delayed until the streamer could catch up to the bitrate demands, pausing the stream for a couple dozen seconds would allow the stream to get more in sync with the viewer again but slowly desync once you start playing again.

Marocco2:
> > Is it possible to add a bitrate target slider? I wanna experimenting some video settings that won't make my stream too much blocky.
> > 
> > Thanks 👍
> 
> I just messed around with the bitrate min, max, and target values. The target bitrate didn't seem to do anything even when streaming 4k footage. The stream was hovering around 4mbps regardless if the target was 5kbps or 20mbps. The min changed the stream to be crystal clear when set to >8mbps with the downside being it constantly using 8mbps even on a still frame. When changing the max bitrate too high, nothing would happen (because of the target bitrate staying around 4mbps regardless of changing it). Setting the max bitrate too low caused blockiness instead of blurriness. Setting the resolution to 480p, 720p, 1080p, 1440p, and Source all had the same bitrate results with the only difference being blurriness the lower in res I went (this is just from having to upscale 480p to 1080p on the viewers end, not bitrate related).
> 
> The trade off of quality comes in latency. When streaming at higher bitrates, the stream would get more and more delayed until the streamer could catch up to the bitrate demands, pausing the stream for a couple dozen seconds would allow the stream to get more in sync with the viewer again but slowly desync once you start playing again.
> 
> 

Thank you for your findings. I'll keep them in mind when I'm going to mess around them 👌

src/plugins/customScreenShare.desktop/goofs.tsx
```diff
@@ -0,0 +1,39 @@
...
+            const hitSound = Object.assign(document.createElement('audio'), { src: 'https://cdn.discordapp.com/attachments/545600181744173068/1090841136458899516/hit.wav', autoplay: true, volume: 0.15 });
+            hitSound.onended = () => {
+                hit.removeChild(hitSound);
+                document.body.removeChild(hit);
+            };
+            hit.appendChild(hitSound);
+            document.body.appendChild(hit);
+        };
+    }, 1000);
+}
```
Vendicated, line 39: wha?

PR #2096, opened 2024-01-09
feat: plugin to fix twitter embeds by using fxtwitter.com

arian81:


MrDiamondDog:
Can be done using TextReplace

arian81:
> Can be done using TextReplace

I didn't know about that plugin. I will close this pr.

PR #2098, opened 2024-01-10
roleColorEverywhere: thread role color (again)

AutumnVN:
why do code keep vanishing 😭

and why is there a patch that put \<span> into color props


Vendicated:
> and why is there a patch that put <span> into color props

why did this make me wheeze lmao

PR #2100, opened 2024-01-10
feat(plugin): Add coin game option to oneko

0xGingi:
Adds a simple coin collection game to oneko

I feel like this setting could be implemented better, so if anyone has any suggestions, please let me know


thororen1234:
you need to make it so the coins get deleted on stop as they just stay there when I tested this.

0xGingi:
> you need to make it so the coins get deleted on stop as they just stay there when I tested this.

Fixed

Vendicated:

thank you regardless!

PR #2101, opened 2024-01-10
feat(plugin): oneko fork plugin that adds coin collection

0xGingi:
Adds a plugin that is a fork of oneko to add coin collection

Not sure which is preferred, so have 2 PRs, one that is a individual plugin, and one that adds an option to the existing plugin


Vendicated:

thank you regardless!

Sqaaakoi:
> this is cool but i think oneko is perfect in its current form
> 
> thank you regardless!

couldn't this be an option for the people who still want it? or do you think this is just a bad idea entirely

Vendicated:
its a cool idea but i don't really want to first-party oneko because it's just unnecessary

PR #2103, opened 2024-01-12
ClearURLs: Add igsh

Bloofield:
Added "igsh" to the list, I believe Instagram may have replaced igshid with igsh, so it may not be necessary to have them both, but better safe than sorry.

image


Vendicated:
tbh editing this plugin every time there's a new param somewhere is not very sustainable.

we could redesign this plugin to where it fetches one or more blocklists from a remote place (optimally github), much similar to how extensions like ublock origin do it. we could make this configurable so people can make and distribute their own lists, but also provide our official blocklist

Issue #655, opened 2023-03-22
[Feature Request] Keybind API


Vendicated:
### Discord Account

the machine

### What is it that you'd like to see?

Currently if a plugin wants to offer keybinds, it has to do the logic itself and hardcode a specific keybind. It would be nice to have something that lets you define keybinds similar to settings:
- Name
- Description
- Default keybind

Then these keybinds would be presented somehow in the settings modal automagically and users can then rebind them

### Request Agreement

- [X] This is not a plugin request

ArjixWasTaken:
Working on it.

highjeans:
Bump on this, I am really looking forward to this being possible.

PR #2104, opened 2024-01-12
Add a delete modifier key in messageClickActions plugin

whos-evan:
Added a modifier for the messageClickActions plugin which allows users to change the modifier key for deleting messages.
Discord_IDK4dzjpYI

### DIFF
--- a/src/plugins/messageClickActions/index.ts
+++ b/src/plugins/messageClickActions/index.ts
@@ -24,15 +24,32 @@ import { findByPropsLazy } from "@webpack";
 import { FluxDispatcher, PermissionsBits, PermissionStore, UserStore } from "@webpack/common";
 
 let isDeletePressed = false;
-const keydown = (e: KeyboardEvent) => e.key === "Backspace" && (isDeletePressed = true);
-const keyup = (e: KeyboardEvent) => e.key === "Backspace" && (isDeletePressed = false);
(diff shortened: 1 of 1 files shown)


src/plugins/messageClickActions/index.ts
```diff
@@ -24,16 +24,33 @@ import { findByPropsLazy } from "@webpack";
...
     },
+    changeDeleteModifier: {
+        type: OptionType.SELECT,
+        description: "Change the modifier for delete on click",
+        default: Modifiers.BACKSPACE,
+        options: [
+            { label: "Backspace Key", value: Modifiers.BACKSPACE, default: true },
+            { label: "Shift Key", value: Modifiers.SHIFT },
+            { label: "Ctrl Key", value: Modifiers.CTRL },
+            { label: "Alt Key", value: Modifiers.ALT },
```
Vendicated, line 51: these have the potential to conflict with the other inbuilt "require modifier" setting. i also think this selection of modifiers leaves much to be desired. the other 3 modifiers hardly make any sense and users will definitely ask for more options

much nicer would be something like https://github.com/Vendicated/Vencord/issues/655

Vendicated:
in the current state this isn't mergeable and you haven't addressed the feedback comment

thus i'll be closing this for now, if you're still interested in finishing this up feel free to just respond

thank you either way!

PR #2105, opened 2024-01-13
feat: Update anonymiseFileNames add option to exclude archive files

cod3nym:
I think this is a nice to have feature, since I don't want to loose the name information for archive files.

One could also implement a list to exclude user chosen extensions.


Vendicated:
this is way too specific and there are way more archive extensions than just this

there is a toggle to individually disable it for specific files, why not just make use of that?

cod3nym:
I see, yeah might be to specific.

Vendicated:
i don't think adding some sort of automatic blacklist to the plugin is a bad idea though

PR #2107, opened 2024-01-14
Add search plugin

Jacob1Tm:
Very simple, but maybe someone other than me will find this useful

### DIFF
--- a/src/plugins/search/index.ts
+++ b/src/plugins/search/index.ts
@@ -1,0 +1,98 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2024 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { ApplicationCommandOptionType } from "@api/Commands";
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+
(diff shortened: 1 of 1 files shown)


src/plugins/search/index.ts
```diff
@@ -0,0 +1,98 @@
...
+            },
+            {
+                type: ApplicationCommandOptionType.STRING,
+                name: "Search query",
+                description: "What do you want to search?",
+                required: true
+            },
+        ],
+        execute(args) {
+            console.log(args);
```
lewisakura, line 75:
```suggestion
```

src/plugins/search/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    name: "Search",
+    authors: [Devs.JacobTm],
+    settings,
+    description: "Generates search links for various search engines.",
+    dependencies: ["CommandsAPI"],
+    commands: [{
+        name: "search",
+        description: "Generates search link for selected engine.",
+        options: [
+            {
```
MrDiamondDog, line 60: I would make this a setting rather than an argument

src/plugins/search/index.ts
```diff
@@ -0,0 +1,98 @@
...
+                type: ApplicationCommandOptionType.STRING,
+                name: "Search query",
+                description: "What do you want to search?",
+                required: true
+            },
+        ],
+        execute(args) {
+            console.log(args);
+            const query = args[1].value.replace(" ", "+");
```
lewisakura, line 76: this is a really naive way of doing this and is vulnerable to querystring injection. use encodeURIComponent instead:
```suggestion
            const query = encodeURIComponent(args[1].value);
```

src/plugins/search/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        name: "search",
+        description: "Generates search link for selected engine.",
+        options: [
+            {
+                type: ApplicationCommandOptionType.STRING,
+                name: "Search engine",
+                description: "Which search engine do you want to use?",
+                required: true,
+                choices: searchEngineChoices,
+            },
```
lewisakura, line 66: you should set a default or a setting like it seems like you were planning to do

lewisakura (changes requested):
despite that this most likely will not be accepted since the plugin itself is so simple (generating a link to then click, rather than actually giving the search results) i will give a review to assist in future plugin development

Jacob1Tm:
> despite that this will not be accepted since the plugin itself is so simple (generating a link to then click, rather than actually giving the search results) i will give a review to assist in future plugin development

Yea I didn't really expect this to be merged but thanks for review

PR #2113, opened 2024-01-15
feat(plugin): RedirectYouTube

katlyn:
This plugin replaces YouTube embeds with sn embed from an alternate third party frontend, such as Invidious or Piped.

This is more of an experiment than anything else, and I'm not really sure if the way that the `patchCspDirective` function is done is a way that fits well with the rest of vencord's codebase - in any case though, I decided that making a draft PR would be good to get feedback and potentially discuss other options.

I'm fairly hesitant about the currently hardcoded instances, perhaps the defaults could be updated every so often if needed. There is an API we could query as well but this is very much a quick implementation so I didn't bother with that yet.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -23,6 +23,14 @@ import { ensureSafePath, getSettings } from "./ipcMain";
 import { IS_VANILLA, THEMES_DIR } from "./utils/constants";
 import { installExt } from "./utils/extensions";
 
+const patchedCspDirectives: Record<string, Set<string>> = {};
+export const patchCspDirective = (directive: string, values: string[]) => {
+    patchedCspDirectives[directive] ??= new Set();
+    for (const value of values) {
+        patchedCspDirectives[directive].add(value);
+    }
+};
+
 if (IS_VESKTOP || !IS_VANILLA) {
     app.whenReady().then(() => {
         // Source Maps! Maybe there's a better way but since the renderer is executed
@@ -95,6 +103,11 @@ if (IS_VESKTOP || !IS_VANILLA) {
                 for (const directive of ["style-src", "connect-src", "img-src", "font-src", "media-src", "worker-src"]) {
                     csp[directive] ??= [];
                     csp[directive].push("*", "blob:", "data:", "vencord:", "'unsafe-inline'");
+                }
+
+                for (const directive of Object.keys(patchedCspDirectives)) {
+                    csp[directive] ??= [];
+                    csp[directive].push(...patchedCspDirectives[directive]);
                 }
 
                 // TODO: Restrict this to only imported packages with fixed version.
--- a/src/plugins/redirectYouTube.desktop/index.ts
+++ b/src/plugins/redirectYouTube.desktop/index.ts
@@ -1,0 +1,37 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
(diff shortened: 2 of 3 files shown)


src/main/index.ts
```diff
@@ -97,6 +105,11 @@ if (IS_VESKTOP || !IS_VANILLA) {
                     csp[directive].push("*", "blob:", "data:", "vencord:", "'unsafe-inline'");
                 }
 
+                for (const directive of Object.keys(patchedCspDirectives)) {
+                    csp[directive] ??= [];
+                    csp[directive].push(...patchedCspDirectives[directive]);
```
Vendicated, line 110: use Object.entries if you need both key and value

src/plugins/redirectYouTube.desktop/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import definePlugin, { OptionType } from "@utils/types";
+
+export default definePlugin({
+    name: "RedirectYouTube",
+    description: "Replace all YouTube embeds with an embed from an third-party YouTube frontend, such as Invidious or Piped.",
+    authors: [Devs.katlyn],
+    settings: definePluginSettings({
+        instances: {
+            type: OptionType.STRING,
+            description: "A comma-separated list of alternate YouTube frontends (including protocol such as https). If multiple frontends are specified, one will be chosen at random.",
```
Vendicated, line 18: what's the point of this? it only seems useful if it's a "try every instance in order until one doesn't error"

src/plugins/redirectYouTube.desktop/index.ts
```diff
@@ -0,0 +1,37 @@
...
+            restartNeeded: true,
+            default: "https://invidious.fdn.fr/, https://vid.puffyan.us/, https://invidious.nerdvpn.de/, https://invidious.projectsegfau.lt/, https://invidious.flokinet.to/"
+        }
+    }),
+    patches: [{
+        find: "case\"YouTube\":",
+        replacement: {
+            match: /(let{src:\i,autoMute:\i.{10,200}src:)(\i)/,
```
Vendicated, line 26: if you don't actually need to match it, just use a lookbehind
```suggestion
            match: /(?<=let{src:\i,autoMute:\i.{10,200}src:)(\i)/,
```

src/plugins/redirectYouTube.desktop/index.ts
katlyn: The thought was that if we provide any default instances, to randomize between them to spread load - since there isn't any official instance it wouldn't be very polite to dump a bunch of vencord users onto one specific instance - that said, we could also just not provide any default instances and make each user provide their own in order to use the plugin.

ashley0143:
didint i say I can make this and um yall just Said use text replace

ashley0143:
> > didint i say I can make this and um yall just Said use text replace
> 
> This replaces *all* embeds instead of just ones you send, so even ones sent by other people will show with a different embed.
> 
> 

ye ik but I would also change the embeds

MrDiamondDog:
> > > didint i say I can make this and um yall just Said use text replace
> > 
> > 
> > This replaces _all_ embeds instead of just ones you send, so even ones sent by other people will show with a different embed.
> 
> ye ik but I would also change the embeds

Then go make a plugin that does that

src/plugins/redirectYouTube.desktop/index.ts
Vendicated: how heavy is the load from people using your invidious? does it not do most work on the client side?

Vendicated: oh damn i just checked and it seems like it uses a lot of resources.. that's interesting i didnt even know

katlyn:
Closing this PR as #2116 covers the primary use case.

PR #2116, opened 2024-01-19
feat(plugin): FixYoutubeEmbeds - fix UMG blocked embeds

CoolElectronics:
Allows embedded youtube videos to be played when they're usually blocked.

I *believe* youtube uses window.location.ancestorOrigins to detect this, and it's unreasonably hard to overwrite it. Instead I just reload the frame which clears ancestorOrigins and shouldn't break randomly if the frontend updates.

### DIFF
--- a/src/plugins/fixYoutubeEmbeds.desktop/index.ts
+++ b/src/plugins/fixYoutubeEmbeds.desktop/index.ts
@@ -9,6 +9,6 @@ import definePlugin from "@utils/types";
 
 export default definePlugin({
     name: "FixYoutubeEmbeds",
-    description: "Fixes youtube embeds from being blocked",
+    description: "Bypasses youtube videos being blocked from display on Discord (for example by UMG)",
     authors: [Devs.coolelectronics]
 });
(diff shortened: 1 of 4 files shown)


CoolElectronics:
umg

src/plugins/fixYoutubeEmbeds.desktop/index.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "FixYoutubeEmbeds",
+    description: "Fixes youtube embeds from being blocked",
```
Vendicated, line 12: i think smth like that would be better
```suggestion
    description: "Bypasses youtube videos being blocked from display on Discord (for example by UMG)",
```

Vendicated:
please add a README.md to your plugin that explains in a bit more detail what this does and maybe show a before and after screenshot

CoolElectronics:
Added!

Vendicated:
thank you!

PR #2122, opened 2024-01-21
SearchFix plugin

Jaxx7594:
Fixes the annoying "We dropped the magnifying glass" error when using the search bar. I very rarely use typescript or javascript, so if there are any issues with this code, please tell me and I'll try to fix it.


src/plugins/searchFix/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "SearchFix",
+    description: "Fixes the annoying \"We dropped the magnifying glass!\" error. This fix isn't perfect, so you may have to reload the search bar to fix issues. Note: Discord only allows a max offset of 5000 (this is what causes the magnifying glass error). This means that you can only see precisely 125000 messages into the past, and 125000 messages into the future (when sorting by old). This plugin just jumps to the opposite sorting method to try get around Discord's restriction, but if there is a VERY large search result, and you try to view a message that is unobtainable with both methods of sorting, the plugin will simply show offset 0 (either newest or oldest message depending on the sorting method).",
```
sunnniee, line 23: keep the description in a single sentence, and then use the `settingsAboutComponent` plugin property to keep the additional info

```ts
export default definePlugin({
    name: "SearchFix",
    description: 'Fixes the annoying "We dropped the magnifying glass!" error.',
    settingsAboutComponent: () => "Discord only allows a max offset of 5000 (this is what causes the magnifying glass error). This means that ..."
```

src/plugins/searchFix/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        {
+            id: 901016640253227059n,
+            name: "Jaxx",
+        },
+    ],
+    patches: [
+        {
+            find: "function(e,t,n){\"use strict\";let i;n.r(t),n.d(t,{default:function(){return O}}),n(\"843762\")",
+            replacement: {
+                match: /(.)\.offset=null!==\((.)=(.)\.offset\)&&void 0!==(.)\?(.):0/i,
```
sunnniee, line 34: hardcoding variable names (& webpack module ids) inside the finder is very very bad. you should use something that wont change the next time discord decides to rebuild. here since you're just patching a store, you can use `.displayName="SearchStore";`

also you should use `\i` to match variable names, it's a custom token by vencord for that purpose

src/plugins/searchFix/index.ts
```diff
@@ -0,0 +1,53 @@
...
+                replace: (_, v, v1,query,v3,v4) => "" +
+                    `${query}.offset > 5000 && ((${query}.sort_order === "asc" ? ${query}.sort_order = "desc" : ${query}.sort_order = "asc"), ${query}.sort_by = "timestamp", ${query}.offset = ${query}.offset > 5000 - 5000 ? 0 : ${query}.offset - 5000),\n` +
+                    `${v}.offset = null !== (${v1} = ${query}.offset) && void 0 !== ${v3} ? ${v4} : 0`
```
sunnniee, line 37: you can add a function to your plugin that does this in a readable way, and then reference it within the patch by using `$self` (you can also make groups larger than a single variable, or use `$&` to reinsert the entire match, to not have to rewrite discord's code within the patch)

```ts
export default definePlugin({
  patches: [
  // ...
  replace: (_, myval) => `foo;$self.doSomething(${myval});bar`
  ],
  doSomething(myval) {
      // ...
  }
```

sunnniee (changes requested):

src/plugins/searchFix/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "SearchFix",
+    description: "Fixes the annoying \"We dropped the magnifying glass!\" error. This fix isn't perfect, so you may have to reload the search bar to fix issues. Note: Discord only allows a max offset of 5000 (this is what causes the magnifying glass error). This means that you can only see precisely 125000 messages into the past, and 125000 messages into the future (when sorting by old). This plugin just jumps to the opposite sorting method to try get around Discord's restriction, but if there is a VERY large search result, and you try to view a message that is unobtainable with both methods of sorting, the plugin will simply show offset 0 (either newest or oldest message depending on the sorting method).",
+    authors: [
```
MrDiamondDog, line 24: Add yourself to devs

MrDiamondDog (commented):
Add a readme.md to your plugin, providing a longer description (like the one you have now) and possibly some images.

Jaxx7594:
Thanks for all the feedback! I'll try to fix all this tomorrow

Jaxx7594:
Should be much more readable, and now has a readme showing before and after gifs

src/plugins/searchFix/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        {
+            find: ".displayName=\"SearchStore\";",
+            replacement: {
+                match: /(\i)\.offset=null!==\((\i)=(\i)\.offset\)&&void 0!==(\i)\?(\i):0/i,
+                replace: (_, v, v1,query,v3,v4) => `$self.main(${query}), ${v}.offset = null !== (${v1} = ${query}.offset) && void 0 !== ${v3} ? ${v4} : 0`
+            }
+        }
+    ],
+    main(query){
+        if (query.offset > 5000) {
```
MrDiamondDog, line 36: try inverting this if statement so you dont nest as much
```suggestion
        if (query.offset <= 5000) return;
```

src/plugins/searchFix/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+            if (query.sort_order === "asc") {
+                query.sort_order = "desc";
+            } else {
+                query.sort_order = "asc";
+            }
```
MrDiamondDog, line 41: This if block would be better using ternary, not so much the other one
```suggestion
            query.sort_order === "asc" ? "desc" : "asc";
```

src/plugins/searchFix/index.tsx
Jaxx7594: Inverting it broke everything for some reason, and its too small of an issue for me to bother debugging it. I did the ternary thing though

Jaxx7594: nvm ternary broke it too.. guess ive gotta do some debugging

Jaxx7594: oml im an idiot, forgot to assign the var lmao

MrDiamondDog: Yeah I forgot to put that lol

Jaxx7594:
bump

Vendicated:
i finally understood what this is supposed to solve but now i don't exactly understand the use case. in what scenario would you go this far back in history and not just sort by old instead?

you might be better off asking discord to instead remove this limitation

Jaxx7594:
@Vendicated I can't remember exactly why I made this, but its probably because its just a pointless limitation that only annoys users, and its pretty annoying having to calculate (for example) what page 6324 while sorting by newest would be when sorting by oldest. As for why you would go so far back in history, for example, a user may be looking for a specific message on a specific date

Vendicated:
yeaahh i really cannot find a use case for this... thank you for the contribution anyway!

PR #2128, opened 2024-01-24
fixYoutubeEmbeds - Multilanguage Support

kbvivi:
Implementation of the fix from earlier expects only English errors. This pull request solves the issue and checks if video error contains link to the YouTube video thus meaning the video is not available to watch in embed.


Vendicated:
thanks!

PR #2132, opened 2024-01-27
feat(plugin) allCallTimers: adds timers to all users in guild voice calls

Max-Herbold:
Builds on the idea of the callTimer plugin but adds timers to all users in guild voice calls.

---
### VIEW

image
image
image

### Settings
image

---

This does not modify any network behavior of Discord it simply grabs updates as they are naturally informed to the client. There is a toggle, `Watch Large Guilds` to subscribe to all guilds (large guilds are not subscribed automatically), when this is disabled large guilds are usually updated roughly every 7-15 minutes (through passive updates). Since a user naturally subscribes to a guild when they interact with it, if the toggle `Watch Large Guilds` is enabled, on client load, it acts as if all guilds have been interacted with by a user.

Please let me know if you have any questions or suggestions for improvement.

fixes https://github.com/Max-Herbold/AllCallTimersDiscordPlugin/issues/22


src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export default definePlugin({
+    name: "AllCallTimeCounter",
+    description: "Add call timer to all users in a server voice channel.",
+    authors: [
+        {
+            id: 0n,
+            name: "Max",
+        },
+    ],
```
D3SOX, line 17: Add yourself to Devs

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+        for (let guildId in guilds) {
+            let guild = guilds[guildId];
+            for (let userId in guild) {
+                users.push(userId);
+            }
+        }
+        return users;
+    },
+
+    update_listings() {
```
D3SOX, line 42: Don't use snake case, please use ESLint

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    } else {
+                        // user is not in the users object
+                        this.users[userId] = {
+                            "channelId": channelId,
+                            "join_time": Date.now()
+                        };
+                    }
+                }
+            }
+        }
```
D3SOX, line 77: You could use array functions like `.map` here to make the code prettier

D3SOX (changes requested):
I think you could use IPC events instead of the interval to handle users joining/switching channels.
As per looks I think it looks kinda bad as the red color doesn't fit very well and it overlaps other things
Your plugin is missing a `README.md`

The data is lost on restart because the plugin is restarted btw :)

src/plugins/allCallTimers/index.tsx
Max-Herbold: thx, old habit

src/plugins/allCallTimers/index.tsx
Max-Herbold: brackets was the linter suggestion seemed less readable to me but listening to the linter removed the underline

Max-Herbold:
> I think you could use IPC events instead of the interval to handle users joining/switching channels

How it is at the moment works; given it's never necessary to update this more than once a second. I agree, where could I find some example how to do/use this? I have 0 experience with this.

>As per looks I think it looks kinda bad as the red color doesn't fit very well and it overlaps other things

I also agree. It's less than ideal to move elements around; there is very limited space. When it is inline, a long username will cause ellipsed usernames. Under the username makes it too confusing about who the timer belongs to. there is not enough space to the left. Where it is now is conditionally free when there are no rendered icons. The icons can still be seen clearly. As for the red, again I agree but if the element doesn't fit cleanly why not go all out?

D3SOX (changes requested):
Some more suggestions:

For the looks maybe use a clock icon and then put the time in the tooltip if it's okay to be only visible on hover

For the IPC implementation just look into other plugins. For example
https://github.com/Vendicated/Vencord/blob/620c127b58e94ce4bf0b78d7e55fbce8e4f376e2/src/plugins/vcNarrator/index.tsx#L160-L188

The Plugin name doesn't need to be camelCase (also see other plugins for reference)

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            return;
+        }
+        const startTime = user.joinTime;
+        return <ErrorBoundary>
+            <this.Timer time={startTime} />
+        </ErrorBoundary>;
+    },
+
+    Timer({ time }: { time: number; }) {
+        const timer = useTimer({});
```
D3SOX, line 114: `timer` is unused here

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+        // start a timeout that runs every second and calls updateListings
+        this.timeout = setInterval(() => this.updateListings(), 1000);
+    },
+
+    stop() {
+        // clear the timeout
+        clearInterval(this.timeout);
+    },
+
+    showInjection(property) {
```
D3SOX, line 90: add type

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            replacement: [
+                {
+                    match: /(render\(\)\{.+\}\),children:)\[(.+renderName\(\),)/,
+                    replace: "$&$self.showInjection(this),"
+                }
+            ]
+        }
+    ],
+
+    allUsers(guilds) {
```
D3SOX, line 26: add type

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+                        channelId: channelId,
+                        joinTime: Date.now()
+                    };
+                }
+            }
+        }
+    },
+
+    start() {
+        this.VoiceStateStore = findStore("VoiceStateStore");
```
D3SOX, line 77: Not sure if it's a problem to have the store in `this`. Other plugins do it differently. Another contributor should probably review this.

D3SOX (changes requested):

src/plugins/allCallTimers/index.tsx
Max-Herbold: it calls itself? I don't understand this inbuilt react timer but it works. Without it, the timer won't refresh

D3SOX: Oh ok, I'm sure there's a better way to make the component re-render but I can't debug it right now

Max-Herbold:
> For the looks maybe use a clock icon and then put the time in the tooltip if it's okay to be only visible on hover

image (mouse is hovering on the user's clock icon)

Also implemented with a toggle switch setting (because I like not having to move my mouse to see the times). The default setting is with the icon + tooltip.

src/plugins/allCallTimers/index.tsx
Max-Herbold: Looks like the standard is to lazily find at the top of the file - switched to this.

D3SOX:
Are you working on the IPC implementation? Or should I do it?

Max-Herbold:
> Are you working on the IPC implementation? Or should I do it? And a quick heads-up the `README.md` is still missing
I was planning to have a look tomorrow but I would love to see what you come up with!

D3SOX:
Done: https://github.com/Max-Herbold/Vencord/pull/1

Max-Herbold:
It occurred to me previously but I only realised it was an issue now. UserState does not dispatch updates for users in a guild that has not been opened (literally viewed on the client side). This plays into the issue I was trying to fix with https://github.com/Vendicated/Vencord/pull/2132/commits/5ebc2c8819b3fa489ce5e3bcaaa8083d00d65d33. I guess on the first open of a guild discord client bulk updates all users in the guide and then updates them when something actually happens. Essentially; if a guild hasn't been opened and a user moves/joins this cannot be tracked with UserState.

D3SOX:
Oh yes good catch haven't noticed that

D3SOX:
> Potentially also there is an issue after a certain period of time of a guild being unopened events stop being dispatched but this is only an intuitive guess that should be checked

I don't think so as this worked for me
image

Max-Herbold:
> Oh yes good catch haven't noticed that. Analyze the format when the client is loaded I don't think the way you do it right now is necessary.

True. I think that change (https://github.com/Vendicated/Vencord/commit/5ebc2c8819b3fa489ce5e3bcaaa8083d00d65d33) simply made the issue more complicated

D3SOX:
Hm but it does seem to work for me

Max-Herbold:
> Hm but it does seem to work for me

There is the case where a user joins/moves channel that hasn't been opened and they end up without a timer (in the `userJoinTimes`) and therefor no visual timer.

Max-Herbold:
reverted the change to use VoiceStateStore but kept everything else.

https://github.com/Max-Herbold/Vencord/pull/2

D3SOX:
> VOICE_STATE_UPDATES does not dispatch updates for users in a guild that has not been opened (literally viewed on the client side)

> There is the case where a user joins/moves channel that hasn't been opened and they end up without a time (in the `userJoinTimes`) and therefor no visual timer.

I can't reproduce this issue with the state of my PR. Can you give me more detailed instructions on how to reproduce it?
I have two Discord accounts open. With the Vencord one I'm just at the friends tab and when I log the voice state updates and search for my user ID I can clearly see the update looking fine on initial load (without `oldChannelId`) and when switching channels

Max-Herbold:
> I can't reproduce this issue with the state of my PR. Can you give me more detailed instructions on how to reproduce it? I have two Discord accounts open. With the Vencord one I'm just at the friends tab and when I log the voice state updates and search for my user ID I can clearly see the update looking fine on initial load (without `oldChannelId`) and when switching channels

After restarting Discord directly into DM's (somewhere other than a guild); I had added a console debug in the IPC only really to look at when it occurrs. Without opening any guilds it never occurred (other than the initial startup). Although this test case was probably exclusive to 10+ servers with 100+ users. After opening some of the guilds and returning to DM's updates could then be seen flowing.

Max-Herbold:
soz for force - wrong branch

D3SOX:
I think I definitely know why it happens and can fix the IPC implementation. Are you on Discord btw?

Max-Herbold:
> I think I definitely know why it happens and can fix the IPC implementation. Are you on Discord btw?

yeah ill msg u

src/plugins/allCallTimers/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    settings,
+
+    patches: [
+        {
+            find: "renderPrioritySpeaker",
+            replacement: [
+                {
+                    match: /(render\(\)\{.+\}\),children:)\[(.+renderName\(\),)/,
+                    replace: "$&$self.showInjection(this),"
```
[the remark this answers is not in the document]
Max-Herbold: fixed with https://github.com/Vendicated/Vencord/pull/2132/commits/9cc5c09fba604cc9d3515d9ca2159590a8b8c02c

D3SOX (approved):

Max-Herbold:
@Vendicated please review!

Max-Herbold:
This is ready to merge.

Inbestigator:
Damn, and here was me thinking that **I** committed a lot

src/plugins/allCallTimers/timerText.tsx
```diff
@@ -0,0 +1,15 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+export function TimerText({ text, className }: Readonly<{ text: string; className: string; }>) {
+    return <div className={`timeCounter ${className}`} style={{
+        marginTop: -6,  // this margin value doesn't change the default size of the user container
```
Loukious, line 9: The test workflow will fail because of this comment and the one below it as well.

Loukious (changes requested):

Max-Herbold (commented):
Do these changes address the issue?

src/plugins/allCallTimers/timerText.tsx
Max-Herbold: Does [de5c4d0](https://github.com/Vendicated/Vencord/pull/2132/commits/de5c4d04477af405bc52ead52b8f7324bf1423cd) address this correctly?

Loukious: > Does [de5c4d0](https://github.com/Vendicated/Vencord/pull/2132/commits/de5c4d04477af405bc52ead52b8f7324bf1423cd) address this correctly?

Yes, thanks.

Loukious (approved):

Loukious (approved):

Max-Herbold:
This is ready to be merged.

kittygirlyy (commented):
meow

Max-Herbold:
> did i do something wrong because the build still doesn't work
Hi, @6faw it looks as though you have copied across the folder `AllCallTimers`. Unfortunately, this is not sufficient because some changes (refactoring/improvements/additions) were made to internal helper functions. Still waiting for some review/merge from @Vendicated.

Max-Herbold:
@Vendicated still awaiting some review.

PR #2135, opened 2024-01-28
FollowUser plugin

D3SOX:
This is the best new plugin idea I could come up with :o


D3SOX:
I don't get whatever VSCode did with the merge here :sob: Gonna fix my JetBrains and force push from there.

Vendicated:
i get the idea but this is borderline selfbotting and kinda creepy, so i think i'll have to deny this one

thanks for the effort anyway!

D3SOX:
This was meant for staying together when in a big server, not for being creepy but I get that

PR #2137, opened 2024-01-31
Added way to intercept http requests

Pixeluted:
Hello, I was working on a plugin that needed me to intercept HTTP requests so I decided to create a native way in vencord to do that.

The API allows you to change what requests are being sent to server and what responses are going to given back to the client, you can add listeners before the request is sent or after request is sent and we got response.

I tested it and everything seems to be working (unless you break your whole client yourself in a plugin)

If you have any questions, please let me know.

### DIFF
--- a/src/api/NetworkHooking.ts
+++ b/src/api/NetworkHooking.ts
@@ -1,0 +1,160 @@
+import { Logger } from "@utils/Logger";
+import { RestAPI } from "@webpack/common";
+
+export type HttpMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE";
+
+/**
+ * You can return a modified query in this listener, which will be sent to the server instead of the original query
+ * (If you return null or undefined, the original query will be sent)
+ * @param url The url that is about to be requested
+ * @param query The query that is about to be sent
+ * @returns The modified query
+ */
+export type NetworkPreSendListener = (url: string, query: string | Array<any>, method: HttpMethod) => any;
+
+/**
+ * Whatever you will return in this listener will be sent back to the client instead of the server response
+ * (If you return null or undefined, the original response will be sent)
+ * @param url The url that was requested
+ * @param response The response from the server
+ * @param ok Whether the request was successful
+ * @param status The status code of the response
+ * @returns The modified response
+ */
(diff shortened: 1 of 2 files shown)


Nuckyz:
I don't know what is Ven opinion on this but it should use regex patches anyways, instead of monkey patching.

However your plugin that needs to modify requests can likely just patch that one request/response which you need, outside of the RestAPI src code

Vendicated:
i don't think this is a good idea. it seems kind of like an XY problem.

could you describe a use case for this? 

in most cases you should just be able to patch the function that does your request to change whatever you need. patching it like this worsens performance and encourages less experienced devs to just patch requests instead of doing it properly and patching the correct code. it also seems kind of dangerous because it's very easy to mess up requests and possibly get your account locked for spam

Pixeluted:
> i don't think this is a good idea. it seems kind of like an XY problem.
> 
> could you describe a use case for this?
> 
> in most cases you should just be able to patch the function that does your request to change whatever you need. patching it like this worsens performance and encourages less experienced devs to just patch requests instead of doing it properly and patching the correct code. it also seems kind of dangerous because it's very easy to mess up requests and possibly get your account locked for spam

Hello you are right about it encouraging inexperienced devs to just patch requests but this is comes from the fact vencord has little documentation, so this feature is result of me having no idea how to find the code to patch and then patch it. You might argue that I can look at existing plugins but that only looks like a weird mess of regex that just does it things magically because there is no explanation on how to find and patch properly.

I've made this feature because in my plugin I hooked a lot of API requests to get my plugin to function how I want, simply because there is no good documentation stating how to find code to patch and how to patch properly and techniques to do so.

Reading through code did taught me a lot about using vencord but definitely not how to properly patch, you might bring up community in discord to help me but either there is no one to respond because I'm asking such "basic question" or because the person responding doesn't bother to explain it further or until I/or someone else understands it.

TheKodeToad:
does this help you: https://github.com/Vendicated/Vencord/blob/main/docs/2_PLUGINS.md ?

Pixeluted:
> does this help you: https://github.com/Vendicated/Vencord/blob/main/docs/2_PLUGINS.md ?

I already read that before typing that

PR #2142, opened 2024-02-02
fixed outdated copyright

pompompur1nn:
i updated all copyright years in the code to 2024. 😊


Vendicated:
there is no such thing as outdated copyright. updating the year is meaningless

PR #2143, opened 2024-02-03
feat(plugin): Added returnToRequests

furSUDO:
This is a straightforward plugin that provides an option in the DM ContextMenu to close the DM and send it back to the Message Requests Queue. This is useful for quickly answering questions and then closing the DM, so that it does not reappear in the sidebar.

image

### DIFF
--- a/src/plugins/returnToRequests/index.tsx
+++ b/src/plugins/returnToRequests/index.tsx
@@ -57,7 +57,7 @@ export function removeContextMenus() {
 
 export default definePlugin({
     name: "ReturnToRequests",
-    description: "Right-click any DM and select 'Move to Message Requests' to return DMs to the Message Requests queue. This works similarly to ignoring incoming message requests, where any new messages from the sender will appear in the Message Requests queue. Will appear in the Message Requests queue.",
+    description: "Right-click any DM and select 'Move to Message Requests' to return DMs to the Message Requests queue. This works similarly to ignoring incoming message requests, where any new messages from the sender will appear in Message Requests.",
     authors: [Devs.SUDO],
 
     start: addContextMenus,
--- a/src/plugins/returnToRequests/index.tsx
+++ b/src/plugins/returnToRequests/index.tsx
@@ -19,7 +19,7 @@
 import { Menu, RestAPI } from "@webpack/common";
 import definePlugin from "@utils/types";
 import { Devs } from "@utils/constants";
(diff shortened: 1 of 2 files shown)


furSUDO:
Overlooked disabling the plugin. Will fix.

src/plugins/returnToRequests/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Menu, RestAPI } from "@webpack/common";
+import definePlugin from "@utils/types";
+import { Devs } from "@utils/constants";
+import { NavContextMenuPatchCallback, addContextMenuPatch, findGroupChildrenByChildId, removeContextMenuPatch } from "@api/ContextMenu";
+
+async function returnDMtoRequests(id: string) {
+    await RestAPI.delete({
```
Vendicated, line 25: this seems to just be the same as closing the dm

furSUDO: The difference is that when the sender sends a new message, it will appear in "Message Requests" rather than in the sidebar with the rest of your DMs.

Closing the DM seemingly just hides the channel.

Closing a DM: `DELETE https://discord.com/api/v9/channels/channelId`
"Ignoring" the DM: `DELETE https://discord.com/api/v9/channels/channelId/recipients/@me`

TheKodeToad:
- Why does the description say it adds the close dm option?
- Perhaps it should be "Move to Message Requests" or something since it seems like it works on all dms, not just those which originally appeared in requests

furSUDO:
- This, to me, was the best way to convey the plugin's behavior of first closing the DM and then returning it to the Message Requests queue. Otherwise, it could be interpreted that the DM will just be moved to the Request queue and still be visible when this is not the case. I will instead elaborate on that in the description to clarify the plugin's behavior. 
- It does indeed work on any DM channel, Thanks for the suggestion. (Changed text to 'Move to Message Requests')

dolfies:
Unfortunately, DMs do not truly return to the message requests queue. When the user DMs you again, once you reload your client, the DM will be back in your private channel list.

TheKodeToad:
Huh, it seems to be using a direct request. That's weird!

furSUDO:
> When the user DMs you again, once you reload your client, the DM will be back in your private channel list.

That is really unfortunate. Will close PR, as that makes this pointless. Apologies for not testing this further before making a PR.

PR #2145, opened 2024-02-03
feat(plugin): Added CommandPalette

ethan-davies:
This plugin allows a user to perform certain actions through a command palette (accessible via `Control + Shift + P`). This plugin can be easily modified to add further functionality. The purpose of this plugin is to quickly be able to perform actions with ease instead of needing to go through multiple menus or click multiple buttons. Also, in the case that this is merged, it will allow for future plugins (or current ones) to implement some of their own commands related to their plugin into the menu. In terms of adding separate functionality you can use the `openSimpleTextInput()` which will ask for a text input or the `openMultipleChoice()` which as the name implies will open a multiple choice menu with your selected options.


ethan-davies:
I'm going to improve the documentation and remove the `onKeyDown={handleKeyDown}` which I have no clue why I added that there xD

docs/3_COMMAND_PALETTE.md
```diff
@@ -0,0 +1,49 @@
+# Command Palette Actions Guide
```
MrDiamondDog, line 1: I wouldn't put this in the top level docs folder, as it's specific to your plugin. Put this file in your plugin's folder, and link to it in your readme

src/plugins/commandPalette/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        default: true
+    }
+});
+
+function recordKeybind() {
+    let keys: Set<string> = new Set();
+    let keyLists: string[][] = [];
+
+    const updateComponentText = () => {
+        const button = document.querySelector(`.${cl("key-recorder-button")}`);
```
MrDiamondDog, line 43: Don't use dom manip, this can easily be done in the component

PR #2149, opened 2024-02-04
feat(api): FormattedMessage

ryan-0324:
Allows the creation and usage of `FormattedMessage` objects.

Example:
```ts
    // logs "There is 1 object."
    const OBJECT_COUNT = new FormattedMessage("There {count, plural, =0 {are no objects} one {is # object} other {are # objects}}.");
    console.log(OBJECT_COUNT.format({ count: 1 });
```


ryan-0324:
Closing in favor of #2446

PR #2157, opened 2024-02-08
fixYoutubeEmbeds - Support HTTP href URLs

Roki100:
So yeah, i use another locale than english and besides last fix it still didn't work on an example video i tried it with, i checked why and it appears that the href of "watch on youtube" was http, so this pr adds support for that too
image


Roki100:
nvm, #2155

PR #2167, opened 2024-02-11
feat(plugin): Add StagingBuild plugin

happyendermangit:
This plugin changes the client build to "staging", it gives access to experiments removes tracking and more...


Roki100:
i havent noticed any interesting difference besides this
image
couldnt find any info online confirming your claims either, also experiments and tracing can both be added/blocked via existing plugins, what's the real benefit of a staging build?

PR #2169, opened 2024-02-12
feat(translate): Auto Translate Received Messages

MrDiamondDog:
Saw this requested a lot so I made it


lewisakura:
this will get you rate limited from the translate api, which is why it isnt in the main mod; it was a conscious choice made to prevent api abuse

Vendicated:
i think this feature is fine to have. i tested the rate limit in the past by sending thousands of requests to google and there doesn't seem to be any sort of rate limit

src/plugins/translate/TranslateIcon.tsx
```diff
@@ -81,7 +81,7 @@ export const TranslateChatBarIcon: ChatBarButton = ({ isMainChat }) => {
                 "aria-haspopup": "dialog"
             }}
         >
-            <TranslateIcon className={cl({ "auto-translate": autoTranslate, "chat-button": true })} />
+            <TranslateIcon className={cl({ "auto-translate": autoTranslate || autoTranslateReceived, "chat-button": true })} />
```
Vendicated, line 84: i think it might be better to use different colours / icon modifiers to signalise the state

src/plugins/translate/index.tsx
```diff
@@ -96,4 +97,17 @@ export default definePlugin({
...
         removeAccessory("vc-translation");
     },
+
+    flux: {
+        MESSAGE_CREATE: async event => {
+            if (!settings.store.autoTranslateReceived) return;
+            if (event.channelId !== Util.getCurrentChannel().id) return;
+            if (event.message.author.id === UserStore.getCurrentUser().id) return;
+
+            console.log(event.message);
```
Vendicated, line 107:
```suggestion
```

src/plugins/translate/index.tsx
```diff
@@ -25,7 +25,8 @@ import { addPreSendListener, removePreSendListener } from "@api/MessageEvents";
 import { addButton, removeButton } from "@api/MessagePopover";
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
-import { ChannelStore, Menu } from "@webpack/common";
+import { ChannelStore, Menu, UserStore } from "@webpack/common";
+import { Util } from "Vencord";
```
Vendicated, line 29: no

Vendicated (changes requested):
i think the current implementation is kinda meh. it would be better to keep separate state per channel. most people will only want to translate specific channels

Vendicated:
i think the best solution to making it configurable per server would be smth like this
image

MrDiamondDog:
Everything you mentioned should now be implemented. I'm not sure what I should do to show when you are translating received messages though.

src/components/PluginSettings/components/SettingSliderComponent.tsx
```diff
@@ -62,6 +63,7 @@ export function SettingSliderComponent({ option, pluginSettings, definedSettings
                 stickToMarkers={option.stickToMarkers ?? true}
                 {...option.componentProps}
             />
+            <Forms.FormText style={{ color: "var(--text-muted)" }}>{option.description}</Forms.FormText>
```
Vendicated, line 66: please use a css class instead of inline style

MrDiamondDog:
hold on i messed up this branch

PR #2170, opened 2024-02-12
feat(plugin): NoBlockedUsers

rushiiMachine:
Rewrite of my own old plugin `NoBlockedMessages` to support removing users from UI as much as possible.

New features:
- Hide blocked users from the currently typing list in chat.
- Hide blocked authors of referenced messages in replies.
- Hide blocked users from chat autocomplete

TODO:
- [ ] Hide from member list


pwall2222:
This PR is stale, but I tried it rebasing it into dev and it works well!
Although now it would also have to change the [plugin name here](https://github.com/Vendicated/Vencord/blob/40c5ade82d2c1a5471a0d053299342acfad76cc5/src/plugins/quickReply/index.ts#L143).

(I'm also just bumping this because it's great QoL upgrade)

rushiiMachine:
IIRC I need to update the patches, it broke a long time ago

pwall2222:
> IIRC I need to update the patches, it broke a long time ago

I rebased it into dev this morning, and it still worked, so no idea.

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+            restartNeeded: true,
+        },
+        hideReferencedAuthor: {
+            description: "Hide blocked authors of referenced messages in replies.",
+            type: OptionType.BOOLEAN,
+            default: false,
+            restartNeeded: true,
+        },
+        hideFromMemberList: {
+            description: "Hide blocked users from the members list.",
```
Vendicated, line 57: no periods
```suggestion
            description: "Hide blocked users from the members list",
```

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+    },
+);
+
+export default definePlugin({
+    name: "NoBlockedUsers",
+    description: "Hides blocked users and their messages from everywhere possible",
+    authors: [Devs.rushii, Devs.Samu],
+    settings,
+    patches: [
+        // Based on canary bd584cf8aedd674f8d65c316892cf5054b7b6fd5
```
Vendicated, line 71: meaningless
```suggestion
```

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+            // Hide blocked message groups from non-DM channels
+            find: ".MESSAGE_GROUP_BLOCKED||",
+            replacement: {
+                match: /(\i)\.type===(?:\i\.)+MESSAGE_GROUP_SPAMMER\)\{/,
+                replace: "$& console.log(arguments);if($1.type === 'MESSAGE_GROUP_BLOCKED') return;",
+            },
+        },
+        ...[
+            // Ignore new messages from blocked users
+            "\"MessageStore\"",
```
Vendicated, line 82:
```suggestion
            '"MessageStore"',
```

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+                match: /(?<=MESSAGE_CREATE:function\((?<props>\i)\){)/,
+                replace: "if($self.isBlocked($<props>.message.id)) return;",
+            },
+        })),
+        {
+            // Hides the author of referenced messages in replies to them
+            find: "isRepliedMessage:!!",
+            predicate: () => settings.store.hideReferencedAuthor,
+            replacement: {
+                match: /function \i\(\i\){/,
```
Vendicated, line 98: too generic

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+                replace: "$& if($self.isBlocked(arguments[0].message.author.id)) return;",
+            },
+        },
+        {
+            // Ignore all TYPING_START events from blocked users
+            find: "\"TypingStore\"",
+            predicate: () => settings.store.ignoreTyping,
+            replacement: {
+                match: /TYPING_START:(?<func>\i)/,
+                replace: "TYPING_START: (e) => { if($self.isBlocked(e.userId))return; $<func>(e) }",
```
Vendicated, line 108: variables introduced into patches should ensure no collisions. what if the callback function just so happens to be called e? e(e) and boom error

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+        hideFromMemberList: {
+            description: "Hide blocked users from the members list.",
+            type: OptionType.BOOLEAN,
+            default: false,
+            restartNeeded: true,
+        },
+    },
+);
+
+export default definePlugin({
```
Vendicated, line 65:
```suggestion
migratePluginSettings("NoBlockedUsers", "NoBlockedMessages")
export default definePlugin({
```

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const settings = definePluginSettings(
+    {
+        hideBlockedMessages: {
+            description: "Hide messages from blocked users.",
+            type: OptionType.BOOLEAN,
+            default: true,
+            restartNeeded: true,
+            onChange: newValue => {
+                if (!newValue) settings.store.ignoreBlockedMessages = false;
```
Vendicated, line 35: this is weird UX if it randomly toggles switches. it would be better to
1. use the second argument of defineSettings to disable the other setting if appropriate
2. combine booleans in your logic

src/plugins/noBlockedUsers/index.ts
rushiiMachine: It's a single method in a module thats somewhat unidentifiable
I could capture `isRepliedMessage` like the find but is it really worth
image

src/plugins/noBlockedUsers/index.ts
rushiiMachine: I fixed #1, what do you mean by #2

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+migratePluginSettings("NoBlockedUsers", "NoBlockedMessages");
+
+export default definePlugin({
+    name: "NoBlockedUsers",
+    description: "Hides blocked users and their messages from everywhere possible",
+    authors: [Devs.rushii, Devs.Samu],
+    settings,
+    patches: [
+        // Based on canary 303229
```
Vendicated, line 68:
```suggestion
```

rushiiMachine: It's for future use when I have to fix so I can reference working modules
I left the same in message logger

rushiiMachine: actually it is kinda pointless

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,136 @@
...
+        {
+            // Hide blocked users from chat autocomplete
+            find: ".ALLOW_EVERYONE_OR_HERE,",
+            replacement: {
+                match: /(queryResults.+?)return\{results:(.+?\))}/,
+                replace:
+                    "$1" +
+                    "let results = $2;" +
+                    "results.users=results.users.filter(res=>!$self.isBlocked(res.user.id));" +
+                    "return { results }",
```
Vendicated, line 115:
```suggestion
                match: /(?<=queryResults.+?)return\{results:(.+?\))}/,
                replace: "return { ...$1, users: $1.users.filter(res=>!$self.isBlocked(res.user.id)); }",
```

src/plugins/noBlockedUsers/index.ts
```diff
@@ -0,0 +1,135 @@
...
+                replace: "if($self.settings.store.ignoreBlockedMessages && $self.isBlocked($1.message.author.id)) return;",
+            },
+        })),
+        {
+            // Hides the author of referenced messages in replies to them
+            find: "isRepliedMessage:!!",
+            predicate: () => settings.store.hideReferencedAuthor,
+            replacement: {
+                match: /return null!=.+?SYSTEM_TAG/,
+                replace: "if($self.isBlocked(arguments[0].message.author.id)) return; $&",
```
Vendicated, line 94: these accesses are dangerous. you should either use optional chaining or destructure them inside your own function with try catch
```suggestion
                replace: "if($self.isBlocked(arguments[0]?.message?.author?.id)) return; $&",
```

src/plugins/noBlockedUsers/index.ts
rushiiMachine: This makes no sense because $2 is an expression and not a variable

Vendicated: then you can do `$self.filterUsers($1)`

pwall2222:
I found a slight bug, in a role category with only 1 user, whom is blocked. It shows the category empty.

Oroborius:
Could you perchance add an update to remove blocked users from voice chats as well? Love this plugin btw.

Oroborius:
This is now busted in the latest versions of Discord/Vencord.

PR #2171, opened 2024-02-13
Create JumpToFirst

cheesesamwich:


Vendicated:
this formatting is very inconsistent with the rest of vencord. please have a look and stay more consistent

how does this work with dms? group dms? how about channels that aren't textable? how about forums/threads?

Issue #2173, opened 2024-02-13
[Bug] FakeNitro doesn't ignore "use external emoji/stickers" permissions in sticker picker


xparadoxical:
tldr: emoji bypass bypasses permissions, sticker bypass does not.

### Discord Account

230v

### What happens when the bug or crash occurs?

FakeNitro respects "use external emoji" and "use external stickers" permissions, not letting you send any sticker you want. Occurs in any place where you can send messages that has the two permissions restricted in any combination.

| Use external emoji | Use external stickers | Server icons<br/>in the sticker picker | Favorite<br/>stickers | Most used<br/>stickers | Other external<br/>stickers |
|:-:|:-:|:-:|:-:|:-:|:-:|
| ✔️ | ✔️ | visible | enabled | enabled | enabled |
| ✔️ | ❌ | visible | **disabled** | **hidden** | **disabled** |
| ❌ | ✔️ | **hidden** | enabled | enabled | **hidden** |
| ❌ | ❌ | **hidden** | **disabled** | **hidden** | **hidden** |

Bold items should be changed.

Stable 265637 (863f550)
Vencord 8b6a403 (Vesktop V1.5.0)
Electron 28.1.3
Chromium 120.0.6099.199

Stable 265637 (863f550)
Vencord 8b6a403 (Web)

![image](https://github.com/Vendicated/Vencord/assets/43796685/27a407c2-16c2-4f14-ac92-8feba8572206)

### What is the expected behaviour?

Being able to get/send links to stickers from any server no matter what's the state of the two permissions, just like with the "Emoji bypass" option. The restriction doesn't make much sense (thus I perceive it as a bug) as it can be trivially worked around by getting annoyed, going to a different channel for a second, getting the link there and sending it in the original channel.

| Use external emoji | Use external stickers | Server icons<br/>in the sticker picker | Favorite<br/>stickers | Most used<br/>stickers | Other external<br/>stickers |
|:-:|:-:|:-:|:-:|:-:|:-:|
| ✔️ | ✔️ | visible | enabled | enabled | enabled |
| ✔️ | ❌ | visible | **enabled** | **enabled** | **enabled** |
| ❌ | ✔️ | **visible** | enabled | enabled | **enabled** |
| ❌ | ❌ | **visible** | **enabled** | **enabled** | **enabled** |

### How do you recreate this bug or crash?

[...]

PR #2174, opened 2024-02-13
Add a mention of permission bypassing to FakeNitro settings

xparadoxical:
I don't think I'd ever figure out that these settings do bypass these permissions if I didn't investigate what's the behavior in order to make #2173. As per the issue, sticker bypass is affected by *both* permissions, so I chose universal wording for both instead of doing an ugly "external emojis/stickers" kind of thing.

Marked as draft until I get any official opinion about that issue.


Vendicated:
tucked this in with https://github.com/Vendicated/Vencord/commit/0f9acba59ed4096eb6fc6a73da3c4c252c494c21, thank you

PR #2177, opened 2024-02-15
Fix Message Link Embeds on @me guilds

VillainsRule:
this fixes bots such as Make a Quote attempting to use @me as a guild, which discord supports (for some reason).
my method for the guildID var was a bit hacky, but there seem to be strict const/let structures.

> this is my first time contributing, so lmk if i did something wrong


Vendicated:
This makes no sense. `@me` is the "guild" id for private channels. links with `@me` point to messages in some private channel

Vendicated:
https://github.com/Vendicated/Vencord/commit/bc0a55053d709533143f66a2bc4b6914326e907d

PR #2185, opened 2024-02-16
ServerProfilesToolbox plugin

D3SOX:
Closes https://github.com/Vencord/plugin-requests/issues/52

# Preview

image

# TODO

- [ ] Copy other stuff than nickname and pronouns


Vendicated:
honestly pretty cool! instead of storing stuff in variables, why not JSON.stringify + copy to clipboard instead? (& make paste button read from clipboard)

that way you can also copy between clients or even back it up somewhere

src/plugins/serverProfilesToolbox/README.md
```diff
@@ -0,0 +1,5 @@
+# ServerProfilesToolbox
+
+Adds a copy/paste/reset button to the server profiles editor
+
+![Screenshot](./screenshot.png)
```
Vendicated, line 5: please do not include screenshots in source code. that will quickly bloat the repo and git is awful at storing binary files! post it in https://github.com/Vendicated/Vencord/issues/1749 to get a github cdn link

src/plugins/serverProfilesToolbox/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+                </Button>
+                <Button onClick={() => {
+                    // reset
+                    setPendingNickname("");
+                    setPendingPronouns("");
+                    if (premiumType === 2) {
+                        setPendingBio("");
+                        setPendingThemeColors([]);
+                        setPendingBanner("");
+                        setPendingAvatar("");
```
Vendicated, line 75: i think you forgot avatar decorations and profile effects

Vendicated: also default value for everything that isn't a simple text input is `null`, not empty string!

D3SOX: Added avatar decorations and profile effects.
However I still can't test it because I don't have Discord Nitro

src/plugins/serverProfilesToolbox/README.md
D3SOX: Resolved

D3SOX:
Changed it to use an object instead and added buttons for clipboard copy/paste

src/plugins/serverProfilesToolbox/index.tsx
Vendicated: > for anything that isn't a simple text input

:p

PR #2186, opened 2024-02-16
feat(MessageLogger): Compact display of history

Kyuuhachi:
Deleted messages take a lot of space, and in many cases they're deleted for a good reason, such as being spam or containing gore. So let's collapse them!

3 deleted messages — Show messages

Not sure if the way I handle the formatted string is any good, but hey, it works.


Kyuuhachi:
Also merged my HistoryModal plugin into this, since its purpose is very similar.

PR #2192, opened 2024-02-18
feat(plugin) FakeNitro: Allow customising hyperlink text

LualtOfficial:
If you use U+2800 then links will essentially be invisible which imo looks way cleaner. Image below is how it should look with the fix.

ksnip_20240218-163048

### DIFF
--- a/src/plugins/fakeNitro/index.ts
+++ b/src/plugins/fakeNitro/index.ts
@@ -162,6 +162,11 @@ const settings = definePluginSettings({
         description: "Whether to use hyperlinks when sending fake emojis and stickers",
         type: OptionType.BOOLEAN,
         default: true
+    },
+    invisibleHyperLInks: {
+        description: "Wether it should hide hyperlinks with an empty character",
+        type: OptionType.BOOLEAN,
+        default: false
     }
 });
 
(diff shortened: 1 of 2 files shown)


EdVraz:
This is a thing of preference. I'd prefer to see the emoji name as link rather than nothing. 

I would suggest making the text of the hyperlink configurable instead, with the default being the emoji name and the user being able to change it to anything.

src/plugins/fakeNitro/index.ts
```diff
@@ -162,6 +162,11 @@ const settings = definePluginSettings({
         description: "Whether to use hyperlinks when sending fake emojis and stickers",
         type: OptionType.BOOLEAN,
         default: true
+    },
+    hyperLinkText: {
+        description: "What text the hyperlink should use, if empty it uses the emoji name",
```
Vendicated, line 167: i don't like the empty approach. instead, support a placeholder of `{{NAME}}`

so
```
:{{NAME}}: -> :foo:
{{NAME}} -> foo
```

then the default would be just `{{NAME}}`

Vendicated:
please do not pull request from the main branch. use features branches

Vendicated:
thanks!

PR #2193, opened 2024-02-18
Collapsible UI

TheKodeToad:
I didn't really do a lot; just opening this so people know about it


zany130:
does this work on the latest discord stable? using vesktop and enabling the plugin doesn't seem to change anything

PR #2196, opened 2024-02-20
VoiceJoinMessages plugin

Sqaaakoi:
This plugin adds client-sided ephemeral message notifications from your friends DMs when they join a voice channel, and also logs when users join/leave/move channels in a guild within the associated channels' text chat.

### DIFF
--- a/src/plugins/voiceJoinMessages/index.ts
+++ b/src/plugins/voiceJoinMessages/index.ts
@@ -175,18 +175,4 @@ export default definePlugin({
             });
         },
     },
-    patches: [
-        // Join messages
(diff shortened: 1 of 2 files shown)


Sqaaakoi:
I think the scope of the (singular) plugin might be getting too big, as I'm thinking of adding a proper "voice channel log" menu. Should I split this into 2 plugins?

Sqaaakoi:
I have split it into 2 plugins but I am questioning if this is a good idea to split it

Sqaaakoi:
that'll be another branch and I'll merge it back over the voiceJoinMessages branch if the plugins being split is better

RobinRMC:
I'm currently using this plugin and have found it pretty useful.

PR #2197, opened 2024-02-20
SpotifyControls: export album cover as CSS variable for themes

Av32000:
I've added (as an option) the cover of the album/track currently playing to the SpotifyControl plugin player
Preview : 
image

### DIFF
--- a/src/plugins/spotifyControls/PlayerComponent.tsx
+++ b/src/plugins/spotifyControls/PlayerComponent.tsx
@@ -349,7 +349,7 @@ function Info({ track }: { track: Track; }) {
     );
 }
 
-export function Player() {
+export function Player({ useBg }: { useBg: Boolean; }) {
     const track = useStateFromStores(
         [SpotifyStore],
         () => SpotifyStore.track,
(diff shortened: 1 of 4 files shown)


Oi-jojo:
thanks, i will be using this lol

Av32000:
> image
> 
> Contrast might become a problem with light covers

Do you have any idea how to solve this problem?

cheesesamwich:
> Do you have any idea how to solve this problem?
you could make the buttons use an invert filter? or something like that

D3SOX:
The other ones were actually custom icons from my CSS but here it's barely readable for the text too. The rotating makes it less bad as it's not always the case when it's not fully white

image

Vendicated:
i don't think this is a good addition. way too opiniated no matter how you implement it. it should imo instead be left to themes. the plugin would just need to expose the album cover url as a css variable on the player, then any theme can do whatever they want

D3SOX:
I like @Vendicated's idea. Would you change this PR to only do this and then provide your CSS as a theme in your own repo.

Av32000:
It's done !
The new css var is `--vc-spotify-track-image`

Vendicated:
thanks! pretty cool. works nicely

```css
#vc-spotify-player {
    background: linear-gradient(
          rgba(0, 0, 0, 0.7), 
          rgba(0, 0, 0, 0.7)
        ), var(--vc-spotify-track-image) !important;
    background-size: cover !important;
}
```

image
image

PR #2198, opened 2024-02-21
feat(plugin): NewUserIndicator

devkvm:
Made this to help moderate a server and thought it may be useful for others.

Suggestions and criticisms appreciated. 

Screenshot_20240220_235935

### DIFF
--- a/src/plugins/newUserIndicator/index.tsx
+++ b/src/plugins/newUserIndicator/index.tsx
@@ -1,0 +1,101 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addBadge, BadgePosition, ProfileBadge, removeBadge } from "@api/Badges";
+import { addDecorator, removeDecorator } from "@api/MemberListDecorators";
+import { addDecoration, removeDecoration } from "@api/MessageDecorations";
+import { definePluginSettings } from "@api/Settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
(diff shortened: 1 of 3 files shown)


src/plugins/newUserIndicator/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const currentDate = new Date();
+    const userCreatedDate = new Date(SnowflakeUtils.extractTimestamp(user.id));
+    const diff = getTimeDiff(currentDate, userCreatedDate);
+    const tooltip = `Account created ${diff} days ago`;
+    const enabled = settings.store[indType] as Boolean;
+
+    if (settings.store.days > diff && enabled) {
+        return <Tooltip text={tooltip}>
+            {(tooltipProps: any) => (
+                <span {...tooltipProps} tabIndex={0}>❗</span>
```
Sqaaakoi, line 35: I feel like this would probably be better implemented as a SVG icon rather than as an emoji

Sqaaakoi (changes requested):

src/plugins/newUserIndicator/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+            )}
+        </Tooltip>;
+    }
+    return null;
+};
+
+const badge: ProfileBadge = {
+    component: u => checkUser(u.user, "badges"),
+    position: BadgePosition.START,
+    shouldShow: _ => true,
```
Sqaaakoi, line 45: You should probably do your checking here

Sqaaakoi:
I'm personally not a maintainer of Vencord but I do hang out in the #progaming channel occasionally trying to make stuff, go take another look at https://github.com/Vendicated/Vencord/blob/main/src/plugins/_api/badges.tsx

src/plugins/newUserIndicator/index.tsx
devkvm: noted
will take a look at the things you brought up after i finish up work

PR #2199, opened 2024-02-21
VcNarrator: Add an option to ignore your own actions

EepyElvyra:
closes https://github.com/Vencord/plugin-requests/issues/95

### DIFF
--- a/src/plugins/vcNarrator/index.tsx
+++ b/src/plugins/vcNarrator/index.tsx
@@ -167,6 +167,7 @@ export default definePlugin({
             for (const state of voiceStates) {
                 const { userId, channelId, oldChannelId } = state;
                 const isMe = userId === myId;
+                if (isMe && Settings.plugins.VcNarrator.ignoreSelf) return;
                 if (!isMe) {
                     if (!myChanId) continue;
                     if (channelId !== myChanId && oldChannelId !== myChanId) continue;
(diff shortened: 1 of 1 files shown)


src/plugins/vcNarrator/index.tsx
```diff
@@ -167,6 +167,7 @@ export default definePlugin({
             for (const state of voiceStates) {
                 const { userId, channelId, oldChannelId } = state;
                 const isMe = userId === myId;
+                if (isMe && Settings.plugins.VcNarrator.ignoreSelf) return;
```
D3SOX, line 170: But when you get moved, you do want it to say something. That's how I read the request

EdVraz: Not sure; the move message triggers both when someone else moves you or when you get moved by someone else.

AuTune: I hadn't accounted for the move event when I proposed this idea. For me at least I don't really care if it still announces moves, initiated by me or not. What if you could individually select which events self ignore? It would add even finer control for those who want it and for those who don't care you can maybe put a big "toggle all" button.

Vendicated: you could detect if it was done by yourself by hooking into the function/flux event responsible for joining a vc

PR #2201, opened 2024-02-23
messageLogger: ignore messages deleted within the next n milliseconds

meadowsys:
Added option to messageLogger to ignore quickly deleted messages, like ones that are deleted as part of pluralkit's proxying. There is a setting added to control the threshold (default 0ms, which is off). I would like help with the description of that setting, if theres a clearer / more concise way to describe it than what I put down :p


InfectedGriffon:
maybe something like "minimum message lifespan in milliseconds before logging" for the description?

Meadowsys:
...minimum age of logged messages?

src/plugins/messageLogger/index.tsx
```diff
@@ -163,6 +163,11 @@ export default definePlugin({
             description: "Comma-separated list of guild IDs to ignore",
             default: ""
         },
+        ignoreQuickDelete: {
```
Vendicated, line 166:
```suggestion
        ignoreDurationsWithinMs: {
```

src/plugins/messageLogger/index.tsx
```diff
@@ -163,6 +163,11 @@ export default definePlugin({
...
+            type: OptionType.NUMBER,
+            description: "Ignore deleted messages if they were deleted quicker than this many milliseconds",
```
Vendicated, line 168: milliseconds is not a very userfriendly unit. would be better to change it to seconds & use a slider instead

Vendicated (commented):
pretty useful! are you still working on this? i think it's already fine in its current state, there isn't really a use case for ignoring quick edits

meadowsys:
yeah you're right, I could do the seperation of delete/edit in another PR if/when I ever get to it. Apart from your review/changes I don't have anything else left to do on this heh

src/plugins/messageLogger/index.tsx
meadowsys: I'm thinking about users might want values that are out of range of whatever slider range gets put in? but yeah, seconds is better for sure, I think a field with seconds input is alright

PR #2206, opened 2024-02-25
feat(plugin): KeyboardClick

Slluxx:
Adds nice clicking sounds when typing.

### DIFF
--- a/src/plugins/keyboardClick/index.ts
+++ b/src/plugins/keyboardClick/index.ts
@@ -1,0 +1,102 @@
+import definePlugin, { OptionType } from "@utils/types";
+import { definePluginSettings } from "@api/Settings";
+
+let sound_click1, sound_click2, sound_click3, sound_clickBackspace;
+
+export const settings = definePluginSettings({
+    allowEscape: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Escape key",
+        restartNeeded: false,
+        default: true
+    },
+    allowTab: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Tab key",
+        restartNeeded: false,
+        default: true
+    },
+    allowCaps: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the CapsLock key",
+        restartNeeded: false,
+        default: true
+    },
+    allowShift: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Shift key",
+        restartNeeded: false,
+        default: true
+    },
+    allowCtrl: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Control key",
+        restartNeeded: false,
+        default: true
+    },
+    allowMeta: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Meta key",
+        restartNeeded: false,
+        default: false
+    },
+    allowAlt: {
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Alt key",
+        restartNeeded: false,
+        default: true
+    },
+});
+
+export default definePlugin({
+    name: "KeyboardClick",
+    description: "Modified BD plugin for some nice clicking sounds when typing.",
+    authors: [
+        {
+            id: 346059939502096396n,
+            name: "Slluxx",
+        },
+    ],
+    settings,
+    patches: [],
+    start() {
+        sound_click1 = new Audio('https://dl.dropboxusercontent.com/s/vfxrfu2u8jiq6xw/click1.wav?raw=1');
+        sound_click2 = new Audio('https://dl.dropboxusercontent.com/s/wtw25tzfctkpers/click2.wav?raw=1');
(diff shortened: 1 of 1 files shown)


CodeF53:
Take off headphones to enjoy this feature without any extra code

Sqaaakoi:
Was this PR sponsored by Opera GX?
Also I'd suggest adding a different sound for space, and also move yourself into the Devs list

src/plugins/keyboardClick/index.ts
```diff
@@ -0,0 +1,102 @@
...
+        description: "Play click when pressing the Alt key",
+        restartNeeded: false,
+        default: true
+    },
+});
+
+export default definePlugin({
+    name: "KeyboardClick",
+    description: "Modified BD plugin for some nice clicking sounds when typing.",
+    authors: [
```
MrDiamondDog, line 54: Add yourself to devs

src/plugins/keyboardClick/index.ts
```diff
@@ -0,0 +1,102 @@
...
+        {
+            id: 346059939502096396n,
+            name: "Slluxx",
+        },
+    ],
+    settings,
+    patches: [],
+    start() {
+        sound_click1 = new Audio('https://dl.dropboxusercontent.com/s/vfxrfu2u8jiq6xw/click1.wav?raw=1');
```
MrDiamondDog, line 63: The click sounds should definitely be configurable

src/plugins/keyboardClick/index.ts
```diff
@@ -0,0 +1,102 @@
+import definePlugin, { OptionType } from "@utils/types";
+import { definePluginSettings } from "@api/Settings";
+
+let sound_click1, sound_click2, sound_click3, sound_clickBackspace;
+
+export const settings = definePluginSettings({
+    allowEscape: {
```
MrDiamondDog, line 7: There has to be a better way to do this (custom settings component maybe?)

src/plugins/keyboardClick/index.ts
```diff
@@ -0,0 +1,102 @@
+import definePlugin, { OptionType } from "@utils/types";
```
MrDiamondDog, line 1: Make sure to run eslint

MrDiamondDog (changes requested):

src/plugins/keyboardClick/index.ts
Sqaaakoi: Why would a custom settings component be useful in this case?

MrDiamondDog: So each key doesn't need an individual setting, would also make checking if the sound should play simpler. Although it might be fine how it is

src/plugins/keyboardClick/index.ts
```diff
@@ -0,0 +1,102 @@
...
+        type: OptionType.BOOLEAN,
+        description: "Play click when pressing the Alt key",
+        restartNeeded: false,
+        default: true
+    },
+});
+
+export default definePlugin({
+    name: "KeyboardClick",
+    description: "Modified BD plugin for some nice clicking sounds when typing.",
```
MrDiamondDog, line 53:
```suggestion
    description: "Adds nice clicking sounds when typing.",
```

src/plugins/keyboardClick/index.ts
Slluxx: > So each key doesn't need an individual setting, would also make checking if the sound should play simpler. Although it might be fine how it is

The goal was not to have settings for each button but to blacklist common modifier and control buttons. The BD plugin has no such option and there is no and was no intent to expand this in any way.

src/plugins/keyboardClick/index.ts
Slluxx: Roger

src/plugins/keyboardClick/index.ts
Slluxx: I dont think eslint is mentioned anywhere. The Plugin Doc specifically says "Once you've made your plugin, make sure you run pnpm test and make sure your code is nice and clean!", which i did. Does the guide need to be changed?

src/plugins/keyboardClick/index.ts
Slluxx: I agree.

src/plugins/keyboardClick/index.ts
verticalsync: eslint will already throw an error for missing header, it's just best to run eslint to also ensure the codebase is the same throughout.

verticalsync: eslint will already throw an error for missing header, it's just best to run eslint to also ensure the codebase is the same throughout.

MrDiamondDog: Your pr is actually blocked from merging if it fails listing. Install the eslint extension in vscode and it will auto-format

Vendicated:
hi, thanks a lot for your contribution!

i don't think this is very useful, too niche. why not just do it on a system level?

PR #2207, opened 2024-02-25
[Plugin] HideMessage

hanzydev:
Adds a context menu option to hide messages.


hanzydev:
> How is this different from the existing delete message button?

The Delete Message button deletes the message you sent from everyone. The difference of this plugin is that it only deletes the message from your client, it does not send any API requests.

Koishukaze:
> > How is this different from the existing delete message button?
> 
> The Delete Message button deletes the message you sent from everyone. The difference of this plugin is that it only deletes the message from your client, it does not send any API requests.

what's the use case for this even

TheKodeToad:
I guess maybe if you're taking screenshots and you want to delete stuff that contains private info or irrelavent stuff?

hanzydev:
> > > How is this different from the existing delete message button?
> > 
> > 
> > The Delete Message button deletes the message you sent from everyone. The difference of this plugin is that it only deletes the message from your client, it does not send any API requests.
> 
> what's the use case for this even

Sometimes our friends and other people can send nsfw or disturbing images/content in a text channel that we don't want to see in a public place. This plugin removes the message from the DOM on demand. Kind of WhatsApp does but without a storage mechanism. When you reload your client the messages will be shown again.

hanzydev:
> I guess maybe if you're taking screenshots and you want to delete stuff that contains private info or irrelavent stuff?

Exactly.

ReXulEc:
Cool stuff. Keep it up!

src/plugins/deleteMessageFromMe/utils.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { classNameFactory } from "@api/Styles";
+
+export const cl = classNameFactory("vc-delete-from-me-");
```
MrDiamondDog, line 9: Just move this into the main file

MrDiamondDog (commented):
I think this could do with a rebranding. Imo, "Delete From Me" doesn't make much sense. I would change it to "Hide Message".

hanzydev:
> I think this could do with a rebranding. Imo, "Delete From Me" doesn't make much sense. I would change it to "Hide Message".

I will do it.

hanzydev:
> I think this could do with a rebranding. Imo, "Delete From Me" doesn't make much sense. I would change it to "Hide Message".

You are absolutely right. I changed the plugin name to Hide Message.

MrDiamondDog (approved):

ethan-davies:
Haven't tried this plugin but I feel, if it's not already implemented, that there should be a button to show all hidden messages without having to reload your client. Maybe this could be a command or something that show when you right click a message.

Ushie:
Hi, I already [opened a PR](https://github.com/Vendicated/Vencord/pull/1568) for the same plugin a while back, should I close it?

hanzydev:
> Haven't tried this plugin but I feel, if it's not already implemented, that there should be a button to show all hidden messages without having to reload your client. Maybe this could be a command or something that show when you right click a message.

Implemented in my latest commit

Discord_4mEKVFYwhp

hanzydev:
I also added support for using Message Logger with Hide Message plugin:

Discord_5z0lDSHZHx

hanzydev:
It's been 2 weeks and no one is looking at PR. Why is that? @Vendicated

nin0-dev:
> It's been 2 weeks and no one is looking at PR. Why is that? @Vendicated

vee reviews PRs when they are motivated to, be patient

src/plugins/hideMessage/index.tsx
```diff
@@ -4,47 +4,165 @@
...
+
+        :is(${elements}) [class^='contents'] [id^='message-content']:after {
+            content: "Hidden content";
+        }
+    ` : `
+        :is(${elements}) {
+            display: none !important;
+        }
+    `;
+};
```
lewisakura: why is this done with CSS rather than a patch?

lewisakura (changes requested):

src/plugins/hideMessage/index.tsx
hanzydev: > why is this done with CSS rather than a patch?

it is much easier to hide and show messages using CSS rather than JavaScript. I would have to delete the element from the DOM but putting it back would be hard and if I have to stick with JavaScript, I have to append a style to the message div that will hide the message but I will eventually use CSS if I do that.

hanzydev:
Currently there is no result and in 6 days, 2 months will have passed.

Ushie:
There is over a hundred PRs with some dating over a year ago, someone will eventually review your PR but until then it's best to just forget about this until you get that magic notification

hanzydev:
> There is over a hundred PRs with some dating over a year ago, someone will eventually review your PR but until then it's best to just forget about this until you get that magic notification, [my PR](https://github.com/Vendicated/Vencord/pull/1568) for the same function took place over 9 months ago

They're literally making fun of people 😂

lewisakura:
Your PR will not be reviewed if you keep asking about it. Please be patient, we have other things to do both for the project and in life.

hanzydev:
> Your PR will not be reviewed if you keep asking about it. Please be patient, we have other things to do both for the project and in life.

A month has passed and the pull request has still not been merged. You say we have things to do in our lives, but a pull request opened 4 hours ago is merged. There are people who have been waiting for a year. If you are not going to approve it, close this pull request and don't drag it out any longer.

lewisakura:
Okay, I don't normally do codereview for plugins anymore, but I'll give a response.

As it stands this plugin is not suited to be merged - creating tons of CSS for hiding messages is not effective and will eventually create a client slowdown. You don't need to manipulate the DOM to do this, you need to create a patch that filters out the hidden messages by ID when the channel view is rendered. I flagged this up previously but it seems like it was dismissed.

Vendicated:
yeah, as pointed out by lewi, this current implementation is not very good. using css for this is hacky and not sustainable. HideAttachments suffers from very much the same problem and honestly needs a rewrite, it's just that it's very old (made within the first two months of vencord existing), that's why it doesn't follow the quality standard as much as other plugins

in the current state this is not mergeable for that reason. here are a few better approaches i can think of:
- patch the message component to render null for hidden messages
- patch somewhere else to filter out undesired messages
- dispatch MESSAGE_DELETE events for the hidden messages

prs are looked at and merged whenever they are looked at and merged. this is a hobby project and there is a long backlog of contributions. this is not a very high priority change. if you are not okay with waiting, you probably shouldn't be contributing

PR #2214, opened 2024-02-28
ShowMeYourName Modification

cheesesamwich:
I made it so that you can individually toggle what to show, as well as change what the main name display shows
image
image

### DIFF
--- a/src/plugins/showMeYourName/index.tsx
+++ b/src/plugins/showMeYourName/index.tsx
@@ -1,6 +1,6 @@
 /*
  * Vencord, a Discord client mod
- * Copyright (c) 2024 Vendicated and contributors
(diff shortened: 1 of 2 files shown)


Vendicated:
this is really spaghetti.. i suggest just letting the user input a template string for the secondary display, similar to VCNarrator

src/plugins/showMeYourName/index.tsx
```diff
@@ -1,6 +1,6 @@
 /*
  * Vencord, a Discord client mod
- * Copyright (c) 2023 rini
+ * Copyright (c) 2024 Vendicated and contributors
```
Vendicated, line 3: do not edit the header

PR #2215, opened 2024-02-28
[Plugin] ToneIndicators

ImpishMoxxie:
ToneIndicators highlights and gives you a tooltip on some common tone indicators like /s, /srs, and more, and also allows you to add custom tone indicators to remember other ones and make your own!

# /srs example
image

# custom ToneIndicator
image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.8.2",
+    "version": "1.8.3",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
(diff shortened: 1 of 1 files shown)


lewisakura (changes requested):
can this be made into a generic initialisms highlighter rather than saying tone indicators?  it seems more useful that i could add things like "ttyl" to the highlighter rather than just tone indicators (especially since there's a built in editor but it's restricted to `/this` which seems a bit counterintuitive in my mind)

ImpishMoxxie:
@lewisakura look at the new pr we fixed it 
- https://github.com/Vendicated/Vencord/pull/2448

PR #2216, opened 2024-02-29
[Plugin] Encryptcord

Inbestigator:
# Encryptcord
Encryptcord allows you to finally have fully secure, end-to-end encrypted conversations with all of your friends. It makes use of a RSA asymmetrical encryption/decryption system. Remain safe with a new key pair being generated every restart.

## /encryptcord start
image
## /encryptcord invite
Invite as many of your friends as you want!

image

## Key exchange system
image

## Sender PoV
image 
## Receiver PoV
image

## Easily toggle between sending encrypted and unencrypted messages
image
image


package.json
```diff
@@ -34,18 +34,21 @@
         "@sapphi-red/web-noise-suppressor": "0.3.3",
         "@vap/core": "0.0.12",
         "@vap/shiki": "0.10.5",
+        "axios": "^1.6.7",
```
Vendicated, line 37: axios is useless bloat, use https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

package.json
```diff
@@ -34,18 +34,21 @@
...
         "eslint-plugin-simple-header": "^1.0.2",
         "fflate": "^0.7.4",
         "gifenc": "github:mattdesl/gifenc#64842fca317b112a8590f8fef2bf3825da8f6fe3",
         "monaco-editor": "^0.43.0",
         "nanoid": "^4.0.2",
+        "node-forge": "^1.3.1",
```
Vendicated, line 43: https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API

Vendicated:
hi, thank you for your efforts!

i don't think this would be a good addition for various reasons:
- it is very niche. most users won't understand what this is or how to use it and might just accidently "spam" with it 
- getting security right is very hard. i'm not sure if i want to bear that responsibility
- it is inherently vulnerable to man in the middle attacks
- if you want end to end encrypted messaging, you're much better off using a proper solution like XMPP or matrix

so i'll reject this, but you may post it in the third party plugins channel if you would like. i just highly recommend getting rid of these additional dependencies first because they are not necessary

Inbestigator:
> hi, thank you for your efforts!
> 
> i don't think this would be a good addition for various reasons:
> 
>     * it is very niche. most users won't understand what this is or how to use it and might just accidently "spam" with it
> 
>     * getting security right is very hard. i'm not sure if i want to bear that responsibility
> 
>     * it is inherently vulnerable to man in the middle attacks
> 
>     * if you want end to end encrypted messaging, you're much better off using a proper solution like XMPP or matrix
> 
> 
> so i'll reject this, but you may post it in the third party plugins channel if you would like. i just highly recommend getting rid of these additional dependencies first because they are not necessary

If I fix the imports and make it more user friendly, could it be considered again? The plugin was originally designed to be a better version or InvisibleChat.

PR #2218, opened 2024-03-01
[Plugin] MoreSearchEngines

ImpishMoxxie:
## Allows you to search in a few different search engines!

**Only searches text from msgs**

plugin request thingy: https://github.com/Vencord/plugin-requests/issues/117

img

yes we added some funny sites :3


src/plugins/MoreSearchEngines/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+                                <img
+                                    style={{
+                                        borderRadius: i >= 3 // Do not round Google, Yandex & SauceNAO
+                                            ? "50%"
+                                            : void 0
+                                    }}
+                                    aria-hidden="true"
+                                    height={16}
+                                    width={16}
+                                    src={new URL("/favicon.ico", Engines[engine]).toString().replace("lens.", "")}
```
Vendicated, line 52:
```suggestion
                                    src={new URL("/favicon.ico", Engines[engine]).toString()}
```
none of your engines contain lens. (anymore?)

src/plugins/MoreSearchEngines/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+                        action={() => search(src, Engines[engine])}
+                    />
+                );
+            })}
+        </Menu.MenuItem>
+    );
+}
+
+const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) => () => {
+    const group = findGroupChildrenByChildId("copy-link", children);
```
Vendicated, line 66:
```suggestion
	if (!message) return;

    const group = findGroupChildrenByChildId("copy-link", children);
```

Vendicated:
double search... why not replace the native discord option instead?

image

Kubuxu:
I was looking at writing a plugin like this but for a slightly different use path.
I wanted to replace the `Search with Google` with my preferred search engine.

Also, could you add Kagi to the list? `https://kagi.com/search?q=`

ImpishMoxxie:
> I was looking at writing a plugin like this but for a slightly different use path. I wanted to replace the `Search with Google` with my preferred search engine.
> 
> Also, could you add Kagi to the list? `https://kagi.com/search?q=`


we made it so that now it's under the "search with Google" and we are working on making it only search the highlighted part.
also ya I can add kagi.com

ImpishMoxxie:
updated it to show under the "search with Google" as https://github.com/Vendicated/Vencord/pull/2218#issuecomment-1972188521 asked

added settings to select what engines to use

only shows now when marking text and uses the marked text for the search

src/plugins/MoreSearchEngines/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+import PluginModal from "@components/PluginSettings/PluginModal";
+import { Devs } from "@utils/constants";
+import { openModal } from "@utils/modal";
+import definePlugin, { OptionType } from "@utils/types";
+import { Menu, SettingsRouter } from "@webpack/common";
+
+const testieWestie: boolean = true;
+
+const Engines = {
+    DuckDuckGo: "https://duckduckgo.com/",
```
Vendicated, line 19: where's Google?

ImpishMoxxie: google still shows this is just extra browsers

image

Vendicated:
> updated it to show under the "search with Google" as #2218 (comment) asked

that's absolutely not what i asked.. i suggested to **replace** the Google search option

PR #2219, opened 2024-03-01
MoreUserTags: If server owner tag is disabled, do not give other tags

Noxillio:
Some who already use a crown to indicate server owner do not want the server owner to have an additional tag placed next to their name as it is redundant. This change will prevent the server owner from receiving any other tag if the `OWNER` tag is disabled in settings.

If needed then an additional setting could be added to make this behavior optional, which would allow server owner to continue receiving `ADMIN` tag and so on.


Vendicated:
cc @sunnniee

Nuckyz:
Thanks for the improvement! Sorry for taking a long time to review

PR #2220, opened 2024-03-01
ModViewBypass plugin

Sqaaakoi:
Unlocks the Mod View menu to be used without moderator permissions, and enables it in every guild (so the user doesn't need to go into experiments to do so)


Vendicated:
could you describe the use case?

Sqaaakoi:
> could you describe the use case?

I often find myself using mod view in guilds I am not a moderator in to see if someone is a troll who has been in the guild for a long time but has sent little to no messages. The built-in filtering of notable permissions and roles with permissions is also nice to have as a member.

ac-tu:
How can you enable it with experiments? Considering this may not get merged. Despite its quality of life use.

ac-tu:
> But this only enables mod view for servers you have mod permissions in. The extension enables this and gives you it in servers you don't have mod permissions in.

Oh unfortunate that you can't enable it despite perms, either way that experiement is useful, thank you!

thewilloftheshadow:
Mod view has been rolled out to all servers now

Sqaaakoi:
> Mod view has been rolled out to all servers now

well they changed something that actually affects the main functionality, I have to fix that patch (I only just noticed today)

Vendicated:
superseded by https://github.com/Vendicated/Vencord/pull/2415

thanks regardless!

PR #2221, opened 2024-03-01
chore: tidy up suggested vscode extensions list

mcpower:
This should be more robust and more maintained than an extension.


Vendicated:
marking this as draft because suggested extensions need some adjustments anyway

Vendicated:
hmm after further testing this setting doesn't work as well as the extension

but i wanted to remove the opinioated extension suggestions anyway and only leave the 100% required ones in, so i'll still merge this

PR #2224, opened 2024-03-02
Add option to toggle the tooltip in memberCount.

cheesesamwich:
https://github.com/Vencord/plugin-requests/issues/128 :3


vrad-exe:
Thanks for adding this! You should probably do the opposite too, and make it so you can turn off the display in the member list and have it only display in the tooltip. I wouldn't personally use that option but it makes sense if you're adding a settings menu.

cheesesamwich:
good idea, ill do it soon.

Nuckyz:
Use an Enum instead of hardcoding numebrs

Sqaaakoi:
> Use an Enum instead of hardcoding numebrs

an array of enums mapped to strings and calling settings.store.views.includes(enum) is a more specific way of doing it
would also be more extensible in the future

cheesesamwich:
> Use an Enum instead of hardcoding numebrs

done :P

src/plugins/memberCount/index.tsx
```diff
@@ -32,6 +33,25 @@ export const ChannelMemberStore = findStoreLazy("ChannelMemberStore") as FluxSto
     getProps(guildId: string, channelId: string): { groups: { count: number; id: string; }[]; };
 };
 
+enum viewTypes {
```
Vendicated, line 36:
```suggestion
const enum ViewTypes {
```

src/plugins/memberCount/index.tsx
```diff
@@ -32,6 +33,25 @@ export const ChannelMemberStore = findStoreLazy("ChannelMemberStore") as FluxSto
...
+    MemberList,
+    Tooltip,
+    Both
+}
+
+const settings = definePluginSettings({
+    views: {
+        type: OptionType.SELECT,
+        description: "Where the member count should be displayed.",
```
Vendicated, line 45:
```suggestion
        description: "Where the member count should be displayed",
```

Vendicated:
i don't like the multiple choice because it is not extendable... what if we add it in a third place? we'd have to change how it works and migrate old settings. it would be better to just use multiple boolean options

Sqaaakoi:
> i don't like the multiple choice because it is not extendable... what if we add it in a third place? we'd have to change how it works and migrate old settings. it would be better to just use multiple boolean options

I think Vencord needs a dedicated "select one or more" settings type for these kinds of things at this point

vrad-exe:
It should probably just be 2 toggles, and just let the user turn all of them off if they really want to do that for some reason

Sqaaakoi:
> just let the user turn all of them off if they really want to do that for some reason

you severely underestimate how dumb some users are, I think a safety check for this is a good idea

Vendicated:
don't add pointless checks. it doesn't hurt to have it on with no features enabled

PR #2225, opened 2024-03-02
Add date to filename in settings backup

00-kat:
When exporting settings, the file name is always `vencord-settings-backup.json`. This small change adds the date to the filename, to make it easier when exporting if there is already a `vencord-settings-backup.json` in the folder chosen.

I tried all the locales available, and `ky` was the only one that printed it in the format `yyyy-mm-dd`. Feel free to change the format if necessary.

### DIFF
--- a/src/utils/settingsSync.ts
+++ b/src/utils/settingsSync.ts
@@ -49,7 +49,8 @@ export async function exportSettings({ minify }: { minify?: boolean; } = {}) {
 }
 
 export async function downloadSettingsBackup() {
-    const filename = "vencord-settings-backup.json";
+    // The date is formatted as yyyy-mm-dd.
+    const filename = `vencord-settings-backup-${(new Date).toLocaleDateString('ky')}.json`;
     const backup = await exportSettings();
     const data = new TextEncoder().encode(backup);
 


src/utils/settingsSync.ts
```diff
@@ -49,7 +49,8 @@ export async function exportSettings({ minify }: { minify?: boolean; } = {}) {
 }
 
 export async function downloadSettingsBackup() {
-    const filename = "vencord-settings-backup.json";
+    // The date is formatted as yyyy-mm-dd.
+    const filename = `vencord-settings-backup-${(new Date).toLocaleDateString('ky')}.json`;
```
Vendicated, line 53: using ky locale here is a bit strange and based on my testing this seems to format with slashes which will cause issues on non windows

moment can easily do custom format
```suggestion
    const filename = `vencord-settings-backup-${moment().format("YYYY-MM-DD")}.json`;
```

Vendicated:
implemented this via https://github.com/Vendicated/Vencord/commit/0f9acba59ed4096eb6fc6a73da3c4c252c494c21, thank you a lot for the contribution!

src/utils/settingsSync.ts
00-kat: > using ky locale here is a bit strange and based on my testing this seems to format with slashes which will cause issues on non windows

Huh, that's weird, since it used dashes on my system (Linux). Maybe it's different on different operating systems, which I did not expect.

Vendicated: i don't think os matters, it's just js :p 

it might be that it automatically changed / -> - cause / isnt allowed in filenames?

image

either way doesnt matter, i already fixed & pushed it. thanks again

PR #2227, opened 2024-03-03
MessageLinkEmbeds: Add option to only load visible links

mcpower:
If a channel has many message links, MessageLinkEmbeds may load many link embeds, even for messages that are not in the user's viewport. As of writing, the plugin loads embeds from the oldest message to the newest message, so the embeds for the most recent messages do not appear until all the old embeds are loaded.

Add an option to only load link embeds in the user's viewport to work around this issue. This will result in less API requests and faster loading for channels with many message links, but may cause embeds to suddenly appear when scrolling through a channel, so this option is off by default.

### DIFF
--- a/src/plugins/messageLinkEmbeds/index.tsx
+++ b/src/plugins/messageLinkEmbeds/index.tsx
@@ -22,6 +22,7 @@ import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants.js";
 import { classes } from "@utils/misc";
 import { Queue } from "@utils/Queue";
+import { useIntersection } from "@utils/react";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy, findComponentByCodeLazy } from "@webpack";
 import {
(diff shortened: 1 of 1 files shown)


mcpower:
This is my first ever time using React, so there may be issues with the code.

I'm not too happy with the verbose names for `MessageEmbedAccessoryMaybeWithIntersection` and `MessageEmbedAccessoryWithIntersection` - what do you think?

Vendicated:
this will very likely mess up scrolling. you could just achieve something similar to this by using queue.unshift instead of queue.push, this way messages will be fetched from bottom to top

Vendicated:
i just noticed you opened another pr just like that. why open multiple prs that achieve the same thing?

mcpower:
Specifically, it will result in less API requests and faster loading for channels. This is independent of the ordering change, but I agree that the main issue - that it takes a while to load recent embeds - is fixed by either commit.

The (default-off) option was added specifically due to the scrolling issue you mentioned. I personally don't mind this and would rather have better performance / less API requests, but I understand that the option may not be useful to most people, and is probably not worth the maintenance cost.

PR #2229, opened 2024-03-03
Make `Queue#unshift` constant time

mcpower:
Before this commit, `Queue#push` was using an constant time array push, while `Queue#unshift` - and most importantly - `Queue#next` were using O(n) array (un)shifts.
However, as both `push()` and `unshift()` called `run()`, which called `next()` - _all_ methods were O(n).

Swap the time complexity of these methods by reversing the order of the underlying array.
`Queue#unshift` is now constant time, and `Queue#push` is now O(n) - which is the same as before, as it previously called an O(n) `run()`. This speeds up users of `Queue#unshift`, such as WhoReacted and ValidUser.

### DIFF
--- a/src/utils/Queue.ts
+++ b/src/utils/Queue.ts
@@ -34,7 +34,7 @@ export class Queue {
     private promise?: Promise<any>;
 
     private next() {
-        const func = this.queue.shift();
+        const func = this.queue.pop();
         if (func)
             this.promise = Promise.resolve()
                 .then(func)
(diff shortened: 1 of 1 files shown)


lewisakura:
does this actually have any real world effect? overoptimizing isn't very useful and makes the code less understandable

Vendicated:
it doesn't no. the queue is meant for async work that needs throttling, especially sending requests to discord. minor details like being O(n) aren't relevant, the bottleneck is the networking

PR #2230, opened 2024-03-03
MessageLinkEmbeds: Load embeds from newest to oldest

mcpower:
Before this commit, MessageLinkEmbeds loaded embeds from oldest to newest. This was especially noticeable in channels with many message links - the most recent embeds would only load after all the old embeds were loaded first.

Additionally, if a channel had many message links, switching to another channel would prevent message link embeds from loading in the newly switched channel until the old channel's embeds are loaded first.

Swap the `Queue#push` with `Queue#unshift` to simulate a FIFO stack instead of a FILO queue. This addresses both of the above issues.


Vendicated:
thanks

PR #2237, opened 2024-03-06
refactor(Webpack): more reliable patching

Nuckyz:
I honestly recommend reviewing this by just looking at files changed and looking at the new file, lots of small changes and stuff like just indentation

- Rework how webpack require is grabbed:
  - Instead of trying to push a fake module to webpackChunkdiscord_app everytime its value is defined again, it now monkey patches the init callback Discord runs to initialize the app after all needed chunks are loaded
  
  - This lets us more reliably get the webpack require as soon as we can, but of course not too soon in a point where needed properties would not be yet defined. It also makes it impossible for us to grab the sentry webpack require instead of the main, in case sentry somehow loaded faster than the main webpack script.
- Add beforeWebpackInit API:
  - By monkey patching the init callback we can also run our own code before the app is intialized. This is specifically used by the code used to grab the require and by the reporter to start searching for chunks and loading them.
- Add factoryListeners API:
  - Also only used by the reporter for now, but allows a callback to run everytime a new module factory is added to the modules array.
- Future proof and fix issues/potential isseus with extractAndLoadChunks & reporter:
  - Makes extractAndLoadChunks and the reporter not rely on the chunk group id being the also the entry point id.
  
  - extractAndLoadChunks now grabs the chunk group id and entry point as two separate capturing groups. I also fixed logic issues with it, which I didn't notice before.
  
  - Reporter no longer depends on the chunk group id being the entry point when looking at the chunks object to load chunks. Now it manually scrapes chunks loaded from every module factory using regex, to grab the chunk group id and the entry point, until every chunk loaded by Discord is manually loaded by us. After that it manually loads the rest of the chunks which Discord is not using, so we can be aware of duplicate modules too.
- Cleanup Webpack patching for more readable code.


Nuckyz:
@CodiumAI-Agent /review

Vendicated:
@github-advanced-security thoughts on this code?

src/plugins/index.ts
```diff
@@ -150,12 +150,14 @@ export const startPlugin = traceFunction("startPlugin", function startPlugin(p:
     }
 
     if (flux) {
+        logger.info("Subscribing to flux events of plugin", name);
```
Vendicated, line 153: probably should be debug calls instead cause otherwise this will get quite spammy

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -273,12 +275,12 @@ export default definePlugin({
                 {
                     // Change the role permission check to CONNECT if the channel is locked
                     match: /ADMINISTRATOR\)\|\|(?<=context:(\i)}.+?)(?=(.+?)VIEW_CHANNEL)/,
-                    replace: (m, channel, permCheck) => `${m}!Vencord.Webpack.Common.PermissionStore.can(${PermissionsBits.CONNECT}n,${channel})?${permCheck}CONNECT):`
```
Vendicated, line 276: what's this for?

src/webpack/patchWebpack.ts
```diff
@@ -18,66 +18,117 @@
...
+// Discord uses it to await for all the chunks to be loaded before initializing the app
+// We monkey patch it to also monkey patch the initialize app callback to get immediate access to the webpack require and run our listeners before doing it
+Object.defineProperty(Function.prototype, "O", {
+    configurable: true,
+
+    set(onChunksLoaded: any) {
+        // When using react devtools or other extensions, or even when discord loads the sentry, we may also catch their webpack here.
+        // This ensures we actually got the right one
+        // this.e (wreq.e) is the method for loading a chunk, and only the main webpack has it
+        if (new Error().stack?.includes("discord.com") && this.e.toString().includes("Promise.all")) {
```
Vendicated, line 66:
```suggestion
        if (new Error().stack?.includes("discord.com") && String(this.e).includes("Promise.all")) {
```

src/webpack/patchWebpack.ts
```diff
@@ -18,66 +18,117 @@
...
+            logger.info("Found main Webpack onChunksLoaded");
```
Vendicated, line 67:
```suggestion
            logger.info("Found main WebpackRequire.onChunksLoaded");
```

src/webpack/patchWebpack.ts
Nuckyz: why use String instead of toString?

src/plugins/index.ts
```diff
@@ -119,6 +119,29 @@ export function startDependenciesRecursive(p: Plugin) {
     return { restartNeeded, failures };
 }
 
+export function subscribePluginFluxEvents(p: Plugin, fluxDispatcher: typeof FluxDispatcher) {
+    if (p.flux && !p.fluxEventsSubscribed) {
+        p.fluxEventsSubscribed = true;
```
Vendicated, line 124: i can't find the unsubscribe counter part. wouldn't this break plugins if you enable -> disable -> enable them?

Vendicated: tbh we should also not store these properties on the plugin. instead we could keep an internal Set()

src/plugins/index.ts
```diff
@@ -119,6 +119,29 @@ export function startDependenciesRecursive(p: Plugin) {
...
+
+        logger.debug("Subscribing to flux events of plugin", p.name);
+        for (const event in p.flux) {
```
Vendicated, line 127:
```suggestion
        for (const [event, handler] of Object.entries(p.flux)) {
```

src/plugins/index.ts
```diff
@@ -119,6 +119,29 @@ export function startDependenciesRecursive(p: Plugin) {
...
+            fluxDispatcher.subscribe(event as FluxEvents, p.flux[event]);
+        }
+    }
+}
+
+export function subscribeAllPluginsFluxEvents(fluxDispatcher?: typeof FluxDispatcher) {
+    if (!fluxDispatcher) {
```
Vendicated, line 134: what's this check for?

src/plugins/index.ts
Nuckyz: It's in the stopPlugin function, but I also forgot to set the prop to false

src/plugins/index.ts
Nuckyz: you are right, if it happens then our function doesn't even get called

src/webpack/patchWebpack.ts
```diff
@@ -18,66 +18,118 @@
...
+                        for (const beforeInitListener of beforeInitListeners) {
+                            beforeInitListener(wreq);
+                        }
 
-                if (_initWebpack(v)) {
-                    logger.info("Successfully initialised Vencord webpack");
-                    // @ts-ignore
-                    delete window[WEBPACK_CHUNK];
-                    window[WEBPACK_CHUNK] = v;
+                        originalCallback();
```
Vendicated, line 90:
```suggestion
                        originalCallback.apply(this, arguments);
```

src/webpack/patchWebpack.ts
```diff
@@ -99,41 +152,41 @@ function patchPush(webpackGlobal: any) {
...
+
         const originalMod = mod;
         const patchedBy = new Set();
 
-        const factory = factories[id] = function (module, exports, require) {
+        const factory = factories[id] = function (module: any, exports: any, require: WebpackInstance) {
+            if (wreq == null) {
+                if (!webpackNotInitializedLogged) {
+                    webpackNotInitializedLogged = true;
+                    logger.error("WebpackRequire was not initialized, running modules without patches instead.");
```
Vendicated, line 177: why?

src/webpack/patchWebpack.ts
```diff
@@ -171,107 +224,127 @@ function patchFactories(factories: Record<string | number, (module: { exports: a
...
-        // when you force load all chunks???
-        factory.toString = () => mod.toString();
+        factory.toString = originalMod.toString.bind(originalMod);
         factory.original = originalMod;
 
+        for (const factoryListener of factoryListeners) {
+            try {
+                factoryListener(originalMod);
+            } catch (err) {
+                logger.error("Error in Webpack factory listener", err);
```
Vendicated, line 239: logging the listener would be useful
```suggestion
                logger.error("Error in Webpack factory listener:\n", err, factoryListener);
```

src/webpack/patchWebpack.ts
Nuckyz: basically most of the entry points of our plugins are done by patches, when also in most cases they use things from the discord webpack, which makes a lot of errors happen when webpack was not initialized.

doing this makes the console way more clear and shows what's the actual issue, instead of spamming errors

PR #2238, opened 2024-03-06
[Plugin] BlockKeywords

CatCraftYT:
Blocks messages containing certain keywords (defined by the user) as if they were sent by a blocked user.

### DIFF
--- a/src/plugins/blockKeywords/index.ts
+++ b/src/plugins/blockKeywords/index.ts
@@ -1,75 +1,36 @@
 import definePlugin, { OptionType } from "@utils/types";
-import { Settings } from "@api/Settings";
+import { Settings, definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import { MessageJSON } from "discord-types/general";
 
-// o: has ._channelMessages and .commit()
-// o is ChannelMessages
-// o is h class
-// foreach iterates through _messages
-// D is RelationshipStore?
+var blockedKeywords: Array<RegExp>;
 
-// o.default.forEach(e=>{o.default.commit(e.reset(e.map(e=>e.set('blocked',$self.containsBlockedKeywords($1)))))})
+const settings = definePluginSettings({
+    blockedWords: {
+        type: OptionType.STRING,
+        description: "Comma-seperated list of words to block",
+        default: "",
+        restartNeeded: true
+    },
+    useRegex: {
+        type: OptionType.BOOLEAN,
+        description: "Use each value as a regular expression when checking message content (advanced)",
+        default: false,
+        restartNeeded: true
+    },
+    caseSensitive: {
+        type: OptionType.BOOLEAN,
+        description: "Whether to use a case sensitive search or not",
+        default: false,
+        restartNeeded: true
+    }
+});
 
 export default definePlugin({
     name: "BlockKeywords",
     description: "Blocks messages containing specific user-defined keywords, as if the user sending them was blocked.",
     authors: [Devs.catcraft],
     patches: [
-        /*
-        {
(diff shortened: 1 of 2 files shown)


Terraman2:
Please make it support regular expressions as was mentioned in this issue: https://github.com/Vencord/plugin-requests/issues/159, I **need** it. Also please make it so it can also replace words in other people's messages from your perspective. & it would be nice if you make it so if the description message of a forum post or thread is hidden, the forum post or thread itself is also hidden.

src/plugins/blockKeywords/index.ts
```diff
@@ -0,0 +1,95 @@
...
+import { MessageJSON } from "discord-types/general";
+
+var blockedKeywords: Array<RegExp>;
+
+const settings = definePluginSettings({
+    blockedWords: {
+        type: OptionType.STRING,
+        description: "Comma-seperated list of words to block",
+        default: "",
+        restartNeeded: true
```
ryan-0324, line 13: Is a restart actually needed for all of these settings? If not, an [onChange](https://github.com/Vendicated/Vencord/blob/d2941281a40cd8b6b4a74f69da6530c843e3128f/src/utils/types.ts#L170) function that updates `blockedKeywords` could instead be added to these setting definitions. Maybe the code in `start()` could be moved to a top-level function that is used as the value of `start` and the `onChange`s.

CatCraftYT:
> Please make it support regular expressions as was mentioned in this issue: [Vencord/plugin-requests#159](https://github.com/Vencord/plugin-requests/issues/159), I **need** it. Also please make it so it can also replace words in other people's messages from your perspective. & it would be nice if you make it so if the description message of a forum post or thread is hidden, the forum post or thread itself is also hidden.

It does support regular expressions, it's in the settings :)

src/plugins/blockKeywords/index.ts
CatCraftYT: I made a new branch on my repo with changes that almost implement this, but it only actually visibly blocks the messages when the user switches channels. The messages in the currently viewed channel aren't updated when the message list is updated for some reason. The block feature gets around this but I'm not sure how rn

Terraman2:
I think there really should be a feature where you can publish filter lists, & there should also be a feature where anyone can make a fork of a list, & the fork can be merged into the original. & anyone should be able to use any version of a filter list & choose whether or not to automatically update it. If they use the newest version, automatic updates would be enabled by default, but if they use any other version, automatic updates would be disabled by default.

j4k0xb:
Can you add multiple regex inputs or at least one per line in a textarea instead of splitting on `,`?
Makes it impossible to use e.g. `\w{1,10}`

CatCraftYT:
> Can you add multiple regex inputs or at least one per line in a textarea instead of splitting on `,`? Currently its impossible to use e.g. `\w{1,10}`
> 
> Here's my attempt: 
> It will merge them into a single regex (and escape the strings): `/steamcommunity\.com.+?\]\(https|test/i`
Awesome, where can I find the source for this?

j4k0xb:
https://github.com/j4k0xb/Vencord/tree/block-keywords/src/plugins/blockKeywords

Messured:
I can't install this could someone tell me how real quick?

Feathers8:
> https://github.com/j4k0xb/Vencord/tree/block-keywords/src/plugins/blockKeywords
edit: deleted fork, waste of time trying to contribute

Huh? Why though? Can you elaborate please? @j4k0xb

j4k0xb:
@Feathers8 because many PRs or even 1-line simple fixes stagnate for months to years and maintainers decide which plugins can exist rather than users/devs, due to everything being centralized.

PR #1783, opened 2023-10-08
feat(userVoiceShow): Add button to view voice channel without joining

LordEliasTM:
title says it
merge if cute
image

PR #2248, opened 2024-03-10
feat(UserVoiceShow): Add custom VoiceChannelSection

Johannes7k75:
This adds a option that can be selected to use  this little prettier Channel Section.
Inspired by https://github.com/Vencord/plugin-requests/issues/173

UserPopout
image

UserProfileModal
image

PrivateChannelProfile
image


RPPSH:
how to install

Johannes7k75:
Follow these [Installation Instructions](https://github.com/Vendicated/Vencord/blob/main/docs%2F1_INSTALLING.md)
But as git clone you use this:
`git clone --branch VoiceActivity_CustomVoiceChannelSection https://github.com/Johannes7k75/Vencord`

D3SOX:
This is basically an improved version of #1783, right?

Johannes7k75:
> This is basically an improved version of #1783, right? I don't know why we would need those two styles (default and custom). I think you should stick with only the custom one
Yeah basically, but on this the buttons at the bottom are switched

D3SOX:
Found a small bug

image

It adds this empty section on top of mutuals (I guess it happens when the user is not in a voice channel)

D3SOX:
Can confirm it being fixed now

src/plugins/userVoiceShow/components/VoiceChannelSection.css
```diff
@@ -1,3 +1,73 @@
...
+    min-width: 32px !important;
+    max-width: 32px !important;
+    padding: 0;
+}
+
+.vc-uvs-join>div {
+    margin: 4px !important;
+}
+
+span:has(.vc-uvs-join) {
```
Vendicated, line 28: this is a very bad selector

src/plugins/userVoiceShow/index.tsx
```diff
@@ -93,22 +93,42 @@ export default definePlugin({
...
+                // paste my fancy custom button below the username
                 replace: "$&$self.patchModal(arguments[0]),",
             }
+        },
+
+        // Private Channel Profile - above Activities
+        {
+            find: "UserProfileTypes.PANEL,useDefaultClientTheme",
+            replacement: {
+                match: /user:(\i).+?voiceGuild,voiceChannel.+?:null,/,
```
Vendicated, line 129: try to make this user:\i part a little more specific

src/plugins/userVoiceShow/index.tsx
```diff
@@ -93,22 +93,42 @@ export default definePlugin({
         );
     },
 
+    patchPrivateChannelProfile({ user }: UserProps) {
```
Vendicated, line 96: these functions should use error boundaries to prevent crashes in case something breaks in thew future

Johannes7k75: Do you mean the patchPrivateChannelProfile function itself or the returned jsx/html.
Cause the VoiceChannelField is already warped in an ErrorBoundary.

Sqaaakoi: > Do you mean the patchPrivateChannelProfile function itself or the returned jsx/html. Cause the VoiceChannelField is already warped in an ErrorBoundary.

that doesnt protect the actual voicechannelfield though??????

Johannes7k75: Ok now i have wraped all pathes in `ErrorBoundary.wrap`.
I hope this would prevent crashes...

src/plugins/userVoiceShow/index.tsx
Johannes7k75: i made it now `(\i){1,2}`

src/plugins/userVoiceShow/components/VoiceChannelSection.css
Johannes7k75: This style wasnt even neede so i removed it

Johannes7k75: its was neede...

i replace it with `.vc-uvs-buttons span` that should be better, i think

Nuckyz:
Superseded by https://github.com/Vendicated/Vencord/commit/bae9f8cbe696b3f4ddfd41f80e0bfdb5ccc3a755, thanks regardless!

PR #2250, opened 2024-03-10
fix(ServerProfile): crop banner to prevent overflow

nakoyasha:
discord doesnt crop the banner server-side, so you can upload a tall image and [cause serverprofile's modal layout to explode.](https://files.catbox.moe/g5w1up.png)
this fixes that by cropping the server banner image in (almost) the same way that discord does
Discord_9P10FLqjKY


Vendicated:
thank you!

PR #2252, opened 2024-03-11
feat(plugin): CustomAppIcons

SerStars:
Add/upload your own custom (In-)App Icon(s)

- Following the steps in the plugin settings ("How to use?" header) should explain it all how you can use the plugin

ScreenShot1

ScreenShot2

ScreenShot3

### DIFF
--- a/src/plugins/CustomAppIcons/AppIconModal.tsx
+++ b/src/plugins/CustomAppIcons/AppIconModal.tsx
@@ -1,0 +1,104 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { localStorage } from "@utils/localStorage";
+import { ModalCloseButton, ModalContent, ModalFooter, ModalHeader, ModalProps, ModalRoot, ModalSize } from "@utils/modal";
+import { findByProps } from "@webpack";
+import { Button, FluxDispatcher,Forms, React, showToast, Text, TextInput, Toasts, useState } from "@webpack/common";
+
+
+interface AppIcon {
+    id: string,
+    iconSource: string,
+    name: string,
+    isPremium: boolean;
+}
+function AppIconModal(props: ModalProps) {
+    const [name, setName] = useState("");
+    const [imageUrl, setimageUrl] = useState("");
+
+    function addAppIcon(name, url) {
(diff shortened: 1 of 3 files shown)


happyendermangit:
Nice plugin (i contributed)

TheKodeToad:
Can't you do this with CS? is it particularly useful?

happyendermangit:
> Can't you do this with CS? is it particularly useful?

You technically can use css, but this add's it to the menu, and css will break faster, because discord css classes names get's changed a lot

Sqaaakoi:
Yeah it can be done with CSS

At this point though we need a proper way of LOCALLY embedding images in custom quick css

Sqaaakoi:
> > Can't you do this with CS? is it particularly useful?
> 
> You technically can use css, but this add's it to the menu, and css will break faster, because discord css classes names get's changed a lot

This is why I believe a CSS postprocessor selector that effectively calls findByProps with all the arguments and returns the value of the first item with all spaces replaced with . so it's a class selector would be useful

Vendicated:
to embed your image, encode it as data uri and use that. or upload it to imgur, github, or whatever else and link it

you can use attribute selectors instead of hardcoding class hashes. this way your css will only break if discord completely rewrites the ui

happyendermangit:
Yea

happyendermangit:
but this make's it easier instead of using css, because for some people they dont know how to write css

Sqaaakoi:
css is not *that* hard

also, the css snippets channel exists

happyendermangit:
Yea, but if you use css, it wont be easy to change the icons, this plugin saves the icons, so you can switch between them anytime

TheKodeToad:
Frequently switching between your own icons a very niche use-case

nyakowint:
Vencord precedent/rule of thumb is to utilize css as opposed to plugins whenever possible, and needing to switch between icons is pretty niche

happyendermangit:
but there is people that wont understand css, having a ui for this is much better & gives a better experience

Vendicated:
if we went by the philosophy of "plugin is easier than css" then we'd have tons of plugins for everything and be extremely bloated

just write an easy to use snippet and tell people how to use it 

if they can't figure it out, skill issue

Vendicated:
just make it user friendly, like this. if someone cant figure that out they probably wouldnt be able to install vencord either


image

happyendermangit:
What about userbg plugin, it can be done using css right? why it got accepted? & what about fake profile themes plugin, it can be also done using css? why it got accepted???

PR #2253, opened 2024-03-12
TypingTweaks: Add option to show username over display name

MannuVilasara:
image


Vendicated:
why not integrate with the ShowMeYourName plugin instead? would allow customisation

MannuVilasara:
> why not integrate with the ShowMeYourName plugin instead? would allow customisation

won't it create issues with typingTweak plugin? like ShowMeYourName plugin will try to change the typing style while typingTweak will do the same

MannuVilasara:
> > why not integrate with the ShowMeYourName plugin instead? would allow customisation
> 
> won't it create issues with TypingTweak plugin? like ShowMeYourName plugin will try to change the typing style while TypingTweak will do the same My plan was to add it in ShowMeYourName plugin but thinking of this I added it in TypingTweaks

Or can integrate in both of them. Will depend upon the user which one they choose

MannuVilasara:
I have added the option in both the plugins which can be used to toggle the same function. is this approach okay? please let me know

PR #2254, opened 2024-03-12
petpet: fix image option failing with "no image specified"

EdVraz:


EdVraz:
Changing the `DRAFT_TYPE` variable to 5 doesn't work; wasn't sure if i should introduce a new variable (and if so, what to name it) so if yes tell me.

Vendicated:
dupe of #2070

PR #2255, opened 2024-03-12
NoGenericFilenames plugin

Sphirye:
# NoGenericFilenames 

A plugin that prevents you from saving generic filenames as "image.png" or "video.mp4" so you dont have to deal with these annoying overwrite confirmation popups again.

image

## Usage
It really just requires to enable the plugin, it will automatically rename the posible cases with a timestamp.

## Options
Captura desde 2024-03-13 18-22-15


src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    const res = await fetch(url);
+    if (res.status !== 200) return;
+
+    return await res.blob();
+}
+
+function isGenericFilename(filename: string) {
+    let rex = /^(bruh)\s*(?:\(([\d]+)\))?$/g;
+    rex = new RegExp(rex.source.replace("bruh", genericFilenames.join("|")));
+    return rex.test(filename);
```
TheKodeToad, line 26: why not just put image|video or use template literal

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        type: OptionType.BOOLEAN,
+        default: true,
+    }
+});
+
+export default definePlugin({
+    name: "NoGenericFilenames",
+    description: "Prevent discord downloads from overwriting files with generic names by adding a timestamp.",
+    settings: settings,
+    authors: [
```
MrDiamondDog, line 56: Add yourself to `devs`

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import definePlugin, { OptionType } from "@utils/types";
+import { saveFile } from "@utils/web";
+
+const genericFilenames: string[] = [
```
MrDiamondDog, line 11: This should be an option

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    rex = new RegExp(rex.source.replace("bruh", genericFilenames.join("|")));
+    return rex.test(filename);
+}
+
+function getFilenameData(filename: string): { name: string, extension: string; } {
+    const regex = /^(.+)(\.[^.]+)$/;
+    const result = regex.exec(filename);
+
+    if (!result) {
+        throw new Error(`Invalid filename: ${filename}`);
```
Kyuuhachi, line 34: Won't this make it crash if the file has no extension?

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    return {
+        name: result[1],
+        extension: result[2].substring(1),
+    };
+}
+
+const settings = definePluginSettings({
+    includeMilis: {
+        name: "Include milis",
```
Kyuuhachi, line 45: Milliseconds is spelled with two L

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            replacement: [
+                {
+                    match: /(?<=SAVE_IMAGE_MENU_ITEM,)action:/,
+                    replace: "action:()=>$self.saveImage(arguments[0]),"
+                },
+            ]
+        }
+    ],
+
+    getCurrentDate(): string {
```
Kyuuhachi, line 75: Why not just `new Date().toISOString()`? Or something from momentjs.

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        const date = new Date();
+        const hour = date.getHours().toString().padStart(2, "0");
+        const minutes = date.getMinutes().toString().padStart(2, "0");
+        const seconds = date.getSeconds().toString().padStart(2, "0");
+        const milis = date.getMilliseconds().toString().padStart(4, "0");
+        const day = date.getDate().toString().padStart(2, "0");
+        const month = (date.getMonth() + 1).toString().padStart(2, "0");
+        const year = date.getFullYear();
+
+        return `${hour}-${minutes}-${seconds}${includeMilis ? `-${milis}` : ""} ${day}-${month}-${year}`;
```
Kyuuhachi, line 87: `HH-MM-SS.mmmm DD-MM-YYYY`? That's a really weird date format.

src/plugins/noGenericFilenames/index.ts
Sphirye: Because im planning to add a input option where people will be able to set their custom matching name, so including it on the regex should be easy as pushing the input value in the genericFilenames array

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    authors: [
+        {
+            id: 858410144349945867n,
+            name: "Kirigamium",
+        },
+    ],
+
+    patches: [
+        {
+            find: 'id:"save-image"',
```
Kyuuhachi, line 65: Why patch the context menu rather than the `saveImage` file it calls? Find string `"Save image method called outside native app"` perhaps.

src/plugins/noGenericFilenames/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            replacement: [
+                {
+                    match: /(?<=SAVE_IMAGE_MENU_ITEM,)action:/,
+                    replace: "action:()=>$self.saveImage(arguments[0]),"
```
Kyuuhachi, line 69: This will leave a stray `action: ()=>..., A }`. Won't cause any issues *now*, but if that variable happens to be named `id` in the next build, that wouldn't be very fun.

Kyuuhachi (commented):
I'm pretty sure this plugin should have `.desktop` or `.discordDesktop` suffix, since it's only applicable in Electron.

src/plugins/noGenericFilenames/index.ts
Sphirye: Resolved in https://github.com/Vendicated/Vencord/pull/2255/commits/c26ff4498631135457e28a1c50331786c5325ce7

src/plugins/noGenericFilenames/index.ts
Sqaaakoi: That's very hard to sort in a file manager too. Use YYYY-MM-DD_hh:mm:ss or similar instead

src/plugins/noGenericFilenames/index.ts
Sphirye: Resolved in https://github.com/Vendicated/Vencord/pull/2255/commits/c8b6062d5bd48be89a16edf74bd061e0cafac4a2

src/plugins/noGenericFilenames/index.ts
Sphirye: It seems colons are not allowed in windows filenames, gonna use YYYY-MM-DD hh-mm-ss instead, hope thats good enough https://github.com/Vendicated/Vencord/pull/2255/commits/bcfe588b21b0cf812aefc2883b732766dbe97856

src/plugins/noGenericFilenames/index.ts
Sphirye: Done, now it should just return the extension as an empty string if it results to be null

src/plugins/noGenericFilenames/index.ts
Sphirye: Done, the array stays as a static value in `FilenameUtil.ts` but it content is being instantiated in the default option value so people can reset their coincidences lists if they want to

imagen

src/plugins/noGenericFilenames/index.ts
Sphirye: May i ask for more details on this? Im not sure about what you mean by that find string thing

Sphirye:
>https://github.com/Vendicated/Vencord/pull/2255#discussion_r1522238885
>https://github.com/Vendicated/Vencord/pull/2255#discussion_r1522237461

@Kyuuhachi May i ask for more details on both of these? I'm not sure about what you mean by that find string thing

Kyuuhachi:
Would be more convenient to discuss that over in #plugin-development if that's ok with you.

src/plugins/noGenericFilenames/index.ts
Sphirye: Resuming this soon, im stuck resolving these two issues

Vendicated:
honestly I don't think this really solves the issue it's trying to solve. timestamp name is still as generic as unknown.png. it'd be better to just name files properly yourself. this is also quite niche anyway

thank you anyway

PR #2256, opened 2024-03-13
feat(plugin) WebhookManager

byeoon:
Manage your webhooks easily; delete, send messages, get detailed info and more. This plugin is very useful for developers, general users, (and for people who scout out scammers who use discord webhooks as credential stealers). I'm planning on slowly adding more features, but this is what I have so far. 

Since the plugin is still new, I'm going to be keeping my eye out on bugs that people report and fixing them as quickly as possible.

## Get Information
image

---

## Send Message
image
(Custom name AND profile pic support, also has TTS and raw support for JSON.)
image

---

## Delete (Works on **any** webhook)
image
image


src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                "Webhook Token: " + response.token + "\n " +
+                                "Channel ID: " + response.channel_id + "\n " +
+                                "Server ID: " + response.guild_id + "\n " +
+                                "Webhook Profile Picture: " + "[Click Me](https://cdn.discordapp.com/avatars/" + response.id + "/" + response.avatar + ".png)" + "\n " +
+                                "Webhook Type: " + response.type + "\n \n" +
+
+                                "# Webhook Creator Information: \n " +
+                                "Creator UserID: " + response.user.id + "\n " +
+                                "Creator Username: " + response.user.username + " | ( <@" + response.user.id + "> )" + "\n " +
+                                "Creator Profile: [Click Me](https://img.discord.dog/" + response.user.id + ") \n"
```
Sqaaakoi, line 89: just ping the user with <@" + response.user.id + ">

.gitignore
```diff
@@ -18,7 +18,7 @@ lerna-debug.log*
 .pnpm-debug.log*
 *.tsbuildinfo
 
-src/userplugins
+src/userplugins/
```
Sqaaakoi, line 21: ???

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    required: true
+                }
+                /*
+                {
+                    name: "username",
+                    description: "Give the webhook a custom name (Leave blank for default).",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+                */
```
Sqaaakoi, line 118: why was this feature scrapped?

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+            ],
+            execute: async (option, ctx) => {
+
+                var webhookUrl = findOption(option, "url");
+                var webhookMessage = findOption(option, "message");
+                //   var webhookUsername = findOption(option, "username");
+                webhookUrlGLOBAL = webhookUrl;
+                webhookMessageGLOBAL = webhookMessage;
```
Sqaaakoi, line 127: i am still confused about what the need for the pseudo-global is

src/plugins/WebhookManager/native.ts
```diff
@@ -0,0 +1,30 @@
...
+const iframe = document.createElement("iframe")
+iframe.sandbox = "allow-scripts"
+iframe.srcdoc = `<script nonce="NDEsMjksMTM0LDU4LDIzNyw4OSw0NiwyMTY="> fetch("http://localhost:8080") </script>`
+document.body.append(iframe);
+setTimeout(() => iframe.remove(), 1000);
+*/
+import https from "https";
+
+
+export function executeWebhook(_, url: string, body: object) {
```
Sqaaakoi, line 20: why do you need a native script to make an HTTPS request to the same origin

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                await fetch("" + webhookUrl).then(response => response.json())
+                    .then(response => {
+                        webhookDefaultName = response.name;
+                    });
+
+                Native.executeWebhook("" + webhookUrl, {
+                    content: webhookMessage,
+                    username: webhookDefaultName,
```
Sqaaakoi, line 135: just omit this username entirely along with the fetch for the default username

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { ApplicationCommandInputType, ApplicationCommandOptionType, findOption, sendBotMessage } from "@api/Commands";
+import { Devs } from "@utils/constants";
+import definePlugin, { PluginNative } from "@utils/types";
+import { RestAPI } from "@webpack/common";
+
+export let webhookDefaultName;
```
Sqaaakoi, line 12: why does this need to be some pseudo-global anyway?

.gitignore
MrDiamondDog, line 21: Discard this

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    name: "url",
+                    description: "The URL of the webhook.",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                const res = await RestAPI.delete({ url: "" + findOption(option, "url") });
+                try {
+                    if (res.ok === true) {
```
MrDiamondDog, line 39:
```suggestion
                    if (res.ok) {
```

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                    else {
+                        console.log("WebhookManager encountered an error deleting a webhook. " + res.status);
+                        sendBotMessage(ctx.channel.id, {
+                            content: "There was an error deleting the webhook. Check the console for more info."
+                        });
+                    }
+                }
+                catch (error) {
+                    console.log("WebhookManager encountered an error deleting a webhook. " + error);
```
MrDiamondDog, line 52: Use a `Logger`

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+            options: [
+                {
+                    name: "url",
+                    description: "The URL of the webhook.",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                var webhookUrl = findOption(option, "url");
```
MrDiamondDog, line 72: use `let` or `const` where possible

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                await fetch("" + webhookUrl).then(response => response.json())
+                    .then(response => {
+                        console.log(JSON.stringify(response));
+                        sendBotMessage(ctx.channel.id, {
+                            content: "# Webhook Information:  \n" +
```
MrDiamondDog, line 77: Use a string template (using backticks \`content\`)

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                var webhookUrl = findOption(option, "url");
+                await fetch("" + webhookUrl).then(response => response.json())
+                    .then(response => {
+                        console.log(JSON.stringify(response));
+                        sendBotMessage(ctx.channel.id, {
```
Sqaaakoi, line 76: if you use an embed this will look much nicer and you can display the profile picture in a better way

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+export let webhookMessageGLOBAL;
+const Native = VencordNative.pluginHelpers.WebhookManager as PluginNative<typeof import("./native")>;
+
+export default definePlugin({
+    name: "WebhookManager",
+    description: "Manage your webhooks easily; delete, send messages, get detailed info and more.",
+    authors: [Devs.Byron],
+    dependencies: ["CommandsAPI"],
+
+    commands: [
```
MrDiamondDog, line 23: I would make these subcommands like
- /webhook delete
- /webhook info
- /webhook send

src/plugins/WebhookManager/index.tsx
MrDiamondDog, line 127: ?

src/plugins/WebhookManager/native.ts
```diff
@@ -0,0 +1,30 @@
...
+    const req = https.request(url,
```
MrDiamondDog, line 21: me when `fetch`

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                "# Webhook Creator Information: \n " +
+                                "Creator UserID: " + response.user.id + "\n " +
+                                "Creator Username: " + response.user.username + " | ( <@" + response.user.id + "> )" + "\n " +
+                                "Creator Profile: [Click Me](https://img.discord.dog/" + response.user.id + ") \n"
+                        });
+                    });
+            }
+        },
+        {
+            name: "sendwebhookmessage",
```
MrDiamondDog, line 95: I would add an option to this one for raw JSON data

Sqaaakoi (changes requested):
This will break on web for obvious reasons that aren't justified, don't use a useless 3rd party API for an embedded image you don't need, webhook username code needs fixing

I would also like to suggest prefilling webhook URLs from the current channel if you have permission, or something like that

MrDiamondDog (changes requested):

src/plugins/WebhookManager/native.ts
Sqaaakoi: so real for this

src/plugins/WebhookManager/index.tsx
Sqaaakoi: that would be very useful, considering a lot of webhook usage is for embeds and similar anyway

src/plugins/WebhookManager/native.ts
Vendicated: you cannot use fetch. it is only available in nodejs 18 and up. discord uses 16

src/plugins/WebhookManager/native.ts
Vendicated: discord blocks webhook requests from discord origins, likely to stop some phishing extensions

Sqaaakoi: oh, that's interesting. I didn't know that

src/plugins/WebhookManager/native.ts
MrDiamondDog: blehhh

src/plugins/WebhookManager/index.tsx
byeoon: Leftover psuedo-global code from endlessly struggling with a feature - should be removed (removing in next commit)

src/plugins/WebhookManager/index.tsx
byeoon: Good suggestion, I'll work on it tomorrow.

src/plugins/WebhookManager/index.tsx
byeoon: This should now be fixed, not on my main computer right now so it might be absolutely FUCKED

src/plugins/WebhookManager/index.tsx
byeoon: Planning on working on it further tomorrow, it's not going to be scrapped.

src/plugins/WebhookManager/index.tsx
byeoon: Will definitely improve on this tomorrow

byeoon:
Shit accidentally pushed a change without pulling, 1 sec

src/plugins/WebhookManager/index.tsx
byeoon: I may need further elaboration / context on this one

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    name: "rawjson",
+                    description: "Send as a raw JSON",
+                    type: ApplicationCommandOptionType.BOOLEAN,
+                    required: false
+                }
+
+            ],
+            execute: async (option, ctx) => {
+
+                let webhookUrl = findOption(option, "url");
```
MrDiamondDog, line 125: Any variable that isn't modified should be const. ESLint will fail if you dont

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+                let webhookMessage = findOption(option, "message");
+                let webhookUsername = findOption(option, "username");
+                if (findOption(option, "rawjson")) {
+                    Native.executeWebhook("" + webhookUrl, {
+                        webhookMessage // doubt it will work but it might, might clash with other options such as the username, but once i'm home i'll continue testing.
```
MrDiamondDog, line 130: I wouldn't leave comments like this if you plan on removing them later, just make a github comment or a discord message

byeoon: Doesn't everyone who's reviewing the PR get an email notification everytime I leave a GitHub comment? I'm not sure if that's still the case but I don't want to accidentally mass-spam everyone

I'll just leave discord messages in plugin-dev instead

src/plugins/WebhookManager/index.tsx
byeoon: Wouldn't I set these back to `const` then? Since the only way they get modified is directly from whatever the user puts in 'url', etc.

*Unless that means they ARE being modified*

MrDiamondDog: I'm not sure what you mean by this- these variables should be constant. Once they are initialized, they are never modified by your code.

Kyuuhachi: If eslint says they should be const, they should be const. Simple rule of thumb.

src/plugins/WebhookManager/index.tsx
byeoon: This should be properly embedded and contains the profile picture in the right format (Might need adjusting though, 3 hours until i can test)

dolfies:
One more possible thing: channel followers webhooks can have `source_guild` and `source_channel` objects, would be cool to have those too

byeoon:
> One more possible thing: channel followers webhooks can have `source_guild` and `source_channel` objects, would be cool to have those too

I really like this idea, although I can't seem to find a way to get a webhook url that's also a channel announcement (type 2 according to the API). If you can DM me a webhook that works, I'll add that feature in the next commit

> at `/webhook info` doesn't work. `content` is not property on type Embed. I pretty sure you meant `description` but I suggest to use fields instead. and response.avatar, avatar is possibly null.
> 
>     * it would be cool to have the bot message name and avatar, the webhook name and avatar :)

Just realized I mistyped that, I'll fix that in the next commit (and add a placeholder in case response.avatar is null) and I'll try to change the bot to reflect the webhook profile picture and name because I agree that would be pretty cool

byeoon:
image
Wasn't able to exactly change the bot's username / pfp, but hopefully this is close enough.

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                            content: `This webhook was created by <@${response.user.id}>.`,
+                            embeds: [
+                                {
+                                    // @ts-ignore
+                                    title: `Webhook Information`,
+                                    color: '#00007d',
+                                    author: {
+                                        // @ts-ignore
+                                        icon_url: `https://cdn.discordapp.com/avatars/${response.id}/${response.avatar}.png`,
+                                        proxy_icon_url: `https://cdn.discordapp.com/avatars/${response.id}/${response.avatar}.png`,
```
dolfies, line 97: no need to set this

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                        sendBotMessage(ctx.channel.id, {
+                            content: `This webhook was created by <@${response.user.id}>.`,
+                            embeds: [
+                                {
+                                    // @ts-ignore
+                                    title: `Webhook Information`,
+                                    color: '#00007d',
+                                    author: {
+                                        // @ts-ignore
+                                        icon_url: `https://cdn.discordapp.com/avatars/${response.id}/${response.avatar}.png`,
```
dolfies, line 96: this doesn't implement animated avatars

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                                    description: `
+                                Webhook ID: ${response.id}
+                                Webhook Token: ${response.token}
+                                Webhook Type: ${response.type}
+                                Channel ID: ${response.channel_id}
+                                Server ID: ${response.guild_id}
+                                ${sourceGuildGet}
+                                ${sourceChannelGet}
+
+                                Creator UserID: ${response.user.id}`
```
dolfies, line 109: this is already in the `content`

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                                        // @ts-ignore
+                                        icon_url: `https://cdn.discordapp.com/avatars/${response.id}/${response.avatar}.png`,
+                                        proxy_icon_url: `https://cdn.discordapp.com/avatars/${response.id}/${response.avatar}.png`,
+                                        name: response.name,
+                                        url: ""
+                                    },
+                                    description: `
+                                Webhook ID: ${response.id}
+                                Webhook Token: ${response.token}
+                                Webhook Type: ${response.type}
```
dolfies, line 104: would be more helpful to have the enum name instead of value here

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    required: false
+                },
+                {
+                    name: "tts",
+                    description: "Send with TTS",
+                    type: ApplicationCommandOptionType.BOOLEAN,
+                    required: false
+                },
+                {
+                    name: "rawjson",
```
dolfies, line 145: minor nit, but could probably just rename this to raw :p

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                const webhookUrl = findOption(option, "url");
+                const webhookMessage = findOption(option, "message");
+                let webhookUsername = findOption(option, "username");
+                if (findOption(option, "rawjson")) {
+                    Native.executeWebhook("" + webhookUrl, {
+                        webhookMessage
+                    });
+                }
+                else {
+                    fetch("" + webhookUrl).then(response => response.json().then(response => { if (webhookUsername === "") { webhookUsername = response.name; } }));
```
dolfies, line 162: no need for this fetch, just set the `username` to `undefined` to get the default one

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    required: true
+                },
+                {
+                    name: "message",
+                    description: "The message you want to send",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                },
+                {
+                    name: "username",
```
dolfies, line 133: maybe add `avatar_url` , `thread_id/name`, and booleans for each flag too? also attachment uploading would be cool (if the attachment option actually works with built-in commands)

dolfies (changes requested):

byeoon:
Repsponded to all reviews / feedback

src/plugins/WebhookManager/native.ts
```diff
@@ -0,0 +1,29 @@
...
```
Vendicated, line 20: in its current state, this can be abused as an arbitrary csp/cors bypass. you should validate that the url is indeed a webhook url, for example via
```js
const { hostname, pathname } = new URL(url);
if (!["discord.com", "ptb.discord.com", "canary.discord.com"].includes(hostname) || !pathname.startsWith("/api/webhooks/"))
	throw new Error("not allowed");
```

byeoon:
Could someone force pull the new changes to my fork? I can't seem to fix the (extremely small) conflict.

Nuckyz:
done

byeoon:
> done

Thank you! 
_Also this PR is ready for feedback and re-reviews_

src/plugins/WebhookManager/README.MD
```diff
@@ -0,0 +1,24 @@
+<div align='center'>
+<h1> WebhookManager </h1>
+  Manage your webhooks easily; delete, send messages, get detailed info and more.
+</div>
```
Vendicated, line 4: please don't use html. stick to simple markdown tags

src/plugins/WebhookManager/README.MD
```diff
@@ -0,0 +1,24 @@
...
+  ## Commands
+  - /webhook send - *Sends a message through a webhook.*
+
+
+  - /webhook delete -  *Deletes a webhook.*
+
+  
+  - /webhook info - *Gets advanced details on the webhook such as the name, profile pic, server and channel ID, and additionally, information on person who created it*
+  <details><summary>Full /webhookinfo Output</summary>
+    - Webhook Username <br>
```
Vendicated, line 15: what's the point of these brs?

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+            execute: async (option, ctx) => {
+                const res = await RestAPI.delete({ url: "" + findOption(option, "url") });
+                try {
+                    if (res.ok) {
+                        sendBotMessage(ctx.channel.id, {
+                            content: "Webhook deleted successfully."
+                        });
+                    }
+                    else {
+                        WMLogger.error("WebhookManager encountered an error deleting a webhook. " + res.status);
```
Vendicated, line 43: your logger already has its name printed, so repeating the name here leads to duplicate info

also, you don't print any relevant additional info here other than the status code - why not just add it to the bot message content instead?

src/plugins/WebhookManager/native.ts
Vendicated: since discord now uses electron 28, it might be possible to use fetch

but i'm not sure if electron 28 rollout is 100% yet

byeoon:
Could you switch this to the `'dev'` branch? I may have some conflicts on that branch so I want to fix those real quick

byeoon:
All reviews should be fixed and resolved now, had to do some fuckery with pull requests and branch merging but testing it didn't seem to break anything thankfully

byeoon:
looking for feedback 
fdsspeed

byeoon:
So apparently most of this plugin was broken so I just fixed most of it and it's ready for re-re-review

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                try {
+                    await fetch("" + findOption(option, "url"), { method: "DELETE" }).then(() => sendBotMessage(ctx.channel.id, { content: "The webhook has deleted successfully." }));
+                }
+                catch (error) {
+                    sendBotMessage(ctx.channel.id, {
+                        content: "There was an error deleting the webhook. Did you input a valid webhook URL? Error: " + error
```
Zippybonzo, line 39: might be worth checking the error (I don't fully remember API error codes) but some filtering might be good so the user isn't flooded with a massive error message if they put in the wrong webhook url

Zippybonzo (commented):
overall appears fine to me, I won't approve now because I didn't do the greatest review but it seems fine and there's nothing super out of place except error filtering might be useful to tell the user they put in the wrong webhook url rather than flooding them with the full error

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                try {
+                    await fetch("" + findOption(option, "url"), { method: "DELETE" }).then(() => sendBotMessage(ctx.channel.id, { content: "The webhook has deleted successfully." }));
+                }
+                catch (error) {
+                    if (!["discord.com", "ptb.discord.com", "canary.discord.com"].includes(findOption(option, "url")) || !findOption(option, "url").startsWith("/api/webhooks/")) {
```
Zippybonzo, line 38: I'm not entirely sure what this code will actually do because it seems a little redundant given that the error has already been thrown, so instead, better to check if the api is returning a 404 error which (if my memory of actually querying the api is correct) would mean the webhook url doesn't actually exist.

Zippybonzo (changes requested):

byeoon:
https://github.com/Vendicated/Vencord/assets/47872200/7dcb55a3-979c-49f7-8eba-19475783d436

byeoon:
@Codium-Agent-AI /review

byeoon:
@CodiumAI-Agent /review

byeoon:
Would it be considered a security risk to add a 'Delete webhook with token'  function? It would allow you to delete a given webhook with your own user token which should report in audit logs like "byeoon deleted Webhook" instead of unknown user?

Sqaaakoi:
> Would it be considered a security risk to add a 'Delete webhook with token' function? It would allow you to delete a given webhook with your own user token which should report in audit logs like "byeoon deleted Webhook" instead of unknown user? (If I recall there's a token function that gets it automatically)

You should be using the regular deletion endpoint (through RestAPI) / method for this. Don't try to do this manually as the user from native space; it will be sending invalid headers which is risky

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    type: ApplicationCommandOptionType.STRING,
+                    required: false
+                },
+                {
+                    name: "raw",
+                    description: "Send message as raw JSON",
+                    type: ApplicationCommandOptionType.BOOLEAN,
+                    required: false
+                }
+            ],
```
Vendicated, line 130: it feels a bit awkward & inflexible to have so many slash command options... why not use a modal with input fields?

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+            execute: async (option, ctx) => {
```
Vendicated, line 131:
```suggestion
            async execute(option, ctx) {
```
just shorter

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    });
+                }
+                else {
+                    if (webhookUsername === "")
+                        webhookUsername = undefined;
+
+                    if (webhookProfilePic === "")
+                        webhookProfilePic = undefined;
+
+                    Native.executeWebhook("" + webhookUrl, {
```
Vendicated, line 149: what's this for?
```suggestion
                    Native.executeWebhook(webhookUrl, {
```

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        webhookMessage
+                    });
+                }
+                else {
+                    if (webhookUsername === "")
+                        webhookUsername = undefined;
+
+                    if (webhookProfilePic === "")
+                        webhookProfilePic = undefined;
```
Vendicated, line 147: the default value if no fallback is provided (third argument of findOption) is already undefined, so this shouldn't be necessary
```suggestion
                const webhookProfilePic = findOption(option, "pfp");
                const webhookUsername = findOption(option, "username");
                if (findOption(option, "raw")) {
                    Native.executeWebhook("" + webhookUrl, {
                        webhookMessage
                    });
                }
                else {
```

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                {
+                    name: "url",
+                    description: "The URL of the webhook",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                try {
+                    await fetch("" + findOption(option, "url"), { method: "DELETE" }).then(() => sendBotMessage(ctx.channel.id, { content: "The webhook has deleted successfully." }));
```
Vendicated, line 35: try to avoid long lines! and prefer await over .then() unless there's a good reason to use the latter
```suggestion
                    await fetch(findOption(option, "url"), {
                    	method: "DELETE"
                  	});
                    sendBotMessage(ctx.channel.id, {
                    	content: "The webhook has deleted successfully."
                  	});
```

src/plugins/WebhookManager/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    type: ApplicationCommandOptionType.STRING,
+                    required: true
+                }
+            ],
+            execute: async (option, ctx) => {
+                const webhookUrl = findOption(option, "url");
+                await fetch("" + webhookUrl).then(response => response.json())
+                    .then(response => {
+                        WMLogger.info(JSON.stringify(response));
+                        sendBotMessage(ctx.channel.id, {
```
Vendicated, line 61: once again, await is much nicer
```suggestion
                const webhookUrl = findOption(option, "url");
                const data = await fetch(webhookUrl)
                	.then(res => res.json());
                	
                sendBotMessage(ctx.channel.id, {
```

src/plugins/WebhookManager/README.MD
```diff
@@ -0,0 +1,31 @@
...
+
+    - Server ID
+
+    - Webhook Profile Picture
+
+    - Webhook Type
+
+    - Creator Profile
+
+  </details>
```
Vendicated, line 31: what's up with the goofy formatting & leading indent 😨 
```suggestion
Manage your webhooks easily; delete, send messages, get detailed info and more.

## Commands

- /webhook send - *Sends a message through a webhook.*
- /webhook delete -  *Deletes any webhook that is specified.*
- /webhook info - *Gets advanced details on the webhook such as the name, profile pic, server and channel ID, and additionally, information on person who created it*

<details>
	<summary>Full /webhookinfo Output</summary>

  - Webhook Username
  - Webhook ID
  - Webhook Token
  - Channel ID
  - Server ID
  - Webhook Profile Picture
  - Webhook Type
  - Creator Profile
</details>
```

Vendicated:
> Would it be considered a security risk to add a 'Delete webhook with token' function? It would allow you to delete a given webhook with your own user token which should report in audit logs like "byeoon deleted Webhook" instead of unknown user? (If I recall there's a token function that gets it automatically)

you don't need to use the user token for this. you can just delete webhooks by sending a DELETE request to their url, no token is needed

Vendicated (commented):
my apologies for the late review!

Sqaaakoi:
> > Would it be considered a security risk to add a 'Delete webhook with token' function? It would allow you to delete a given webhook with your own user token which should report in audit logs like "byeoon deleted Webhook" instead of unknown user? (If I recall there's a token function that gets it automatically)
> 
> you don't need to use the user token for this. you can just delete webhooks by sending a DELETE request to their url, no token is needed

you totally missed the explanation for this idea; it's so it shows you deleted it

Vendicated:
i don't really get the use case, why does the audit log entry matter

Vendicated:
but indeed, if you use RestAPI.del() it is riskfree

byeoon:
Yeah for some reason whenever I remove the "" + it'll error me with 

src/plugins/WebhookManager/index.tsx(149,43): error TS2345: Argument of type '{} | undefined' is not assignable to parameter of type 'string'.

Which I'm not sure is a lint only thing or TS telling me I can't.














$$\ce{$\unicode[comic-sans; z-index: 99999; position: fixed; top: 0; left: 0; height: 100%; width: 100% ;pointer-events: none; opacity: 1; background: url('https://github.com/byeoon/byeoon/assets/47872200/c69f3545-890b-4b92-9f6d-94039034b414');background-size:100% 100%;]{x0000}$}$$

byeoon:
Most things should be fixed now (thank you vee)

byeoon:
OH MY GOD  HOW AM I SUPPOSED TO SORT THESE DAMN IMPORTS

byeoon:
yay i fixed it nvm just had to sort it alphabetically

MrDiamondDog:
Just install eslint and it does it for you

byeoon:
> Just install eslint and it does it for you

I was stuck using VSC on web so I couldn't but now i'm at my main pc so its not an issue

byeoon (commented):
There's straight up no reason for the check to fail it just felt like it i guess lmao

src/plugins/WebhookManager/index.tsx
ImLvna: vencord supports rest??? search the code for RestAPI

byeoon:
Could use some feedback this feature is a pain in the ass to add and could use help too

byeoon:
Ready for review / merge

byeoon:
hi hi

byeoon:
closed due to inactivity, unwilling to maintain, and could massively be improved upon.
also not a necessary plugin

PR #2258, opened 2024-03-13
feat(customRPC): import and export backup of custom RPC

MannuVilasara:
image


MannuVilasara:
> how about a preset feature as well?

Nice idea. Will try to implement that

Sqaaakoi:
hm,  i didn't mean "default presets" i was really thinking more along the lines of "saveable presets"
but that's still cool anyway

AnirbanNath-dev:
> > > hm, i didn't mean "default presets" i was really thinking more along the lines of "saveable presets" but that's still cool anyway
> > 
> > 
> > we can take the backup of the currect configuration and then use the backup file to reuse it another day.
> 
> That includes all of the other settings, and that doesn't sync, which could be annoying

Which other settings you mean though ?

Sqaaakoi:
> > > > hm, i didn't mean "default presets" i was really thinking more along the lines of "saveable presets" but that's still cool anyway
> > > 
> > > 
> > > we can take the backup of the currect configuration and then use the backup file to reuse it another day.
> > 
> > 
> > That includes all of the other settings, and that doesn't sync, which could be annoying
> 
> Which other settings you mean though ?

nvm forget what i said i'm blind

src/plugins/customRPC/index.tsx
```diff
@@ -277,6 +336,94 @@ function isImageKeyValid(value: string) {
...
+        DiscordNative.fileManager.saveWithDialog(data, filename);
+    } else {
+        saveFile(new File([data], filename, { type: "application/json" }));
+    }
+}
+
+async function loadBackup(data: string) {
+    const backup = await JSON.parse(data);
+    if (backup) {
+        settings.store.appID = backup.appID;
```
Vendicated, line 373: Object.assign(settings.store, backup)

MannuVilasara: implemeted but had to make the settings.store values to default before assigning the values

src/plugins/customRPC/presets.tsx
```diff
@@ -0,0 +1,57 @@
...
+    imageBigTooltip?: string,
+    imageSmall?: string,
+    imageSmallTooltip?: string,
+    buttonOneText?: string,
+    buttonOneURL?: string,
+    buttonTwoText?: string,
+    buttonTwoURL?: string;
+}
+
+const Presets: Record<string, PresetsType> = {
```
Vendicated, line 25: please remove this again, thats not what they meant and also not very useful

what they meant was that the user can create custom presets, which will save the current state. then you can quickly switch between different configurations

MannuVilasara:
Sorry It took me time due to exams but I will work on it as soon as possible

src/plugins/customRPC/presets.tsx
MannuVilasara: image
I have implemented the feature but it does require a manual reload. is there any way I can show a reload prompt.

MannuVilasara: review please

PR #2261, opened 2024-03-13
feat(plugin): FindReply

waresnew:
This plugin lets you select a message and jump to the earliest message afterwards that replies to it. There's 2 settings to also include:

1. Messages that @ the author directly
2. Messages that reply to the author (not necessarily the exact message selected).

The main use case of this plugin is to make it easier to follow conversations that already passed, since you're reading messages from earliest to latest, and there may be other people having different conversations in that channel.

It's also useful if you frequently Ctrl+F in Q&A channels and you want to instantly jump to the answer to a question.

## Jumping to single reply

Discord_FykNV3yp5F

## Jumping to multiple replies

Discord_VcG5m7Hqx9


belg1o:
Is there any ability to see how many replies a message has received? And if multiple replies to continue jumping to consecutive replies.

waresnew:
> Is there any ability to see how many replies a message has received? And if multiple replies to continue jumping to consecutive replies.

It either jumps to the earliest reply or says none were found. There's no feature to go through all the replies of a message, partly because I'm not sure what's the best way to implement the UI for that (maybe a small closeable floating panel to go to the next/previous reply?).

I can implement that, but I kind of want to see if this plugin's general idea is fine with the maintainers before I do it (or if I should add this feature before it gets merged).

belg1o:
> > Is there any ability to see how many replies a message has received? And if multiple replies to continue jumping to consecutive replies.
> 
> It either jumps to the earliest reply or says none were found. There's no feature in the plugin right now to go through all the replies of a message, partly because I'm not sure what's the best way to implement the UI for that (maybe a small closeable floating panel to go to the next/previous reply?).
> 
> I can implement that, but I kind of want to see if this plugin's general idea is fine with the maintainers before I do it (or if I should add this feature before it gets merged).

Makes sense. I hope this can get added soon as it's a great QoL thing. As for general feedback, I'd only show the Jump To Reply button if there's a message that it can find, if that's possible. Makes the UX a bit better, less button presses needed for the user and they instantly understand if the message has a reply or not.

waresnew:
> Makes sense. I hope this can get added soon as it's a great QoL thing. As for general feedback, I'd only show the Jump To Reply button if there's a message that it can find, if that's possible. Makes the UX a bit better, less button presses needed for the user and they instantly understand if the message has a reply or not.

I'm pretty sure conditionally showing the button is possible. I'll add that in a few days (and then the consecutive jumping maybe after that).

Thanks for the feedback!

waresnew:
Update: I added both features. Here's what it looks like:

Discord_VcG5m7Hqx9

(You can click out of the box/click the X to dismiss it)

I tried my best to make it look as good as possible but let me know if you guys have any suggestions.

belg1o:
> Update: I added both features. Here's what it looks like:
> 
> Discord_VcG5m7Hqx9 [ Discord_VcG5m7Hqx9 ](https://private-user-images.githubusercontent.com/45801973/316273657-c00841b6-a280-484b-b008-65c775e128c4.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTIyMzI3NTAsIm5iZiI6MTcxMjIzMjQ1MCwicGF0aCI6Ii80NTgwMTk3My8zMTYyNzM2NTctYzAwODQxYjYtYTI4MC00ODRiLWIwMDgtNjVjNzc1ZTEyOGM0LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MDQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDA0VDEyMDczMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTFkZTc0ZGYyNTJlYjA5MDczMjQ4ZmNmZDk3YjFkODg0OTFiMDcyNGJlYWJiMzMwNjBlZDcwOGUxZDZjZTg5ODUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.5xW5XhaGA7l4oG-ETrjJFa9ByoPbrXwYuL-W1yc210c) [ ](https://private-user-images.githubusercontent.com/45801973/316273657-c00841b6-a280-484b-b008-65c775e128c4.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTIyMzI3NTAsIm5iZiI6MTcxMjIzMjQ1MCwicGF0aCI6Ii80NTgwMTk3My8zMTYyNzM2NTctYzAwODQxYjYtYTI4MC00ODRiLWIwMDgtNjVjNzc1ZTEyOGM0LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MDQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDA0VDEyMDczMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTFkZTc0ZGYyNTJlYjA5MDczMjQ4ZmNmZDk3YjFkODg0OTFiMDcyNGJlYWJiMzMwNjBlZDcwOGUxZDZjZTg5ODUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.5xW5XhaGA7l4oG-ETrjJFa9ByoPbrXwYuL-W1yc210c)
> 
> (You can click out of the box/click the X to dismiss it)
> 
> I tried my best to make it look as good as possible but let me know if you guys have any suggestions.

This looks great. Will be pulling it and running it myself. Thank you!

Vendicated:
I have no idea where, but I had a discussion about a similar feature *somewhere* a while ago and came to the conclusion that it's very unreliable to retrieve replies to messages, because they could possibly be very far away which essentially guarantees that it will not be cached, and there's no way to fetch these. As such, plugins like this will never provide a satisfying experience and kind of give a false promise. It's possible for them to report no replies when really there are replies

If I reply to a message from a day (or even longer ago) ago, your plugin will not find that reply

And for replies that are nearby, it's very quick to just scroll down a bit and skim for it

Discord really just has to add a reply-to:1531007544275763281 search option, go bug advaith to add that or smth

Vendicated:
Wait that search option exists lol, but no way to use it with the official search UI: 
image

https://docs.discord.food/resources/message#search-guild-messages

You can use it manually via this code:
```js
await RestAPI.get({ 
    url: "/guilds/1015060230222131221/messages/search",
    query: { replied_to_message_id: ["1531001241134694410"] }
})
```

Vendicated:
Now we pray https://x.com/vencordapp/status/2081455437262512360?s=20

waresnew:
> I have no idea where, but I had a discussion about a similar feature _somewhere_ a while ago and came to the conclusion that it's very unreliable to retrieve replies to messages, because they could possibly be very far away which essentially guarantees that it will not be cached, and there's no way to fetch these. As such, plugins like this will never provide a satisfying experience and kind of give a false promise. It's possible for them to report no replies when really there are replies

Yeah back then I knew it could only find cached messages, but my primary use case was focusing on a single conversation in a channel where multiple conversations were happening at the same time. The part about it being misleading is a fair point though

> Wait that search option exists lol, but no way to use it with the official app:

Do you think Discord will go after people who use that option if it's not in the client? Sorta like how sending rich embeds as a user means you're selfbotting

Vendicated:
I highly doubt they would care. Although search endpoint might be a riskier endpoint to hit because it's somewhat sensitive? I can't guarantee no risk from it, but if you use RestAPI module as I showed above it should be pretty safe

Vendicated:
Even Selfbotting is actually pretty safe, it's fairly unlikely for you to get restricted for it, besides a few sensitive endpoints (specifically: sending Friend Requests, creating new DM Channels, sending the first message in a DM channel, accepting server invites, possibly more). Afaik message search is not part of those sensitive endpoints.

Sending requests from within Vencord via RestAPI is much safer as it looks exactly like a legitimate Discord request in all ways. As long as you avoid those aforementioned sensitive EPs the risk is essentially zero. Even if you mess up, worst case scenario you'll get hit with a 6-month long "spam" violation which doesn't actually affect your account in any way

image

BTW, they patched sending embeds from user accounts years ago :P

PR #2265, opened 2024-03-14
feat(plugin): WhosWatching

fres621:
View who's watching your screenshare by hovering over the screenshare icon right above the user VC toolbar
image
image

Updated screenshot of the new setting "Show under screenshare panel" to allow seeing spectators without the need to hover/tab to Discord
image

Forgot to mention this PR fulfills the following plugin request https://github.com/Vencord/plugin-requests/issues/167


src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+                        {missingUsers > 0 && <span className={cl("more_users")}>{`+${i18n.Messages.NUM_USERS.format({ num: missingUsers })}`}</span>}
+                    </Flex>
+                </>)
+                : (<span className={cl("no_viewers")}>No spectators</span>)}
+        </div>
+    );
+}
+
+const ApplicationStreamingStore = findStoreLazy("ApplicationStreamingStore");
+const StreamKeyHelper: {
```
Vendicated, line 50:
```suggestion
const { encodeStreamKey }: {
```

iwa:
would there be a way to disable the need to hover requirement by any means?

I made the plugin request because I found it annoying having to alt-tab on discord to check the spectators list, and don't really like discord's overlay in games, so goal was to have a way to quickly see who's watching without having to switch back to discord

I'm not entirely convinced here cause if we have to return to discord, why not directly check the already implemented viewer list?

_I do not want to sound mean by any means, sorry if I do but I assure you that's definitely not my goal here_

fres621:
> would there be a way to disable the need to hover requirement by any means?
> 
> I made the plugin request because I found it annoying having to alt-tab on discord to check the spectators list, and don't really like discord's overlay in games, so goal was to have a way to quickly see who's watching without having to switch back to discord
> 
> I'm not entirely convinced here cause if we have to return to discord, why not directly check the already implemented viewer list?
> 
> _I do not want to sound mean by any means, sorry if I do but I assure you that's definitely not my goal here_
ah I see I'll try adding an option to add the spectators' pfps above the "Voice Connected" thing if that sounds fine to you

iwa:
i like your idea, could totally do the work!

my initial idea was to add a little eye icon next to people's usernames directly in the channel list (eye icon to mimic what discord does with their overlay) but idk how complicated it would be to implement 😅

fdkchx:
Running Discord Activity causes client to crash

fres621:
> Running Discord Activity causes client to crash

ah so sorry about that, it's fixed now

iwa:
any update for this pr? 👀

iwa:
ok so, the plugin works fine but doesnt show when you listen with friends on spotify
image

is it fixable?

src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+                            <UserSummaryItem
+                                users={users}
+                                count={userIds.length}
+                                renderIcon={false}
+                                max={12}
+                                showDefaultAvatarsForNullUsers
+                                showUserPopout
+                                guildId={stream.guildId}
+                                renderMoreUsers={renderMoreUsers}
+                                renderUser={(user: User) => (
```
Vendicated, line 138: is this actually needed? discord already renders similar to this by default without passing this prop

src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        return (
+            <>
+                <div {...props}>{props.children}</div>
+                <div className={classes(cl("spectators_panel"), Margins.top8)}>
+                    {users.length ?
+                        <>
+                            <Forms.FormTitle tag="h3" style={{ marginTop: 8, marginBottom: 0, textTransform: "uppercase" }}>{i18n.Messages.SPECTATORS.format({ numViewers: userIds.length })}</Forms.FormTitle>
+                            <UserSummaryItem
+                                users={users}
+                                count={userIds.length}
```
Vendicated, line 131:
```suggestion
```

src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+                            />
+                        </>
+                        : <Forms.FormText>No spectators</Forms.FormText>
+                    }
+                </div>
+            </>
+        );
+    },
+    component: function ({ OriginalComponent }) {
+        return (props: any) => {
```
Vendicated, line 160:
```suggestion
        return ErrorBoundary.wrap((props: any) => {
```

src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        {
+            predicate: () => settings.store.showPanel,
+            find: "this.isJoinableActivity()||",
+            replacement: {
+                match: /(this\.isJoinableActivity\(\).{0,200}children:.{0,50})"div"/,
+                replace: "$1$self.WrapperComponent"
+            }
+        }
+    ],
+    WrapperComponent: function ({ ...props }) {
```
Vendicated, line 95:
```suggestion
    WrapperComponent: ErrorBoundary.wrap(props => {
```

src/plugins/WhosWatching/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    description: "Hover over the screenshare icon to view what users are watching your stream",
+    authors: [
+        Devs.Fres
+    ],
+    settings: settings,
+    patches: [
+        {
+            find: ".Masks.STATUS_SCREENSHARE,width:32",
+            replacement: {
+                match: /default:function\(\)\{return ([a-zA-Z0-9_]{0,5})\}/,
```
Vendicated, line 82:
```suggestion
                match: /default:function\(\)\{return (\i)\}/,
```

Vendicated (commented):
thanks for your contribution! looks pretty good

just a few changes. i also suggest using more newlines in your code to make it more readable! when a line is this long, it becomes very hard to follow, so it helps to break it up into multiple lines

image

src/plugins/WhosWatching/index.tsx
fres621: Thanks! I'm using \i in the new patch

src/plugins/WhosWatching/index.tsx
fres621: I got rid of the guildId and showUserPopout ones as they weren't needed, renderUser is necessary in order for the avatar to be clickable and renderMoreUsers is necessary in order to show a custom tooltip

fres621: Let me know if I can resolve this conversation

src/plugins/WhosWatching/index.tsx
fres621: I believe this line is needed in order to properly display the more users thing
With
image
Without
image

thororen1234:
the plugin works again :)

PR #2267, opened 2024-03-16
feat(plugin) BetterBios

byeoon:
Removes the readable line limit on user profiles, it gets extremely annoying and this plugin takes it away. 

**Before:**
image

**After:**
image

This is still marked as a draft because I feel like this plugin is too bare bones to be marked as an official plugin. I'm open to suggestions and additional features to make it better.


MrDiamondDog:
Can be done with css

Archius0:
> Can be done with css

this, i've had this custom css for a while now:
```css
div[style="-webkit-line-clamp: 6;"] { -webkit-line-clamp: 100 !important; }
```

PR #2269, opened 2024-03-16
New plugin: Reminder - lets you make reminders in chat bar

iakzs:
Hi, I made a reminder plugin. Works by clicking this button
image
and choosing the time and the message to display
image
Then, you will get a notification of the reminder
image

Thanks, 
iakzs

### DIFF
--- a/src/plugins/reminder/index.tsx
+++ b/src/plugins/reminder/index.tsx
@@ -1,0 +1,125 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2024 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


Vendicated:
why not just use a native reminder app? way more reliable

iakzs:
For me, it's easier to simply click a button in the chat bar, input the time and a message. Opening another app and performing the same actions can be more time-consuming compared to using my plugin.

Vendicated:
with that mindset we'd turn Vencord into Discord OS and integrate every single feature imaginable into discord

i think this feature is much better done by a proper dedicated app for it that can also provide features like sync, phone/email notifs, etc, having it in discord makes very little sense

thank you for your contribution regardless

PR #2271, opened 2024-03-18
WhoReacted: fix scroll jumping when rendering users

### LINKED ISSUE #952: [Bug] WhoReacted causing scrolling issues
### Discord Account

_No response_

Whenever you scroll in a channel with reactions, every time WhoReacted loads a new icon the scroll position jumps, making it very hard to continue scrolling.

Scrolling should be smooth as when WhoReacted isn't enabled.

1. Enable WhoReacted
2. Open a channel with a frequent reactions
3. Scroll up
4. Notice scroll jitter

### Errors

_No response_

### Request Agreement

waresnew:
Fixes #952 

## Before

https://github.com/Vendicated/Vencord/assets/45801973/adc5ca2c-8e1b-4b13-861d-c7d789072ecd

(it mainly dies in the second half because most of the recent messages have 10+ reactions so it doesn't render)

## After

https://github.com/Vendicated/Vencord/assets/45801973/1a94c9dd-f418-4bc3-8545-b854b3f5fb6a

(There's still a bit of jumping when Discord's loading new messages, but that also happens when the plugin is disabled.)

### Problem Explanation

The reason why Discord interrupts your scrolling when rendering new users is because they change the height of the message. Rendering users on a reaction bar leads to a 4 pixel increase in height because the profile pictures are 24 pixels tall while the reaction emoji is 20 pixels tall. One way to fix this is to just change the style so the profile pictures aren't bigger than the emoji: `style={{ marginLeft: "0.5em", transform: "scale(0.8)", height: "1rem", display: "flex", alignItems: "center" }}`.

But there's also the problem where a new row of reactions may be created because the profile pictures widen the reaction bars.

I did some poking around and found that Discord uses a custom scroll anchoring implementation to keep track of a message to anchor to. When the height of the message list changes, Discord "fixes" the scroll position by scrolling to that anchor, causing the jump. If you keep the scroll anchor set to `null`, then the jumping won't happen anymore.

### Fix

However, I didn't want to just completely disable anchoring since that might create side effects.

Since Discord tries to `findAnchor()` (which searches through every cached message) repeatedly during scrolling, one way to fix this was to just set the current anchor to `null` once a WhoReacted component re-renders. This way, when it re-renders (to add new users, potentially changing the height), the `ResizeObserver` will fire and Discord will attempt to scroll to the stored anchor (causing the jump) but fail. The next time the user scrolls (usually immediately after), Discord will find a new scroll anchor so it hopefully won't disrupt the rest of the app too much.

I think a better solution would be to cancel `fixScrollPosition()` if the `ResizeObserverEntry` shows that a new avatar was rendered, but since it returns the entire message window element, I think it'd be too messy (+ might break on updates?) to search through it.

Another thing to note is that if you scroll a lot, WhoReacted will get rate limited and you'll have to wait a while for it to start rendering users on the message that you scrolled to. You could probably just make it only process messages that are currently in view, but that'd probably fit in another issue.


src/plugins/whoReacted/index.tsx
```diff
@@ -105,7 +105,19 @@ export default definePlugin({
             match: /(?<=CONNECTION_OPEN:function\(\){)(\i)={}/,
             replace: "$&;$self.reactions=$1"
         }
-    }],
+    },
+    {
+
+        find: "cleanAutomaticAnchor(){",
+        replacement: {
+            match: /constructor\(.{1,2}\){/,
```
Vendicated, line 113: please make this a little more specific to be safer

Vendicated:
thank you a lot for your thorough research

waresnew:
Thanks for the review! I changed the regex match to `/this\.automaticAnchor=null,this\.messageFetchAnchor=null,/`.

Also, I made it only remove the anchor if the user is currently scrolling, since that's the only time when the bug occurs. It's also because the scroll anchor will be null for some time if lots of avatars are rendering at once, even if the user isn't scrolling, which may cause the message list to slowly scroll on its own. The message list shouldn't drift anymore.

FichteFoll:
Thanks a lot for the fix. I was very excited when I read this in the changelog and even more when it actually worked.

PR #2272, opened 2024-03-18
new plugin: OverrideForumDefaults

Inbestigator:
OverrideForumDefaults lets you set the default layout/sort order of a forum instead of relying on moderator choices. You can still change the settings on a per-channel basis.

image

[Original plugin suggestion](https://github.com/Vencord/plugin-requests/issues/195)


MrDiamondDog (commented):
Imo I would make it per-channel, and have a button in the context menu to change it

Inbestigator:
> Imo I would make it per-channel, and have a button in the context menu to change it

The pull request said that they wanted it to be global, and I think that's probably a good option, you're still able to change the view on a per-channel basis. The plugin is just overriding the default.

MrDiamondDog:
Oh ok

Vendicated:
thanks!

PR #2273, opened 2024-03-18
Center profile

HarshAr2416:
this is the code for enabling center profile in discord

### DIFF
--- a/center profile
+++ b/center profile
@@ -1,0 +1,12 @@
+/* Center Profile */
+[class^="UserProfile_"] {
+    text-align: center;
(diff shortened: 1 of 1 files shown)


Sqaaakoi:
this is not a CSS snippets repository. go to the vencord server and submit it to css snippets instead. you may have to open a modmail ticket and request access.

Sqaaakoi:
also, have you even tested your CSS?
it really does look like you've shoved ????? on top of ?????? because you don't fully know how CSS works yet.

PR #2279, opened 2024-03-22
Fix discord-types package not being updated to the most recent version

zastlx:
pretty self-explanatory, vencord currently uses version 1.3.26 from the npm registry, but the newest version of discord-types is only on GitHub and not on the npm registry. you can solve this by just adding the GitHub repo as the source of the package instead of the npmjs version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
         "@typescript-eslint/eslint-plugin": "^5.59.1",
         "@typescript-eslint/parser": "^5.59.1",
         "diff": "^5.1.0",
-        "discord-types": "^1.3.26",
+        "discord-types": "github:Swishilicous/discord-types",
         "esbuild": "^0.15.18",
         "eslint": "^8.46.0",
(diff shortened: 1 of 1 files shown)


zastlx:
so I overlooked the fact that there is a typo in the discord-types package which breaks it anyway, honestly @Vendicated you should probably just fork the repo, fix it, and use your own for this :woman_facepalming:.

(I'm going to leave this PR open until vee sees it, after that it might as well be closed)

zastlx:
> so I overlooked the fact that there is a typo in the discord-types package which breaks it anyway, honestly @Vendicated you should probably just fork the repo, fix it, and use your own for this 🤦‍♀️.
> 
> (I'm going to leave this PR open until vee sees it, after that it might as well be closed)

typo in question - https://github.com/Swishilicous/discord-types/blob/7ece571cd9f5c9e238eb09c5635d4468ede050d2/general/User.d.ts#L12C33-L12C34

PR #2281, opened 2024-03-22
fix(PinDMs): display properly when there are 0 unpinned dms

Syncxv:
When you add all your dms to a category it causes everything to disappear. This happens because of this:
https://github.com/Vendicated/Vencord/blob/ca18b6e0446b6f1858a82de47489bb01f925ce25/src/plugins/pinDms/index.tsx#L203

this line of code assumes an empty `channelIds` means dms havent been fetched from discord, but it doenst account for the case where dms have been fetched and all the private dms have been added to one or more categories.

### DIFF
--- a/src/plugins/pinDms/index.tsx
+++ b/src/plugins/pinDms/index.tsx
@@ -69,7 +69,6 @@ export default definePlugin({
         {
             find: ".privateChannelsHeaderContainer,",
             replacement: [
-                // Init
                 {
                     // Filter out pinned channels from the private channel list
                     match: /(?<=\i,{channels:\i,)privateChannelIds:(\i)/,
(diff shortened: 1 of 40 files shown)


Syncxv:
done

src/plugins/pinDms/index.tsx
```diff
@@ -62,10 +70,6 @@ export default definePlugin({
             find: ".privateChannelsHeaderContainer,",
             replacement: [
                 // Init
```
Nuckyz, line 72:
```suggestion
```

Nuckyz (changes requested):

Vendicated:
this still seems to rely on length. what happens if you use this on an account with 0 dms?

Syncxv:
i dont think we need to check if its initialized.
https://github.com/Vendicated/Vencord/blob/ca18b6e0446b6f1858a82de47489bb01f925ce25/src/plugins/pinDms/index.tsx#L156-L160

https://github.com/Vendicated/Vencord/blob/ca18b6e0446b6f1858a82de47489bb01f925ce25/src/plugins/pinDms/data.ts#L35-L40

the init function forceUpdates the component on connection open (when channels have been fetched), meaning that check is kinda useless.

Vendicated:
thanks!

PR #2282, opened 2024-03-22
Add support for Vendroid-only plugins

nin0-dev:
This PR adds support for adding Vendroid-exclusive pluginsby adding the `.mobile` suffix to the plugin folder, just like other platforms.

It also adds a new build script- `pnpm buildMobile`, which is copied from the buildWeb script but includes mobile plugins. This can be useful for a future VendroidEnhancements plugin that would be directly added in vencord codebase, removing the need to inject custom plugins directly in vendroid code.

It also adds a `webDesktop` suffix, which builds the plugin on PC web platforms (aka any web that isnt vendroid).


nin0-dev:
closed in favor of custom patches directly integrated in the app

PR #2283, opened 2024-03-22
Array support for find + ResurrectHome: View Server Home Button on Server Guide

Nuckyz:
Adds a button to view the Server Home in the Server Guide channel row. Also adds support for array finds because it was needed to match the right module.

### DIFF
--- a/src/components/VencordSettings/PatchHelperTab.tsx
+++ b/src/components/VencordSettings/PatchHelperTab.tsx
@@ -33,8 +33,8 @@ if (IS_DEV) {
     var differ = require("diff") as typeof import("diff");
 }
 
-const findCandidates = debounce(function ({ find, setModule, setError }) {
-    const candidates = search(find);
+const findCandidates = debounce(function ({ finds, setModule, setError }) {
+    const candidates = search(...finds);
     const keys = Object.keys(candidates);
     const len = keys.length;
     if (len === 0)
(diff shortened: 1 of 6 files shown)


src/components/VencordSettings/PatchHelperTab.tsx
```diff
@@ -180,15 +180,15 @@ function ReplacementInput({ replacement, setReplacement, replacementError }) {
 
     return (
         <>
-            <Forms.FormTitle>replacement</Forms.FormTitle>
+            <Forms.FormTitle className="">replacement</Forms.FormTitle>
```
Vendicated, line 183: is this empty className intentional?

src/plugins/resurrectHome/index.tsx
```diff
@@ -92,14 +145,37 @@ export default definePlugin({
...
+            find: "69386d_5",
+            replacement: {
+                match: /location:"69386d_5".+?;/,
+                replace: "$&$self.useDisableViewServerHome();"
+            }
         }
     ],
 
+    ViewServerHomeButton: ErrorBoundary.wrap(({ text }: { text?: string; }) => {
+        if (text == null || text !== i18n.Messages.SERVER_GUIDE) return null;
```
Vendicated, line 168:
```suggestion
        if (text !== i18n.Messages.SERVER_GUIDE) return null;
```

Vendicated:
sooo this pr seems to patch the Guild thingie row component and check if `props.text` is GuildHome, then renders its own component

you could just directly patch the parent instead. the component takes a `badge` property that renders exactly in the position you want. you could just use that property and have a much cleaner patch without the hacky find for the super tiny generic module

image

src/components/VencordSettings/PatchHelperTab.tsx
Nuckyz: yes, Discord codes adds a class I don't want when you dont have a className set

PR #2284, opened 2024-03-22
feat(plugin) FakeNitro:  Add soundboard bypass options

Loukious:
## This adds to options:

- Allows the use of all servers' soundboards (others can't hear it).
- Allows the use of all sounds in the server's soundboard no matter what the server's boost level is.

### DIFF
--- a/src/plugins/fakeNitro/index.tsx
+++ b/src/plugins/fakeNitro/index.tsx
@@ -156,6 +156,18 @@ const settings = definePluginSettings({
         default: true,
         restartNeeded: true
     },
+    enableSoundboardBypass: {
+        description: "Allows the use of all servers' soundboards (others can't hear it)",
+        type: OptionType.BOOLEAN,
+        default: true,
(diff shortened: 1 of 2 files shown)


Vendicated:
what's the point if noone can hear it

MrDiamondDog:
The second one should definitely be included, but the first one is useless. I myself have been wanting something like the second option, but I didn't think it was possible.

Loukious:
> The second one should definitely be included, but the first one is useless. I myself have been wanting something like the second option, but I didn't think it was possible.

You can't download sounds from other servers easily without the first option. It allows you to listen and/or download sounds from other servers easily instead of manually switching to each server individually just to listen to them or download them.

NinDevs:
This is a great idea if done well - perhaps by having an input that soundboard audio can be played through (set it as mic output) and soundboard sounds can be played through your mic, like with voicemeeter for example. Hmmmm

PR #2285, opened 2024-03-22
[Plugin] linkProfilePicture

Loukious:
Makes profiles pictures a clickable link which opens it up in your default browser in full size.


Vendicated:
exists as ViewIcons

PR #2286, opened 2024-03-23
Blur removed spoiled images

### LINKED ISSUE #1815: [Bug] Message Logger will unblur spoiled images.
### Discord Account

_No response_

When an image is caught by the MessageLogger, if it contains an image that is spoiled it won't be blurred. 

If MessageLogger finds an image that is spoiled it should be blurred. 

1. Upload an image.
2. Tag as spoiled.
3. Send message
4. Delete message
5. Spoiled image will no longer be blurred but still have "Spoiler" in front of it.

### Errors

_No response_

### Request Agreement

k26pl:
Fixes #1815

### DIFF
--- a/src/plugins/messageLogger/messageLogger.css
+++ b/src/plugins/messageLogger/messageLogger.css
@@ -8,7 +8,7 @@
     .emoji,
     [data-type="sticker"],
     iframe,
-    .messagelogger-deleted-attachment,
+    .messagelogger-deleted-attachment:not([class*="hiddenAttachment_"]),
     [class|="inlineMediaEmbed"]
 ) {
     filter: grayscale(1) !important;
--- a/src/plugins/messageLogger/messageLogger.css
+++ b/src/plugins/messageLogger/messageLogger.css
@@ -36,6 +36,6 @@
     opacity: 0.5;
 }
 
-.hiddenSpoiler__57186 .hiddenAttachment_dfecc7 .messagelogger-deleted-attachment {
+.messagelogger-deleted-attachment [class^="hiddenAttachment_"] {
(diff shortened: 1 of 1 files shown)


src/plugins/messageLogger/messageLogger.css
```diff
@@ -35,3 +35,7 @@
 .theme-light .messagelogger-edited {
     opacity: 0.5;
 }
+
+.hiddenSpoiler__57186 .hiddenAttachment_dfecc7 .messagelogger-deleted-attachment {
```
waresnew, line 39: Try not to hardcode entire class names because the part after the `_` might change when Discord updates (maybe do something like `[class^="hiddenSpoiler_"]` instead)

waresnew (changes requested):

k26pl:
i have removed hiddenSpoiler as there are to different classes with same name.  hiddenAttachment appears to be on all images too, so this should still work.

src/plugins/messageLogger/messageLogger.css
```diff
@@ -35,3 +35,7 @@
 .theme-light .messagelogger-edited {
     opacity: 0.5;
 }
+
+.messagelogger-deleted-attachment [class^="hiddenAttachment_"] {
```
Vendicated, line 39: for me these two classes are on the same element, so this doesn't actually work
image

what works for me is changing line 11:
```diff
- .messagelogger-deleted-attachment,
+ .messagelogger-deleted-attachment:not([class*="hiddenAttachment_"]),
```

k26pl: Thanks! It is a better solution

k26pl:
Also we should disable buttons on removed polls, but that probably should be in separate issue

PR #2287, opened 2024-03-24
new plugin UnlockedAvatarZoom: allows crop zooming in further

nakoyasha:
Unlocks and allows you to use custom multipliers for the zoom slider in the Edit Avatar modal as the default 2 is "barely enough"

### DIFF
--- a/src/plugins/unlockedAvatarZoom/index.ts
+++ b/src/plugins/unlockedAvatarZoom/index.ts
@@ -20,7 +20,7 @@ const settings = definePluginSettings({
 
 export default definePlugin({
     name: "UnlockedAvatarZoom",
-    description: "Unlocks the zoom mutliplier on the avatar modal",
+    description: "Unlocks the avatar zoom slider on the edit avatar modal",
     authors: [Devs.nakoyasha],
     settings,
     patches: [
--- a/src/plugins/unlockedAvatarZoom/index.ts
+++ b/src/plugins/unlockedAvatarZoom/index.ts
@@ -5,15 +5,13 @@
  */
 
 import { definePluginSettings } from "@api/Settings";
-import { makeRange } from "@components/PluginSettings/components";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 
 const settings = definePluginSettings({
     zoomMultiplier: {
-        type: OptionType.SLIDER,
+        type: OptionType.NUMBER,
         description: "Zoom multiplier",
-        markers: makeRange(2, 16),
         default: 2,
     },
 });
--- a/src/plugins/unlockedAvatarZoom/index.ts
+++ b/src/plugins/unlockedAvatarZoom/index.ts
@@ -1,0 +1,35 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { makeRange } from "@components/PluginSettings/components";
+import { Devs } from "@utils/constants";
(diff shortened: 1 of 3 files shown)


src/plugins/unlockedAvatarZoom/index.ts
```diff
@@ -0,0 +1,35 @@
...
+});
+
+export default definePlugin({
+    name: "UnlockedAvatarZoom",
+    description: "Unlocks the avatar zoom slider on the edit avatar modal",
+    authors: [Devs.nakoyasha],
+    settings,
+    patches: [
+        {
+            find: "\"aria-label\":I.default.Messages.FORM_LABEL_AVATAR_SIZE",
```
MrDiamondDog, line 28: Don't have obfuscated variables names in your finds. Otherwise you will need to change this every time discord updates.

`.Messages.FORM_LABEL_AVATAR_SIZE` would likely work.

MrDiamondDog (changes requested):

src/plugins/unlockedAvatarZoom/index.ts
nakoyasha: Another component seems to use .Messages_FORM_LABEL_AVATAR_SIZE, so using that causes the patch to not have any effect.

I could probably use either `keyboardStep` or `asValueChanges` instead, as neither seem to be present elsewhere

MrDiamondDog: It only matters for variables name `I`, `w`, etc. These variables are obfuscated and change with discord updates.

src/plugins/unlockedAvatarZoom/index.ts
```diff
@@ -0,0 +1,35 @@
...
+export default definePlugin({
+    name: "UnlockedAvatarZoom",
+    description: "Unlocks the avatar zoom slider on the edit avatar modal",
+    authors: [Devs.nakoyasha],
+    settings,
+    patches: [
+        {
+            find: ".Messages.AVATAR_UPLOAD_EDIT_MEDIA",
+            replacement: {
+                match: /(maxValue:(\w))/,
```
Vendicated, line 30:
```suggestion
                match: /maxValue:\d/,
```

src/plugins/unlockedAvatarZoom/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import { makeRange } from "@components/PluginSettings/components";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+
+const settings = definePluginSettings({
+    zoomMultiplier: {
+        type: OptionType.SLIDER,
+        description: "Zoom multiplier",
+        markers: makeRange(2, 16),
+        default: 2,
```
Vendicated, line 17: the default should definitely not be 2. otherwise, people will enable the plugin and wonder why it does nothing at all

is there any point adding a limit? why not just always have it as 16?

Vendicated:
please add a README.md file!

src/plugins/unlockedAvatarZoom/index.ts
nakoyasha: it was limited since I thought a slider would be way nicer, and it didn't make sense to use higher values anyway (as it got too unstable to be actually usable)

replaced it with a number value anyway - also increased the default to 4

Vendicated: you can keep it as a slider, that's fine, especially if you say it becomes too unstable

just the default should be higher, at least 4

nakoyasha: yea on second thoughts it wasnt really the best idea to turn it into a number - i feel like people will try to deliberately break it (by using really big numbers or negative numbers) just to complain :/

Vendicated:
thank you! :3

PR #2288, opened 2024-03-24
feat(plugin): AltKrispSwitch

waresnew:
Fulfills plugin request Vencord/plugin-requests#214


liliantdn:
any update on this?

waresnew:
idk if adding new plugins is their priority rn but for now you can just use this as a 3rd party plugin

liliantdn:
> for now you can just use this as a 3rd party plugin

Okay I tried and it looks like it doesn't work anymore :/ too bad

PR #2290, opened 2024-03-25
new plugin: AutomodContext

JohnyTheCarrot:
# Problem I'm solving with this plugin:

To figure out if someone tried to bypass our automod or to see the surrounding context, I have to search for their user's message history and scroll down to the rough timestamp it happened at.
This is very tedious, so this plugin adds a button to automod messages to allow you to jump to the surrounding messages of when the message was flagged/blocked.

https://github.com/Vendicated/Vencord/assets/24935987/393a6312-409b-4418-a6d7-eaeb5f3d0281


src/plugins/automodContext/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { disableStyle, enableStyle } from "@api/Styles";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+import { Button, NavigationRouter, SelectedGuildStore, Text } from "@webpack/common";
+const Kangaroo = findByPropsLazy("jumpToMessage");
+
+import style from "./style.css?managed";
```
Vendicated, line 15: your styles don't alter things that don't belong to your plugin, so it doesn't need to be managed.

```suggestion
import "./style.css";
```

katlyn:
An addition that could be neat for this plugin is injecting the automodded message into the message cache similar to how messagelogger does with deleted messages - that way it's possible to see the message actually in context rather than just seeing messages around the same time.

programminglaboratorys:
> An addition that could be neat for this plugin is injecting the automodded message into the message cache similar to how messagelogger does with deleted messages - that way it's possible to see the message actually in context rather than just seeing messages around the same time.

that sounds like something similar of how BetterAutomod works [#2300](https://github.com/Vendicated/Vencord/pull/2300)

JohnyTheCarrot:
Appreciate your clean-up. This is my first plugin so I'll read up your changes to see how I can improve.

PR #2293, opened 2024-03-26
CustomRPC Toggle

SzyZuu:
adds a button next to your mic so you can quickly toggle CustomRPC without going to the settings :D

### DIFF
--- a/src/plugins/customRPC/index.tsx
+++ b/src/plugins/customRPC/index.tsx
@@ -417,14 +417,8 @@ async function setRpc(disable?: boolean) {
 
 function toggleRPC() {
     console.log(activeRPC);
-
-    if (activeRPC) {
-        setRpc(false);
-        activeRPC = false;
-    } else {
-        setRpc(true);
-        activeRPC = true;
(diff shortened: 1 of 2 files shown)


Vendicated:
https://vencord.dev/plugins/GameActivityToggle

SzyZuu:
> https://vencord.dev/plugins/GameActivityToggle

I'm sorry but my changes have a different functionality than game activity toggle. See, if you wanted to use game activity toggle with customrpc, customrpc has to be on all the time, meaning rpc from games won't work properly. The customrpc toggle gives you the ability to activate the customrpc only when you need it. Example:
You have set up customrpc for "Playing Piano", so you want it to be active only when you actually play, but when you're NOT playing you still want others to see your games rpc.

Vendicated:
yes, i was just about to suggest the VencordToolbox. it's a plugin that adds a toolbar button in the top right and lets plugins specify actions there. just add a Custom Rpc toggle there

SzyZuu:
I see, thanks

PR #2296, opened 2024-03-26
new plugin ReplyTimestamp: show timestamps of replied messages

Kyuuhachi:
image


src/plugins/replyTimestamp/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function ReplyTimestamp({
+    referencedMessage,
+    baseMessage,
+}: {
+    referencedMessage: { state: number, message?: Message },
+    baseMessage: Message;
+}) {
+    if(referencedMessage.state === 0) {
```
Vendicated, line 32:
```suggestion
    if (referencedMessage.state !== 0) return null;
```

src/plugins/replyTimestamp/README.md
```diff
@@ -0,0 +1,3 @@
+# ReplyTimestamp
+
+Shows timestamps on the previews of replied-to messages. Pretty simple.
```
Vendicated, line 3: add a screenshot 

see https://github.com/Vendicated/Vencord/issues/1749

src/plugins/replyTimestamp/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+function ReplyTimestamp({
+    referencedMessage,
+    baseMessage,
+}: {
+    referencedMessage: { state: number, message?: Message },
+    baseMessage: Message;
+}) {
+    if (referencedMessage.state !== 0) return null;
+    const refTimestamp = referencedMessage.message!.timestamp as any;
```
Vendicated, line 33: why type it as optional above only to then immediately non null assert? 

it would be better to either explicitly check for non null or adjust the type to make it non nullable

src/plugins/replyTimestamp/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+            compact={isSameDay(refTimestamp, baseTimestamp)}
+            timestamp={refTimestamp}
+            isInline={false}
+        >
+            <Sep>[</Sep>
+            { isSameDay(refTimestamp, baseTimestamp)
+                ? dateFormat(refTimestamp, "LT")
+                : calendarFormat(refTimestamp)
+            }
+            <Sep>]</Sep>
```
Vendicated, line 48: i'm a bit confused. these separators don't show up in your screenshots. am i misunderstanding their purpose? or are your screenshots outdated

in any case, why not do this with `::before` and `::after` in css?

Kyuuhachi: Mostly to imitate vanilla `Timestamp`'s behavior if not setting children. I think it's for accessibility and/or copypaste reasons.

Vendicated: got it, this isn't actually visible.. thanks!

Vendicated:
thank you a lot!!

PR #2297, opened 2024-03-26
feat(plugin): ImageLink

Kyuuhachi:
i don't like that discord hides the link to images if they're the only thing in the message so i made it not


src/plugins/imageLink/index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "ImageLink",
+    description: 'Suppresses the hiding of links for "simple embeds"',
```
Vendicated, line 12: this description is hard to understand imo. maybe something like
```suggestion
    description: "Never hide image links in messages, even if it's the only content",
```
?

Vendicated:
please add a README!

Vendicated:
thanks a lot!

PR #2298, opened 2024-03-26
Add TeX plugin

Kyuuhachi:
Renders $\LaTeX$ formulas. `` `$\sin x$` `` renders as $\sin x$, `` `$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$` `` renders as $\displaystyle x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$. The latter is intentionally set to an inline-block instead of a block, to save space.

This plugin requires external libraries, which will not work in the userscript version without additional workarounds.

### DIFF
--- a/src/plugins/tex/README.md
+++ b/src/plugins/tex/README.md
@@ -1,0 +1,5 @@
+# TeX
+
+Renders $\LaTeX$ formulas. `` `$\sin x$` `` renders as $\sin x$, `` `$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$` `` renders as $\displaystyle x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$. The latter is intentionally set to an inline-block instead of a block, to save space.
+
+This plugin requires external libraries, which will not work in the userscript version of Vencord without additional workarounds.
--- a/src/plugins/tex/index.tsx
+++ b/src/plugins/tex/index.tsx
@@ -1,0 +1,108 @@
+import "./style.css";
+
+import { Devs } from "@utils/constants";
+import { makeLazy } from "@utils/lazy";
+import { classes } from "@utils/misc";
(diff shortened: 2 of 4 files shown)


byeoon:
math is pretty scary ngl

but this plugin seems helpful

waresnew:
Do you think it's possible to add a command to send latex as a rendered image as well? So you can share math stuff with people without the plugin

Vendicated:
thanks for the contribution!

a few thoughts:
- how safe is this? using dangerouslySetInnerHTML is sketch. we don't want any XSS happening
- there are some commands that may load external files, such as
  ```tex
  \includegraphics[height=0.8em, totalheight=0.9em, width=0.9em, alt=KA logo]{https://katex.org/img/khan-academy.png}
  ```
  have you made sure these are disabled?
- does this prevent infinite loops? e.g. 
  ```tex
  \def\x{Hello!\par\x}
  \x
  ```

i'm also not sure how useful a plugin like this is in light of the recent userapps. tex inside discord is already fairly niche, and in an era where you can just install an app to your account and use /tex anywhere, it doesn't seem that useful

Kyuuhachi:
Those are good points. I personally have no issue with dangerouslySetInnerHTML, but I understand that it can be seen as sketchy. I'm pretty sure there's an option in katex to set a time limit, and there probably is one for external urls too; I haven't looked.

And yeah the value of this plugin is pretty low. It came up as a plugin request a while ago, and it was fun to make. Perhaps it's better off remaining as a userplugin. I don't mind if you choose to close this.

lewisakura:
please make sure you validate that [you have configured the security options in katex](https://katex.org/docs/security) to prevent people from abusing it

Vendicated:
will be closing as too niche. if you'd like to post it in third party plugins, feel free to do so after making sure the security concerns are addressed!

sorry and thank you anyway

PR #2300, opened 2024-03-27
new plugin: BetterAutomod

programminglaboratorys:
This plugin echoes the Automod logs real time into the location of the flagged message, and let's you test your rules in the automod settings
preview-no-test
preview-test
preview-bad-boi


byeoon:
very cool and helpful plugin!! +1 from me

Kyuuhachi:
Probably just set the wrong branch to merge into.

programminglaboratorys:
> Seems really cool, but it also seems like this repo has edits to other plugins

yeah i am not sure how I did that. I only used git one time in my life, this is the second

programminglaboratorys:
added a README and improved/minified the code

programminglaboratorys:
this plugin no longer fit my standards and not so user friendly, and I believe it could been executed better. if someone wants to pick it up feel free to do so.

PR #2303, opened 2024-03-27
fix(VoiceMessages): Make Chrome record audio as ogg

### LINKED ISSUE #1512: VoiceMessages: Web recorder records webm instead of ogg
Like the title says, the recording is webm, not ogg. This causes voice messages to be unplayable on ios.

Unfortunately, chrome does not support ogg, so we will likely have to use a library like https://github.com/chris-rudmin/opus-recorder

waresnew:
Fixes #1512 

Works on Chrome extension + Chrome userscript

I ending up having to import the web worker file from unpkg because I was having trouble with bundling it into the userscript (Worker constructor needs a URL so I tried storing a blob of the entire file in the userscript, but that increased size by 70%)

Let me know if you prefer not fetching from unpkg (I tried bundling the `encoderworker.min.js` into the `third-party` folder and it worked for the extension but idk how to do it for userscript)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
         "gifenc": "github:mattdesl/gifenc#64842fca317b112a8590f8fef2bf3825da8f6fe3",
         "monaco-editor": "^0.43.0",
         "nanoid": "^4.0.2",
+        "opus-recorder": "^8.0.5",
         "virtual-merge": "^1.0.1"
     },
     "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
 patchedDependencies:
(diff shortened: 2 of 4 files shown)


Vendicated:
> I tried bundling the encoderworker.min.js into the third-party folder and it worked for the extension but idk how to do it for userscript

you can't. userscripts are single files

waresnew:
Oh ok, so does that mean it's fine to import from unpkg in this case?

lewisakura:
If it can be lazy-loaded remotely and it's genuinely that large then set it up as a dynamically loaded dependency: https://github.com/Vendicated/Vencord/blob/main/src/utils/dependencies.ts

waresnew:
> If it can be lazy-loaded remotely and it's genuinely that large then set it up as a dynamically loaded dependency similar to rnnoise: https://github.com/Vendicated/Vencord/blob/main/src/utils/dependencies.ts

it was already lazy-loaded because i only load it if the plugin's enabled

i moved the url to dependencies.ts though bc i think that's the convention

mikethebos:
@Vendicated thoughts here?

PR #2309, opened 2024-03-28
Fix DisableCallIdle

TheKodeToad:
yes
(not tested but it's annoying to)


Nuckyz:
Cherry picked your commit, thanks

PR #2310, opened 2024-03-28
fix AlwaysTrust, BetterGifPicker, Experiments

AutumnVN:
omg popup jumpscare


Nuckyz:
I manually cherry picked only the second commit, thanks

PR #2313, opened 2024-03-29
translate: fix language generation script

lonelil:
Fixes the language generation script for the translate plugin.


nyakowint:
nuh uh

Vendicated:
it's already correct

lonelil:
there's a typo in the script 
image
it's trying to query select from temp1

Vendicated:
that's correct.

PR #2314, opened 2024-03-29
new plugin: NoMouseNavigation

Moondarker:
NoMouseNavigation allows users to disable the default handling of mouse side buttons (3 and 4).

Many people use side buttons for PTT. By default, Discord will handle those by navigating forward/backward through visited pages while the app is in focus, causing severe butthurt for PTT users. This plugin simply prevents the default behavior, PTT continues working normally.

P.S. I'm not sure if adding myself into the dev list in `constants.ts` is appropriate, but looking through other plugins, I couldn't find any files where authors are added in the way described by `2_PLUGINS.md`, so I assumed this is the new undocumented standard? Anyways, I can always revert that, just lmk.

### DIFF
--- a/src/plugins/noMouseNavigation/index.ts
+++ b/src/plugins/noMouseNavigation/index.ts
@@ -1,0 +1,29 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NoMouseNavigation",
+    description: "Disables forward/back navigation with mouse side buttons",
+    tags: ["FFMSB", "sidebuttons"],
+    authors: [Devs.Vaker],
+
+    start() {
+        window.addEventListener("mouseup", this.event);
+    },
+
+    stop() {
+        window.removeEventListener("mouseup", this.event);
+    },
+
+    event(e: MouseEvent) {
+        if ([3, 4].includes(e.button)) {
+            e.preventDefault();
+        }
+    }
+});
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -490,6 +490,10 @@ export const Devs = /* #__PURE__*/ Object.freeze({
         name: "ScattrdBlade",
         id: 678007540608532491n
     },
+    Vaker: {
+        name: "Vaker",
+        id: 92999669617000448n
+    },
 } satisfies Record<string, Dev>);
 
 // iife so #__PURE__ works correctly


src/plugins/noMouseNavigation/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NoMouseNavigation",
+    description: "Aka FFMSB. Disables forward/back navigation with mouse side buttons",
+    authors: [Devs.Vaker],
+
+    start() {
+        window.addEventListener("mouseup", this.event);
```
Vendicated, line 16: instead of adding your own listener, it would be better to simply noop discords listeners, to avoid having unnecessary listeners

Moondarker: I believe it must be handled by the Electron itself, the same way it's hardcoded into Chrome. I'll double check tho

Moondarker: Yeah, based on my experience with Chromium and the fact that I couldn't find anything explicitly handling this from Discord's side, I'm assuming this is pretty much the only way to implement this - but I'll be happy to be proven wrong here. It must be a native thing inherited from Chromium

If Discord had its own listeners for `mouseup`, `e.preventDefault()` wouldn't affect those, and we'd need to call `e.stopPropagation()`/`e.stopImmediatePropagation()`, which is not the case, @Vendicated

src/plugins/noMouseNavigation/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NoMouseNavigation",
+    description: "Aka FFMSB. Disables forward/back navigation with mouse side buttons",
```
Vendicated, line 12: only very little users will know what FFMSB stands for. you can add it via the `tags:` property instead
```suggestion
    description: "Disables forward/back navigation with mouse side buttons",
```

Moondarker: Agreed, changed according to suggestion

Nuckyz (dismissed):

Nuckyz:
I tested and it seems like mouseup is not being fired for my side buttons, which makes the plugin not work

Vendicated:
works for me. on linux, maybe this is platform dependant?

Moondarker:
> I tested and it seems like mouseup is not being fired for my side buttons, which makes the plugin not work

Which OS are you on? If you're on Linux, which distro and are you on X11 or Wayland?

Nuckyz:
> > I tested and it seems like mouseup is not being fired for my side buttons, which makes the plugin not work
> 
> Which OS are you on? If you're on Linux, which distro and are you on X11 or Wayland?

windows haha

Moondarker:
> > > I tested and it seems like mouseup is not being fired for my side buttons, which makes the plugin not work
> > 
> > 
> > Which OS are you on? If you're on Linux, which distro and are you on X11 or Wayland?
> 
> windows haha

Can't reproduce at all... But it should work the same way everywhere. Do you want to say that it navigates on `mousedown` for you? What if you replace the handled event to `mousedown`?

PR #2318, opened 2024-03-31
ReviewDB: Fix website url

Kodarru:
Since URL used to have an `/` it would redirect to https://reviewdb.mantikafasi.dev//api/redirect instead.

This fix will help with confusion for the average non-techy Vencord user


Vendicated:
wow good catch! this was working flawlessly before but we moved hosting, so i guess the new hosting doesn't normalise paths

Kodarru:
I now need contributor 😡😡

mantikafasi:
I actually fixed this in my own pr but meh better to get this fixed faster

PR #2319, opened 2024-03-31
TypingIndicator: Add an option to show user avatars

Sqaaakoi:
This adds an option to show user avatars to the left of the typing indicator.
Default behaviour is to show only the typing dots, so for existing users nothing will change.
Example

<sub>yes I did leave this for a month and forget about it, yes it has been stable for that whole time.</sub>


Vendicated:
what does this look like when there are multiple people typing? what if a lot of people are typing? like 10 or more

you can experiment by dispatching fake typing events with flux dispatcher

Sqaaakoi:
> what does this look like when there are multiple people typing? what if a lot of people are typing? like 10 or more
> 
> you can experiment by dispatching fake typing events with flux dispatcher

I tried to keep the same limit of users as the indicator under the chatbox.
Also, are the avatars too small? Discord doesn't support size 20 for UserSummaryItem so I picked 16 instead. I know that WhoReacted has a scale CSS hack to get around this but I would rather not do that as the bounding box doesn't scale with the contents.

Sqaaakoi-31-03-2024-23-05-34
Sqaaakoi-31-03-2024-23-05-57
Sqaaakoi-31-03-2024-23-04-56

src/plugins/typingIndicator/index.tsx
```diff
@@ -16,20 +16,28 @@
...
 
 import { buildSeveralUsers } from "../typingTweaks";
 
 const ThreeDots = findExportedComponentLazy("Dots", "AnimatedDots");
+const UserSummaryItem = findComponentByCodeLazy("defaultRenderUser", "showDefaultAvatarsForNullUsers");
 
 const TypingStore = findStoreLazy("TypingStore");
 const UserGuildSettingsStore = findStoreLazy("UserGuildSettingsStore");
 
+enum IndicatorMode {
```
Vendicated, line 36:
```suggestion
const enum IndicatorMode {
```

src/plugins/typingIndicator/style.css
```diff
@@ -0,0 +1,18 @@
...
+.vc-typing-indicator-avatars {
+    margin-left: 6px;
+}
+
+.vc-typing-indicator-dots {
+    margin-left: 6px;
+    height: 16px;
+    display: flex;
+    align-items: center;
+    z-index: 0;
```
Vendicated, line 16: is this actually needed?

Vendicated:
looks really cool

src/plugins/typingIndicator/style.css
Vendicated: eh who cares

Inbestigator:
fr

PR #2321, opened 2024-04-01
EmoteCloner: fix low quality; don't count managed emojis

byeoon:
This pull request simply improves the quality of stickers so that they get cloned to 2048. 

**Before**:
image

**After**:
image

### DIFF
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -132,10 +132,11 @@ function getGuildCandidates(data: Data) {
         for (const emoji of emojis) {
             if (emoji.animated === isAnimated) {
                 count++;
+                if (emoji.managed === true) {
+                    count--;
+                }
             }
-            if (emoji.managed === true) {
-                count--; // twitch emojis do not count towards the limit
-            }
+
         }
         return count < emojiSlots;
     }).sort((a, b) => a.name.localeCompare(b.name));
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -56,7 +56,7 @@ function getUrl(data: Data) {
     if (data.t === "Emoji")
         return `${location.protocol}//${window.GLOBAL_ENV.CDN_HOST}/emojis/${data.id}.${data.isAnimated ? "gif" : "png"}`;
 
-    return `${window.GLOBAL_ENV.MEDIA_PROXY_ENDPOINT}/stickers/${data.id}.${StickerExt[data.format_type]}?size=2048`;
+    return `${window.GLOBAL_ENV.MEDIA_PROXY_ENDPOINT}/stickers/${data.id}.${StickerExt[data.format_type]}?size=4096`;
 }
 
 async function fetchSticker(id: string) {
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -363,7 +363,7 @@ export default definePlugin({
     name: "EmoteCloner",
     description: "Allows you to clone Emotes & Stickers to your own server (right click them)",
     tags: ["StickerCloner"],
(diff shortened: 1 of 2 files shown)


src/plugins/emoteCloner/index.tsx
```diff
@@ -56,7 +56,7 @@ function getUrl(data: Data) {
     if (data.t === "Emoji")
         return `${location.protocol}//${window.GLOBAL_ENV.CDN_HOST}/emojis/${data.id}.${data.isAnimated ? "gif" : "png"}`;
 
-    return `${window.GLOBAL_ENV.MEDIA_PROXY_ENDPOINT}/stickers/${data.id}.${StickerExt[data.format_type]}`;
+    return `${window.GLOBAL_ENV.MEDIA_PROXY_ENDPOINT}/stickers/${data.id}.${StickerExt[data.format_type]}?size=2048`;
```
kyrie25, line 59: I've been stumbling upon this issue too and I think it'd be nice to use the highest quality possible to preserve fidelity, and since this is the highest option Discord provides afaik

```suggestion
    return `${window.GLOBAL_ENV.MEDIA_PROXY_ENDPOINT}/stickers/${data.id}.${StickerExt[data.format_type]}?size=4096`;
```

byeoon: the higher the better, just pushed that change rn

Sqaaakoi: don't stickers cap out at 320px anyway?

byeoon: > don't stickers cap out at 320px anyway?

Not sure, but I don't think it does

byeoon:
Something I thought of - should there be an optional setting for clone quality? (Or just keep it as is)

byeoon:
Somewhat of a new fix added to this PR: 
If you had externally added emotes or ones that were managed outside of discord (twitch, etc) they would count up towards the emote limit, which would cause the plugin to sometimes not include certain servers to allow cloning even if they had the correct space. 
Now it properly counts without the external emotes being added.

src/plugins/emoteCloner/index.tsx
```diff
@@ -129,8 +129,14 @@ function getGuildCandidates(data: Data) {
         const { emojis } = EmojiStore.getGuilds()[g.id];
 
         let count = 0;
-        for (const emoji of emojis)
-            if (emoji.animated === isAnimated) count++;
+        for (const emoji of emojis) {
+            if (emoji.animated === isAnimated) {
+                count++;
+            }
+            if (emoji.managed === true) {
```
Vendicated, line 136: this is not correct this will sometimes deduct the count even if it wasn't previously incremented

just merge it into the above if statement

src/plugins/emoteCloner/index.tsx
```diff
@@ -129,8 +129,15 @@ function getGuildCandidates(data: Data) {
...
-            if (emoji.animated === isAnimated) count++;
+        for (const emoji of emojis) {
+            if (emoji.animated === isAnimated) {
+                count++;
+                if (emoji.managed === true) {
+                    count--;
+                }
+            }
+
+        }
```
ryan-0324, line 140:
```suggestion
        for (const emoji of emojis)
            if (emoji.animated === isAnimated && !emoji.managed)
                count++;
```

ryan-0324: Please correct me if I'm wrong, but is this not equivalent?

byeoon: > Please correct me if I'm wrong, but is this not equivalent?

Should be, since it'll still skip over the ones that are managed, but I'll double check just in case

byeoon: Yep, that works as well, thank you for the suggestion 👍

Vendicated:
thanks!

PR #2324, opened 2024-04-02
PinDMs: Fix unexpected behaviours when using last message sort

### LINKED ISSUE #2322: [Bug] PinDMs: alt up/down doesn't work correctly when using recent message sort order
### Discord Account

_No response_

It just doesn't navigate in the correct order

It should have the correct order

1. Set sort order to recent message
2. Pin a few dms
3. Have bottom pin message you so it goes to top
4. Try using keyboard navigation
5. Observe that it doesn't behave correctly

### Errors

_No response_

### Request Agreement

puppy-girl:
Fixed alt + up/down not working as expected and fixed the channel preview when collapsing a category being incorrect when using the most recent message sort option

Resolves #2322

### DIFF
--- a/src/plugins/pinDms/data.ts
+++ b/src/plugins/pinDms/data.ts
@@ -106,12 +106,12 @@ export function categoryLen() {
 }
 
 export function getAllUncollapsedChannels() {
-    return categories.filter(c => !c.collapsed).map(c => {
-        if (settings.store.pinOrder === PinOrder.LastMessage)
-            return PrivateChannelSortStore.getPrivateChannelIds().filter(channel => c.channels.includes(channel));
+    if (settings.store.pinOrder === PinOrder.LastMessage) {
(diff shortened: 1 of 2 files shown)


Vendicated:
cc @Syncxv

src/plugins/pinDms/data.ts
```diff
@@ -106,7 +106,12 @@ export function categoryLen() {
 }
 
 export function getAllUncollapsedChannels() {
-    return categories.filter(c => !c.collapsed).map(c => c.channels).flat();
+    return categories.filter(c => !c.collapsed).map(c => {
```
Vendicated, line 109: i know this was already this way before your changes, but `.map(...).flat()` is better written as `.flatMap(...)`
```suggestion
    return categories.filter(c => !c.collapsed).flatMap(c => {
```

Nuckyz:
Thanks for the fix! But in the next time please don't pr from main as it makes it harder to checkout your pr and test.

PR #2326, opened 2024-04-03
chore: allow external code to be properly licenced

sexybanana99:
I am working on a plugin that would require an external library, but that library is too small to guarantee a dependency.
Also, dynamically loading that library from npm/unpkg is not possible, since it loads a pollyfil it assumes is also a dependency.

So, I decided to simply copy the 4 typescript files I needed, to the folder of my plugin.
The eslint configuration forces all files to have `GPL-3.0-or-later` as the licence, meaning `MIT` code will be improperly licenced.

So I added the licence info as a template pattern to allow other licences on non-vencord files.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,7 +30,10 @@
             "error",
             {
                 "files": ["scripts/header-new.txt", "scripts/header-old.txt"],
-                "templates": { "author": [".*", "Vendicated and contributors"] }
+                "templates": {
+                    "author": [".*", "Vendicated and contributors"],
+                    "licence": [".*", "GPL-3.0-or-later"]
+                }
             }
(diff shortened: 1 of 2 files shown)


Vendicated:
hi arjix

sexybanana99:
Hi

Vendicated:
i would just disable the rule for the file, see https://github.com/Vendicated/Vencord/blob/main/src/api/DataStore/index.ts as an example

sexybanana99:
oooh!

that is much better!

Vendicated:
> hi arjix

i unblocked you so you can just use your main in the future. i think enough time passed

sexybanana99:
gracias

PR #2328, opened 2024-04-03
fix(ShowMeYourName): Update names in typing indicator

### LINKED ISSUE #2129: [Bug] ShowMeYourName not updating name in "is typing..."
### Discord Account

ezid

The ShowMeYourName plugin doesn't update the name in the "<user> is typing..." field at the bottom of Discord.
This is especially confusing with the option 'Username only' as you'll see names there that you wouldn't recognise from the chat.

([Initially reported in the support channel.](https://discord.com/channels/1015060230222131221/1026515880080842772/1199801287181619200))

When a user is typing, the nickname is replaced with the username (or display name in case that toggle is on). (The `(...)` for the alternative name seems overkill here.)

1. Enable the ShowMeYourName plugin.
2. Set display to 'Username only'.
3. Observe a user with a nickname different from the username typing.
4. In the bottom of Discord, it'll show "<nickname> is typing..." instead of "<username> is typing...".

### Errors

_No response_

### Request Agreement

refact0r:
first contribution, hopefully i didnt fuck anything up

image

### DIFF
--- a/src/plugins/showMeYourName/index.tsx
+++ b/src/plugins/showMeYourName/index.tsx
@@ -20,10 +20,6 @@ interface UsernameProps {
     userOverride?: User;
 }
 
-interface NickUser extends User {
-    nick: string;
-}
-
 const settings = definePluginSettings({
     mode: {
         type: OptionType.SELECT,
@@ -99,7 +95,7 @@ export default definePlugin({
         }
     },
 
-    renderTypingNames(props: any, users: NickUser[], children: any) {
+    renderTypingNames(props: any, users: User[], children: any) {
         if (!Array.isArray(children)) return children;
 
         let index = 0;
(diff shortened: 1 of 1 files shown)


Nuckyz:
Polluting the user object with a nick property is bad

refact0r:
copied the GuildMemberStore/RelationshipStore method to get nicks from the TypingTweaks plugin.

Vendicated:
is this compatible with TypingTweaks?

refact0r:
ok i just checked and it is not. how should i make it compatible?

i assume i should just add a predicate that disables the patch if typingtweaks is enabled.

ideally if both are enabled at once the typingtweaks names would display in the showmeyourname format. is cross-plugin behavior like that even allowed?

lewisakura:
> is cross-plugin behavior like that even allowed?

Plugins should be naturally compatible with each other (that's why patches are written in a specific way, to preserve functionality), and ideally we shouldn't be adding checks to see if other plugins are enabled. If it's absolutely necessary, then so be it, but please prefer finding a solution where both can work at once without special handling.

PR #2329, opened 2024-04-04
add phishReport plugin

Stefanuk12:
This adds a plugin that makes it super easy to report phishing links. It adds a new menu item within the context menu whenever you right click a link. [^1] Within the settings, you are able to specify your API key and base. [^2]

## Note

~~Due to CORS, a proxy is needed. You can find the source code to the default proxy [here](https://github.com/Stefanuk12/phishreport-proxy).~~

## Footnotes

[^1]: image
[^2]: image

### DIFF
--- a/src/plugins/phishReport/index.tsx
+++ b/src/plugins/phishReport/index.tsx
@@ -20,77 +20,12 @@ import { findGroupChildrenByChildId, NavContextMenuPatchCallback } from "@api/Co
 import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import { ModalCloseButton, ModalContent, ModalFooter, ModalHeader, ModalProps, ModalRoot, ModalSize, openModal } from "@utils/modal";
-import definePlugin, { OptionType } from "@utils/types";
+import definePlugin, { OptionType, PluginNative } from "@utils/types";
 import { Button, Forms, Menu, Text } from "@webpack/common";
 
-const debug = false;
+import { ResponseType } from "./types";
 
-enum ResponseType {
-    Success,
-    Error,
-}
-
-interface SuccessResponse {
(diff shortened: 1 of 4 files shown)


Sqaaakoi:
you should implement a native module where it is supported to get around CORS instead of using a proxy

Sqaaakoi:
also the proxy will probably have to be removed for security reasons (for example; what if you log people's API tokens)

Vendicated:
1. like sqaak said, please do not use a proxy. for security and privacy reasons, we don't allow third party apis that aren't trusted
2. requiring an api key is very bad user experience. does the service you're using offer a way to report stuff without login via their Web dashboard? if yes, you can reverse engineer how the site does it (hint: look at the network tab), and use that method
3. this seems very niche. why would the average discord user need this? please explain the use case a little

Stefanuk12:
> you should implement a native module where it is supported to get around CORS instead of using a proxy

the only issue is that i couldn't find a way how to, so i did that instead. i don't imagine there being a way around it without using a third party http client that doesn't depend on the browser

Vendicated:
look at openInApp

Stefanuk12:
> 2. requiring an api key is very bad user experience. does the service you're using offer a way to report stuff without login via their Web dashboard? if yes, you can reverse engineer how the site does it (hint: look at the network tab), and use that method

if no api key is provided, the user is limited to an "analysis" which does do too much. the takedown process allows you to do more things like reporting to other sites automatically

Zippybonzo (commented):
Seems fine, but needing an API key seems a bit inconvenient.

Vendicated:
yeah sorry but with that in mind this is not very useful :( not to mention that this is way too niche

as such, i'll be rejecting this. thank you a lot for your contribution regardless!

PR #2333, opened 2024-04-05
 (feat) Oneko: allow customization for the speed of the cat :3 && migrate from the decrypted define plugins to the new one

not-varram:
P.S. check commit https://github.com/Vendicated/Vencord/commit/5f7c18a5666f5ecdb768d484b451e0cca087310f or the one with the comment "fix issues based on comments :3" to see the latest modifications to the core plugin (oneko)

### DIFF
--- a/src/plugins/oneko/index.ts
+++ b/src/plugins/oneko/index.ts
@@ -21,46 +21,39 @@ import { definePluginSettings } from "@api/Settings";
 import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
-export const settings = definePluginSettings({
+function runner() {
+    fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
+        .then(x => x.text())
+        .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${onekoSettings.store.speed};`))
+        .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
+            .replace("(isReducedMotion)", "(false)"))
+        .then(eval);
+}
+export const onekoSettings = definePluginSettings({
     speed: {
         description: "Speed of Da Cat :3",
-        type: OptionType.NUMBER,
+        type: OptionType.SLIDER,
         default: 10,
-        isValid: (value: number) => {
-            if (value && value < 0) return "The number must be bigger than 0";
-            return true;
-        },
+        markers: [1, 10, 15, 20, 25, 30, 35, 40, 45, 50],
         onChange: () => {
-            // note: cant call the start() function from here. so i just copy pasted it (This was pointed out in the last commit i made. So this is to just clear stuff up for any future devs that work on this :D )
             if (Settings.plugins.oneko.enabled) {
                 document.getElementById("oneko")?.remove();
-                fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
-                    .then(x => x.text())
-                    .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${onekoSettings.store.speed};`))
-                    .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
-                        .replace("(isReducedMotion)", "(false)"))
-                    .then(eval);
+                runner();
             }
             return;
         }
     },
 });
-
 export default definePlugin({
     name: "oneko",
     description: "cat follow mouse (real)",
     // Listing adryd here because this literally just evals her script
     authors: [Devs.Ven, Devs.adryd, Devs.varram],
 
-    settings,
+    settings: onekoSettings,
 
     start() {
-        fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
-            .then(x => x.text())
-            .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${onekoSettings.store.speed};`))
-            .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
-                .replace("(isReducedMotion)", "(false)"))
-            .then(eval);
+        runner();
     },
 
     stop() {
--- a/src/plugins/oneko/index.ts
+++ b/src/plugins/oneko/index.ts
@@ -16,7 +16,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-
 import { definePluginSettings } from "@api/Settings";
 import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
@@ -29,7 +28,7 @@ function runner() {
             .replace("(isReducedMotion)", "(false)"))
         .then(eval);
 }
-export const onekoSettings = definePluginSettings({
+export const settings = definePluginSettings({
     speed: {
         description: "Speed of Da Cat :3",
         type: OptionType.SLIDER,
@@ -50,7 +49,7 @@ export default definePlugin({
     // Listing adryd here because this literally just evals her script
     authors: [Devs.Ven, Devs.adryd, Devs.varram],
 
-    settings: onekoSettings,
+    settings,
 
     start() {
         runner();
--- a/src/plugins/oneko/index.ts
+++ b/src/plugins/oneko/index.ts
@@ -21,7 +21,7 @@ import { definePluginSettings } from "@api/Settings";
 import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
-export const onekoSettings = definePluginSettings({
+export const settings = definePluginSettings({
     speed: {
         description: "Speed of Da Cat :3",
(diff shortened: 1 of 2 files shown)


src/plugins/oneko/index.ts
```diff
@@ -16,18 +16,48 @@
...
+        onChange: () => {
+            // note: cant call the start() function from here. so i just copy pasted it (This was pointed out in the last commit i made. So this is to just clear stuff up for any future devs that work on this :D )
+            if (Settings.plugins.oneko.enabled) {
+                document.getElementById("oneko")?.remove();
+                fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
+                    .then(x => x.text())
+                    .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${onekoSettings.store.speed};`))
+                    .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
+                        .replace("(isReducedMotion)", "(false)"))
+                    .then(eval);
```
Nuckyz, line 42: Make this a function and re-use inside the sart method

src/plugins/oneko/index.ts
```diff
@@ -16,18 +16,48 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+
+import { definePluginSettings } from "@api/Settings";
+import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
+export const onekoSettings = definePluginSettings({
```
Nuckyz, line 24:
```suggestion
export const settings = definePluginSettings({
```

src/plugins/oneko/index.ts
```diff
@@ -16,18 +16,48 @@
...
+});
 
 export default definePlugin({
     name: "oneko",
     description: "cat follow mouse (real)",
     // Listing adryd here because this literally just evals her script
-    authors: [Devs.Ven, Devs.adryd],
+    authors: [Devs.Ven, Devs.adryd, Devs.varram],
+
+    settings: onekoSettings,
```
Nuckyz, line 55:
```suggestion
    settings,
```

src/plugins/oneko/index.ts
```diff
@@ -16,18 +16,48 @@
...
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
+export const onekoSettings = definePluginSettings({
+    speed: {
+        description: "Speed of Da Cat :3",
+        type: OptionType.NUMBER,
+        default: 10,
+        isValid: (value: number) => {
+            if (value && value < 0) return "The number must be bigger than 0";
```
Nuckyz, line 30: You can just make this a slider instead

Nuckyz (changes requested):

not-varram:
@Nuckyz Made all the changes as directed :D

not-varram:
@Nuckyz its ready to be commited :D

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
...
+    speed: {
+        description: "Speed of Da Cat :3",
+        type: OptionType.SLIDER,
+        default: 10,
+        markers: [1, 10, 15, 20, 25, 30, 35, 40, 45, 50],
+        onChange: () => {
+            if (Settings.plugins.oneko.enabled) {
+                document.getElementById("oneko")?.remove();
+                runner();
+            }
```
lewisakura, line 41: Can this be written in a way that doesn't cause it to re-eval the script every time? e.g. inject a way of modifying the speed without restarting script, or expose the speed variable to window and modify it here.

lewisakura (changes requested):

src/plugins/oneko/index.ts
not-varram: Good point; But the variable onekoSpeed is declared as "const" in the og script, Meaning the value cant be modified; +this is really inconsequential when coming to performance;

not-varram:
@lewisakura Good point; But the variable onekoSpeed is declared as a "const" in the OG script, Meaning the value cant be modified; +This is really inconsequential when coming to performance;

not-varram:
Also i tried that initally and it did not work for some reason :/

not-varram:
I think its ready to be commited now (i have been waiting for a few months 🙏)

src/plugins/oneko/index.ts
Sqaaakoi: I would remove the definition for `nekoSpeed` and replace all references to it with `Vencord.Plugins.plugins.oneko.settings.store.speed`

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
...
+import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
-
+import definePlugin, { OptionType } from "@utils/types";
+function runner() {
+    fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
+        .then(x => x.text())
+        .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${settings.store.speed};`))
+        .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
```
Sqaaakoi, line 27: What's with the multiple chained .then for replacing text, when the line below just continues this replacement statement?

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
...
+        description: "Speed of Da Cat :3",
+        type: OptionType.SLIDER,
+        default: 10,
+        markers: [1, 10, 15, 20, 25, 30, 35, 40, 45, 50],
+        onChange: () => {
+            if (Settings.plugins.oneko.enabled) {
+                document.getElementById("oneko")?.remove();
+                runner();
+            }
+            return;
```
Sqaaakoi, line 42: Remove this useless return, or remove the entire onChange function if you find my suggestion above would make it work without needing to restart it

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
...
 export default definePlugin({
     name: "oneko",
     description: "cat follow mouse (real)",
     // Listing adryd here because this literally just evals her script
-    authors: [Devs.Ven, Devs.adryd],
+    authors: [Devs.Ven, Devs.adryd, Devs.varram],
+
+    settings,
 
     start() {
```
Sqaaakoi, line 54: You could just use `start: runner,`

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import { definePluginSettings } from "@api/Settings";
+import { Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
-
+import definePlugin, { OptionType } from "@utils/types";
```
Sqaaakoi, line 22: where did the the empty line go

Sqaaakoi (changes requested):

src/plugins/oneko/index.ts
not-varram: the main issue is that the core code for the oneko plugin (the cat) isn't mine and I don't wanna repost it to make changes (cuz I don't wanna upset the og creator of oneko)

src/plugins/oneko/index.ts
not-varram: just for organization;+ (remember we are talking the script from another creator and modifying it slightly to work with discord) + I added like 26 bu the other .then statements where already there

src/plugins/oneko/index.ts
not-varram: as started previously  the only way would be to take the code of og creator (which we eval from their script) make our own fork of it and modify it (which I don't know if they will  be ok with).
+ we don't need to go through all that hastle as this is a very **very** minor performance improvement that is not worth the hastle

src/plugins/oneko/index.ts
not-varram: wym?

src/plugins/oneko/index.ts
```diff
@@ -16,21 +16,43 @@
...
+function runner() {
+    fetch("https://raw.githubusercontent.com/adryd325/oneko.js/8fa8a1864aa71cd7a794d58bc139e755e96a236c/oneko.js")
+        .then(x => x.text())
+        .then(s => s.replace("const nekoSpeed = 10;", `const nekoSpeed = ${settings.store.speed};`))
+        .then(s => s.replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
+            .replace("(isReducedMotion)", "(false)"))
+        .then(eval);
```
ryan-0324, line 29:
```suggestion
        .then(s => eval(s.replace(/(?<=const\s+nekoSpeed\s*=\s*)\S+?(?=\s*;)/, `${settings.store.speed}`)
            .replace("./oneko.gif", "https://raw.githubusercontent.com/adryd325/oneko.js/14bab15a755d0e35cd4ae19c931d96d306f99f42/oneko.gif")
            .replace("(isReducedMotion)", "(false)")));
```

ryan-0324: I recommend making the search value for the `nekoSpeed` replace more robust, as it seems the author of oneko.js occasionally makes changes to the script.
Also, I believe that, as @Sqaaakoi previously mentioned, the extra chained `then()` methods are unnecessary, since the `replace()` method of String does not return a Promise object.

Vendicated:

thank you regardless!

not-varram:
@Vendicated the feature request to add an option to change oneko speed has been there for a while all this does is allow customization and it also moves it to the oneko plugin from the decrypted plugin decoration thing to the new one.

PR #2336, opened 2024-04-06
PinDMs: Fixes and improvements to hidden channels

puppy-girl:
Updated the hidden channel function to show unread channels in collapsed categories; better mimics the behaviour of collapsing categories in servers and gives you more visibility on chats you may want to respond to, along with the option to leave a chat visible by marking messages as unread :3

image

Also fixed a few issues in relation to hidden channels:
- Active channels in the DMs section were being hidden if you collapsed DMs
- Alt + up/down was including DMs while the DMs section was collapsed
- Alt + up/down now uses visible channels over uncollapsed channels to work with unreads

Thank you! ❤️


Vendicated:
avoid having a lot of code in patch replacements. prefer creating a function on your plugin and using that

puppy-girl:
cut down the size of a few patch replacements, let me know if there's anything I can do better :]

PR #2337, opened 2024-04-06
new plugin ValidReply ~ fix "Message could not be loaded"

waresnew:
Basically same thing as ValidUser but for replies

Should I just merge this into ValidUser instead of making a new plugin actually? Might have to rebrand the name then 

Discord_nwGtV4zYbv


cheesesamwich:
this is awesome

Golonchy:
very nice
will be an epic merge

src/plugins/validReply/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    channel: Channel
+    referencedMessage: { state: number } // 0 = normal, 1 = couldn't load, 2 = deleted
+    compact: boolean
+    isReplyAuthorBlocked: boolean
+}
+const fetching = new Map<string,string>();
+let ReplyStore:any;
+export default definePlugin({
+    name: "ValidReply",
+    description: "Fixes \"Message could not be loaded\" upon hovering over the reply",
```
Vendicated, line 24:
```suggestion
    description: 'Fixes "Message could not be loaded" upon hovering over the reply',
```

Vendicated:
since this is gonna be a very rare occurence, i don't think having the hover requirement is necessary

src/plugins/validReply/index.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+        fetching.set(message, channel);
+        RestAPI.get({ url: `/channels/${channel}/messages`,
+            query: {
+                limit: 1,
+                around: message
+            },
+            retries: 2
+        }).then(res => {
+            const reply:Message|undefined = res?.body?.[0];
```
Vendicated, line 58: u should verify that u got the right message

waresnew:
> since this is gonna be a very rare occurence, i don't think having the hover requirement is necessary

In one of the servers I'm in, ctrl+F'ing for a certain term guarantees all the replies in the search results to be unloaded (maybe bc the results are spaced apart in terms of timestamp)

So I'm worried if not having a hover check would send like 10 requests at once if the user starts scrolling through the search results (or I could just use a queue)

waresnew:
if it can't fetch the message, it'll update the reply to say it was deleted now (i tried manually looking for a reply that couldn't be loaded by the plugin and it turns out it was indeed deleted)

this is mainly a UX thing

Discord_I268lLW8t8

Vendicated:
thanks, looks good!

your formatting doesn't follow the project's configured formatter. make sure you're using vscode & have the suggested extensions installed! also in general i recommend using more newlines to separate different sections/"paragraphs" of the code

waresnew:
> thanks, looks good!
> 
> your formatting doesn't follow the project's configured formatter. make sure you're using vscode & have the suggested extensions installed! also in general i recommend using more newlines to separate different sections/"paragraphs" of the code

thank you! I'm kinda busy rn but I'll update the formatting of my other prs when I have time.

in terms of the formatting, I use jetbrains and it has eslint + editorconfig installed, so idk why it didn't autoformat

PR #2338, opened 2024-04-06
fix: PatchHelper not autofilling match field

waresnew:
Before:

Discord_XzB83B35BL

After:

Discord_cEkrvhtRTp

I made it so modifying `initialValue` from the parent will cause a re-render anyway, since otherwise, you would only be able to change the textbox's contents through `onChange`

I think `externalValue` would be a better name now bc you can change the value after setting it

I'm pretty sure a separate state was originally made for `value` to let people type in the type box even if there's an error, but you can still do that after my change (so it should be fine)


MrDiamondDog:
I made this exact change when I made the full patch input- it was reverted and I was told this is not the correct solution

[I suggested an alternative solution here](https://discord.com/channels/1015060230222131221/1015063227299811479/1220578508485562428), which seems correct.

waresnew:
> [I suggested an alternative solution here](https://discord.com/channels/1015060230222131221/1015063227299811479/1220578508485562428), which seems correct.

(we talked about it on Discord, but basically it has a side effect where you can't type anything if there's an error [prob a TextInput feature], and we both agreed that you prob have to modify CheckedTextInput to fix this bug)

@MrDiamondDog told me that ven said that this change causes `onChange` to fire without the user typing anything

I think it's bc useEffect hooks run when you first render them (eg. when opening up settings page)

So I just made it so you don't fire onchange if it was bc of an external change (I tested it and no non-user-input onChange events were fired)

Hopefully it's good now

lewisakura:
Using an effect should really be unnecessary here and the actual problem needs to be found rather than using a bandaid solution

waresnew:
now that f74da73 has been added, i'm pretty sure this can be fixed by using the same code that the new regex find uses for the match field as well

@Nuckyz do you wanna add that or should i?

Nuckyz:
> now that [f74da73](https://github.com/Vendicated/Vencord/commit/f74da73086b2b361490e028c999c80b3ac2ba76a) has been added, i'm pretty sure this can be fixed by using the same code that the new regex find uses for the match field as well
> 
> @Nuckyz do you wanna add that or should i?

Can you clarify which code?

waresnew:
> > now that [f74da73](https://github.com/Vendicated/Vencord/commit/f74da73086b2b361490e028c999c80b3ac2ba76a) has been added, i'm pretty sure this can be fixed by using the same code that the new regex find uses for the match field as well
> > @Nuckyz do you wanna add that or should i?
> 
> Can you clarify which code?

mainly the `onFindBlur` function and the `findError`/`setFindError`

and then instead of using CheckedTextInput for the match field we can use TextInput and reuse the functions above

Nuckyz:
I mean yeah, but the problem of CheckedTextInput still needs to be fixed I guess

waresnew:
> I mean yeah, but the problem of CheckedTextInput still needs to be fixed I guess

we could just not use CheckedTextInput

since the Find field just uses TextInput and it achieves the same purpose (validating the regex and displaying an error)

Nuckyz:
Okay I did the changes, super small but can someone review?

waresnew:
it was crashing if you typed out an invalid regex (firing onMatchChange) so i added regex validation during the canoncalizeMatch() part

this way they can still type if it's invalid regex

Nuckyz:
good catch, for some reason it worked fine in my tests

Nuckyz (approved):

Nuckyz:
Thanks for the effort of fixing it!

PR #2339, opened 2024-04-07
Better error for primitives on proxyLazy + fix StartupTimings

waresnew:
`findByPropsLazy` makes the properties of the lazy object also lazy, but that doesn't work for primitives bc `Reflect.get` can't be run on primitives

Before:

image

After:

image

### DIFF
--- a/src/plugins/startupTimings/index.tsx
+++ b/src/plugins/startupTimings/index.tsx
@@ -17,9 +17,9 @@
 */
 
 import { Devs } from "@utils/constants";
-import { LazyComponent } from "@utils/react";
 import definePlugin from "@utils/types";
 
+import StartupTimingPage from "./StartupTimingPage";
 export default definePlugin({
     name: "StartupTimings",
     description: "Adds Startup Timings to the Settings menu",
@@ -31,5 +31,5 @@ export default definePlugin({
             replace: '{section:"StartupTimings",label:"Startup Timings",element:$self.StartupTimingPage},$&'
         }
     }],
-    StartupTimingPage: LazyComponent(() => require("./StartupTimingPage").default)
+    StartupTimingPage
 });
--- a/src/plugins/startupTimings/index.tsx
+++ b/src/plugins/startupTimings/index.tsx
@@ -20,6 +20,7 @@ import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
 
 import StartupTimingPage from "./StartupTimingPage";
+
 export default definePlugin({
     name: "StartupTimings",
     description: "Adds Startup Timings to the Settings menu",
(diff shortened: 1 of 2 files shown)


src/utils/lazy.ts
```diff
@@ -116,8 +116,12 @@ export function proxyLazy<T>(factory: () => T, attempts = 5, isChild = false): T
...
                     true
                 );
-
-            return Reflect.get(target[kGET](), p, receiver);
+            const obj = target[kGET]();
+            if (typeof obj === "object" || typeof obj === "function") {
+                return Reflect.get(obj, p, receiver);
+            } else {
+                return ()=>obj; // don't call Reflect.get if it's a primitive
+            }
```
Nuckyz, line 124:
```suggestion
            const lazyTarget = target[kGET]();
            if (typeof lazyTarget === "object" || typeof lazyTarget === "function") {
                return Reflect.get(lazyTarget, p, receiver);
            }

            const val = lazyTarget[p];
            return typeof val === "function" ? val.bind(lazyTarget) : val;
```

Nuckyz (changes requested):

waresnew:
Thanks @Nuckyz for telling me how to actually fix startuptimings

src/utils/lazy.ts
```diff
@@ -116,8 +116,11 @@ export function proxyLazy<T>(factory: () => T, attempts = 5, isChild = false): T
                     attempts,
                     true
                 );
-
-            return Reflect.get(target[kGET](), p, receiver);
+            const lazyTarget = target[kGET]();
+            if (typeof lazyTarget === "object" || typeof lazyTarget === "function") {
+                return Reflect.get(lazyTarget, p, receiver);
+            }
+            throw new Error("proxyInner called on a primitive value");
```
Nuckyz, line 123:
```suggestion
            throw new Error("proxyLazy called on a primitive value");
```

src/plugins/startupTimings/index.tsx
```diff
@@ -17,9 +17,9 @@
 */
 
 import { Devs } from "@utils/constants";
-import { LazyComponent } from "@utils/react";
 import definePlugin from "@utils/types";
 
+import StartupTimingPage from "./StartupTimingPage";
```
Nuckyz, line 22:
```suggestion
import StartupTimingPage from "./StartupTimingPage";

```

Nuckyz (changes requested):

src/utils/lazy.ts
waresnew: 💀 mbbbbb

Nuckyz (approved):

PR #2342, opened 2024-04-08
feat(plugin): CustomIdle

waresnew:
Fulfills Vencord/plugin-requests#199

Main feature is changing the timer until your status gets set to idle (or preventing auto-idle forever)

image

There's also a confirmation notification where it'll let you stay idle, even if you come back to Discord

image

You'll stay idle until you click "Exit idle"


waresnew:
ngl i think it'd be better to do like a "You are currently still idle. Click the button to exit idle" bar instead of the grace period

lemme change that

Vendicated:
keep in mind that this idle timer is coupled with mobile notifications. so if you set it to a very high number or disable it outright, it might break mobile notifications

waresnew:
> keep in mind that this idle timer is coupled with mobile notifications. so if you set it to a very high number or disable it outright, it might break mobile notifications

oh ok i'll make a more precise patch for disabling then

waresnew:
aight it won't affect mobile notifications now (it'll respect the setting in vanilla discord)

you can see here:

image

^ I'm online (disabled auto-idle) but I'm afk

Also changed the "grace period" thing to this:

image

Golonchy:
Cool af

Wraaath:
Bump

R00tB33rMan:
Did some testing with this and feel it’s ready for use. Vesktop currently does not handle Idling very well and this as a workaround seems like a fair consideration to me. Could not produce any bugs with this branch. At this time, I have to run four Discord tabs, instead of two to keep my Discord online at all times, which is incredibly inefficient. How ready would you say this is or what’s blocking this from being added to the plugin library, @Vendicated?

JosTheDude:
Requesting addition here, works well

Kromilan:
I was deperatly looking for a plugin like this. When I have Discord open on my second monitor and have an eye on it, I am there. Hands on my keyboard and mouse, and always an eye on Discord. But I still get set to idle when I don't actually click into the client for too long and then they think I'm not available and nobody messages me any more :( I'm so glad there's a plugin in development to fix this!!
I have a question though. I am absolutely not familiar with github though, so sorry if this is a stupid question: can this already be installed and if so, how? I saw some people in the comments tested it, so I guess they installed it somehow, but then there are also comments suggesting it's not available yet?

waresnew:
> can this already be installed and if so, how?

it can be installed, but it's not an official plugin yet (vee needs to approve it first) so you need to install the [dev version of Vencord](https://github.com/Vendicated/Vencord/blob/main/docs/1_INSTALLING.md). then you download the index.ts in this pull request and you move it to `src/userplugins/customidle` (create a folder)

however you won't get any support with dev versions/unofficial plugins like this one so you have to figure it out yourself in the end

UglukOrc:
I'm so happy you've made this dream come true! I understand it isn't officially released, But, I found it amazing how BetterDiscord had something like it, and I was just wondering beyond wonder why Vencord didn't have any option for it! Thanks @waresnew

Vendicated:
> How ready would you say this is or what’s blocking this from being added to the plugin library, @Vendicated?

@R00tB33rMan there's nothing blocking it from being merged, it's just that i haven't gotten to it yet! there's quite the backlog of pull requests so it sadly takes a while sometimes. anyway ill look at this some time in the coming few days

you (and anyone reading this) can always help by testing open pull requests early and leaving constructive feedback / finding bugs yourself

for example in this pr it'd be good to test & report if this causes any issues / edge cases with mobile notifications or if those work flawlessly

Vendicated:
please add a readme file

waresnew:
ok it looks like a lot of people use this plugin to partially solve idling while the window is in the foreground

i think i'll be able to add that as a separate setting at some point

in terms of formatting, i can't access vscode rn so i'll fix it in like 10 hours

i'll add a readme in a moment

src/plugins/customidle/index.ts
```diff
@@ -0,0 +1,85 @@
...
+            sentNotif = true;
+            Notices.showNotice("Welcome back! Click the button to go online. Click the X to stay idle until reload.", "Exit idle", () => {
+                Notices.popNotice();
+                FluxDispatcher.dispatch({
+                    type: "IDLE",
+                    idle: false
+                });
+                sentNotif = false;
+            });
+        }
```
Nuckyz, line 76:
```suggestion

        const backOnlineMessage = "Welcome back! Click the button to go online. Click the X to stay idle until reload.";
        if (Notices.noticesQueue.some(([, noticeMessage]) => noticeMessage === backOnlineMessage)) return;

        Notices.showNotice(backOnlineMessage, "Exit idle", () => {
            Notices.popNotice();
            FluxDispatcher.dispatch({
                type: "IDLE",
                idle: false
            });
        });
```

Nuckyz (changes requested):

Nuckyz:
The way this is currently done has a big problem. The same function that handles setting mobile notifications to true is the function that sets the idle. It is ran using a setInterval as seen here:
image

Modifying IDLE_DURATION to MAX_SAFE_INTEGER causes "C" to run on loop non stop. But also, modifying it means you are still making setting mobile notifications depend on the new idle timeout, because it changes how often it does the checking

Nuckyz:
I have applied my suggestion of a better of handling showing the notice, and also better formatting. But the issues I mentioned are still present

waresnew:
ok i just realized that setInterval's delay has to be a 32 bit integer

so MAX_SAFE_INTEGER was making it overflow to 0 :skull:

technically this made it so the setinterval was still being run but this probably caused lag

Vendicated:
so this still needs work if i'm understanding correctly? marking as draft accordingly

waresnew:
ok it won't prevent you from being AFK on web installs now

i'm gonna unmark it as draft now

src/plugins/customidle/index.ts
```diff
@@ -0,0 +1,94 @@
...
+                match: /(IDLE_DURATION:function\(\){return )\i/,
+                replace: "$1$self.getIdleTimeout()"
+            }
+        },
+        {
+            find: 'type:"IDLE",idle:',
+            replacement: [
+                {
+                    match: /Math\.min\((\i\.AfkTimeout\.getSetting\(\)\*\i\.default\.Millis\.SECOND),\i\.IDLE_DURATION\)/,
+                    replace: "$1" // Decouple idle from afk (phone notifications will remain at user setting)
```
Nuckyz, line 48:
```suggestion
                    replace: "$1" // Decouple idle from afk (phone notifications will remain at user setting or 10 min maximum)
```

src/plugins/customidle/index.ts
```diff
@@ -0,0 +1,94 @@
...
+            FluxDispatcher.dispatch({
+                type: "IDLE",
+                idle: false
+            });
+        });
+    },
+
+    getIdleTimeout() { // milliseconds, default is 6e5
+        const { idleTimeout } = settings.store;
+        return idleTimeout === 0 ? Number.MAX_SAFE_INTEGER : idleTimeout * 60000;
```
Nuckyz, line 92:
```suggestion
        return idleTimeout === 0 ? Infinity : idleTimeout * 60000;
```

Nuckyz (changes requested):

Nuckyz:
Great fix, just these two nitpicks and it should be ready

src/plugins/customidle/README.md
```diff
@@ -0,0 +1,6 @@
+# CustomIdle
+
+Lets you change the time until your status gets automatically set to idle. You can also prevent idling altogether.
+
+![](https://github.com/Vendicated/Vencord/assets/45801973/4e5259b2-18e0-42e5-b69f-efc672ce1e0b)
+
```
Nuckyz, line 6:
```suggestion
![Plugin Configuration](https://github.com/Vendicated/Vencord/assets/45801973/4e5259b2-18e0-42e5-b69f-efc672ce1e0b)
```

Nuckyz (changes requested):

waresnew:
okk made the changes 👍

Nuckyz:
Thank you for the new plugin!

UglukOrc:
> Thank you for the new plugin!

Hol up! This working now?? @Nuckyz

Nuckyz:
When it gets released to the main branch, yes

R00tB33rMan:
Thank you all for your hard work. I really appreciate everyone's interest in improving the user experience and am glad that this was merged through everyone's availability and persistence.

DanielVALAT:
Hello
The plugin doesn't work anymore :(
I remain inactive when I lock my computer (after 0 secs)

Discord_OkDnD5FOFP
Discord_YsCYE134lw

Vendicated:
use our support channel instead of necroing old threads.

PR #2021, opened 2023-11-30
new plugin WatchTogetherAdblock: block ads in youtube activity

ImLvna:
Finally got around to pr'ing this

PR #2343, opened 2024-04-09
Add adblock plugin

pogoamerica:
Adds a plugin that just follows the [example usage of the ghostery adblocker for electron](https://github.com/ghostery/adblocker/blob/master/packages/adblocker-electron/README.md).

Currently the only activity I know of with ads is the WatchTogether one, but in theory this should also block ads in other activities since it's applied to the default electron session

Apart from the included .ts files, this adds a few dependencies
```
$ pnpm install --save-dev cross-fetch
$ pnpm install --save-dev @cliqz/adblocker-electron
$ pnpm install --save-dev @cliqz/adblocker-electron-preload
```

So far I've just tested this by building/injecting and watching the WatchTogether activity. I know next to nothing about electron/js in general so if there's anything else I need to test/change, please let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,8 @@
         "virtual-merge": "^1.0.1"
     },
     "devDependencies": {
+        "@cliqz/adblocker-electron": "^1.27.0",
+        "@cliqz/adblocker-electron-preload": "^1.27.0",
         "@types/chrome": "^0.0.246",
         "@types/diff": "^5.0.3",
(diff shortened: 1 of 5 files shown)


Vendicated:
duplicate of https://github.com/Vendicated/Vencord/pull/2021

you just added half a megabyte of code (bundled & minified). please don't introduce new dependencies so carelessly

image
image

PR #2344, opened 2024-04-09
feat(plugin): enables lootboxes back again

happyendermangit:
Discord as of today disabled lootboxes, so i made this plugin that enables them back
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.7.7",
+    "version": "1.7.8",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
     "bugs": {


mufaroxyz (commented):
LGTM

src/plugins/LootBoxes/LootBoxes.js
```diff
@@ -0,0 +1,44 @@
...
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "LootBoxes",
+    description: 'Enables lootboxes back.',
+    authors: [Devs.HAPPY_ENDERMAN],
+    patches: [
+        /** enable the badge */
+        {
+            find: "Messages.PACKAGES_BADGE",
```
MrDiamondDog, line 29: Dont use \w+, use \i (custom selector made by us)

Vendicated: no it isn't. variable names may contain characters that are not matched by \w, such as $

Vendicated:
i'm not sure how useful this is.

won't discord just remove the api / code for this very soon? longevity of this plugin seems unlikely

why not post it as a js snippet instead?

happyendermangit:
The api will never be removed, discord will just forget about this, a js snippet is not possible...

MrDiamondDog:
It is getting removed very soon https://discord.com/channels/1015060230222131221/1105582084237570088/1227358383468838965

byeoon (changes requested):
If this were to be a update that was pushed upstream, this can't be the README.

PR #2346, opened 2024-04-09
add Native settings implementation

MachineMuse:
It's not much but it will help a bit with the type signatures and create a more obvious place for things to go in the future if applicable.

It might be better to use the `mergeDefaults` in misc.tsx but it crashes due to the `isMobile` definition in there (`ReferenceError: navigator is not defined`) and I don't want to move around exports that might be used by e.g. vesktop.


Vendicated:
u can just extract mergeDefaults to src/utils/mergeDefaults.tx

Vendicated:
looks good, thanks!

PR #2349, opened 2024-04-11
Plugin: Media downloader

MachineMuse:
Depends on #2346

Addresses plugin request https://github.com/Vencord/plugin-requests/issues/188

Many TODOs still but it's working pretty well for what it is so I figured I could put it in the review queue and if I get more time to work on it I'll just keep adding on.

Features (each can be enabled/disabled individually):
- (background) A 'safe' native interface that allows the user to select their media download folder without exposing direct file writing capabilities to the renderer
- A button in the message hover toolbar ('accessories') which downloads all embedded and attached media to the configured folder in a single click
- A menu item in the context menu when clicking on an image (in a message or when enlarged) which downloads the image to the preselected folder in one click
  - This currently can only download proxied media and still previews of videos, just like the 'save as' menu option. It might be possible to improve this behaviour but it'll be more invasive.
- A download hover button over image attachments, in the top-right, like the one over video attachments
  - Only attached images, not embedded ones. Embedded ones will need a separate, more invasive patch.
- An option to hijack aforementioned hover button to one-click download to the configured media folder (for videos, images, and audio) instead of opening it in the browser

### DIFF
--- a/src/plugins/mediaDownloader.desktop/index.tsx
+++ b/src/plugins/mediaDownloader.desktop/index.tsx
@@ -1,0 +1,59 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addButton, removeButton } from "@api/MessagePopover";
+import { Devs } from "@utils/constants";
+import definePlugin, { PluginNative } from "@utils/types";
+import { ChannelStore } from "@webpack/common";
+
+import { DownloadImagesIcon } from "./DownloadImagesIcon";
+import { settings } from "./settings";
+
+const Native = VencordNative.pluginHelpers.MediaDownloader as PluginNative<typeof import("./native")>;
+
+
+export default definePlugin({
+    name: "MediaDownloader",
+    description: "Tool(s) to quickly save images and videos",
+    authors: [Devs.Eloelle],
+    settings: settings,
+
+    // TODO: Add other interactions such as floating download button on individual media
+    start() {
+        addButton("vc-imagedownload", message => {
+            if (message.attachments.length === 0 && message.embeds.length === 0) return null;
+            return {
+                label: "Download All Images & Videos",
+                icon: DownloadImagesIcon,
+                message,
+                channel: ChannelStore.getChannel(message.channel_id),
+                onClick: async () => {
+                    // console.log("Attachments:");
+                    message.attachments.forEach(attachment => {
+                        // console.log(attachment);
+                        Native
+                            .downloadFile(attachment.url, attachment.proxy_url)
+                            .then(success => console.log(success))
+                            .catch(error => console.log(error));
+                    });
+                    // console.log("Embeds:");
+                    message.embeds.forEach(embed => {
+                        // console.log(embed);
+                        const target = embed.image || embed.video || null;
+                        if (!target) return false;
+                        Native
+                            .downloadFile(target.url, target.proxyURL || "")
+                            .then(success => console.log(success))
+                            .catch(error => console.log(error));
+                    });
+                }
+            };
+        });
+    },
+    stop() {
+        removeButton("vc-imagedownload");
+    }
+});
--- a/src/plugins/mediaDownloader.desktop/index.tsx
+++ b/src/plugins/mediaDownloader.desktop/index.tsx
@@ -60,6 +60,38 @@ export default definePlugin({
     contextMenus: {
         "message": messageContextMenuPatch,
         "image-context": imageContextMenuPatch
-    }
-    // TODO: Add floating download button on individual media
+    },
+    patches: [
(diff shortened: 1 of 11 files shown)


src/plugins/mediaDownloader.desktop/settings.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { definePluginSettings } from "@api/Settings";
+import { OptionType, PluginNative } from "@utils/types";
+import { Button, Flex, Forms, TextInput, Toasts, useEffect, useState } from "@webpack/common";
+
+const Native = VencordNative.pluginHelpers.MediaDownloader as PluginNative<typeof import("./native")>;
+
+function DirectoryPickerComponent(props: { setValue(v: any): void; }) {
+    const [value, setValue] = useState("");
+    useEffect(() => { Native.getMediaFolder().then(setValue).catch(console.log); });
```
Vendicated, line 15: this is a bit strange, it'd be better to do something like:
```js
const [signal, forceUpdate] = useForceUpdater();
const [value, error, isLoading] = useAwaiter(Native.getMediaFolder, {
	deps: [signal],
	defaultValue: null
})
```

this will properly handle the promise & error handling for you and calling forceUpdate will re-run it

src/plugins/mediaDownloader.desktop/native.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!dir) return false;
+    if (dir === "C:\\") return false;
+    if (dir.startsWith("/dev")) return false;
+    if (dir.startsWith("/bin")) return false;
+    // TODO more reasonable safety checks
+    return true;
+}
+
+function getNativeSettings() {
+    if (!NativeSettings.store.plugins) NativeSettings.store.plugins = {};
```
Vendicated, line 24:
```suggestion
    NativeSettings.store.plugins ??= = {};
```

src/plugins/mediaDownloader.desktop/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+        return Native
+            .downloadFile(sourceURI, proxyURI)
+            .then(success => console.log(success))
+            .catch(error => console.log(error));
+    },
+    downloadFileSafe(uri: string) {
+        return Native
+            .downloadFile(uri, uri)
+            .then(success => console.log(success))
+            .catch(error => console.log(error));
```
Vendicated, line 94: these should definitely notify the user in some other way, e.g. a toast

src/plugins/mediaDownloader.desktop/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+                match: /!(\w{1,3}&&)"AUDIO"===(\w{1,3}\|\|)/,
+                replace: (m, z, v) => `${m}!${z}"IMAGE"===${v}`
+            },
+            predicate: () => settings.store.addHoverButtonToImageAttachments
+        },
+        // Hijack download hoverbutton to use media downloader
+        {
+            find: "AnalyticEvents.MEDIA_DOWNLOAD_BUTTON_TAPPED",
+            replacement: {
+                match: /href:(\w{1,3}),onClick:(\w{1,3}),target:(\w{1,3}),rel:(\w{1,3}),className:(\w{1,3}),"aria-label":(\w{1,3}).default.Messages.DOWNLOAD,/g,
```
Vendicated, line 78: Vencord has a custom \i escape sequence that matches identifiers!
```suggestion
                match: /href:(\i),onClick:(\w{1,3}),target:(\w{1,3}),rel:(\w{1,3}),className:(\w{1,3}),"aria-label":(\w{1,3}).default.Messages.DOWNLOAD,/g,
```

src/plugins/mediaDownloader.desktop/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+        addButton("vc-imagedownload", message => {
+            if (message.attachments.length === 0 && message.embeds.length === 0) return null;
+            if (!settings.store.showInMessageHoverMenu) return null;
+            return {
+                label: "Download All Images & Videos",
+                icon: DownloadImagesIcon,
+                message,
+                channel: ChannelStore.getChannel(message.channel_id),
+                onClick: async () => {
+                    // console.log("Attachments:");
```
Vendicated, line 35: please remove all these debug lines once you're done

also might i suggest using the debugger instead? it's underrated! you can just search some code of your plugin in devtools (ctrl shift f), jump to the source code (make sure you have js source maps enabled) and put breakpoints

Vendicated (commented):
Wow, this looks great so far, especially for a first time contributor!

This is only a partial review for now, mainly cause i feel bad for taking so long to get to it

I'll have to do a more thorough review soon, hopefully tomorrow, due to the size & potential security issues there could be

I still see a bunch of TODO comments... Do you need help with those or are you still working on them?

MegaCornflakes:
Is this still being developed? I'm working on a similar plugin  that supports multiple folders/image sources, but I don't want to duplicate something that already exists.

PR #2352, opened 2024-04-13
fixes the readallnotifications button's css

Golonchy:
Hi, i fixed the css for the readallnotifications button, making it not transparent and making it towards the center and left and right margin not touch other elements of discord.
here is a image,
Screenshot 2024-04-13 at 21 14 26

Thank you -@golonchy

### DIFF
--- a/src/plugins/readAllNotificationsButton/style.css
+++ b/src/plugins/readAllNotificationsButton/style.css
@@ -1,11 +1,8 @@
 .vc-ranb-button {
-    color: var(--interactive-normal);
-    padding: 0 0.5em;
+    background-color: rgb(30, 46, 226);
+    padding: 0.3em;
     margin-bottom: 0.5em;
-    width: 100%;
     box-sizing: border-box;
-}
-
-.vc-ranb-button:hover {
-    color: var(--interactive-active);
+    position: relative; 
+    margin-left: 5.5px;
 }


Golonchy:
i forgot to include a before picture, silly me
here is before with the previous style
Screenshot 2024-04-13 at 20 13 23

Tr0llie:
This actually fixes my issue with that button right now, but the color is too vibrant, maybe change the color of the button to the normal discord color `#7785cc` 
is just an suggestion anyway.
also another thing, i should make this in plugin request but what if users can custom pick the color.

Golonchy:
> This actually fixes my issue with that button right now, but the color is too vibrant, maybe change the color of the button to the normal discord color `#7785cc` is just an suggestion anyway. also another thing, i should make this in plugin request but what if users can custom pick the color.

i have it the dark blue color so it's easy to see.
to make the color customizable i think i would need to change the `index.ts` and idk typescript lol

Tr0llie:
> > This actually fixes my issue with that button right now, but the color is too vibrant, maybe change the color of the button to the normal discord color `#7785cc` is just an suggestion anyway. also another thing, i should make this in plugin request but what if users can custom pick the color.
> 
> i have it the dark blue color so it's easy to see. to make the color customizable i think i would need to change the `index.ts` and idk typescript lol

tbh we dont really need a color picker, is just a small button

Vendicated:
i deliberatyle made it transparent via https://github.com/Vendicated/Vencord/commit/8ab56f5bcfe720b8e6d03f85b5f4ada1a18bfc37 to be less obtrusive 

if you think this is now too unnoticeable, you can propose an alternative solution but your current pr is very ugly. it would probably be ideal to make it in a similar style as these. as in, use the same colour and stuff 
image

this pr also hardcodes a rgb value which is always bad because it won't adapt to themes. stick to discord css variables

PR #2353, opened 2024-04-13
new plugin MessageLatency: indicator for other people's latency

arHSM:
# MessageLatency

Displays an indicator for messages that took ≥n seconds to send.

> **NOTE**
>
> -   This plugin only applies to messages received after opening the channel
> -   False positives can exist if the user's system clock has drifted.
> -   Grouped messages only display latency of the first message

### Chat View

chat-view

### Clock -ve Drift

pissbot-on-top

### Clock +ve Drift

dumb-ai

### Connection Delay

who-this

### Icons

icons

### DIFF
--- a/src/plugins/messageLatency/index.tsx
+++ b/src/plugins/messageLatency/index.tsx
@@ -1,0 +1,147 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import { isNonNullish } from "@utils/guards";
+import definePlugin, { OptionType } from "@utils/types";
+import { waitFor } from "@webpack";
+import { SnowflakeUtils, Tooltip } from "@webpack/common";
+import { Message } from "discord-types/general";
+
+type FillValue = ("status-danger" | "status-warning" | "text-muted");
+type Fill = [FillValue, FillValue, FillValue];
+type DiffKey = keyof Diff;
+
+interface Diff {
+    days: number,
+    hours: number,
+    minutes: number,
+    seconds: number;
+}
+
+let HiddenVisually;
+waitFor("HiddenVisually", mod => {
+    HiddenVisually = mod.HiddenVisually;
+});
+
+export default definePlugin({
+    name: "MessageLatency",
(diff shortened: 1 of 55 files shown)


Grafaffel:
this is really well made

Vendicated:
so what exactly is the use case? why would a user want this feature?

Tr0llie:
> so what exactly is the use case? why would a user want this feature?

i can see many use cases as giveaways or other important stuff that needs exact timing. i would use this.
as per my opinion

arHSM:
image

src/plugins/messageLatency/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+        // Also thanks dziurwa
+        // 2 minutes
+        const TROLL_LIMIT = 2 * 60;
+        const { latency } = this.settings.store;
+
+        const fill: Fill = delta >= TROLL_LIMIT || ahead ? ["text-muted", "text-muted", "text-muted"] : delta >= (latency * 2) ? ["status-danger", "text-muted", "text-muted"] : ["status-warning", "status-warning", "text-muted"];
+
+        return abs >= latency ? { delta: stringDelta, ahead: abs !== delta, fill } : null;
+    },
+    Tooltip(message: Message) {
```
Vendicated, line 89: please wrap this in a noop-ErrorBoundary to avoid crashing in the worst case scenario

```suggestion
    Tooltip: ErrorBoundary.wrap(({ message }: { message: Message }) => {
```

(you'll also have to slightly adjust ur patch since now it expects a proper props object as first arg)

src/plugins/messageLatency/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+    stringDelta(delta: number) {
+        const diff: Diff = {
+            days: Math.round(delta / (60 * 60 * 24)),
+            hours: Math.round((delta / (60 * 60)) % 24),
+            minutes: Math.round((delta / (60)) % 60),
+            seconds: Math.round(delta % 60),
+        };
+
+        const str = (k: DiffKey) => diff[k] > 0 ? `${diff[k]} ${k}` : null;
+        const keys: DiffKey[] = ["days", "hours", "minutes", "seconds"];
```
Vendicated, line 60:
```suggestion
        const keys = Object.keys(diff);
```

Vendicated (commented):
Looks good so far!

Vendicated:
thanks pookie

PR #2355, opened 2024-04-14
pronounDB: Update to API v2

EdVraz:
this probably sucks but it works :3


src/plugins/pronoundb/pronoundbUtils.ts
```diff
@@ -140,21 +141,23 @@ async function bulkFetchPronouns(ids: string[]): Promise<PronounsResponse> {
...
-    // If it is set to lowercase and a special code (any, ask, avoid), then just return the capitalized text
-    else if (
-        pronounsFormat === PronounsFormat.Lowercase
-        && ["any", "ask", "avoid", "other"].includes(pronouns)
-    ) return PronounMapping[pronouns];
-    // Otherwise (lowercase and not a special code), then convert the mapping to lowercase
-    else return PronounMapping[pronouns].toLowerCase();
+
+    if (pronouns.length === 1) {
+        if (pronounsFormat === PronounsFormat.Capitalized || ["any", "ask", "avoid", "other", "unspecified"].includes(pronouns[0]))
```
Vendicated, line 157: what's this [...].includes() for?

would be good to add a short comment explaining it

src/plugins/pronoundb/types.ts
```diff
@@ -26,31 +26,41 @@ export interface UserProfilePronounsProps {
...
+    they: "They/Them",
     any: "Any pronouns",
     other: "Other pronouns",
     ask: "Ask me my pronouns",
     avoid: "Avoid pronouns, use my name",
-    unspecified: "Unspecified"
+    unspecified: "No pronouns specified.",
+} as const;
+
+export const ShortPronounMapping = {
```
Vendicated, line 56: why is this map needed? seems to just map lowercase -> titlecase, you could just do
```js
const toTitle = (s: string) => s[0].toUpperCase() + s.slice(1)
```
instead

Vendicated:
looks fine other than the two reviews, thanks for migrating the plugin! :3

Vendicated:
thank youuu <3

PR #2360, opened 2024-04-17
feat(plugin): AmITyping

MrDiamondDog:
Shows if you are typing. Works with all other typing-related plugins

### DIFF
--- a/src/plugins/amityping/README.md
+++ b/src/plugins/amityping/README.md
@@ -1,0 +1,3 @@
+# AmITyping
+
+Simply shows if you are typing. This is almost exactly what it will look like for other users, so it may show after you have finished typing.
--- a/src/plugins/amityping/index.ts
+++ b/src/plugins/amityping/index.ts
@@ -1,0 +1,24 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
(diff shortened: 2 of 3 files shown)


Kodarru:
Tbh, I don't see a point in this

Koishukaze:
Is there text in your message input? congrats, you are typing!

Zippybonzo:
Seems useless IMO.

byeoon:
I think this plugin is nice, accidentally typed a character in a channel you were lurking in and now you can't tell if others can see that or not

MrDiamondDog:
I made this plugin so you have a very clear indicator of when other people can see you typing, which I personally find useful

Kodarru:
Why don't you add this to third-party plugins instead

SerStars:
> I made this plugin so you have a very clear indicator of when other people can see you typing, which I personally find useful

I understand your point but you can view in the members list if you are typing and even in the network logs, of course it's hard to see yourself if its a long members list with all the scrolling drama, but i think this is more a 3rd party plugin on my personal opinion

SerStars:
> I made this plugin so you have a very clear indicator of when other people can see you typing, which I personally find useful

I understand your point but you can view in the members list if you are typing and even in the network logs, of course it's hard to see yourself if its a long members list with all the scrolling drama, but i think this is more a 3rd party plugin on my personal opinion

PR #2361, opened 2024-04-17
Plugin: VoiceCommands

aequabit:
Adds various options to user and channel context menus, which allow the user to quickly manage the voice channels created by popular bots on large servers (example: https://wiki.mee6.xyz/en/plugins/temporary-channels).

Channel context menu User context menu

I was not sure if naming it "MEE6 Voice Channel Shortcuts" was a smart idea due to trademarks, so I chose a neutral name.


Sqaaakoi:
plugin would be better if it wasn't designed for one specific bot (that is very much widely hated)

also use slash commands instead

Vendicated:
yeah this is way too niche, sorry

especiiiially not mee6 i want nothing to do with that spawn of hell in the mod

(also for future reference, usage of their name in the plugin name or description would fall under fair use, as long as you don't make it seem like you're officially endorsed by mee6. "CommandsForMee6" would be fine, something like "Mee6Pro" probably wouldn't)

Vendicated:
now if you made this more generic to work for any bot, that could possibly be considered

like being able to add custom command shortcuts to various context menus

Sqaaakoi:
> now if you made this more generic to work for any bot, that could possibly be considered
> 
> like being able to add custom command shortcuts to various context menus

potential idea for this could be a mod bot that allows you to quickly lock/unlock a channel

however i had this as an idea before (but with internal slash commands), where it would just overwrite the channel permissions directly, instead of using a mod bot (which would be... too overcomplicated?)

yes i am yapping again

aequabit:
> plugin would be better if it wasn't designed for one specific bot (that is very much widely hated)

This was exactly my thought, but the plugin settings seemed too limited for that (i.e. only single-line textbox). Maybe a JSON format with presets for popular bots and a web UI to generate them would be a good idea - I'm open for any suggestions.

> also use slash commands instead

For some reason they didn't work, maybe because I used the wrong message type.

TheKodeToad:
I'm pretty sure you don't use the message creation endpoint, but the interaction creation endpoint.
However, multiple slash commands could have the same name and it would be a pain to deal with :/

PR #2362, opened 2024-04-18
Update patch-worker.js comment

sebilune:
Multiline comment simplicity


Vendicated:
what? this is a vendored file and comments are compiled away, why would we change anything about this

sebilune:
> what? this is a vendored file and comments are compiled away, why would we change anything about this

It an attempt to get that cool contributor badge 😔

lewisakura:
Contributor badges are not granted because you made a single PR, nor are they granted automatically, they are for significant contributions to the project as a token of gratitude. Please do not try and spoil that goodwill for others by making these sorts of PRs.

PR #2369, opened 2024-04-19
fix ValidUser

EdVraz:


gonsodany:
so was this merged or not? and does this fix it? because im almost pretty sure discord added some kind of privacy block (the same say you can see channels you dont have access to) now you cant see users you've never interacted with, whenever you used to see numbers, now you'll be blocked from seeing the user profile all together, which is raally stupid, essentially made the userid useless for everyone who actually used it. I might be wrong tho (hopefully)

byeoon:
yes, i believe they did add a userID privacy block for whatever fucking reason

this might be my "exploit vulnerabilities in discord api  because those fuckers at discord HQ pissed me off by adding a dumb feature" arc

gonsodany:
what a useless fucking "security" feature, anyone that used the userid to be able to check people by their ids cant do that anymore, its so fucking anoying, i tested with an account, i mentioned them with the <@id> thing, for my friend they could click the profile, but for me it said "unknown user" and i was locked out of their profile, even tho we had 9 servers in common mind you.

I had to go to one of the servers we have in common, search for them and then click on their profile, after doing that the mention starts working again.

Why the fuck am i locked out of a user's profile who i have 9 servers in common? what "security" does exactly that bring?

I run a community server now if anyone comes to me reporting a user id i cant even check their profile by mentioning to check what servers we have in common. Fucking useless

gonsodany:
LMK if we ever find a workaround it, because as of now you can still get a profile through the api, not sure about actually opening a profile tho

lewisakura:
There is no point complaining about Discord's measures here. Use the suggestions website for that.

Vendicated:
what even is this rant. discord changed absolutely nothing, other than changing the display of mentions of uncached users from `<@1234567890>` -> `@unknown-user`

For the average Discord user who doesn't know what a user id is or how to use them, this is objectively more user friendly. Not to mention that this makes it more consistent with role or channel mentions, which have always rendered as `@deleted-role` and `#unknown-channel`

This plugin still works exactly the same way, it just needs a minor code fix, which is the purpose of this pull request. You can still access the underlying mention user id, by copying the message text. You're getting upset about a minor improvement of user experience

gonsodany:
> what even is this rant. discord changed absolutely nothing, other than changing the display of mentions of uncached users from `<@1234567890>` -> `@unknown-user`
> 
> For the average Discord user who doesn't know what a user id is or how to use them, this is objectively more user friendly. Not to mention that this makes it more consistent with role or channel mentions, which have always rendered as `@deleted-role` and `#unknown-channel`
> 
> This plugin still works exactly the same way, it just needs a minor code fix, which is the purpose of this pull request. You can still access the underlying mention user id, by copying the message text. You're getting upset about a minor improvement of user experience

the thing is, that with this plugin (validUser) when you hovered by the "<@1234567890>" it would change to the user profile, so essentially by just having the userid you could open anyone's profile, cached or not, and now it doesnt work, like i explained before, tested with someone that had 9 servers in common, and just because i never opened their profile it now says "unknown-user", i run a community server and used "validUser" on a daily basis to check on users reported by community mods or other users, basically now the userid is essentially useless for any power-user

Vendicated:
wrong.

Vendicated:
thanks elvy!

PR #2370, opened 2024-04-19
Add HideServers

bepvte:
Small plugin that adds a context menu option to hide specific servers from the guilds list and quick switcher

### DIFF
--- a/src/plugins/_api/serverList.ts
+++ b/src/plugins/_api/serverList.ts
@@ -33,10 +33,15 @@ export default definePlugin({
         },
         {
             find: "Messages.SERVERS,children",
-            replacement: {
+            replacement: [{
                 match: /(?<=Messages\.SERVERS,children:)\i\.map\(\i\)/,
                 replace: "Vencord.Api.ServerList.renderAll(Vencord.Api.ServerList.ServerListRenderPosition.In).concat($&)"
+            },
+            {
+                match: /guildDiscoveryRef.{0,300}\{\}\)\]\}\)\]/,
+                replace: "$&.concat(Vencord.Api.ServerList.renderAll(Vencord.Api.ServerList.ServerListRenderPosition.Below))"
             }
+            ]
         }
     ]
 });
--- a/src/plugins/hideServers/components/HiddenServersButton.tsx
+++ b/src/plugins/hideServers/components/HiddenServersButton.tsx
@@ -1,0 +1,28 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import "./style.css";
+
+import { classNameFactory } from "@api/Styles";
+import { Button, ButtonLooks } from "@webpack/common";
+
+const cl = classNameFactory("vc-hideservers-");
+
+export default function ({ count, onClick, }: { count: number; onClick: () => void; }) {
+    return (
+        <div className={cl("button-wrapper")}>
+            {count > 0 ? (
+                <Button
+                    className={cl("button")}
+                    look={ButtonLooks.BLANK}
+                    size={Button.Sizes.MIN}
+                    onClick={onClick} >
+                    {count} Hidden
+                </Button>
+            ) : null}
+        </div>
+    );
+}
--- a/src/plugins/hideServers/components/modal.tsx
+++ b/src/plugins/hideServers/components/modal.tsx
@@ -1,0 +1,112 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { classNameFactory } from "@api/Styles";
+import { classes } from "@utils/misc";
+import {
+    closeModal,
+    ModalCloseButton,
+    ModalContent,
+    ModalFooter,
+    ModalHeader,
+    ModalProps,
+    ModalRoot,
+    ModalSize,
+    openModal,
+} from "@utils/modal";
+import { findByPropsLazy } from "@webpack";
+import { Button, Forms, IconUtils, Text, useState } from "@webpack/common";
+import { Guild } from "discord-types/general";
+
+import { removeHidden } from "../settings";
+
+const cl = classNameFactory("vc-hideservers-");
+const IconClasses = findByPropsLazy("icon", "acronym", "childWrapper");
+
+function HiddenServersModal({
+    servers,
+    modalProps,
+    close,
+}: {
+    servers: Guild[];
+    modalProps: ModalProps;
+    close(): void;
+}) {
+    const [serverList, setServerList] = useState(servers);
+    function removeServer(id: string) {
+        removeHidden(id);
+        setServerList(serverList.filter(x => x.id !== id));
+    }
+
+    return (
+        <ModalRoot {...modalProps} size={ModalSize.LARGE}>
+            <ModalHeader>
+                <Text variant="heading-lg/semibold" style={{ flexGrow: 1 }}>
+                    Hidden Servers
+                </Text>
+                <ModalCloseButton onClick={close} />
+            </ModalHeader>
+
+            <ModalContent>
+                <div className={cl("list")}>
+                    {serverList.length > 0 ? (
+                        serverList.map(server => (
+                            <div key={server.id} className={cl("row")}>
+                                <div className={cl("guildicon")}>
+                                    {server.icon
+                                        ? <img height="48" width="48"
+                                            src={IconUtils.getGuildIconURL({
+                                                id: server.id,
+                                                icon: server.icon,
+                                                canAnimate: true,
+                                                size: 240,
+                                            })} />
+                                        : <div
+                                            aria-hidden
+                                            className={classes(
+                                                IconClasses.childWrapper,
+                                                IconClasses.acronym
+                                            )}>
+                                            {server.acronym}
+                                        </div>
+                                    }
+                                </div>
+                                <Forms.FormTitle className={cl("name")}>
+                                    {server.name}
+                                </Forms.FormTitle>
+                                <Button
+                                    className={"row-button"}
+                                    color={Button.Colors.PRIMARY}
+                                    onClick={() => removeServer(server.id)}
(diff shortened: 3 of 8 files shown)


src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+            action={() => addHidden(guild)}
+        />
+    );
+};
+
+export default definePlugin({
+    name: "HideServers",
+    description:
+        "Allows you to hide servers from the guild list and quick switcher by right clicking them",
+    authors: [
```
MrDiamondDog, line 54: Add yourself to Devs

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+            id: 0n,
+        },
+    ],
+    tags: ["guild", "server", "hide"],
+    contextMenus: {
+        "guild-context": Patch,
+        "guild-header-popout": Patch,
+    },
+    patches: [
+        // i tried a few different things like filtering the getGuildsTree area but all of them end up resetting scroll position sometimes
```
MrDiamondDog, line 66: I would make this comment say what the patch does rather than this

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+        {
+            find: '("guildsnav")',
```
MrDiamondDog, line 68: If you have 2 patches with the same find then you can make `replacement` an array and add both in one object

MrDiamondDog (commented):
As far as I can tell, there's no way to un-hide individual servers. Either this should be implemented or I am blind

bepvte:
I don't really know how or where the UI for removing individual servers should be. I could cram it all into a settings component, but no other plugin seems to do that.

MrDiamondDog:
> I don't really know where the UI for removing individual servers should be. I could cram it all into a settings component, but no other plugin seems to do that.

I would have a button at the bottom of the server list that opens a modal to all the servers you've hidden

Golonchy:
how do i still view hidden servers?

Tr0llie:
> how do i still view hidden servers?

as mentioned maybe there could be a button in which lets you view all the server that you have hidden but there could be Ui issue if you try to view it from devices that does not have vencord installed.
there could be errors such as the servers being messy etc
this could be fixed by putting the servers in a separated folders when a user hides a channel 
and the button will also let you view all the server that u've hidden but just in a better ui and not the normal folder one smth

Golonchy:
a ye,
thanks bro

Sqaaakoi:
> > I don't really know where the UI for removing individual servers should be. I could cram it all into a settings component, but no other plugin seems to do that.
> 
> I would have a button at the bottom of the server list that opens a modal to all the servers you've hidden

I implemented a horrible hack to do something similar in pure CSS and using folder names starting with . (before they did some class rerolls and it broke and I never updated it)
I just showed those folders when the guild discovery page was selected.

...although this could be a great use case for a concept I had for a guilds list page where it would show detailed cards about all your guilds you are in

bepvte:
image
image

bepvte:
@Vendicated I was wondering if this could be merged anytime soon... It looks like it slipped through the cracks

src/plugins/_api/serverList.ts
```diff
@@ -33,10 +33,15 @@ export default definePlugin({
         },
         {
             find: "Messages.SERVERS,children",
-            replacement: {
+            replacement: [{
```
Sqaaakoi, line 36: start the object on a new line and fix indentation

src/plugins/hideServers/components/HiddenServersButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default function ({ count, onClick, }: { count: number; onClick: () => void; }) {
+    return (
+        <div className={cl("button-wrapper")}>
+            {count > 0 ? (
+                <Button
+                    className={cl("button")}
+                    look={ButtonLooks.BLANK}
+                    size={Button.Sizes.MIN}
+                    onClick={onClick} >
```
Sqaaakoi, line 22: put the closing > on a new line

src/plugins/hideServers/components/modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        ))
+                    ) : (
+                        <Text variant="heading-sm/medium">
+                            No hidden servers
+                        </Text>
+                    )}
+                </div>
+            </ModalContent>
+
+            <ModalFooter></ModalFooter>
```
Sqaaakoi, line 97: Remove the footer if you aren't going to use it

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+type qsResult = {
+    type: "GUILD" | string;
+    record?: {
+        id?: string;
+        guild_id?: string;
+    };
+};
+
+let rerenderRef: undefined | WeakRef<Function>;
+export function rerender() {
```
Sqaaakoi, line 42: I have no idea what is going on here but I can tell that there is probably a better way of doing whatever you are trying to achieve

src/plugins/hideServers/settings.tsx
```diff
@@ -0,0 +1,82 @@
...
+        type: OptionType.BOOLEAN,
+        description: "Show menu to unhide servers at the bottom of the list",
+        default: true,
+        onChange: val => {
+            if (val) {
+                addIndicator();
+            } else {
+                removeIndicator();
+            }
+        }
```
Sqaaakoi, line 65: No manual open manager option within the settings?

src/plugins/hideServers/settings.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+import { addIndicator, removeIndicator, rerender } from ".";
+
+const SortedGuildStore = findStoreLazy("SortedGuildStore");
+
+const DB_KEY = "HideServers_servers";
+
+export let hiddenGuilds: Set<string> = new Set();
+
+export function addHidden(guild: Guild) {
```
Sqaaakoi, line 22: Ah. This is why you needed that. You should convert this into a store (take a look at the SpotifyStore added by SpotifyControls for a good example on doing this) then use the useStateFromStore hook inside of a component to rerender whenever changes are emitted from the store.

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+    addServerListElement(ServerListRenderPosition.Below, serverListFunction);
+}
+
+export function removeIndicator() {
+    removeServerListElement(ServerListRenderPosition.Below, serverListFunction);
+}
+
+export default definePlugin({
+    name: "HideServers",
+    description:
```
Sqaaakoi, line 84: fix this to be on the same line

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+            ],
+        },
+        {
+            find: "QUICKSWITCHER_PROTIP.format",
+            replacement: {
+                match: /(?<=renderResults\(\){)let{query/,
+                replace:
+                    "this.props.results = $self.filterResults(this.props.results);$&",
+            },
+        },
```
Sqaaakoi, line 116: Do all of these patches include the ctrl+alt+up/down navigation? That would be my main use case for using a plugin like this, so if it doesnt, please consider it

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+    );
+};
+
+function serverListFunction() {
+    // if youve left a server dont show it in the count
+    const actuallyHidden = Array.from(hiddenGuilds).filter(x => GuildStore.getGuild(x)).length;
+    return <HiddenServersButton
+        count={actuallyHidden}
+        onClick={() => openHiddenServersModal({ servers: hiddenGuildsDetail() })}
+    ></HiddenServersButton>;
```
Sqaaakoi, line 70: use self closing tag

src/plugins/hideServers/components/style.css
```diff
@@ -0,0 +1,47 @@
...
+    font-size: 60%;
+    background-color: var(--background-primary);
+    color: var(--header-secondary);
+}
+
+.vc-hideservers-list {
+    flex: 1 1 auto;
+}
+
+/* copied from blocked row */
```
Sqaaakoi, line 18: Why not just import the original class? Do note that the class names being lazy loaded is a valid excuse, I hate doing lazy loading for modals

src/plugins/hideServers/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+        {
+            find: '("guildsnav")',
+            replacement: [
+                {
+                    match: /(?<=Messages\.SERVERS,children:)(.+?\i)(\.map\(\i\))/,
+                    replace: "$1.flatMap($self.filterGuilds)$2",
+                },
+                {
+                    match: /let{disableAppDownload.{0,10}isPlatformEmbedded/,
+                    replace:
```
Sqaaakoi, line 104: is your IDE doing this? (this should be rewritten anyway for reasons described in other places)

Sqaaakoi (commented):
The UI rerendering doesn't seem to be taking advantage of React properly;
Fixing that would involve converting most of settings.tsx into a store.

Your plugin should also patch over up/down navigation keyboard shortcuts (do note I didn't actually test if that is the case but it doesn't look like it)

src/plugins/hideServers/index.tsx
bepvte: Any other problems with this particular patch besides the formatting? Just trying to make sure I get what you are saying here

Vendicated:
sorry but this is not useful. you can just leave the servers or shove them in some folder (and maybe hide the folder with css if you really want)

definitely doesn't need a plugin

thanks anyway for your contribution

src/plugins/hideServers/index.tsx
bepvte: Oh I get what you mean, I'll convert it to be a store

bepvte:
I think I remember seeing a request for it, and I find its really useful, rejoining servers that you need a break from is kind of a pain and annoying and visible. CSS cant really hide stuff from the quick switcher and my impulses.. Either way, I appreciate you all taking a look.

PR #2375, opened 2024-04-21
feat(ShowTimeouts): Also show invites disabled tooltip

dolfies:
## Summary

- Renames the plugin to be more generic
- Also shows invites disabled tooltip now
- Adds settings for everything

### DIFF
--- a/src/plugins/showHiddenThings/README.md
+++ b/src/plugins/showHiddenThings/README.md
@@ -5,7 +5,7 @@ Displays various moderator-only elements regardless of permissions.
 ## Features
 
 - Show member timeout icons in chat
-![](https://camo.githubusercontent.com/65fc9ee6175e63daaf9acc0ba2c01cff1cf419847569b49197537a85646a49d7/68747470733a2f2f692e646f6c66692e65732f3543506a446b576772742e706e673f6b65793d31754e786368385a3359364c4955)
+![](https://github.com/Vendicated/Vencord/assets/47677887/75e1f6ba-8921-4188-9c2d-c9c3f9d07101)
 
 - Show the invites paused tooltip in the server list
-![](https://camo.githubusercontent.com/7d8983d531c63df54095cd24c19dfb99e00088922fd2ec703f9cf7781204258d/68747470733a2f2f692e646f6c66692e65732f36685472756e4d4867512e706e673f6b65793d344274534d524f46726b77476357)
+![](https://github.com/Vendicated/Vencord/assets/47677887/b6a923d2-ac55-40d9-b4f8-fa6fc117148b)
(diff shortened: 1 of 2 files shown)


Kodarru:
very useful

Sqaaakoi:
Is "ShowHiddenThings" descriptive enough? There's a lot of plugins that "show hidden things" in Vencord, and this could quickly get confusing

Also, I had previously worked on a concept extending this plugin (show a detailed timer on hovering over the timeout icon) that would be more appropriate to keep the name as ShowTimeouts. Unfortunately I gave up on it (due to react not wanting to remove stuff after the channel was unloaded). Would such an idea be best reborn as its own plugin?

Vendicated:
image

maybe not "Actions" but thats already a lot better

dolfies:
> Is "ShowHiddenThings" descriptive enough? There's a lot of plugins that "show hidden things" in Vencord, and this could quickly get confusing

Pretty sure only other similar one is ShowHiddenChannels. I guess they could *maybe* be merged but this seems a little different. Something like ShowModActions could work, though maybe this extends beyond just mod stuff at some point?

> Also, I had previously worked on a concept extending this plugin (show a detailed timer on hovering over the timeout icon) that would be more appropriate to keep the name as ShowTimeouts. Unfortunately I gave up on it (due to react not wanting to remove stuff after the channel was unloaded). Would such an idea be best reborn as its own plugin?

I think that would probably be it's own plugin, something like BetterTimeouts.

Sqaaakoi:
oh also, i forgot to mention
you should migrate the old settings (literally just the built-in enabled boolean) to the new settings for convenience on upgrading

src/plugins/showHiddenThings/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Displays various moderator-only elements regardless of permissions.
+
+## Features
+
+- Show member timeout icons in chat
+![](https://camo.githubusercontent.com/65fc9ee6175e63daaf9acc0ba2c01cff1cf419847569b49197537a85646a49d7/68747470733a2f2f692e646f6c66692e65732f3543506a446b576772742e706e673f6b65793d31754e786368385a3359364c4955)
+
+- Show the invites paused tooltip in the server list
+![](https://camo.githubusercontent.com/7d8983d531c63df54095cd24c19dfb99e00088922fd2ec703f9cf7781204258d/68747470733a2f2f692e646f6c66692e65732f36685472756e4d4867512e706e673f6b65793d344274534d524f46726b77476357)
```
Vendicated, line 11: goofy ahh camo (idek what that means) link are you sure this won't randomly expire?

when in doubt you can upload to our issue thread #1749

dolfies: yeah idk either, fixed :p

Vendicated:
tbh still not happy with the name but i cant come up with anything better

Vendicated:
<:bleh:1047549509745524777>

PR #2378, opened 2024-04-23
feat(plugins/quickSearch) quickly search stuff from context menus

CatGirlDShadow:
Adds context menus to channels, users, messages to quickly search stuff.
1. Requires review cuz probably bad
3. https://github.com/Vencord/plugin-requests/issues/390
image
image


src/plugins/quickSearch/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+                            include_nsfw: true,
+                        };
+                        ELEM_INFO.forEach(element => {
+                            if (queryObject[element.name]) {
+                                if (element.queryName) {
+                                    query[element.queryName] = element.value;
+                                }
+                            }
+                        });
+                        const searchElem = document.getElementsByClassName("DraftEditor-editorContainer");
```
Vendicated, line 145: dont do this

you can likely use Flux events

CatGirlDShadow: Yes, but it was difficult for me to understand even what search state dispatch needed. Now fixed

PR #2382, opened 2024-04-24
ReviewDB: add more context menu shortcuts to view reviews

Sqaaakoi:
Small patch to add the "View Reviews" menu option to a few places

User profile modal ... menu
User profile modal ... menu screenshot

Guild context menu
Guild context menu screenshot

User context menu
User context menu screenshot

I made this mostly so I can click on a reviewers profile then see their reviews


src/plugins/reviewDB/index.tsx
```diff
@@ -47,14 +47,29 @@ const guildPopoutPatch: NavContextMenuPatchCallback = (children, props: { guild:
     );
 };
 
+const userContextPatch: NavContextMenuPatchCallback = (children, props: { user?: User, onClose(): void; }) => {
```
Vendicated, line 49:
```suggestion
const userContextPatch: NavContextMenuPatchCallback = (children, { user }: { user?: User, onClose(): void; }) => {
```

then you dont need the non null assertions

Sqaaakoi: i don't understand how destructuring will fix that? if you look at the commit history you will see why the check was even implemented in the first place

Sqaaakoi: image

Sqaaakoi: ImageZoom also has a similar issue

Vendicated: i resolved it myself

Vendicated:
very useful, thanks a lot!

PR #2383, opened 2024-04-24
Plugin Settings: fix text overflow for long plugin names

HAHALOSAH:
Fix text overflow for long titles in the plugin list


zt64 (approved):

src/components/VencordSettings/AddonCard.tsx
```diff
@@ -42,6 +42,8 @@ interface Props {
 }
 
 export function AddonCard({ disabled, isNew, name, infoButton, footer, author, enabled, setEnabled, description, onMouseEnter, onMouseLeave }: Props) {
+    var title = useRef(null);
+    var titleContainer = useRef(null);
```
Vendicated, line 46:
```suggestion
    const titleContainer = useRef<HTMLDivElement>(null);
```

Vendicated:
thanks :3

PR #2388, opened 2024-04-25
feat(USRBG): update to new API

katlyn:
USRBG has moved storage to a new dedicated platform, allowing better integration and more control over the data that can be made available for plugins to use. This has allowed creation of a new API that has several benefits:
- The API's data is updated immediately upon a user's background being approved
- Data sent over-the-pipe can be reduced, only sending a list of user ids and a small bit of metadata (~800kB instead of the current 1.7MB)
- Dead links are no longer a possibility and do not need to be filtered out of the dataset

### DIFF
--- a/src/plugins/usrbg/index.tsx
+++ b/src/plugins/usrbg/index.tsx
@@ -24,9 +24,15 @@ import definePlugin, { OptionType } from "@utils/types";
 
 import style from "./index.css?managed";
 
-const BASE_URL = "https://raw.githubusercontent.com/AutumnVN/usrbg/main/usrbg.json";
+const API_URL = "https://usrbg.is-hardly.online/users";
+const cachebust = Date.now();
 
-let data = {} as Record<string, string>;
+interface UsrbgApiReturn {
+    endpoint: string
+    bucket: string
+    prefix: string
+    users: string[]
+}
 
 const settings = definePluginSettings({
     nitroFirst: {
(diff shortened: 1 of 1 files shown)


src/plugins/usrbg/index.tsx
```diff
@@ -104,24 +108,42 @@ export default definePlugin({
...
     },
 
     shouldShowBadge({ displayProfile, user }: any) {
-        return displayProfile?.banner && (!data[user.id] || settings.store.nitroFirst);
+        return displayProfile?.banner && (!this.userHasBackground(user.id) || settings.store.nitroFirst);
+    },
+
+    userHasBackground(userId: string) {
+        if (this.data === null) return false;
+        return this.data.users[userId] !== undefined;
```
Vendicated, line 124: could just change this entire func to
```suggestion
        return !!this.data?.users[userId];
```

src/plugins/usrbg/index.tsx
```diff
@@ -104,24 +108,42 @@ export default definePlugin({
...
+
+    getImageUrl(userId: string): string|null {
+        if (this.data == null) {
+            return null;
+        }
+
+        const etag = this.data.users[userId];
+        if (etag === undefined) {
+            return null;
+        }
```
Vendicated, line 135: why not replace the above checks with
```ts
if (!this.userHasBackground(userId)) return null;
```

Vendicated:
looks good, thank you!!

PR #2390, opened 2024-04-25
MsgClickActions: control ping via shift & NoReplyMention plugin

sunnniee:
Vencord/plugin-requests#409

### DIFF
--- a/src/plugins/messageClickActions/index.ts
+++ b/src/plugins/messageClickActions/index.ts
@@ -84,7 +84,7 @@ export default definePlugin({
                     const EPHEMERAL = 64;
                     if (msg.hasFlag(EPHEMERAL)) return;
 
-                    const NoReplyMention = Vencord.Plugins.plugins.NoReplyMention as import("../noReplyMention");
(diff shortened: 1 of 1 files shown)


src/plugins/messageClickActions/index.ts
```diff
@@ -85,11 +85,14 @@ export default definePlugin({
                     const EPHEMERAL = 64;
                     if (msg.hasFlag(EPHEMERAL)) return;
 
+                    let shouldMention = !Settings.plugins.NoReplyMention.enabled;
```
Vendicated, line 88: this old logic is actually outdated because the plugin now has whitelist/blacklist. 

try adapting the logic from https://github.com/Vendicated/Vencord/blob/main/src/plugins/quickReply/index.ts#L162-L171

veygax (approved):

veygax:
misclick

Vendicated:
Looks like you got the logic messed up a little and forgot to check if NoReplyMention is even enabled

Thanks anyway, fixed it up!

PR #2391, opened 2024-04-25
Plugin: betterThemes Revamped Theme Tab

soulfireneedusername:
Easily add buttons to switch themes to personalize your Discord interface. You can also create these buttons yourself and even add images to them. You can scale, edit them and add your own CSS rules for personalization. This way, all your favourite themes will be saved in the themes tab and it is incredibly easy to switch between them. You can also use the Themes Diashow function and set it to a custom interval to automatically change between themes. If you want to add a theme or browse online themes quickly, you can go to the online themes tab to browse through a preselection of the most popular themes of BetterDiscord. This way, you won’t need to add a picture or copy any links.


Sqaaakoi:
# This will be denied for many reasons

Devs are **always** stored in the Devs constant
Vencord uses React and JSX
You should be editing the existing themes code
Do not hardcode randomly generated CSS class names
**Do not blindly port BD code to Vencord. It looks like you copied code that DevilBro made**

<sub>I genuinely can't tell if this is worse than the time someone submitted AI generated code. FakeNitro is commonly said to be unreadable, this is literally 10 times worse</sub>

Also, please learn to use Git locally.

Inbestigator:
Not to mention the 6,635 lines of code that this is

soulfireneedusername:
I knew that the readability would be a problem. But i just don't know how a code with those functionality's could work otherwise. I worked too many hours on this so unless you say it's not worth to try finishing this, please tell me. 

1. I can surely do that (change how information gets store)
2. I can rename the names (which I named because of complexity, but I understand) 
3. I don't know what you mean by port BD code. I am just importing the theme links. I found that a very useful feature but I can remove this feature if it's against some rules
4. I'm not completely sure what you mean by the existing theme code.

I actually optimized it from 9k lines. And is it possible to make this so much shorter? And yes, it's my first time ever to use typescript. But i would really like to make this code acceptable unless you say this will not work.

Sqaaakoi:
> I knew that the readability would be a problem. But i just don't know how a code with those functionality's could work otherwise. I worked too many hours on this so unless you say it's not worth to try finishing this, please tell me.
> 
> 1. I can surely do that (change how information gets store)
We already have IndexedDB support through `DataStore`
> 2. I can rename the names (which I named because of complexity, but I understand)
You should be removing the complexity as it isn't needed, see below about using CSS classes
> 3. I don't know what you mean by port BD code. I am just importing the theme links. I found that a very useful feature but I can remove this feature if it's against some rules
this is a joke about how DevilBro writes awful BD plugins, again, convert your code to TSX and use CSS classes
> 4. I'm not completely sure what you mean by the existing theme code.
Vencord's built in theme support.  I would highly recommend looking at `src/components/VencordSettings/ThemesTab.tsx` as it uses TSX and demonstrates use of React.
For loading themes, take a look at `src/utils/quickCss.ts` (might be a bad example?)
> 
> I actually optimized it from 9k lines. And is it possible to make this so much shorter? And yes, it's my first time ever to use typescript. But i would really like to make this code acceptable unless you say this will not work.
Import CSS files inside of TSX code and use class names to refer to the CSS instead of writing it inline


tl;dr: the major issue is how much inline CSS there is inside of your script

soulfireneedusername:
Ok thank you. I will do that. But I still think it will not go under 1.5-2k.

Vendicated:
hi! thank you for your efforts but this is not even remotely mergeable for various reasons:
- way too large. 6k lines of code is just too much. there's no reason a plugin like this should need that much code. and even if it did, you should split your code across multiple files to make it more maintainable
- vencord uses react & patches. raw dom manipulation is to be avoided
- we have many apis for doing things like adding css (literally just import a css file) and using the indexeddb. this plugin makes no use of any of the provided apis
- i still don't really understand the purpose of this plugin. your description is very confusing. based on your plugin request, i think you meant to port ThemeRepo? if so:
  - someone else has already worked on this in the past
  - this should be natively included in the theme settings component code instead of being a plugin that patches vencord itself

as such i will be closing this, sorry

soulfireneedusername:
> hi! thank you for your efforts but this is not even remotely mergeable for various reasons:
> 
> - way too large. 6k lines of code is just too much. there's no reason a plugin like this should need that much code. and even if it did, you should split your code across multiple files to make it more maintainable
> 
> - vencord uses react & patches. raw dom manipulation is to be avoided
> 
> - we have many apis for doing things like adding css (literally just import a css file) and using the indexeddb. this plugin makes no use of any of the provided apis
> 
> - i still don't really understand the purpose of this plugin. your description is very confusing. based on your plugin request, i think you meant to port ThemeRepo? if so:
> 
>   - someone else has already worked on this in the past
> 
>   - this should be natively included in the theme settings component code instead of being a plugin that patches vencord itself
> 
> 
> 
> as such i will be closing this, sorry

So, should we complete this and upload it for native inclusion in Vencord? Or do you say that it will not be worth the work or the functionality's are not likely to be accepted?

PR #2393, opened 2024-04-27
feat(plugin): Quests completer

happyendermangit:
This plugin completes quests without having the game. **this is useful for quests that rewards you avatar decorations**

The "Complete Quest" button is enabled only if:
- You have uncompleted quests
- You are streaming something
- Someone is watching your stream (your alt or your friend.)

For Web, this requires to install an [extension](https://github.com/SerStars/QuestCompleterExtension/tree/plugin) (only for chrome)

## Screenshots:
image
image
image

### DIFF
--- a/src/plugins/QuestCompleter/index.tsx
+++ b/src/plugins/QuestCompleter/index.tsx
@@ -43,6 +43,7 @@ let interval;
 let quest;
 let questHeroBarUrl;
 let ImagesConfig = {};
+const streamingUtils = findByProps("getCurrentUserActiveStream")
 
 export default definePlugin({
     name: "QuestCompleter",
@@ -81,7 +82,7 @@ export default definePlugin({
         window.currentUserId = currentUserId; // this is here because discord will lag if we get the current user id every time
     },
     renderQuestButton() {
-        const currentStream: Stream | null = findByProps("getCurrentUserActiveStream").getCurrentUserActiveStream();
+        const currentStream: Stream | null = streamingUtils.getCurrentUserActiveStream();
         let shouldDisable = !!interval;
         const { Divider } = findByProps("Divider", "Icon");
 
@@ -137,7 +138,7 @@ export default definePlugin({
     openCompleteQuestUI() {
         // check if user is sharing screen and there is someone that is watching the stream 
 
-        const currentStream: Stream | null = findByProps("getCurrentUserActiveStream").getCurrentUserActiveStream();
+        const currentStream: Stream | null = streamingUtils.getCurrentUserActiveStream();
         const encodedStreamKey = findByProps("encodeStreamKey").encodeStreamKey(currentStream);
         quest = getLeftQuests();
         ImagesConfig = {
--- a/src/plugins/QuestCompleter/index.tsx
+++ b/src/plugins/QuestCompleter/index.tsx
@@ -83,6 +83,7 @@ export default definePlugin({
     renderQuestButton() {
         const currentStream: Stream | null = findByProps("getCurrentUserActiveStream").getCurrentUserActiveStream();
         let shouldDisable = !!interval;
+        const { Divider } = findByProps("Divider", "Icon");
 
         if (!currentStream) {
             shouldDisable = true;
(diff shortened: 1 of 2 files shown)


src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,201 @@
+/*
```
MrDiamondDog, line 1: Use the new header

src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+export default definePlugin({
+    name: "QuestCompleter",
+    description: "A plugin to complete quests without having the game.",
+    authors: [Devs.HAPPY_ENDERMAN, Devs.SerStars],
+    patches: [
+        {
+            find: `"invite-button"`,
+            replacement: {
+                match: /(function .+?\(.+?\){let{inPopout:.+allowIdle.+?}=.+?\.usePreventIdle\)\("popup"\),(.+?)=\[\];if\(.+?\){.+"chat-spacer"\)\)\),\(\d,.+?\.jsx\)\(.+?,{children:).+?}}/,
```
MrDiamondDog, line 55: Use \i for variable names

src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    <Text variant="text-lg/bold">
+                        This plugin won't work right now. Please download
+                        <a href="" target="_blank">our extension</a>
+                        to make the plugin work on web.
+                    </Text>
+            }
+        </>);
+    },
+    start() {
+        const currentUserId: string = findByProps("getCurrentUser").getCurrentUser().id;
```
MrDiamondDog, line 80:
```suggestion
        const currentUserId: string = UserStore.getCurrentUser().id;
```

src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+        window.currentUserId = currentUserId; // this is here because discord will lag if we get the current user id every time
```
MrDiamondDog, line 81: Don't pollute the `window` object like this. If you do what the above comment says then it shouldn't lag

src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+    },
+    renderQuestButton() {
+        const currentStream: Stream | null = findByProps("getCurrentUserActiveStream").getCurrentUserActiveStream();
```
MrDiamondDog, line 84: `findByProps` calls should be put into a variable instead of calling it every time you need something from it. This would call every time the quest button is rendered and that would be very laggy

MrDiamondDog (changes requested):
Afaik you don't need someone else watching for it to work as the js snippet works if no one else is there

verticalsync:
> Afaik you don't need someone else watching for it to work as the js snippet works if no one else is there

As far from what I've heard and experienced myself, you now DO need to have someone watching.

SerStars:
> Afaik you don't need someone else watching for it to work as the js snippet works if no one else is there

It doesn't work since yesterday or 2 days ago, they patched so it only works in the desktop app.
By checking the user agent if it has the "Electron/" string and they also started checking how many people are in the VC, I'm pretty they will add "how many people are watching your stream" soon.

src/plugins/QuestCompleter/index.tsx
happyendermangit: Yea, i get your idea, because of react re rendering.

happyendermangit: Lemme make it a variable

src/plugins/QuestCompleter/index.tsx
happyendermangit: \i on my case, didnt work.

src/plugins/QuestCompleter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+                shouldDisable = true;
+            }
+        }
+        if (!getLeftQuests()) {
+            shouldDisable = true;
+        }
+
+
+
+        const ToolTipButton = findByProps("CenterControlButton").default;
```
MrDiamondDog, line 106: You have many more instances of constant `findByProps` calls you should probably fix also

Vendicated:
i don't think having this is a good idea

discord is very clearly cracking down on quest scripts, aamia had to change her script multiple times. so this plugin would possibly require frequent changes and unnecessarily upset discord

this is also more of a one time use thing than something you'd actually need as a plugin. just using the snippet that's floating around makes more sense

PR #2396, opened 2024-04-29
fix(plugins/vcDoubleClick): conflict with MessageClickActions

arHSM:
After enabling both `VoiceChatDoucleClick` and `MessageClickActions`, double clicking a voice channel mention in a message would both edit/reply to the message as well as join the voice channel, adding a `evt.stopPropagation()` fixes this behavior

### DIFF
--- a/src/plugins/vcDoubleClick/index.ts
+++ b/src/plugins/vcDoubleClick/index.ts
@@ -52,7 +52,7 @@ export default definePlugin({
             replacement: {
                 match: /onClick:(\i)(?=,.{0,30}className:"channelMention".+?(\i)\.inContent)/,
                 replace: (_, onClick, props) => ""
-                    + `onClick:(vcDoubleClickEvt)=>$self.shouldRunOnClick(vcDoubleClickEvt,${props})&&${onClick}()`,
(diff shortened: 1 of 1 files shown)


src/plugins/vcDoubleClick/index.ts
```diff
@@ -52,7 +52,7 @@ export default definePlugin({
             replacement: {
                 match: /onClick:(\i)(?=,.{0,30}className:"channelMention".+?(\i)\.inContent)/,
                 replace: (_, onClick, props) => ""
-                    + `onClick:(vcDoubleClickEvt)=>$self.shouldRunOnClick(vcDoubleClickEvt,${props})&&${onClick}()`,
+                    + `onClick:(vcDoubleClickEvt)=>$self.shouldRunOnClick(vcDoubleClickEvt,${props})(vcDoubleClickEvt.stopPropagation(),${onClick}())`,
```
Vendicated, line 55: this looks wrong, shouldn't it be this?
```suggestion
                    + `onClick:(vcDoubleClickEvt)=>$self.shouldRunOnClick(vcDoubleClickEvt,${props})&&(vcDoubleClickEvt.stopPropagation(),${onClick}())`,
```

Vendicated: also it seems the `find` for this patch is no longer unique and needs updating

PR #2397, opened 2024-04-29
Cloud Sync: Stop discarding unknown root level values

Sqaaakoi:
This will merge local root object values over existing remote values that may not exist in the client.

For example with a change I am making to the core right now where I have added a `cssSnippets` object to the root, this is so that clients with this patch will not remove that from the cloud data if they don't support it.
This is also useful for people running their own fork that adds more to cloud sync like I have, so they don't accidentally delete that data when using the upstream build (for example, when setting up a new device, using the browser extension, or even when testing if something is broken by their own dev build.)


Vendicated:
thanks for the pr!

this seems like something that should be done on the server instead to avoid the unnecessary request, so I suggest opening an issue [there](https://github.com/Vencord/Vencloud] instead to discuss making this change

PR #2399, opened 2024-04-30
feat(plugin): Character count

fres621:
Add Character Count plugin
Always displays the character count on the bottom right of the chat input, followed by the user's max character count
image


fres621:
i have made some changes because otherwise it collides with the slowmode text
image
^ currently like that however i do plan to improve it and make it pretty i just cant do it for a day or so i don't know if i can put the PR on hold or whatever for a while

AZthemute:
Found a bug: it shows the wrong character count in the About Me section.
image

PR #2400, opened 2024-04-30
LastfmRichPresence: Add option to use album name as status name

KK2-5:
Defaults to custom status when album name is not set.

image


Vendicated:
thank you!

PR #2401, opened 2024-04-30
fix(fakeNitro): Use getEmojiURL as emoji.url was removed

HAHALOSAH:
It appears that `emoji.url` no longer exists, breaking fakeNitro. This should fix it.

### DIFF
--- a/src/plugins/fakeNitro/index.tsx
+++ b/src/plugins/fakeNitro/index.tsx
@@ -933,7 +933,7 @@ export default definePlugin({
 
                 hasBypass = true;
 
-                const url = new URL(emoji.url);
+                const url = new URL(IconUtils.getEmojiURL({ id: emoji.id, animated: emoji.animated, size: s.emojiSize }));
                 url.searchParams.set("size", s.emojiSize.toString());
                 url.searchParams.set("name", emoji.name);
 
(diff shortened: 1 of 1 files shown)


vstudiocode:
+1

Cynosphere (approved):

src/plugins/fakeNitro/index.tsx
```diff
@@ -900,7 +900,7 @@ export default definePlugin({
 
                     const emojiString = `<${emoji.animated ? "a" : ""}:${emoji.originalName || emoji.name}:${emoji.id}>`;
 
-                    const url = new URL(emoji.url);
+                    const url = new URL(`https://cdn.discordapp.com/emojis/${emoji.id}.${emoji.animated ? "gif" : "webp"}?quality=lossless`);
```
Nuckyz, line 903:
```suggestion
                    const url = new URL(IconUtils.getEmojiURL({ id: emoji.id, animated: emoji.animated, size: s.emojiSize }));
```

Nuckyz: You also need to edit in the bottom, in the edit listener

Nuckyz (changes requested):

Nuckyz (approved):

PR #2406, opened 2024-05-02
fix(roleColorEverywhere): fixed patch

minemalox:
fixed the patch for voice users to work with the current stable version of discord


Nuckyz:
Fixed in dev already, sorry! https://github.com/Vendicated/Vencord/commit/7af733c7c81e2b5d40fe0ec12c2525336e113716

PR #2409, opened 2024-05-03
refactor(Webpack): Immediate finds using waitFor and complete rewrite of module patching

Nuckyz:
A complete refactor of the Vencord Webpack API. This rewrites how Webpack finds and module patches are done.

## Changes to Webpack finding:

Completely deprecates the `lazy` way that we currently do Webpack finds. Instead, now the default will be a unified way which is a combination of the plain `cacheFind` for already loaded modules, and usage of `waitFor` for when modules were not yet loaded.

The way this works is:
 - If you use the find methods in the top level, you will leverage the performance gain of using the `waitFor` alternative. By using waitFor, everytime a new module is loaded, Vencord will check if that module matches any of the finds requested, and if it does, it sets the result of that find to the module. This avoids having to iterate through the module cache for every find.
 > By using the waitFor alternative, the result of calling the find methods is a Proxy, which has a value inside which points to the module. When the module is found that value is set, and now the Proxy acts as the module found.
 
 > Trying to access a Proxy value before the module value is set will cause an error, which shows the filter that failed. This allows users to post better logs for us.
 - However, if you don't use in the top level, you will mostly likely just find the module using cacheFind, which does need to iterate through the module cache. But of course, if the module is not in the cache it will still register a waitFor, to find the module in case it is loaded later.
 
 By having the unified way, we no longer need a `lazy` alternative of the methods, so the following changes to the API occurred:
- Old `find` -> `cacheFind`
- New `find` (wrapper around `waitFor` and `cacheFind`)
- `findLazy` -> `find` (wrapper)
- `findByProps`, `findByPropsLazy` -> `findByProps`
- `findByCode`, `findByCodeLazy` -> `findByCode`
- `findStore`, `findStoreLazy` -> `findStore`
- `findComponent`, `findComponentLazy` -> `findComponent`
- `findExportedComponent`, `findExportedComponentLazy` -> `findExportedComponent`
- `findComponentByCode`, `findComponentByCodeLazy` -> `findComponentByCode`
- `mapMangledModule`, `mapMangledModuleLazy` -> `mapMangledModule`

The following were added for additional filters and refactors purposes:
- `findProp` -> Useful as an alternative to a top level destructure. See it's JSDoc documentation for more information.
- `findComponentByFields` -> Used for finding a class component using its prototype fields
- `findByFactoryCode` -> Use for finding all the exports of a module using its factory code

The following were changed for more appropriate names:
- `findAll` -> `cacheFindAll`
- `findBulk` -> `cacheFindBulk`
- `findModuleId` -> `cacheFindModuleId`

The following were changed for less weird names:
- `proxyLazyWebpack` -> `webpackDependantLazy`
- `LazyComponentWebpack` -> `webpackDependantLazyComponent`

 `findModuleFactory` now uses the same tactic as the normal finds, and its old version using cache is now `cacheFindModuleFactory`
 
 > [!WARNING]
> Destructuring Webpack finds at top level is now deprecated. It used a very hacky tricky which turns out to not always have the expected result (pshhhh Firefox)
> mapMangledModule is the only exception for this
>  Consider switching to mapMangledModule or findProp

 Important additions:
 - mapMangledModule can now properly find and wrap components, simply use a component filter in one of its mappings

Additional:
- Lots of cleanups related to the main Webpack finding API and lazy stuff, and some plugins practices regarding modules
- Webpack commons and a lot of plugins finds were changed for more stability and clean code
- Mildly refactor to reporter Webpack find testing, taking advantage of the way the new finds work, so it no longer has to re-search everything
- As stated before, better error for modules not found
- Migrated all plugins to use the new settings API

> [!IMPORTANT]
> This change is fully backwards compatible for now. Old methods were kept and just point to the new APIs and top level destructuring is kept with a warning.
> The plan is to remove the deprecated methods/features in the future.

## Changes to Webpack module patching:

Centralizes Webpack patching around a single thing: Patching the module factories object of the WebpackRequire (wreq.m). This wraps the modules object with a proxy to intercept the addition of new modules, and allow us to decide what to do with them.

A new internal setting was added, eagerPatches, and it decides what to do with the newly added modules. With it enabled, we patch the factories as soon as they are set, and with it disabled we only patch them once they are accessed for the first time (default).

For correctly handling patching, because of multiple WebpackInstances, and to allow for patching factories only when they are accessed, our proxy defines getters and setters for each module factory added, instead of doing a simple Reflect.set on the module factories object.

Additionally, these were done:
- Patch timings testing! Reporter now says if a patch took a long time to complete (we don't want slow patches)
- A lot of clean up
- Harder, yet stable conditions for choosing what to do with the WebpackInstances found, and whether it is the main instance, used for Webpack finding
> I analyzed years old instances of webpack to see what was less likely to change and rely on.
- Complete typings and documentation of the WebpackRequire. All of the code now properly uses the correct typings.
- A more future proof and memory efficient patchedFactory wrapper. It includes a fallback to set the main WebpackInstance in case our other approach somehow fails, and also keeps less things into its closure, allowing the garbage collector to act.
- Small reporter refactor to fit the removal of the beforeInitListeners API


Nuckyz:
Documented the changes

Nuckyz:
@CodiumAI-Agent /review

Nuckyz:
@CodiumAI-Agent /improve

Nuckyz:
@CodiumAI-Agent /improve

src/webpack/common/menu.ts
```diff
@@ -16,17 +16,14 @@
...
 
-waitFor(["MenuItem", "MenuSliderControl"], m => Menu = m);
+export const Menu = findByProps<t.Menu>("MenuItem", "MenuSliderControl");
 
-export const ContextMenuApi: t.ContextMenuApi = mapMangledModuleLazy('type:"CONTEXT_MENU_OPEN', {
+export const ContextMenuApi: t.ContextMenuApi = mapMangledModule('type:"CONTEXT_MENU_OPEN', {
     closeContextMenu: filters.byCode("CONTEXT_MENU_CLOSE"),
     openContextMenu: filters.byCode("renderLazy:"),
-    openContextMenuLazy: e => typeof e === "function" && e.toString().length < 100
+    openContextMenuLazy: e => typeof e === "function" && String(e).length < 100
```
Scyye, line 28: Ive noticed this change in a few places, just wanted to ask what the point of using `String(val)` over `val.toString()` is?

src/webpack/common/types/fluxEvents.d.ts
```diff
@@ -32,9 +32,9 @@ function makeFluxEventList() {
         events.add(event);
     }
 
-    return Array.from(events, e => JSON.stringify(e)).sort().join("|");
+    return Array.from(events, e => JSON.stringify(e)).sort().join(" | ");
```
Scyye, line 35: Was this lint auto fix? or did you reformat commented out code?

src/webpack/patchWebpack.ts
```diff
@@ -1,362 +1,507 @@
...
+        define(this, "e", {
+            enumerable: false,
 
+            set(this: WebpackRequire, ensureChunk: WebpackRequire["e"]) {
+                define(this, "e", { value: ensureChunk });
                 clearTimeout(setterTimeout);
-                if (bundlePath !== "/assets/") return;
 
-                logger.info(`Main Webpack found in ${fileName}, initializing internal references to WebpackRequire`);
+                logger.info("Main WebpackInstance found" + interpolateIfDefined` in ${fileName}` + ", initializing internal references to WebpackRequire");
```
Scyye, line 71: Whats the point of this change? It just makes it less intuitive what is going on, and it mixes formatted and concatting strings
```suggestion
                logger.info(`Main WebpackInstance found ${interpolateIfDefined} in ${fileName}, initializing internal references to WebpackRequire`);
```
I *think* thats the same thing? again, its really hard to read what that string is.

src/webpack/common/menu.ts
Nuckyz: consistency and avoid errors when nullish

src/webpack/common/types/fluxEvents.d.ts
```diff
@@ -32,9 +32,9 @@ function makeFluxEventList() {
...
+// 61.8kb worth of events ???????
```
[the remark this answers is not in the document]
Nuckyz: gotta keep the number up to date💪

src/webpack/common/types/fluxEvents.d.ts
Nuckyz: intentional change

src/webpack/patchWebpack.ts
Nuckyz: interpolateIfDefined returns an empty string if any one of the variables used inside of it is nullish, so that's not quite the same

src/utils/lazy.ts
```diff
@@ -1,129 +1,167 @@
...
             }
+
             return proxyDummy[SYM_LAZY_CACHED];
-        }
+        },
+        [SYM_LAZY_CACHED]: void 0 as T | undefined
     });
 
-    return new Proxy(proxyDummy, {
+    const proxy = new Proxy(proxyDummy, {
```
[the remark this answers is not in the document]
Nuckyz: actually it's used above in

```js
proxy.toString = proxyDummy[SYM_LAZY_CACHED].toString.bind(proxyDummy[SYM_LAZY_CACHED]);
```

src/debug/loadLazyChunks.ts
```diff
@@ -109,32 +101,33 @@ export async function loadLazyChunks() {
...
-            searchAndLoadLazyChunks(factory.toString()).then(() => isResolved = true);
+            searchAndLoadLazyChunks(String(factory))
+                .then(() => isResolved = true)
+                .catch(() => isResolved = true);
 
             chunksSearchPromises.push(() => isResolved);
-        });
+        }
 
+        Webpack.factoryListeners.add(factoryListener);
```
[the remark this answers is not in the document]
Nuckyz: factoryListener is manually called under too

src/Vencord.ts
```diff
@@ -125,7 +125,7 @@ async function init() {
         const pendingPatches = patches.filter(p => !p.all && p.predicate?.() !== false);
         if (pendingPatches.length)
             PMLogger.warn(
-                "Webpack has finished initialising, but some patches haven't been applied yet.",
+                "Webpack has finished initializing, but some patches haven't been applied yet.",
```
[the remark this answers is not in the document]
Nuckyz: :x:

Nuckyz:
Closing as I will likely revive this in a different way from scratch

Pharaoh2k:
Look VERY promising! Hopefully will get revived and pushed real soon! 👍

PR #2413, opened 2024-05-03
NewGuildSettings: add push notifs, highlights & events

GabiRP:


src/plugins/newGuildSettings/index.tsx
```diff
@@ -31,6 +31,15 @@ const settings = definePluginSettings({
         type: OptionType.BOOLEAN,
         default: true
     },
+    messages: {
+        description: "Server Notification Settings",
+        type: OptionType.SELECT,
+        options: [
+            { label: "All messages", value: 0 },
+            { label: "Only @mentions", value: 1 },
+            { label: "Nothing", value: 2, default: true },
```
Vendicated, line 40: the default should probably be Only mentions

src/plugins/newGuildSettings/index.tsx
```diff
@@ -77,8 +96,11 @@ export default definePlugin({
         updateGuildNotificationSettings(guildId,
             {
                 muted: settings.store.guild,
+                message_notifications: settings.store.messages,
                 suppress_everyone: settings.store.everyone,
-                suppress_roles: settings.store.role
+                suppress_roles: settings.store.role,
+                mute_scheduled_events: settings.store.events,
+                notify_highlights: settings.store.highlights ? 1 : 2
```
Vendicated, line 103: wouldn't this make more sense as HighlightSettings.NULL? I assume thats the default value
```suggestion
                notify_highlights: settings.store.highlights ? 1 : 0
```

src/plugins/newGuildSettings/index.tsx
GabiRP: done :p

src/plugins/newGuildSettings/index.tsx
GabiRP: Just checked again and you're right, NULL is the default value, don't know why I used that tbh

src/plugins/newGuildSettings/index.tsx
```diff
@@ -31,6 +31,15 @@ const settings = definePluginSettings({
         type: OptionType.BOOLEAN,
         default: true
     },
+    messages: {
+        description: "Server Notification Settings",
+        type: OptionType.SELECT,
+        options: [
+            { label: "All messages", value: 0 },
+            { label: "Only @mentions", value: 1, default: true },
+            { label: "Nothing", value: 2 },
```
Vendicated, line 40: @dolfies raised a good point: it would be good to add a fourth option "Server default" which does not change message_notifications as all, and use that as the default

Vendicated:
really cool additions! feel free to add yourself as a Dev of the plugin if you'd like

src/plugins/newGuildSettings/index.tsx
GabiRP: Done, I don't know if there's a better way to do it though

Vendicated:
looks good now, thanks a lot for your contribution and the super swift feedback fixes

PR #2417, opened 2024-05-05
ImageLink: Fix extra ugly embed

Sqaaakoi:
Before (image courtesy of @Vendicated)
screenshot showing ugly embed

After
screenshot without ugly embed

How I fixed it: Replaced 1 of the 2 places the original patched function is called from with false directly, because the other place the function was being called from was causing that embed.

Also, this shares a find with ValidReply that is technically overwritten, is that an issue?


Nuckyz (approved):

Nuckyz (approved):

Nuckyz:
Thanks!!

PR #2418, opened 2024-05-05
feat(Translate): add toggle for chat bar icon

Cats1337:
Felt bad for pining Vee to add this so I did it myself.


Vendicated:
thanks!

PR #2422, opened 2024-05-08
ThemeAttributes: add data-author-username to messages

OvercastWarmth:
Very minor change, so we didn't add ourselves to the author field.
We use a modified version of the plugin locally because using IDs is unreliable for our use case (theming PluralKit messages) and having the username exposed makes it possible.
Thought we might as well open up a PR for it rather than let it sit alone in a gist somewhere.


Vendicated:
thanks!

PR #2424, opened 2024-05-08
MoreUserTags: Add chat moderator tag

eepyfemboi:
added a trial mod tag because some servers only allow new mods to timeout members

### DIFF
--- a/src/plugins/moreUserTags/index.tsx
+++ b/src/plugins/moreUserTags/index.tsx
@@ -270,11 +270,11 @@ export default definePlugin({
 
     start() {
         // To prevent existing users from crashing on lines 379-380 due to the settings entry not existing (sry that its kinda sloppy)
-        if (settings.store.tagSettings && !settings.store.tagSettings['TRIAL_MODERATOR']) {
-            settings.store.tagSettings['TRIAL_MODERATOR'] = {
-                'text': 'T Mod', 
-                'showInChat': true, 
-                'showInNotChat': true
+        if (settings.store.tagSettings && !settings.store.tagSettings.TRIAL_MODERATOR) {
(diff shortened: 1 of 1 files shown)


Vendicated:
i like the idea of this, but i don't think calling it a "Trial Moderator" is gonna be accurate for many users. Perhaps "Chat Moderator" would be more accurate?

cc @sunnniee

src/plugins/moreUserTags/index.tsx
```diff
@@ -263,7 +269,16 @@ export default definePlugin({
...
+        // To prevent existing users from crashing on lines 379-380 due to the settings entry not existing (sry that its kinda sloppy)
+        if (settings.store.tagSettings && !settings.store.tagSettings['TRIAL_MODERATOR']) {
+            settings.store.tagSettings['TRIAL_MODERATOR'] = {
+                'text': 'T Mod', 
+                'showInChat': true, 
+                'showInNotChat': true
+            };
+            return;
+        }
+        else if (settings.store.tagSettings) return;
```
Vendicated, line 281: i think the old migration has existed for long enough now, so feel free to just remove it

Vendicated: actually i guess we still need the `settings.store.tagSettings = defaultSettings`, but the legacy migration is obsolete now

eepyfemboi: Is there anything i should do about it? or..
Also i changed the name to "Chat Mod" like u mentioned earlier

Vendicated: i did it for you!

Vendicated (approved):
looks good, thanks!

PR #2425, opened 2024-05-08
feat(plugin): CatPictures

Jannniss:
This plugin adds a command to get a cat picture from cataas api cause everyone likes cats.


cheesesamwich:
add yourself to the devs construct if you're PRing a plugin

src/plugins/catPictures/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+                if (response != null) {
+                    let file = new File([response], "cat.jpeg", { type: "image/jpeg" })
+                    setTimeout(() => UploadHandler.promptToUpload([file], ctx.channel, draft_type), 10);
+                }
+            }
+        },
+    ]
+});
+
+function returnFloatWithParameter(number, parameter) {
```
MrDiamondDog, line 133: This function should be named better (i can tell from the strings in this function that this should have a different name)

Also add types to the parameters, we're not in js

src/plugins/catPictures/index.tsx
```diff
@@ -0,0 +1,239 @@
+/*
```
MrDiamondDog, line 1: Use new header

src/plugins/catPictures/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+                        name: "negate"
+                    },
+                    {
+                        label: "custom",
+                        value: "custom",
+                        name: "custom (use custom options)"
+                    }]
+                },
+                {
+                    name: "saturation-multiplier",
```
MrDiamondDog, line 76: Realistically, no one is going to use some of these options. Consider removing some of them

Jannniss: removed color options and second brightness option

Vendicated:
hi, thank you a lot for your contribution!

this doesn't seem very useful and could easily be replaced with a simple discord bot (or better yet a [user installable app](https://discord.com/developers/docs/tutorials/developing-a-user-installable-app))

as such i'll be rejecting this, sorry

PR #2427, opened 2024-05-09
new plugin YoutubeDescription

arHSM:
Adds descriptions to youtube video embeds

demo

### DIFF
--- a/src/plugins/youtubeDescription/index.tsx
+++ b/src/plugins/youtubeDescription/index.tsx
@@ -5,7 +5,12 @@
  */
 
 import { Devs } from "@utils/constants";
+import { isNonNullish } from "@utils/guards";
 import definePlugin from "@utils/types";
+import { useState } from "@webpack/common";
+import { Embed } from "discord-types/general";
+
+interface ToggleableDescriptionProps { embed: Embed, original: () => any; }
 
 export default definePlugin({
     name: "YoutubeDescription",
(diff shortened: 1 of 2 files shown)


bignutty:
LGTM

Cynosphere:
I would personally prefer having it unexpanded by default, showing the first line, and then being able to click to show the rest.

arHSM:
@Cynosphere added

arHSM:
out

src/plugins/youtubeDescription/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    ],
+    ToggleableDescription({ embed, original }: ToggleableDescriptionProps) {
+        const [isOpen, setOpen] = useState(false);
+
+        console.warn(embed);
+
+        return !isNonNullish(embed.rawDescription)
+            ? null
+            : embed.rawDescription.length > 20
+                ? <div
```
Vendicated, line 36: this nested ternary isn't too great.. it'd be way nicer for readability to convert it into simple if statements, like
```js
if (!embed.rawDescription)
	return null;
if (embed.rawDescription.length <= 20)
	return original();

return (
	<div>
		....
```

src/plugins/youtubeDescription/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    patches: [
+        {
+            find: ".default.Messages.SUPPRESS_ALL_EMBEDS",
+            replacement: {
+                match: /case (\i\.MessageEmbedTypes\.VIDEO):(case \i\.MessageEmbedTypes\.\i:)*break;default:(\i)=(?:(this\.renderDescription)\(\))\}/,
+                replace: "$2 break; case $1: $3 = $self.ToggleableDescriptionWrapper({ embed: this.props.embed, original: $4.bind(this) }); break; default: $3 = $4() }"
+            }
+        }
+    ],
+    ToggleableDescription({ embed, original }: ToggleableDescriptionProps) {
```
Vendicated, line 28: this should be wrapped in an ErrorBoundary, then you also wouldn't need your Wrapper function
```suggestion
    ToggleableDescription: ErrorBoundary.wrap(({ embed, original }: ToggleableDescriptionProps) => {
```

src/plugins/youtubeDescription/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        const [isOpen, setOpen] = useState(false);
+
+        console.warn(embed);
```
Vendicated, line 31: please remove this console.log
```suggestion
```

Vendicated:
maybe an improvement to the collapsing could be to collapse on the first newline instead of just after some characters? with a sanity limit of like 200 characters or something

src/plugins/youtubeDescription/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default definePlugin({
+    name: "YoutubeDescription",
+    description: "Adds descriptions to youtube video embeds",
+    authors: [Devs.arHSM],
+    patches: [
+        {
+            find: ".default.Messages.SUPPRESS_ALL_EMBEDS",
+            replacement: {
+                match: /case (\i\.MessageEmbedTypes\.VIDEO):(case \i\.MessageEmbedTypes\.\i:)*break;default:(\i)=(?:(this\.renderDescription)\(\))\}/,
```
D3SOX, line 23:
```suggestion
            find: ".Messages.SUPPRESS_ALL_EMBEDS",
            replacement: {
                match: /case \i\.\i\.VIDEO:(case \i\.\i\.\i:)*break;default:(\i=this\.renderDescription\(\))\}/,
```

patch broke because of the recent doomsday update, here is the fixed patch

D3SOX:
I would also suggest here to make the default state an option like I did in https://github.com/D3SOX/Vencord/commit/34fa4aa32e8f7ee0ef51649b97ff0a6be098d7b8

PR #2428, opened 2024-05-09
MessageLogger: use discord variables instead of hardcoded colors

AnubisNekhet:
Colors now use discord variables for better theme compatibility

### DIFF
--- a/src/plugins/messageLogger/deleteStyleText.css
+++ b/src/plugins/messageLogger/deleteStyleText.css
@@ -1,16 +1,16 @@
 /* Message content highlighting */
 .messagelogger-deleted [class*="contents"] > :is(div, h1, h2, h3, p) {
-    color: var(--red-430, #f04747) !important;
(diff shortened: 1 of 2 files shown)


src/plugins/messageLogger/deleteStyleText.css
```diff
@@ -1,19 +1,19 @@
 /* Message content highlighting */
 .messagelogger-deleted [class*="contents"] > :is(div, h1, h2, h3, p) {
-    color: #f04747 !important;
+    color: var(--red-430, #f04747) !important;
```
Vendicated, line 3: why not just use --status-danger? since that's what the delete message button uses
image

AnubisNekhet: Updated, overlay style still uses `--red-430-hsl` for hsla

Vendicated:
thank you!!

PR #2430, opened 2024-05-09
Add empty userplugins dir

k26pl:
This will allow for easier creation of plugins for unexperienced people

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,8 @@ lerna-debug.log*
 .pnpm-debug.log*
 *.tsbuildinfo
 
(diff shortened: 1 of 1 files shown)


EdVraz:
if you lack the skill to create a directory you should NOT create any plugins.

nyakowint:
It should probably be reiterated that userplugins *aren't for unexperienced people* in the first place

Vendicated:
yeah i agree with the aforesaid, and it would add additional noise to the source code

thank you a lot for the contribution regardless!

PR #2432, opened 2024-05-10
feat(plugin): SVGEmbed

aamiaa:
This plugin makes SVG files embed like normal images. It supports svg files uploaded directly to Discord, as well as svg files linked via a `discord.com` or `cdn.discordapp.com` url.

It patches the existing image embed logic to recognize `.svg` files as valid images and have them embed the same way (i.e. via `img src` attribute), so there should be no XSS incidents.


src/plugins/svgEmbed/index.ts
```diff
@@ -0,0 +1,198 @@
...
+            const domain = new URL(url).hostname;
+            if (domain === "cdn.discordapp.com") {
+                const size = await getFileSize(url);
+                if (!size || size / 1024 / 1024 > MAX_SVG_SIZE_MB) continue;
+            }
+
+            const { width, height } = await getSVGDimensions(url);
+            // @ts-ignore
+            message.embeds.push({
+                id: "embed_1",
```
Vendicated, line 183: will this not cause problems when there are multiple embeds with this same id?

Vendicated (commented):
honestly looks good but i'm still scared about potential XSS D:

have u tested this with some sussy svgs to make sure they are properly handled?

src/plugins/svgEmbed/index.ts
aamiaa: From what I've tested, the final id is random anyways, so I just put it as 1

aamiaa:
> honestly looks good but i'm still scared about potential XSS D:
> 
> have u tested this with some sussy svgs to make sure they are properly handled?

I've tested it against a random SVG to which I've manually appended `<script>` tags, as well as a random one I've found online by googling "svg xss". No javascript was executed in either case.

aamiaa:
Happy 1 year anniversary of the pr! 🥳
Closing due to lack of interest in merging.

PR #2435, opened 2024-05-11
GoogleThat

cheesesamwich:
A simple plugin that adds a command to send a google search link, useful for helping people find things out by themselves


Vendicated:
thank you but text replace can do this! you just miss out on url encoding but u can just use + instead of spaces and thats all you need

i had an idea ages ago to add basic scripting to textreplace to allow it to url encode and such

PR #2440, opened 2024-05-11
Use a tiny character (˞) instead of emoji's name

JagTheFriend:
The message itself will appear cleaner.


Grafaffel:
just change it in the plugins setting

Sqaaakoi:
...this will break stuff more.

Vendicated:
yeah you changed the wrong code

and this is already a native feature of the plugin, just change the format in its settings 
image

JagTheFriend:
> yeah you changed the wrong code
> 
> and this is already a native feature of the plugin, just change the format in its settings image

Ohk, I didn't know

PR #2441, opened 2024-05-11
ShowTimeoutDuration: Simplify tooltip style, allow changing style without reload

Sqaaakoi:
The tooltip looked too big and crowded, so I made it simpler

screenshot from my test server

Please do clean up stuff if it is a mess


Vendicated:
could just replace the original tooltip option instead

Sqaaakoi:
> could just replace the original tooltip option instead

would you recommend that? i'll start working on that now if you'd rather it that way

Vendicated:
why's this still draft?

Sqaaakoi:
yeah i'm only stupid enough to think "hm, maybe this could do with more features" then get overwhelmed trying to reasonably figure out how to make it

Sqaaakoi:
just merge this i'll get around to it in 3-5 business months

Nuckyz:
Very clean! Thanks for the improvements. I took the opportunity to also fix the huge space between the icon and the countdown text

PR #2447, opened 2024-05-12
new api: ExpressionPickerTabsAPI

programminglaboratorys:
Tablist API. let's you create panels for the expression picker an example of it would be the [cookie clicker](https://github.com/programminglaboratorys/Vencord/blob/fun/src/plugins/cookieClicker/cookieClickerPreview.png?raw=true6) plugin. 

it will be awesome to have some plugins moving to the expression picker, like SendTimestamps as adding a button only for timestamps is kinda annoying imo.

the API still needs some improvements. any feedback is welcome!

> [!NOTE]  
> I should opened this PR request two weeks ago as I was thinking to add another feature which its called headless Panels the problem with it that it would ruin performance, so I scrapped the idea, originally it was a Panel without a tab button, its trigger only through code such a small thing to add


Sqaaakoi:
I think I've mentioned this before, but "Tablist" is a confusing name for this API

Something like ExpressionPickerTabAPI or like that would be more clear to anyone building upon this

programminglaboratorys:
> I think I've mentioned this before, but "Tablist" is a confusing name for this API
> 
> Something like ExpressionPickerTabAPI or like that would be more clear to anyone building upon this

don't you think that a bit too long? I will see others opinions, there is probably better namings then TablistApi as you said.

lewisakura:
Naming convention would be to make it clear what your API does rather than care about the length of the name, so I agree that "TablistAPI" is really undescriptive

Something like "ExpressionsAPI" would be more appropriate, since your API involves adding new types of expressions to the picker.

programminglaboratorys:
all done. awaiting review

src/api/ExpressionPickerTabs.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export type ExpressionPickerButtonComponent = (props: ExpressionPickerButtonProps) => JSX.Element | null;
+export type ExpressionPickerPanelComponent = (props: ExpressionPickerPanelProps) => JSX.Element | null;
+
+
+export interface ExpressionPickerTabItem {
+    tab: string,
+    Component: ExpressionPickerPanelComponent;
+    autoFocus?: boolean;
```
Jesus-QC, line 33: this property is never used once declared

Jesus-QC:
patch no longer works

src/api/ExpressionPickerTabs.tsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+export function* TabPanels(selectedTab: string, channel: Channel) {
+    for (const [id, { Component }] of ExpressionPickerComponents) {
+        if (id !== selectedTab) { continue; }
+        const PanelComponent: ExpressionPickerPanelComponent = Component;
+        yield (<ErrorBoundary><PanelComponent selectedTab={selectedTab} channel={channel} /></ErrorBoundary>);
+    }
+}
```
Nuckyz, line 61: Make these not use generator functions

Nuckyz (changes requested):
Recently we got shortcuts in the plugins definition to use most apis, this one should add those shortcuts too

look at renderChatBarButton for an example

programminglaboratorys:
> Recently we got shortcuts in the plugins definition to use most apis, this one should add those shortcuts too
> 
> look at renderChatBarButton for an example

 I was not able to find renderChatBarButton, can you explain those new changes and what am I required to do?

Sqaaakoi:
> > Recently we got shortcuts in the plugins definition to use most apis, this one should add those shortcuts too
> > look at renderChatBarButton for an example
> 
> I was not able to find renderChatBarButton, can you explain those new changes and what am I required to do?

Instead of plugins calling the API in their start/stop methods, plugins should have a property with metadata about the tabs they will add

Take a look at #3028 where this was implemented for all the other APIs

Define a type for this in `src/utils/types.ts` and register/unregister it in `src/plugins/index.ts` using your existing functions

PR #2449, opened 2024-05-12
ThemeAttributes: add larger avatar url variables to avatars

BoardTM:
As was advised in https://github.com/Vendicated/Vencord/pull/1976 I have now made an addition to the ThemeAttributes plugin.
image
image
The plugin now adds a CSS variable called --large-avatar-url to the img element in the AvatarStack discord uses for the profile pop-outs. This can be used in themes that want to make the profile picture larger without losing image quality as a consequence.

### DIFF
--- a/src/plugins/themeAttributes/index.ts
+++ b/src/plugins/themeAttributes/index.ts
@@ -13,7 +13,7 @@ import { Message } from "discord-types/general";
 const settings = definePluginSettings({
     imgSize: {
         type: OptionType.SELECT,
-        description: "The resolution of the avatar image",
+        description: "The resolution of the image in the --large-avatar-url CSS variable",
         options: ["300", "512", "1024", "2048", "4096"].map(n => ({ label: n, value: n, default: n === "300" })),
 
     }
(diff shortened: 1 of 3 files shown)


Vendicated:
does this also apply to chat avatars? ive seen that use case also come up

BoardTM:
> does this also apply to chat avatars? ive seen that use case also come up
Should now also apply for chat avatars!

src/plugins/themeAttributes/index.ts
```diff
@@ -4,15 +4,27 @@
...
+import definePlugin, { OptionType } from "@utils/types";
 import { UserStore } from "@webpack/common";
 import { Message } from "discord-types/general";
 
+const settings = definePluginSettings({
+    imgSize: {
+        type: OptionType.SELECT,
+        description: "The resolution of the image in the --large-avatar-url CSS variable",
+        options: ["300", "512", "1024", "2048", "4096"].map(n => ({ label: n, value: n, default: n === "300" })),
+
```
Vendicated, line 18:
```suggestion
```

src/plugins/themeAttributes/index.ts
```diff
@@ -32,6 +44,24 @@ export default definePlugin({
...
             }
+        },
+
+        // add --large-avatar-url css variable to avatar img elements
+        // popout profiles
+        {
+            find: ".LABEL_WITH_ONLINE_STATUS",
+            replacement: {
+                match: /src:null!=\i\?(\i).{1,50}"aria-hidden":!0/,
+                replace: "$&,style:{\"--large-avatar-url\":\"url(\"+$1.replace(/\\d+$/,$self.settings.store.imgSize)+\")\"}"
```
Vendicated, line 55: try to avoid having too much code inside your replacements, it's hard to maintain and easy to mess up! plus doing it this way allows you to reuse the same func for both avatar patches

```suggestion
                replace: "$&,style:$self.getAvatarStyles($1)"
```

BoardTM:
That should have resolved all of your requested changes.

Vendicated (approved):
looks good, thanks!

PR #2452, opened 2024-05-13
feat: Allow finds to use regex

Nuckyz:


src/webpack/patchWebpack.ts
```diff
@@ -257,7 +257,7 @@ function patchFactories(factories: Record<string, (module: any, exports: any, re
         for (let i = 0; i < patches.length; i++) {
             const patch = patches[i];
             if (patch.predicate && !patch.predicate()) continue;
-            if (!code.includes(patch.find)) continue;
+            if (typeof patch.find === "string" ? !code.includes(patch.find) : !patch.find.test(code)) continue;
```
Vendicated, line 260: imo this ternary inside an if (essentially nested conditions) is kinda hard to parse. it'd be a lot nicer to do like
```js
const moduleMatches = typeof patch.find === "string"
					      ? code.includes(patch.find)
						  : patch.find.test(code);
if (!moduleMatches) continue;
```

looks good other than that

PR #2456, opened 2024-05-13
FakeNitro: allow using subscription-locked emojis

nakoyasha:
Patches FakeNitro to allow posting subscription-locked emojis, *hopefully* shouldn't cause issues 🙏

### DIFF
--- a/src/plugins/fakeNitro/index.tsx
+++ b/src/plugins/fakeNitro/index.tsx
@@ -24,7 +24,7 @@ import { getCurrentGuild } from "@utils/discord";
 import { Logger } from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy, findStoreLazy, proxyLazyWebpack } from "@webpack";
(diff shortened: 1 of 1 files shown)


src/plugins/fakeNitro/index.tsx
```diff
@@ -801,7 +809,20 @@ export default definePlugin({
...
+        const userRoles = GuildMemberStore.getSelfMember(e.guildId).roles;
         if (e.require_colons === false) return true;
+        // check if we have any of the required roles
+        // if we don't do a nested check, then ts will complain about the below code being unreachable
+        if (e.roles.length > 0) {
+            if (e.roles.some(role => userRoles.includes(role))) {
+                return true;
+            } else {
+                return false;
+            }
```
MrDiamondDog, line 822:
```suggestion
            return e.roles.some(role => userRoles.includes(role))
```

nakoyasha: That makes Typescript think the code below is "unreachable" (when it is reachable), I changed it to `if (e.roles.some(role => userRoles.includes(role)) !== true) return false;` instead in the latest commit, which avoids that by explicitily returning

PR #2458, opened 2024-05-13
EmoteCloner: allow cloning from reactions

axiand:
Small patch that allows cloning an emoji from a message reaction. 

Would solve this request: Vencord/plugin-requests#179


Vendicated:
looks good, thanks!

PR #2459, opened 2024-05-14
add option to disable CheckedTextInput

FireStreaker2:
very small change to add an option to disable the ``CheckedTextInput`` component, along with enabling the option for the one located in the vencord cloud section because it kinda bugged me

<details>
  <summary>images</summary>
  
  before:  
  image

  after:  
  image

</details>

### DIFF
--- a/src/components/VencordSettings/CloudTab.tsx
+++ b/src/components/VencordSettings/CloudTab.tsx
@@ -135,6 +135,7 @@ function CloudTab() {
                 </Forms.FormText>
                 <CheckedTextInput
                     key="backendUrl"
+                    disabled={!settings.cloud.authenticated}
                     value={settings.cloud.url}
                     onChange={v => { settings.cloud.url = v; settings.cloud.authenticated = false; deauthorizeCloud(); }}
                     validate={validateUrl}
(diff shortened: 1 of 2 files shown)


src/components/VencordSettings/CloudTab.tsx
```diff
@@ -135,6 +135,7 @@ function CloudTab() {
                 </Forms.FormText>
                 <CheckedTextInput
                     key="backendUrl"
+                    disabled={!settings.cloud.authenticated}
```
Vendicated, line 138: this doesn't seem right

this form is for changing the backend url, and this also affects the enable cloud integrations option

the ui is a bit confusing i guess, the backend url form should be above the enable option

FireStreaker2: > and this also affects the enable cloud integrations option  

ah ok, didnt know about that feature, just seemed weird that you could edit the backend url even though you didnt enable the feature

should i close the pr then?

Vendicated: the enable switch is moreso for opting into the feature & authorising via discord oauth

if you change the backend url option, the switch will authorise with that backend instead. so yes this pr in its current state doesn't make a lot of sense...

but it did make clear that the ui is a bit confusing, the backend url option would make more sense on top i think?

FireStreaker2: alright understood, but yeah i think it would make more sense if it was at the top - might create another pr later

PR #2461, opened 2024-05-15
Fix: PetPet Plugin's Upload Image Option

ScattrdBlade:
The image option for the PetPet command has not been working for about 5 months now, as noted in the closed #2070 pull request. This finally fixes it so it actually works again. This also includes clearing the slash command draft after you run it, so if you run the slash command again, the previously uploaded file from the previous slash command run will not be there in the upload area. The code to clear draft is placed in more than one area within PetPet's code to make sure the slash command draft is cleared even if the command fails and such. Also made DraftType use webpack common and implemented it into the `fakeNitro` and `crashHandler` plugins. Along with DraftType, I also added UploadModule (which is the thing that allows draft clear to happen) to webpack common to make it easier for future use.

### DIFF
--- a/src/plugins/crashHandler/index.ts
+++ b/src/plugins/crashHandler/index.ts
@@ -135,8 +135,11 @@ export default definePlugin({
         try {
             const channelId = SelectedChannelStore.getChannelId();
 
-            DraftManager.clearDraft(channelId, DraftType.ChannelMessage);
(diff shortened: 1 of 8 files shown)


Nuckyz:
Thank you for the fix! I also pushed a commit with just a small clean up and sneaked in a fix for things I did not realize earlier.

PR #2465, opened 2024-05-16
New Plugin: File Share

ScattrdBlade:
Adds a plugin called "File Share" which adds a `/fileshare` slash command that allows you to upload files to gofile.io and then pastes the link into chat box to share with others. Since its a quick and easy way to share files, this would be very convenient for people who want to upload big files but don't have Discord nitro. Plus, it doesn't require other users to directly download the file if they'd like to view the file without downloading (such as a PDF).

### DIFF
--- a/src/plugins/messageLogger/HistoryModal.tsx
+++ b/src/plugins/messageLogger/HistoryModal.tsx
@@ -39,7 +39,7 @@ export function HistoryModal({ modalProps, message }: { modalProps: ModalProps; 
         <ModalRoot {...modalProps} size={ModalSize.LARGE}>
             <ModalHeader className={cl("head")}>
                 <Text variant="heading-lg/semibold" style={{ flexGrow: 1 }}>Message Edit History</Text>
-                <ModalCloseButton onClick={close} />
+                <ModalCloseButton onClick={modalProps.onClose} />
             </ModalHeader>
 
             <ModalContent className={cl("contents")}>


Golonchy:
sounds very cool, nice

src/plugins/fileShare/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    return null;
+}
+
+export default definePlugin({
+    name: "File Share",
+    description: "Upload your files to gofile.io and share the link in chat via /fileshare slash command.",
+    authors: [Devs.ScattrdBlade],
+    dependencies: ["CommandsAPI"],
+    target: "DESKTOP",
```
Sqaaakoi, line 31: Do not do this, name your plugin folder `fileShare.desktop` instead

Sqaaakoi: Also, is this even required to be marked as "desktop only"

ScattrdBlade: Honestly, I wasn't sure originally if it'd work on the web and never ended up removing it later

fdkchx:
add tweaks for token interaction and more serious customizations

ScattrdBlade:
> add tweaks for token interaction and more serious customizations

The only thing I could really add is token input so users can upload files to their own gofile accounts if they want to, an option to upload to specific folders/naming folders specific names, and deleting previously uploaded files. I can only do so much because most of the gofile API requires premium accounts. But I'll do my best to implement the features I just listed as customizations within the next week or so.

Vendicated:
what's the point of such a plugin? why not just use sharex or manually upload?

with the new higher upload limit there's basically no use case for having this plugin

ScattrdBlade:
> what's the point of such a plugin? why not just use sharex or manually upload?
> 
> with the new higher upload limit there's basically no use case for having this plugin

Even though Discord has increased its upload limits, I've still noticed that many users (especially those without nitro) still run into size limitations. Why not use ShareX? As a ShareX user myself, I love it, but it requires linking accounts or setting up API keys (or a custom domain), which many people don't want to do. On the other hand, gofile.io allows for temporary account creation for uploads (similar to Pastebin) which simplifies the upload process significantly, especially since they have nothing to do with the users and the users just receive a link from gofile.io that says the file was uploaded to a temporary account (no actual account creation is done by the user). As for manual uploads to file-sharing sites, they involve many additional steps that lots of users prefer avoiding. Another big plus with gofile.io is that you can preview files online before downloading them (even files that don't render directly on Discord), which can come in handy for those who are cautious about downloading potentially harmful files (since files on gofile.io have no way of acting maliciously until they are downloaded). The plugin gets rid of the extra steps that come with manual uploads or using tools like ShareX, such as account/key creation, linking, repeated signing in, etc. If someone runs into a file size issue on Discord or prefers not to upload directly onto Discord for certain reasons, the plugin makes it super easy and convenient, right under your fingertips. It's just like uploading to Discord but through a simple slash command, saving a ton of time and effort. I've even had a few people show interest in this plugin already, and someone even reached out to me regarding the plugin because they somehow found my repo for it and started using it. So I'd say it can be quite useful and there are definitely people out there that do want it. Sure, this could just be a third-party plugin, but having it as an official plugin makes it so much more convenient for everyone. After all, it's mainly the average (non-dev build) users who often don't use manual file uploaders or ShareX.

PR #2468, opened 2024-05-16
Add NoDefaultHangStatus plugin

D3SOX:
Resolves https://github.com/Vencord/plugin-requests/issues/564


src/webpack/common/types/fluxEvents.d.ts
```diff
@@ -37,4 +37,1730 @@ function makeFluxEventList() {
...
+export type FluxEvents =
```
Curve, line 40: did you mean to format this?

src/plugins/noDefaultHangStatus/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import { FluxDispatcher } from "@webpack/common";
+
+export default definePlugin({
+    name: "NoDefaultHangStatus",
+    description: "Disable the default hang status when joining voice channels",
+    authors: [Devs.D3SOX],
+
+    flux: {
+        UPDATE_HANG_STATUS: ({ status, saveAsDefault }: { status: string, saveAsDefault?: boolean }) => {
+            if (saveAsDefault === undefined && status) {
```
Curve, line 18: out of curiosity, would it also be possible to just discard the event instead of dispatching a new one?

src/webpack/common/types/fluxEvents.d.ts
Nuckyz, line 1: Don't update this file, another pr is already targetted for it

src/plugins/noDefaultHangStatus/index.ts
Nuckyz: I agree, is it possible to just avoid the update hang status being fired for when you join the channel?

Nuckyz (changes requested):

src/plugins/noDefaultHangStatus/index.ts
D3SOX: I wanted to do that too but I don't know if there's a way to cancel events :o

src/webpack/common/types/fluxEvents.d.ts
D3SOX: I already waited for this comment yeah should probably format back

src/plugins/noDefaultHangStatus/index.ts
Nuckyz: What is sending that event? The client or the server websocket?

D3SOX: I'm certain the client dispatches it

src/webpack/common/types/fluxEvents.d.ts
D3SOX: Should I just revert my changes?

Nuckyz: for this file yes

src/plugins/noDefaultHangStatus/index.ts
Nuckyz: okay but based on what? what is the root cause that causes this to be fired? ws event? joinChannel function?

D3SOX:
Now it's using a patch to cancel the event

Nuckyz (approved):

Nuckyz:
Thanks for the new plugin!

PR #2472, opened 2024-05-17
Fix: Resurrect Home Plugin's Cached Messages Bug

ScattrdBlade:
This fixes a bug in the Resurrect Home plugin where recent messages would disappear, leaving only older and real-time messages visible when returning to a channel after viewing an old thread within it.

(I specifically discovered the bug when clicking on the [oldest thread](https://discord.com/channels/1015060230222131221/1059989000204583002) created in `#core-development` in the Vencord server, and then going back to the regular `#core-development` channel)

### DIFF
--- a/src/plugins/resurrectHome/index.tsx
+++ b/src/plugins/resurrectHome/index.tsx
@@ -134,8 +134,8 @@ export default definePlugin({
         {
             find: '"MessageActionCreators"',
             replacement: {
-                match: /(?<=focusMessage\(\i\){.+?)(?=focus:{messageId:(\i)})/,
-                replace: "before:$1,"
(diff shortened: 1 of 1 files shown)


Nuckyz:
Is force: true or only cache is enough? The before was there because it would clear the previous cache, but it was a hackfix

ScattrdBlade:
> Is force: true or only cache is enough? The before was there because it would clear the previous cache, but it was a hackfix

Using only one of them **does** work, however, I decided to use both just to make sure it works in all cases. Not using either of them just scrolls the channel up to where the thread was created.

Nuckyz:
The patch does not apply, which results in the message request spam in homes happening again. And also, cache: true and force: true are not options which you can use, how did you find them?

PR #2479, opened 2024-05-18
feat(plugin): OnlineTyping

vegardev:
A plugin that allows users to appear invisible until they start typing.

- Features two kinds of delays -- a delay useful for if you mistype and don't want to appear online, as well as a delay for switching back to invisible after not typing.

- Users may pick between Online, Idle and Do Not Disturb as their desired status.

- The chat bar may have a toggle if preferred, and may also switch to your desired status if you decide to toggle the plugin off.
<details>
<summary>Screenshot of plugin settings</summary>

image

</details>

### DIFF
--- a/src/plugins/onlineTyping/index.tsx
+++ b/src/plugins/onlineTyping/index.tsx
@@ -1,0 +1,174 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addChatBarButton, ChatBarButton, removeChatBarButton } from "@api/ChatButtons";
+import { definePluginSettings } from "@api/Settings";
+import { makeRange } from "@components/PluginSettings/components";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByCode } from "@webpack";
+import { FluxDispatcher } from "@webpack/common";
+
+// Delay before appearing online when typing
+let statusChangeDelay: NodeJS.Timeout | null = null;
+
+// Delay before status switches back to invisible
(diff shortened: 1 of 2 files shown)


src/plugins/onlineTyping/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    name: "OnlineTyping",
+    description: "Automatically appear online when you are typing, most effective when current status is invisible",
+    settings,
+    authors: [Devs.vegard],
+    dependencies: ["ChatInputButtonAPI"],
+    patches: [
+        {
+            find: 'this,"handleTextareaChange",',
+            replacement: {
+                match: /\((\w+),(\w+),(\w+)\)=>\{let{keyboardModeEnabled:(\w+),channel:{id:(\w+)}}=([^;]+);/,
```
MrDiamondDog, line 127: We have a special regex symbol, `\i`, that matches variable names. Use this instead of `\w+`

MrDiamondDog (commented):
ngl I don't really see any good use case for this plugin. At best it would just look weird when people see you going online and offline constantly

src/plugins/onlineTyping/index.tsx
vegardev: ahh I didn't realize, thanks

vegardev:
> ngl I don't really see any good use case for this plugin. At best it would just look weird when people see you going online and offline constantly

I totally see that, although it could have its use cases for people that don't want to seem online all the time, while not active particularly at the same time.
It's natural I don't want this rejected, but if anything it's a learning experience making this :3

Scyye:
Seems spammy, and I dont see the actual use case for this

vegardev:
> Seems spammy, and I dont see the actual use case for this

Yes, I agree. I've thought about it and could be considered a form of API abuse

PR #2480, opened 2024-05-18
feat(plugin): MediaDownloader

TheColorman:
This is a plugin that adds a `yt-dlp` command. You pass a url, it downloads it and readies it as an attachment. See the gif for a quick example.
based

It supports any URL supported by https://github.com/yt-dlp/yt-dlp.

> ## ~~Warning:~~
> ~~This plugin has 2 dependencies: `yt-dlp` and `ffmpeg`. `yt-dlp` gets automatically downloaded and stored locally, but `ffmpeg` does not, and thus needs to be installed by the user.~~

I have not yet tested this on Linux, ~~and I might also add support for converting directly to gifs~~, but feel free to experiment and find bugs.


src/plugins/YtDlp/index.ts
```diff
@@ -0,0 +1,191 @@
...
+        showNotification({
+            title: "yt-dlp",
+            body: "ffmpeg not found. yt-dlp requires ffmpeg to work."
+        });
+    }
+    return res;
+}
+
+const settings = definePluginSettings({
+    additionalArguments: {
```
MrDiamondDog, line 85: why not make this just a plain string and pass it directly to ytdlp?

TheColorman: The way `child_process.spawn` passes arguments means that each string is a single argument, even if it contains spaces. E.g. if you were to try and pass `"-o output.mp4"` as an argument, you would get an error as `yt-dlp` as no arguments called `o output.mp4`.

I thought about splitting the string on spaces, but then you can't include multi-word arguments, unless I *also* implement some quote parser.

rozbrajaczpoziomow:
Doesn't work under Linux (probably missing e**x**ecute.
image

Also, why redownload yt-dlp if it is already installed on the system? (like on mine)

Vendicated:
i love this plugin idea! 

however based on a quick skim of the code, the current implementation is very questionable. why is ytdlp stored in indexeddb? you need it on the native side, not in the browser

i honestly don't think the plugin should download it for you. why not simply test if it's available and if not show the user a popup with instructions how to install it?

installing programs from a plugin is not something I want to do and it might trip anti malware engines, not to mention that it also just seems a little sussy. and on Linux this should really be handled via package manager anyway

we could just have the user install python themselves and then use pip to install it if it's missing

TheColorman:
@rozbrajaczpoziomow 
> Doesn't work under Linux (probably missing e**x**ecute perm)
Nice catch, weird that it's interrupthing though, I thought I had error handling everywhere. I'll have to take a look at that.

> Also, why redownload yt-dlp if it is already installed on the system? (like on mine)
Good point, should probably add a check.

@Vendicated 
> however based on a quick skim of the code, the current implementation is very questionable. why is ytdlp stored in indexeddb? you need it on the native side, not in the browser
I wanted to store it somewhere and wasn't sure where else to do it..

> i honestly don't think the plugin should download it for you. why not simply test if it's available and if not show the user a popup with instructions how to install it?
Yeah that's fair enough. I wanted to make it completely seamless, but halfway through I realised that doing the same to download ffmpeg would be a pain in the ass, so now it's in a bit of a weird state where half of the dependencies auto-install. This definitely makes more sense.

> we could just have the user install python themselves and then use pip to install it if it's missing
I'm not sure that would add it to `PATH`, would it? I've always had to manually add yt-dlp to the `PATH` unless I entered a Python venv.

Vendicated:
anything installed via pip should be in path. when you install python on windows there's an option to add python stuff to the PATH, you need to make sure to have it checked

in any case, you could just do `python3 -m yt-dlp` or even write a small python script that imports the module and uses the api from python

Sqaaakoi:
> Doesn't work under Linux (probably missing e**x**ecute perm) image
> 
> Also, why redownload yt-dlp if it is already installed on the system? (like on mine)

it's running the windows version??

TheColorman:
I'm pretty happy with the state of the plugin now, but I'm happy with any feedback!

src/plugins/YtDlp/index.ts
```diff
@@ -0,0 +1,195 @@
...
+async function checkDependencies() {
+    return await checkffmpeg() && await checkytdlp();
+}
+
+const settings = definePluginSettings({
+    additionalArguments: {
+        type: OptionType.STRING,
+        description: "Additional arguments to pass to yt-dlp. Format: JSON-parsable array of strings, e.g. [\"--format\", \"bestvideo+bestaudio\"]",
+        default: "[]",
+        restartNeeded: false,
```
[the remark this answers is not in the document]
TheColorman: Removed in [a2b8d91](https://github.com/Vendicated/Vencord/pull/2480/commits/a2b8d91cf35daeaf4693621e813bd348736639c4).

src/plugins/YtDlp/constants.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+export const Extensions = ["webm", "mp4", "mp3", "gif"] as const;
```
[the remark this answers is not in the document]
TheColorman: Removed in [6c7e639](https://github.com/Vendicated/Vencord/pull/2480/commits/6c7e639cfd8779d5341f885935947bc2b95dcd6d).

TheColorman:
And now I've exhausted all the extra ideas I had for the plugin. I haven't encountered any videos it couldn't handle yet, but be sure to let me know if it does happen.

src/plugins/YtDlp/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+        description: "Show a warning message if ffmpeg is not installed.",
+        default: true,
+    },
+    defaultGifQuality: {
+        type: OptionType.NUMBER,
+        description: "The quality level to use if no value is specified when downloading gifs. A number between 1 and 5.",
+        default: 3,
+    }
+}, {
+    additionalArguments: {
```
Scyye, line 168: This was probably asked already, but why does this have to be json? why cant it just be a string?

src/plugins/YtDlp/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+            return await download(ctx.channel, {
+                url,
+                format,
+                gifQuality,
+                addArgs
+            });
+        }
+    }, {
+        description: "test",
```
Scyye, line 246: I wouldnt include this in the release

src/plugins/YtDlp/native.ts
```diff
@@ -0,0 +1,290 @@
...
+    error: string;
+    logs: string;
+}> {
+    logs_global = "";
+    try {
+        const m = await metadata(opt);
+        const f = genFormat(m, opt);
+        const d = await download(f, opt);
+        const r = await remux(d, opt);
+        const u = upload(r);
```
Scyye, line 252: Name variables properly 🥺

src/plugins/YtDlp/index.tsx
Sqaaakoi: Probably because it's an array. Array of arguments only really matters on Unix platforms where arguments are *real*. Also this doesn't even check if it's an array of strings anyway so this needs to be fixed regardless of what outcome will be implemented

src/plugins/YtDlp/native.ts
Scyye:
```diff
- f
+ format
- d
+ download
- r
+ remux
- u
+ upload
```

Scyye:
Also, does this work on the web version?

src/plugins/YtDlp/index.tsx
Scyye: could something similar to TextReplace work? where it adds more options as you fill them in?

Scyye:
> lemme make you think about it for 1 second - if it has to download yt-dlp/ffmpeg locally, run child_process...

that was my point lol

src/plugins/YtDlp/index.tsx
TheColorman: As @Sqaaakoi mentiond, it's due to the way arguments are passed to the child process. This causes problems even on Windows, where I noticed that passing a single string in the form "--output file.mp4" threw an exception, but passing two seperate strings, ["--output", "file.mp4"] worked fine. The best solution would probably just be to create a parser that converts a single string into an array though.

Sqaaakoi: i forgot it's still very cursed on windows

src/plugins/YtDlp/index.tsx
TheColorman: Yikes, seems a couple test changes made it through. Could have sworn I'd removed them though 😖

src/plugins/YtDlp/native.ts
TheColorman: These too 😩
https://github.com/Vendicated/Vencord/pull/2480#discussion_r1615329758

TheColorman:
@Scyye 
> Also, does this work on the web version?; if not, id mark it as such.

How do I mark it as desktop only? I think I've seen a plugin or two using the `.desktip.(ts/tsx)` extension, is that sufficient?

src/plugins/YtDlp/index.tsx
Scyye: just.. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

Scyye:
> @Scyye
> 
> > Also, does this work on the web version?; if not, id mark it as such.
> 
> How do I mark it as desktop only? I think I've seen a plugin or two using the `.desktip.(ts/tsx)` extension, is that sufficient?

rename the folder to <pluginname>.desktop im pretty sure

src/plugins/YtDlp/native.ts
TheColorman: [f8f7405](https://github.com/Vendicated/Vencord/pull/2480/commits/f8f7405e1a959c06344d3a13e84d2e5162227580)

src/plugins/YtDlp/index.tsx
TheColorman: [72c2576](https://github.com/Vendicated/Vencord/pull/2480/commits/72c257685cf76342738d531cd0ff7a3b6882b87b)

src/plugins/YtDlp/index.tsx
TheColorman: Went with an argparser, makes it much easier to use now: [5205530](https://github.com/Vendicated/Vencord/pull/2480/commits/5205530e1f611dbe72f0d1f2dd2aab14fcc121da)

TheColorman:
With 8bc793c8ce1b30e36d9daa287d91a10e8c4ea8a1 it should only build for desktop.

src/plugins/YtDlp/index.ts
TheColorman: Fixed by 5205530e1f611dbe72f0d1f2dd2aab14fcc121da

src/plugins/YtDlp/index.tsx
TheColorman: > just.. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

Wanted to support arguments with spaces too, as those do come up from time to time 😄

src/plugins/YtDlp/native.ts
```diff
@@ -156,41 +160,47 @@ async function download({ format, videoTitle }: { format: string; videoTitle: st
...
+    let outfile: string;
+    let ext: string;
     switch (format) {
         case "audio":
-            ffmpegArgs = ["-i", p(file), "-b:a", `${kilobits}k`, "-maxrate", `${kilobits}k`, "-bufsize", "1M", "-y", p("remux.mp3")];
+            baseArgs = ["-i", p(file), "-b:a", `${kilobits}k`, "-maxrate", `${kilobits}k`, "-bufsize", "1M", "-y"];
+            outfile = "remux.mp3";
             ext = "mp3";
             break;
         case "video":
```
Scyye: Why define this if its just.. nothing?

src/plugins/YtDlp/index.tsx
```diff
@@ -272,20 +262,25 @@ export default definePlugin({
...
 }) {
     const promise = Native.execute({
         url,
         format,
         gifQuality,
-        additional_arguments: [
-            ...parseAdditionalArgs(settings.store.additionalArguments),
-            ...parseAdditionalArgs(addArgs)
+        ytdlpArgs: [
+            ...argParse(settings.store.ytdlpArgs || ""),
```
Scyye: Why arent you just splitting it at spaces? whats the point of making the enduser type out a parsable array?

I still think a system similar to that of TextReplace would work better.

TheColorman: They don't anymore. I updated the argument parser to essentially split at spaces, with some additional logic to ensure it wouldn't split in the middle of a quoted string. E.g.
```
-a x -b "multi word argument" 'quote\'s escape\'s'
```
turn into 
```
["-a", "x", "multi word argument", "quote's escape's"]
```
So now you can just use a string as if you were giving arguments directly in the cli.

src/plugins/YtDlp/native.ts
TheColorman: Are you referring to the outfile variable or something more? I've removed outfile in 0e96e9300df1a0b0841ae73d09fe0881347c931e, but let me know if there's something else.

Scyye: the "video" case, unless im missunderstanding whats going on, as im not the best typescript coder out there, it looks like literally nothing will happen if its a video?

TheColorman: Ah, that's a JavaScript feature called [Fall-Through](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#breaking_and_fall-through). You can basically name multiple cases right after each other, and all of them will execute the same code.

```js
switch (variable) {
    case "a":
    case "b":
        // do stuff - get's executed both in case "a" and case "b"
        break; // important to prevent fall-through down to case "c"
    case "c":
    default:
        // do stuff - gets executed both in case "c" and all other undefined cases.
}
```

I could remove `video` and it would have the same logic, but I prefer keeping it to make it clear that it's video logic.

Scyye: ahhh, okay, makes sense.
im used to java where the syntax is 
```java
switch (variable) {
    case "a", "b":
        // stuff for both "a" & "b"
}
```

src/plugins/ytdlp.desktop/index.tsx
```diff
@@ -133,7 +133,7 @@ async function openDependencyModal() {
 const settings = definePluginSettings({
     showProgress: {
         type: OptionType.BOOLEAN,
-        description: "Send a Clyde message with the download progress.",
+        description: "Send a Clyde message with the download progress. Warning: hiding progress will also hide the \"Cancel download\" button.",
```
Scyye, line 136: Is there any reason this cant be its own setting?

showing the cancel button?

cheesesamwich:
this is pretty cool, maybe you could add other sites like spotify, twitter, reddit etc for ease of access

src/plugins/ytdlp.desktop/index.tsx
TheColorman: The cancel button is tied to a message, which is currently the progress message. But yeah, this makes sense, it should probably just show an empty message with a cancel button if the show progress setting is disabled.

cheesesamwich:
> this is pretty cool, maybe you could add other sites like spotify, twitter, reddit etc for ease of access

nevermind, i just checkedout the yt-dlp support info and they already support a ton of shit, i assumed because of the name it was only youtube lmao

TheColorman:
> this is pretty cool, maybe you could add other sites like spotify, twitter, reddit etc for ease of access

Since it uses yt-dlp under the hood, it does support all of those sites. See [this](https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md) for the like 1500+ supported websites.

cheesesamwich:
> Since it uses yt-dlp under the hood, it does support all of those sites.

Yeah i noticed that as soon as you replied, you should definitely mention the url or maybe even have an expandable/searchable list of the supported sites in the plugin about

src/plugins/ytdlp.desktop/index.tsx
TheColorman: 094195a

TheColorman:
I've been thinking about renaming the plugin to something like "Video Downloader" to disconnect it a bit from the word "yt-dlp", as it's both more technical and only implies YouTube downloads. Thoughts?

Scyye:
> I've been thinking about renaming the plugin to something like "Video Downloader" to disconnect it a bit from the word "yt-dlp", as it's both more technical and only implies YouTube downloads. Thoughts?

"Media Embedder"?
"Media Sender"?
not sure, id not say "Video", as it supports more than video.

cheesesamwich:
> I've been thinking about renaming the plugin to something like "Video Downloader" to disconnect it a bit from the word "yt-dlp", as it's both more technical and only implies YouTube downloads. Thoughts?

I'd reccomend MediaDownloadCommand, or something similar. It's a really mistakable plugin so you want a good name

src/plugins/ytdlp.desktop/native.ts
```diff
@@ -0,0 +1,300 @@
...
+    return { format: format_string, videoTitle };
+}
+async function download({ format, videoTitle }: { format: string; videoTitle: string; }, { ytdlpArgs, url, format: usrFormat }: DownloadOptions) {
+    cleanVideoFiles();
+    const baseArgs = ["-f", format, "-o", "download.%(ext)s", "--force-overwrites", "-I", "1"];
+    const remuxArgs = ffmpegAvailable
+        ? usrFormat === "video"
+            ? ["--remux-video", "webm>webm/mp4"]
+            : usrFormat === "audio"
+                ? ["--remux-audio", "mp3"] // @ TODO THIS DOESNT EXIST
```
verticalsync, line 160:
```suggestion
                ? ["--extract-audio", "--audio-format", "mp3"]
```
you can replace it with this and it'll work, however it downloads the video, extracts the audio and then deletes the video

TheColorman: Oh crap, good catch. Either it was copilot or my brain fried, because `--remux-audio` is not a real flag lol.

verticalsync: > Oh crap, good catch. Either it was copilot or my brain fried, because `--remux-audio` is not a real flag lol.

it was mentioned as a suggestion in a github issue a while ago, perhaps thats why copilot suggested it

TheColorman: 67bf826

TheColorman:
Decided on MediaDownloader for the name, feel like it captures what it does pretty well. I know it doesn't mention the sending part, but I feel like it's implied.

TheColorman:
Maybe MediaUploader makes more sense as that's essentially the final goal of using the plugin 🤔

Scyye:
you should rename the PR to patch the plugin name

TheColorman:
One day...

ImLvna:
this can be accomplished with a user app

https://discord.com/application-directory/search?q=yt

TheColorman:
> this can be accomplished with a user app
> 
> https://discord.com/application-directory/search?q=yt

You could say that about a lot of the Vencord plugins. Almost all plugins that interact with the way chat messages work could be apps, from custom application commands to message encryption to textReplace. This just skips the third party and does everything on-device, which is both beneficial for privacy reasons, allows for bigger filesizes and allows more customization and post-processing as done with ffmpeg.

Vendicated:
no they are right, this is indeed best done via a bot. TextReplace is different because you want it passive without a command, but this plugin already is a command. so bot better

TheColorman:
dang

TheColorman:
> i love this plugin idea!

Initial enthusiasm was really the only reason I developed this past a proof of concept to begin with, since I assumed it was the green light for the idea.

Vendicated:
sorry about that, i didn't think about user apps being better. you can still turn this into a user app and you will have much more control over it cause you're not constrained to vencord limitations. put it into app directory and maybe it will become popular

PR #2484, opened 2024-05-19
ShowHiddenThings: more effectively explode Algolia filters

dolfies:
Current patch doesn't always work as there's multiple functions for fetching Algolia results, might be related to an experiment

### DIFF
--- a/src/plugins/showHiddenThings/index.ts
+++ b/src/plugins/showHiddenThings/index.ts
@@ -88,11 +88,11 @@ export default definePlugin({
             }
         },
         {
-            find: "\">200\"",
+            find: "MINIMUM_MEMBER_COUNT:",
             predicate: () => settings.store.disableDiscoveryFilters,
             replacement: {
-                match: /">200"/,
-                replace: "\">0\""
+                match: /MINIMUM_MEMBER_COUNT:function\(\)\{return \i}/,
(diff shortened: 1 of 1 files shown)


src/plugins/showHiddenThings/index.ts
```diff
@@ -80,11 +80,19 @@ export default definePlugin({
...
             predicate: () => settings.store.disableDiscoveryFilters,
             replacement: {
-                match: /filters:\i\.join\(" AND "\),facets:\[/,
-                replace: "facets:["
+                match: /\{"auto_removed:.*?\}/,
+                replace: "{}"
+            }
+        },
+        {
+            find: "\">200\"",
```
Nuckyz, line 91: Patch the getter for MINIMUM_MEMBER_COUNT instead. I know in the past I said it's bad to patch exports, but in this case it's better than targetting a nunber

```suggestion
            find: "MINIMUM_MEMBER_COUNT:",
```

Nuckyz (changes requested):

src/plugins/showHiddenThings/index.ts
dolfies: Done!

Vendicated:
thankies

src/plugins/showHiddenThings/index.ts
Vendicated: clueless

PR #2485, opened 2024-05-20
refactor(Webpack): Centralize and refactor Webpack patching

Nuckyz:
Centralizes Webpack patching around a single thing: Patching the module factories object of the WebpackRequire (`wreq.m`). This wraps the modules object with a proxy to intercept the addition of new modules, and allow us to decide what to do with them.

A new internal setting was added, `eagerPatches`, and it decides what to do with the newly added modules. With it enabled, we patch the factories as soon as they are `set`, and with it disabled we only patch them once they are accessed for the first time.

For correctly handling patching, because of multiple WebpackInstances, and to allow for patching factories only when they are accessed, our proxy defines getters and setters for each module factory added, instead of doing a simple Reflect.set on the module factories object.

Obtaining the main WebpackRequire instance is now also handled by the same way we patch `wreq.m`, as inside of the wreq.O setter we define logic to check if this `wreq` instance is the main one, and initialize our references if it is.

Additionally, this PR does a lot of clean up related to the main Webpack patching things. This includes changes to the reporter code (which was also modified to fit the removal of the beforeListeners API), variable names, stringifying of modules and QOL changes for better DevTools analyzing.

I also took the opportunity to correctly document all of the WebpackRequire instance, including full typings of it and docs of its internal methods (all of the ones present in the main Discord WebpackRequire). Our code utilizes the newer and correct typings now.

The patchedFactory wrapper is more future proof now and has a fallback logic if we fail to acquire the main WebpackRequire instance, for whatever the reason is. The first time it is called, it will attempt to use the value of the require argument of the factory as our internal reference to the WebpackRequire.

~~and yes, this completely removes the refactor I did 2 months ago~~


Nuckyz:
@CodiumAI-Agent /review

Nuckyz:
@CodiumAI-Agent /improve

Nuckyz:
@CodiumAI-Agent /improve

Nuckyz:
I updated the PR description to explain what changed

Nuckyz:
Superseded by https://github.com/Vendicated/Vencord/pull/2409

PR #2486, opened 2024-05-20
[New Plugin] TeX Renderer

dsnsgithub:
Allows TeX math equations to be sent using a slash command.

image
image

### DIFF
--- a/src/plugins/texRenderer/index.ts
+++ b/src/plugins/texRenderer/index.ts
@@ -36,7 +36,7 @@ type APIResponse = SuccessResponse | FailureResponse;
 
 export default definePlugin({
     name: "TeX Renderer",
-    description: "Allows you to send math equations using TeX.",
+    description: "Allows you to send math equations with slash commands using TeX (/tex).",
     authors: [Devs.DSNS],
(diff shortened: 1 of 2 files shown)


Golonchy:
noice

src/plugins/texRenderer/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+                const LATEX = `
+                \\documentclass{article}
+                \\begin{document}
+                $${equation}$
+                \\pagenumbering{gobble}
+                \\end{document}
+                `;
+
+                const res = await fetch("https://rtex.probablyaweb.site/api/v2", {
```
lewisakura, line 65: does this really need an external API? why can't it be rendered using KaTeX or something similar?

lewisakura (changes requested):
semi-duplicate of https://github.com/Vendicated/Vencord/pull/2298, so please read discussions from there

Vendicated:
will be closing for the same reason mentioned in 2298 and since this unnecessarily uses an external api. a user installed app can easily solve this use case

thank you for the contribution regardless!

PR #2487, opened 2024-05-20
MultiAccount Notification Indicator

pwall2222:


pwall2222:
So far I have implemented the most basic way to check notifications. But in theory everything appart that (Gateway, GUI) is done.

pwall2222:
The only thing missing is dealing with the new style of unread bubbles and I'm struggling to find much info on the:

`READY->notification_settings.flags`
`READY->user_guild_settings[id].flags`
`READY->user_guild_settings[id].channel_overrides[id].flags`

So I would appreciate the help!

pwall2222:
Okay so now the only thing missing to support would be `Mark as unread` (which is a severe PITA), along unread (without mention) markers at boot.

A nice feature to add would be to render which account is the one with the new ping/unread channel. But that's way out of my scope of knowledge.

pwall2222:
> connecting to the discord gateway like this is selfbotting. and doing it manually is extremely risky
Couldn't we use the builtin Discord Gateway code tho?

lewisakura:
I think it's mainly just the risk of not properly impersonating a Discord client leading to userbot suspicions. I think it's best leaving it out just because of that danger. Reusing the client gateway code may cause unprecedented bugs when trying to connect to two accounts at once.

Vendicated:
exactly

you can use one of these to connect to ur other accounts, listen to messages and then show desktop notifications whenever u receive a ping or dm. much much lighter than doing it inside discord and also safer cause these libraries are well crafted (well idk anything about the last one but it seems very popular so its probably decent quality)

https://github.com/diamondburned/arikawa
https://github.com/dolfies/discord.py-self
https://github.com/aiko-chan-ai/discord.js-selfbot-v13

qtlunya:
So, due to "concerns about selfbotting", you're suggesting people to use... selfbot libraries. 💀

qtlunya:
Yeah, my bad for not being clear. It's totally understandable for you to not want to include code like that in Vencord. It just came across as a bit self-contraditory how you said it's "extremely risky" (when in practice there's very little risk with gateway connections today) and yet went on to suggest people to do it anyway (just outside Vencord). But then again, Discord *could* in theory decide to suddenly start a crackdown overnight, and even if they don't, it's understandable that you want to keep the mod reasonably ToS-friendly. (Also, technically it's not *my* code but OP's, but that's irrelevant.)

But maybe this can be maintained as a third party plugin outside Vencord, because personally I'd much rather have it within the client rather than set up janky third party listeners with bad UX and make sure to keep them running at all times.

src/plugins/multiAccountNotifications/gateway.ts
```diff
@@ -0,0 +1,222 @@
...
+            d: {
+                token: this.token,
+                intents: 4609, // Intents for GUILDS, GUILD_MESSAGES, and DIRECT_MESSAGES
+                properties: {
+                    $os: 'windows',
+                    $browser: 'chrome',
+                    $device: 'desktop'
+                }
+            }
+        };
```
Vendicated, line 122: to give an immediate example of why you should not use your own solution: this identify payload is very wrong. user accounts do not use intents, they use capabilities. this is missing a shit ton of properties that a real client would send, e.g. presence, client state, referrer, os version, discord version, and much much more

thus this identify payload is very suspicious and could easily get your account flagged

here's a proper identify properties payload to show just how off this implementation is (very outdated please do not use):
```go
var WindowsProperties = IdentifyProperties{
	Os:                "Windows",
	OsVersion:         "10.0.19044",
	OsArch:            "x64",
	Browser:           "Discord Client",
	SystemLocale:      "en-US",
	ClientVersion:     "1.0.9006",
	ClientBuildNumber: 141635,
	ReleaseChannel:    "stable",
	ClientEventSource: nil,
}
```

PR #2488, opened 2024-05-20
fix(replaceGoogleSearch): trim LF character

Mylloon:
Trim LF character when doing a search


Vendicated:
what problem exactly does this solve?

Mylloon:
Sometimes when doing a search there is a `%0A` at the end

j4k0xb:
There can also be a new line at the start

> what problem exactly does this solve?

duckduckgo errors, most other search engines ignore it

lewisakura:
> There can also be a new line at the start

Shouldn't it be `.trim()` then?

Vendicated:

PR #2178, opened 2024-02-15
BetterActivities plugin

D3SOX:
Closes https://github.com/Vencord/plugin-requests/issues/41
Closes https://github.com/Vencord/plugin-requests/issues/258

# Preview

image
image
image

# Settings

image

PR #2491, opened 2024-05-21
Refactor FakeNitro

D3SOX:
Let me know what you think about this approach

And yes I did this all because of [this comment](https://github.com/Vendicated/Vencord/pull/2178#discussion_r1605246459)
image


LZeugirdor:
Not sure this is the right place to ask, but are soundboard sounds server side or client side, and if they are client side, could fakenitro be expanded to support them?

schoeneBiene:
> Not sure this is the right place to ask, but are soundboard sounds server side or client side, and if they are client side, could fakenitro be expanded to support them?

No, to all of those questions.

D3SOX:
I assume I wasted my time here as FakeNitro had a lot of changes recently which would require me to go over this agin

PR #2492, opened 2024-05-21
[Plugin] fakeBlocked

henmalib:
This plugin allows to completely hide all user messages from a chat.

I don't like the idea of a query after removal or adding to a list, but without updating (like hovering, for example), the message wouldn't be updated and hidden, but patching works well for new messages :3

Some users have requested this plugin previously: https://github.com/Vencord/plugin-requests/issues/466 https://github.com/Vencord/plugin-requests/issues/484, so it might be really useful for someone.


src/plugins/hideMessages/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    settings,
+    contextMenus: {
+        "user-context": userProfileContextPatch,
+        "user-profile-actions": userProfileContextPatch
+    },
+    patches: [
+        {
+            find: ".messageListItem",
+            replacement: {
+                match: /(.{1,30})\.messageListItem,/,
```
schoeneBiene, line 71: You should probably use \i for matching identifiers
```suggestion
                match: /(\i)\.messageListItem,/,
```

schoeneBiene (commented):
Plugins need a README.md

src/plugins/hideMessages/index.tsx
henmalib: Is it better now?

src/plugins/hideMessages/README.md
```diff
@@ -0,0 +1,7 @@
+# HideMessages
+
+Allow you to hide annoying messages without blocking their author
+
+![Before](https://github.com/henmalib/Vencord/assets/68553709/6b74018a-77c9-4937-b504-a70b2cee1d65)
+![After](https://github.com/henmalib/Vencord/assets/68553709/59ec88d7-f9a4-4b07-8775-db2fb67a8ad2)
```
schoeneBiene, line 6: These assets should probably be on the main repo.

schoeneBiene, line 6: It's not really descriptive what these mean, maybe something like
```suggestion
## Before
![Before](https://github.com/henmalib/Vencord/assets/68553709/6b74018a-77c9-4937-b504-a70b2cee1d65)
## After
![After](https://github.com/henmalib/Vencord/assets/68553709/59ec88d7-f9a4-4b07-8775-db2fb67a8ad2)
```
?

henmalib:
Honestly this thing needs a big rewrite, especially after a doomsday and I'm too lazy to do so

[There is a much more better version of this made by Samwich](https://discord.com/channels/1015060230222131221/1032200195582197831/1247901460419711087), if they want to, they'll make a pr which will be much better as this, so I'll better close this one

PR #2493, opened 2024-05-22
Fix Alt + F4 on QuickCSS window closing all windows

fres621:
I think the expected behaviour should be that only the QuickCSS window gets closed when you press Alt + F4 while it is focused.

**Old behaviour:**
 Alt + F4 while the Quick CSS window is focused closes both the Quick CSS window and the main Discord window.
 **New behaviour:**
 Alt + F4 while the Quick CSS window is focused will only close the Quick CSS window.

### DIFF
--- a/src/main/ipcMain.ts
+++ b/src/main/ipcMain.ts
@@ -138,6 +138,12 @@ ipcMain.handle(IpcEvents.OPEN_MONACO_EDITOR, async () => {
             sandbox: false
         }
     });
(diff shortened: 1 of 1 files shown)


Vendicated:
hmm this doesn't seem like the correct solution.. especially because the shortcut is different on linux & mac

Vendicated:
converting to issue since you yourself said this PR was mainly to bring light to the issue https://github.com/Vendicated/Vencord/issues/2548

PR #2494, opened 2024-05-22
Update React error docs in reactErrorDecoder

cloudskies:
Umm maybe I did it wrong I'm unsure


lewisakura:
this doesn't actually work- react.dev stopped using ?invariant, you just put the error number in the path (react.dev/errors/1 for example)

Vendicated:
image

image

PR #2496, opened 2024-05-22
feat(plugin): WakaTime submodule

BaalEvan:
I have added a submodule to the user plugin officially supported by WakaTime, which was created by NeonGamerBot-QK.

I don't think this should require a DevBuild to install and will be much more accessible when added to the main repository. (I'm not affiliated in any way with this plugin or WakaTime. I'm only submitting a submodule pull request.)

Related: Plugin Request https://github.com/Vencord/plugin-requests/issues/369

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/plugins/wakaTime"]
+	path = src/plugins/wakaTime
+	url = https://github.com/wakatime/vencord-wakatime.git


Vendicated:
hi, thansk for your pr! we don't vendor plugins like this, every plugin should be first party

out of respect i would also not merge it unless the original author desires it

thus i'll be rejecting this, sorry!

alanhamlett:
@Vendicated could we re-open this? I'm the owner of [vencord-wakatime](https://github.com/wakatime/vencord-wakatime) so that should satisfy the first party requirement? Thanks :)

PR #2498, opened 2024-05-22
QuickStar plugin

williamistGitHub:
This plugin adds a new button to the message popover that allows you to quickly "star" react a message:
image

This is intended to be used with starboard bots, but the emoji that is reacted is customizable per-server so you could use it for other things as well:
image
image

I'm quite interested in seeing what you think of this.

### DIFF
--- a/src/plugins/quickStar/index.tsx
+++ b/src/plugins/quickStar/index.tsx
@@ -1,0 +1,176 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2024 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { NavContextMenuPatchCallback } from "@api/ContextMenu";
+import { DataStore } from "@api/index";
+import { addButton, removeButton } from "@api/MessagePopover";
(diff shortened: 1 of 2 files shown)


src/plugins/quickStar/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    ReactionManager.addReaction(
+                        message.channel_id,
+                        message.id,
+                        await emojiForServer(channel.guild_id)
+                    );
+                }
+            };
+        });
+
+        // there's no easy way to detect when the user has left a server (without patches, at least)
```
MrDiamondDog, line 159: I personally wouldn't do this, if you rejoin the server i would expect to have the same settings

MrDiamondDog (commented):
:star:

src/plugins/quickStar/index.tsx
williamistGitHub: Ah, that makes sense. I just saw the "clean up no longer used entries" in the comment for the datastore and figured it was a good idea. I suppose it's one big entry anyway so that doesn't even make much sense to do. I can remove that if you think it'll be better that way.

Vendicated: that's mainly if you're storing large amount of data. in your case its less than a kb per server so its not a big deal

williamistGitHub: Alright, I will remove it. (and also some `console.log`s i left in the code oops)

Scyye:
I'd change the name to `QuickReact` or something similar, but thats just preference.

TheKodeToad:
if you favourite the star it already shows up in the *context menu*

Sqaaakoi:
> I'd change the name to `QuickReact` or something similar, but thats just preference.

This isn't the first time I've said this, but..
1: This isn't the first time someone's tried to make this plugin, iirc @nin0-dev made something like this before
2: The context menu reaction thing is already named quick react (you can see the context menu ID for proof) and as I've made a plugin to improve that (it's stuck in PR hell btw), users would get confused with "QuickReact" and "BetterQuickReact" plugins very quickly.

williamistGitHub:
Yeah, the reason it's named QuickStar is because it's intended for starboards. The only reason you can customize what emoji it uses per server is because one server I'm in doesn't use the default star emoji, and I figure that might be true in some other servers as well.

Vendicated:
too niche, just add star to your favourites

PR #2502, opened 2024-05-24
ctrlEnterSend: use cmd+enter on macOS

lumap:
On macOS, the ctrl key is generally used as a secondary alt key for shortcuts. The cmd (⌘) key is the "main" modifier key, the equivalent of ctrl on windows/linux.

### DIFF
--- a/src/plugins/ctrlEnterSend/index.ts
+++ b/src/plugins/ctrlEnterSend/index.ts
@@ -18,7 +18,7 @@ export default definePlugin({
             type: OptionType.SELECT,
             options: [
                 {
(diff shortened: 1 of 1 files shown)


src/plugins/ctrlEnterSend/index.ts
```diff
@@ -54,7 +54,7 @@ export default definePlugin({
                 result = event.shiftKey;
                 break;
             case "ctrl+enter":
-                result = event.ctrlKey;
+                result = process.platform === "darwin" ? event.metaKey : event.ctrlKey;
```
Vendicated, line 57: we're in the browser not nodejs

Vendicated:

PR #2504, opened 2024-05-25
fix(MessageLatency): ignore bots

rushiiMachine:
I've only ever seen bots send messages without a nonce or an incorrect one :trolley:


qtlunya:
Yeah, or selfbots, but well, not much you can do about that, aside from maybe checking if it's a valid snowflake.

Vendicated:

PR #2508, opened 2024-05-26
Plugin: animatedBackgrounds

soulfireneedusername:
Allows you to set a YoutTube video as an animated background. I recommend using a long loop Video. A background supporting theme must be installed. Just paste any YouTube Video URL in the link Input in the plugin settings.

I looked and there should be no other similar request. I hope, its not because its a dumb thing. (this will be so rejected) 
Plugin: animatedBackgrounds  Issue #628

### DIFF
--- a/src/plugins/animatedBackgrounds/index.ts
+++ b/src/plugins/animatedBackgrounds/index.ts
@@ -1,0 +1,68 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Devs } from "@utils/constants";
+import { OptionType } from "@utils/types";
+import definePlugin from "@utils/types";
+import './styles.css';
+
+export default definePlugin({
+    name: "AnimatedBackgrounds",
+    description: "Allows you to use a YouTube video URL as an animated background.",
+    authors: [Devs.soul_fire_],
+    patches: [],
+    options: {
+        source: {
+            description: "Source URL to replace the background (a background-supporting theme must be installed first)",
+            type: OptionType.STRING,
+            restartNeeded: true,
+        }
+    },
+    start() {
+        const appMount = document.querySelector('#app-mount.appMount_c99875');
+        if (appMount) {
+            const source = Vencord.Settings.plugins.AnimatedBackgrounds.source;
+            try {
+                const url = new URL(source);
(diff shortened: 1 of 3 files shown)


src/plugins/animatedBackgrounds/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    patches: [],
+    options: {
+        source: {
+            description: "Source URL to replace the background (a background-supporting theme must be installed first)",
+            type: OptionType.STRING,
+            restartNeeded: true,
+        }
+    },
+    start() {
+        const appMount = document.querySelector('#app-mount.appMount_c99875');
```
MrDiamondDog, line 37: Don't use dom manipulation, use patches

src/plugins/animatedBackgrounds/index.ts
```diff
@@ -0,0 +1,68 @@
...
+                const iframeSrc = `https://www.youtube.com/embed/${videoId}?autoplay=1&mute=1&controls=0&showinfo=0&autohide=1&loop=1&playlist=${videoId}&vq=highres`;
+                const iframe = this.createIframe(iframeSrc);
+                appMount.style.background = 'none';
+                appMount.appendChild(iframe);
+            } catch (error) {
+            }
+        }
+    },
+    createIframe(src) {
+        const container = document.createElement('div');
```
MrDiamondDog, line 57: Again no dom manipulation and use a react component for this

src/plugins/animatedBackgrounds/index.ts
```diff
@@ -0,0 +1,68 @@
...
+import './styles.css';
+
+export default definePlugin({
+    name: "AnimatedBackgrounds",
+    description: "Allows you to use a YouTube video URL as an animated background.",
+    authors: [Devs.soul_fire_],
+    patches: [],
+    options: {
+        source: {
+            description: "Source URL to replace the background (a background-supporting theme must be installed first)",
```
MrDiamondDog, line 31:
```suggestion
            description: "Source URL to replace the background",
```

I would use a settings about component to say this, and provide an basic example theme

src/plugins/animatedBackgrounds/index.ts
```diff
@@ -0,0 +1,68 @@
+/*
```
MrDiamondDog, line 1: Use new header

MrDiamondDog (changes requested):

src/plugins/animatedBackgrounds/index.ts
soulfireneedusername: I've tried for a very long time now. I searched the modules with appMount and attempted to patch it there, but it didn't work. Additionally, I couldn't find any plugin that does something similar, such as patching or appending elements to existing Discord elements.

src/plugins/animatedBackgrounds/index.ts
Vendicated: this is actually a rare case where dom manipulation is acceptable

Vendicated:
hi thanks for your pr!

i don't really understand the purpose of it however.. why not just download the youtube video, convert it to an animated webp/avif (for minimal file size) and use that from css?

soulfireneedusername:
Because it works quite well and is significantly easier. Additionally, it should prevent looping issues. Also, 10-hour loops wouldn't be easy to download, and this method saves many steps. Anyway, do you think I should still convert it, or add it as an additional option?

src/plugins/animatedBackgrounds/index.ts
soulfireneedusername: Do I still need to include the start function in patches?

Vendicated:
closing as users should just use css for this

PR #2510, opened 2024-05-26
Plugin: DisableDiscordRestart

burlone0:
Disables the Ctrl + R shortcut to prevent Discord from restarting.

### DIFF
--- a/src/plugins/disableDiscordRestart/index.tsx
+++ b/src/plugins/disableDiscordRestart/index.tsx
@@ -1,0 +1,87 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { definePluginSettings } from "@api/Settings";
+import definePlugin, { OptionType } from "@utils/types";
+import { Devs } from "@utils/constants";
+
+import './styles.css';
+
+const settings = definePluginSettings({
+    showAlert: {
+        type: OptionType.BOOLEAN,
+        default: true,
+        description: "Show alert when Ctrl + R is disabled",
+        restartNeeded: true,
+    },
+});
+
+const disableRestartShortcut = (event) => {
+    if (event.ctrlKey && event.key === 'r') {
+        event.preventDefault();
+        event.stopPropagation();
+        console.log("Ctrl + R shortcut disabled");
+        if (settings.store.showAlert) {
+            showAlertPopup("Ctrl + R shortcut disabled", "DisableDiscordRestart - Made by Burlone");
+        }
+    }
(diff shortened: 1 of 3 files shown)


src/plugins/disableDiscordRestart/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        event.stopPropagation();
+        console.log("Ctrl + R shortcut disabled");
+        if (settings.store.showAlert) {
+            showAlertPopup("Ctrl + R shortcut disabled", "DisableDiscordRestart - Made by Burlone");
+        }
+    }
+};
+
+const showAlertPopup = (message, title) => {
+    const modal = document.createElement('div');
```
MrDiamondDog, line 46: Don't use dom manipulation

src/plugins/disableDiscordRestart/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+};
+
+const stopPlugin = () => {
+    document.removeEventListener('keydown', disableRestartShortcut);
+};
+
+export default definePlugin({
+    name: "DisableDiscordRestart",
+    authors: [Devs.Burlone],
+    description: "Disables the Ctrl + R shortcut to prevent Discord from restarting.",
```
MrDiamondDog, line 82: The plugin makes a pop up instead of fully disables the keybind. The description should reflect that

MrDiamondDog (commented):
I don't really see a use for this plugin but I'll leave this here

src/plugins/disableDiscordRestart/index.tsx
burlone0: That's the general description, that's what the showAlert setting description says

src/plugins/disableDiscordRestart/index.tsx
burlone0: wdym

burlone0: What do you mean? What should I use?

Inbestigator (commented):
This seems pretty useless, and the execution feels kind of clunky.

src/plugins/disableDiscordRestart/index.tsx
Scyye: react i believe

Scyye (commented):
This doesnt seem useful in the slightest; what's the usecase?

nyakowint:
If anything this can just be an addition to the NoF1 plugin

nyakowint (changes requested):
You should get rid of all this dom manipulation (which u shouldn't be using anyways) and replace it with a simple toast or modal (in webpack commons)

src/plugins/disableDiscordRestart/index.tsx
MrDiamondDog: DOM manipulation is bad when you have patches and react components at your disposal. Have a look at some other plugins for reference

src/plugins/disableDiscordRestart/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+    },
+});
+
+const disableRestartShortcut = (event) => {
+    if (event.ctrlKey && event.key === 'r') {
+        event.preventDefault();
+        event.stopPropagation();
+        console.log("Ctrl + R shortcut disabled");
+        if (settings.store.showAlert) {
+            showAlertPopup("Ctrl + R shortcut disabled", "DisableDiscordRestart - Made by Burlone");
```
schoeneBiene, line 40: does your name really need to be here?

Inbestigator: Yeah, shoving “made by …” down people’s throats feels icky. Also, tbh the whole title for that modal could be removed.

Vendicated:
this is way too specific and not very useful in its current state, so i'll be rejecting this. have you tried adding `ctrl + r` as unbinded in discord keybind settings instead? might work

thanks for your contribution nevertheless

PR #2513, opened 2024-05-27
Create smallSpeak

jsh3d:


Vendicated:
TextReplace can do this if you really want it. But it's also just not very useful and bad for accessibility

So I'll be rejecting this, sorry

jsh3d:
> TextReplace can do this if you really want it. But it's also just not very useful and bad for accessibility
> 
> So I'll be rejecting this, sorry

ok 👍

PR #1414, opened 2023-07-13
delete uwuifier plugin

exhq:
**goodbye. you will be missed.  
you may have a few questions**

# why?
uwuifier was a useless and a meme plugin. it was never meant to be serious. with vencord getting more and more mainstream and meme/useless plugins now having their own place (#third-party-plugins channel on discord) 
Discord_03OTu2SyV5
# what happens next?
i will 100% work on either a textreplace rule or a third party plugin (hopefully better than the current uwuifier). do not worry. i dont want to completely destroy uwuifier, it just seems like the correct thing to do
# now what?
continue uwuing manually and i'll eventually make a comeback, i promise :3

PR #2516, opened 2024-05-28
Plugin: UwUify

WackyModer:
It adds the /uwuify command that just makes your message more "cute" based off of a long word and letter replacement list and some suffixes.

It does not have any super special characters in the suffixes or replacements so it shouldn't get automodded in some servers with those "English only" automods.

### What it looks like:
UwUify in action

### Some options for it:

(sorry for it being so big btw)

Settings menu with some options for the UwUify plugin

### DIFF
--- a/src/plugins/dearrow/index.tsx
+++ b/src/plugins/dearrow/index.tsx
@@ -182,8 +182,8 @@ export default definePlugin({
 
             // add dearrow button
             {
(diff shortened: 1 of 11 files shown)


Nuckyz:
Vencord already had a uwuify plugin in the past and it was removed, see https://github.com/Vendicated/Vencord/pull/1414. This will not be accepted.

Sorry!

Nuckyz:
no problem! thanks for the efforts either way!

PR #2517, opened 2024-05-28
ezCalculate

jsh3d:
Its just a simple way to calculate stuff via Discord Commands. If this gets accepted, I'll add a lot more calculations.


lewisakura:
This plugin has multiple security issues with its math implementation (which is basically just `eval`) and is very niche. I don't think this is a good idea to be included in the mod. Thank you, though!

Vendicated:
you don't need a client mod for slash commands. you can use user installabe apps, they work anywhere on discord (including mobile)

image

PR #2518, opened 2024-05-28
cleanerChannels

jsh3d:
Removes channel icons to give discord a more simple look. Will add customization and a readme.md soon, this is just to see if this will get added.

### DIFF
--- a/src/plugins/cleanerChannels/index.ts
+++ b/src/plugins/cleanerChannels/index.ts
@@ -1,0 +1,34 @@
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "Remove Channel Icons",
+    description: "Removes the # symbol next to text channels, the blowhorn icon next to announcement channels, and the rules icon.",
+    authors: [Devs.jsh4d],
+    patches: [
+        {
+            find: /iconHash:\{[^}]*icon:/,
+            replacement: () => "",
+            match: /iconHash:\{[^}]*icon:"#"/,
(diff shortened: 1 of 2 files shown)


jsh3d:
yo can someone close this and tell me why or add it

src/plugins/cleanerChannels/index.ts
```diff
@@ -0,0 +1,34 @@
...
+            replace: "iconBlowhorn:{}",
+        },
+        {
+            find: /iconRules:\{[^}]*icon:/,
+            replacement: () => "",
+            match: /iconRules:\{[^}]*icon:"rules"/,
+            replace: "iconRules:{}",
+        },
+    ],
+    start() {
```
Sqaaakoi, line 28: ?????? so useless these aren't mandatory to have at all

Sqaaakoi: also yes like others said just use CSS

jsh3d:
> if you want to close a pr you can do it yourself. just click here:
> 
> 
> 
> this likely won't be merged since it can be done using a CSS snippet, which doesn't require any plugin. general "policy" with vencord plugins are that they shouldn't be made if you can do it with CSS

ik i just need a response from the admins or janitors

MrDiamondDog:
> ik i just need a response from the admins or janitors

If bet they're going to say the same thing as we did

Vendicated:
as others have stated, this can easily be achieved with css, so a plugin is not necessary for this

thank you for your contribution nevertheless!

jsh3d:
👍

PR #2525, opened 2024-05-31
Add autoSender plugin

NcxyZero:
new file:   src/userplugins/autoSender/index.ts

It's a simple plugin that automates message sending. Provide the channel id, desired delay and the message in plugin's settings and enable it


programminglaboratorys:
there is a reason src/userplugins is in gitignore so ppl don't accidently upload their userplugins. and automate? isn't that the main definition of a botting (aka selfbotting)

NcxyZero:
> there is a reason src/userplugins is in gitignore so ppl don't accidently upload their userplugins. and automate? isn't that the main definition of a botting (aka selfbotting)

That makes sense but I don't know if this is a problem when as stated in README.md in disclaimer section that even vencord is against tos of discord

src/plugins/autoSender/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export default definePlugin({
+    name: "Auto Sender",
+    description: "Send messages automaticaly to specifed channel",
+    authors: [
+        {
+            id: 322076647589412864n,
+            name: "Noxy",
+        },
+    ],
```
nin0-dev, line 79: Put yourself in the constants.ts file

nin0-dev (changes requested):
For the sake of it I reviewed but don't expect this to be merged, that will likely get you terminated. 
That's just a glorified selfbot

nin0-dev (commented):
Since I forgot some other comments here they are, take that as general advice if you ever want to make other plugins

k26pl:
> That makes sense but I don't know if this is a problem when as stated in README.md in disclaimer section that even vencord is against tos of discord
Discord doesnt care about mods, but selfbots are bannable

NcxyZero:
> > That makes sense but I don't know if this is a problem when as stated in README.md in disclaimer section that even vencord is against tos of discord
> 
> Discord doesnt care about mods, but selfbots are bannable

In that case I think I'm closing this

PR #2526, opened 2024-05-31
feat(plugin): VirusChecker

Jannniss:
A small plugin that adds a button to attachments, allowing users to scan files via Hybrid Analysis.


verticalsync:
shouldnt this be marked as a desktop only plugin since it uses native

Vendicated:
thank you for your pr!

this is not very useful, especially considering users have to manually enter an api key. if you're downloading exes off discord, you're doing it wrong anyway so there's really no reason one should need this

as such i'll be rejecting this, sorry

PR #2528, opened 2024-06-01
feat: translation v2

lewisakura:
This is an updated version of the old translation system, now using JSON files instead of Fluent.

This also includes i18n Ally as a plugin recommendation to make it easier to translate Vencord.


lewisakura:
why on earth does this include 42307cc??

lewisakura:
todo:
- [ ] add pluralisation via [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules)
- [ ] convert all core mod strings to localized forms
- [ ] (optional) convert all plugin strings to localized forms
  - optional task because we need to work out how to localize plugin descriptions (not names, since they have to be preserved), most likely will just be to change description to point to a localization key so it's backwards compatible ([$t returns the string key if it's not found](https://github.com/Vendicated/Vencord/blob/c0111169b890aee4610f6ce69ba598e2150fe79b/src/utils/translation.ts#L91))

lewisakura:
blocked on translation component (requires ability to interpolate components into i18n strings, similar to react-i18next but simpler)

how should it work?
```tsx
const embeddedComponent = <Link to="/def">another link</Link>;

<Translate i18nKey="vencord.example" variables={{ cutie: "vennie" }}>
    {{cutie}} is so cute! Here's a <Link to="/abc">link</Link> and {embeddedComponent}!
</Translate>
```
or
```tsx
<Translate i18nKey="vencord.example" variables={{ cutie: "vennie" }}>
    <Link to="/abc" />
    {embeddedComponent}
</Translate>
```

lewisakura:
Putting this up for review. Please disregard any missing strings in the translation, they will be done eventually. Reviews should be primarily focused on the implementation of translation itself. We will be moving translation files to `.asar` files eventually, so do not worry about the ESBuild plugin or anything relating to the retrieval of translations.

lumap:
bored, will translate to fr*nch once this is merged

lewisakura:
superseded by #3688

PR #2531, opened 2024-06-01
feat(ClearURLs): Add whitelist functionality

mekdimd:
Added a whitelist feature to the ClearURLs plugin. Useful when some links have URL parameters that conflict with the default rules. Pretty simple feature, <20 lines added.

A good example is some **Google Calendar links**
- https://calendar.google.com/calendar/u/0?cid=veryimportantcalendarid. 
- In this context, cid is a Calendar ID, and not a tracking parameter. But cid is also a common tracking parameter, so it's included in the default rules.


mekdimd:
I'm open to any criticism! This is my first contribution.

src/plugins/clearURLs/index.ts
```diff
@@ -104,6 +114,13 @@ export default definePlugin({
             return match;
         }
 
+        // Check if the current URL is whitelisted
+        const { whitelist } = settings.store;
+        const whiteListSplit = whitelist.split(",").map(item => item.trim());
+        if (whiteListSplit.some((hostname: string) => url.hostname.includes(hostname))) {
```
Vendicated, line 120: i dont think this `includes` makes sense

if you were to whitelist `x.com`, it might affect many other urls

mekdimd: Thanks for the comment :) I have changed it to a direct comparison

mekdimd:
Btw for future reference, should I be merging into the dev branch instead of main? The contribution guide didn't specify

thororen1234:
> Btw for future reference, should I be merging into the dev branch instead of main? The contribution guide didn't specify

Yes dev is the branch you want.

PR #2532, opened 2024-06-01
Add a new translate command.

anotherkai:
This pull request introduces a new `translate` command to the moreCommands plugin. It uses mymemory's (by translated labs) API to retrieve text and either send to the current channel or "preview" the result. There is currently a 50k character limit per day on this due to the API being free.

### DIFF
--- a/src/plugins/moreCommands/index.ts
+++ b/src/plugins/moreCommands/index.ts
@@ -20,6 +20,38 @@ import { ApplicationCommandInputType, findOption, OptionalMessageOption, Require
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
 
+const settings = definePluginSettings({
+    show_match: {
+        type: OptionType.BOOLEAN,
+        description: "Shows a match number (like 0.85) to other sources.",
+        default: false,
+        restartNeeded: false,
+    },
+});
+
+const translateText = async (sourceLang, targetLang, text) => {
+    try {
+        const url = `https://api.mymemory.translated.net/get?q=${text}&langpair=${sourceLang}|${targetLang}&de=translated-abuse@littlekai.co.uk`;
+        const response = await fetch(url);
+        const jsonResponse = await response.json();
+        
+        if (jsonResponse.responseStatus === 200) {
+            const translatedText = jsonResponse.responseData.translatedText;
+            const quotaFinished = jsonResponse.quotaFinished;
+            const match = jsonResponse.responseData.match || 0;
+
+            return {
+                translated: translatedText,
+                quotaFinished: quotaFinished,
+                match: match
+            };
+        } else {
+            throw new Error('Translation failed');
+        }
+    } catch (error) {
+        throw new Error(`Error: ${error}`);
+    }
+};
 
 function mock(input: string): string {
     let output = "";
@@ -34,6 +66,7 @@ export default definePlugin({
     description: "echo, lenny, mock",
     authors: [Devs.Arjix, Devs.echo, Devs.Samu],
     dependencies: ["CommandsAPI"],
+    settings.
     commands: [
         {
             name: "echo",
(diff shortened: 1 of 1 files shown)


MrDiamondDog:
This should really be apart of the translate plugin, not this one

anotherkai:
> This should really be apart of the translate plugin, not this one

See I would if I knew how to work with Modals, I'm only just learning how to write functional plugins so for the meanwhile, I will be trying to get this added the moreCommands plugin.

src/plugins/moreCommands/index.ts
```diff
@@ -18,10 +18,43 @@
...
+        } else {
+            throw new Error('Translation failed');
+        }
+    } catch (error) {
+        throw new Error(`Error: ${error.message}`);
+    }
+};
 
-function mock(input: string): string {
+function mock(input) {
```
MrDiamondDog, line 57: why

src/plugins/moreCommands/index.ts
```diff
@@ -18,10 +18,43 @@
 
 import { ApplicationCommandInputType, findOption, OptionalMessageOption, RequiredMessageOption, sendBotMessage } from "@api/Commands";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import { definePluginSettings } from "@api/Settings";
+import definePlugin, { definePluginSettings, OptionType } from "@utils/types";
```
MrDiamondDog, line 22: ?

src/plugins/moreCommands/index.ts
```diff
@@ -18,10 +18,43 @@
...
 
+const settings = definePluginSettings({
+    show_match: {
+        type: OptionType.BOOLEAN,
+        description: "Shows a match number (like 0.85) to other sources.",
+        default: false,
+        restartNeeded: false,
+    },
```
MrDiamondDog, line 30: What does this option do? the description doesn't make sense

MrDiamondDog (changes requested):
The reason I suggested it be moved is because then you can use the same translating api as the translate plugin, which uses google translate and thus does not have a quota

src/plugins/moreCommands/index.ts
anotherkai: I actually didn't know I did this :sob:

Vendicated:
yeah u should add this as a command to the Translate plugin instead, it already has a translate function so you just need to add the command

src/plugins/moreCommands/index.ts
```diff
@@ -65,9 +57,8 @@ function mock(input) {
 export default definePlugin({
     name: "MoreCommands",
     description: "echo, lenny, mock, translate",
-    authors: [Devs.Arjix, Devs.echo, Devs.Samu],
+    authors: [Devs.Arjix, Devs.echo, Devs.Samu, {name: "Kai :3", id: 1205727693811879997n}],
```
Scyye, line 60: Add yourself to the devs constant

src/plugins/moreCommands/index.ts
```diff
@@ -54,7 +46,7 @@ const translateText = async (sourceLang, targetLang, text) => {
     }
 };
 
-function mock(input) {
+function mock(input: string): string {
```
Scyye, line 49: I- why??

Scyye (changes requested):
This should 100% not be in moreCommands. Just move it over to the translation plugin.

src/plugins/moreCommands/index.ts
anotherkai: Did I seriously forget to symbol link :-:

anotherkai: i did :facepalm:

anotherkai:
Closed as I will be moving it over to translate plugin, will make a new pr soon.

PR #2533, opened 2024-06-02
new plugin NoOnboardingDelay: skip long onboarding animations

nekohaxx:
NoOnboardingDelay removes the slow and annoying animation from the onboarding screen


Inbestigator:
Can be disabled with experiments

Nuckyz:
what's the experiment

src/plugins/noOnboardingDelay/index.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
```
Vendicated, line 3: you might want to fix ur calendar btw

PR #2535, opened 2024-06-02
viewIcons: support new simplified profile

AutumnVN:


Sqaaakoi (changes requested):
duplicates code; I ~~would~~ have personally done an array map for the find property like  `...[].map(find => ({`

Sqaaakoi:
> duplicates code; I ~would~ have personally done an array map for the find property like `...[].map(find => ({`

https://github.com/AutumnVN/Vencord/pull/1  I have tried to fix that for you but I also found another issue that needs to be fixed

PR #2540, opened 2024-06-03
MutualGroupDMs: Simplified profiles support

Sqaaakoi:
Screenshot of the Mutual Groups tab in the new simplified profiles UI


src/plugins/mutualGroupDMs/index.tsx
```diff
@@ -58,9 +59,32 @@ export default definePlugin({
...
+                },
+                {
+                    match: /(?<=(\i)===\i\.UserProfileSections\.MUTUAL_GUILDS?.{0,150}\}\):)/,
+                    replace: '$1==="MUTUAL_GDMS"?$self.renderMutualGDMs(arguments[0]):'
+                },
+            ]
+        },
     ],
 
+    useGDMCount(userId: string) {
```
Vendicated, line 83: i think making this a hook is fairly pointless

it's not that important to update the count if a new dm gets added and just adds extra complexity. you save yourself one patch if you just don't

src/plugins/mutualGroupDMs/index.tsx
```diff
@@ -58,9 +59,32 @@ export default definePlugin({
...
+                {
+                    match: /(?<=let\{user:(\i),.{0,700})(?=return\(null!=)/,
+                    replace: "let vencordMutualGroupsCount=$self.useGDMCount($1.id);"
+                },
+                {
+                    match: /(?<=(\i\.push)\(\{section:\i\.UserProfileSections\.MUTUAL_GUILDS,text:.{0,250}\}\)\)\}\))/,
+                    replace: ',$1({section:"MUTUAL_GDMS",text:vencordMutualGroupsCount})'
+                },
+                {
+                    match: /(?<=(\i)===\i\.UserProfileSections\.MUTUAL_GUILDS?.{0,150}\}\):)/,
```
Vendicated, line 76: all of these matches seem like they could cause slow runtime due to the lookbehinds

please check the Tracer timings to see if they are slow (anything under ~2ms) and if yes try to fix them, eg by moving lookbehinds after the normal match or not using lookbehind

Vendicated (commented):
thanks for the fix!

src/plugins/mutualGroupDMs/index.tsx
Sqaaakoi: i've finally gotten around to this, i've tested the patches in power saving mode and the first patch ` /(?<=let\{user:(\i),.{0,700})(?=return\(null!=)/` is super slow (14.5ms)
I will fix this; do you think the `.{0,700}` is the issue?

Sqaaakoi: ok I fixed the lag by just splitting it into 2 patches where the first one sets a variable, is this an appropriate solution?

src/plugins/mutualGroupDMs/index.tsx
Sqaaakoi: it looks like all the other tabs use a hook to calls useStateFromStores to their respective stores anyway

we have 4 options
- leave as-is witb 2 patches
- replace my new hacky variable with the hook variable (i will try this now)
~~- move hook into a condition that is never changed (break the rules of hooks?)~~ obsoleted idea while I was typing this comment
- remove hook

src/plugins/mutualGroupDMs/index.tsx
Sqaaakoi: forget that see other comment thread

Sqaaakoi (commented):
Sqaaakoi-09-06-2024-15-46-59
how will we deal with this? it's starting to clip off the right side of the tab bar, and it will get even worse when ReviewDB (maybe even PermissionsViewer too?) gets updated to work with the simplified profiles

Vendicated:
superseded

PR #2543, opened 2024-06-03
SpotifyControls: fix no artists on local files

notsuu:
Because artists array filters out ID-less artist objects, which local files just so happen to return

| before | after |
| --- | --- |
| who is this by bro | thanaks now i know |


src/plugins/spotifyControls/index.tsx
```diff
@@ -77,6 +77,13 @@ export default definePlugin({
                 match: /repeat:"off"!==(.{1,3}),/,
                 replace: "actual_repeat:$1,$&"
             }
+        },
+        {
+            find: 'P.SpotifyResourceTypes.TRACK',
```
schoeneBiene, line 82: Don't hardcode variable names in your find.

src/plugins/spotifyControls/index.tsx
```diff
@@ -77,6 +77,13 @@ export default definePlugin({
...
+            replacement: {
+                match: /\(0,(.{1,2})\.isNotNullish\)\((.{1,2})\.id\)&&/,
```
schoeneBiene, line 84: Use \i for matching identifiers

```suggestion
                match: /\(0,(\i)\.isNotNullish\)\((\i)\.id\)&&/,
```

Vendicated:
looks good, thanks for the fix!

PR #2544, opened 2024-06-03
formatting changes

sketchycrypt:
formatting changes to the artist and track commands for the spotifyShareCommands plugin


Vendicated:
thanks for your pr! but this is just worse imo. also what if the track/album name contains ]? it will break the link

Sqaaakoi:
> thanks for your pr! but this is just worse imo.

couldn't this be configurable?

PR #2547, opened 2024-06-05
feat(plugin): Copy Icons

Scyye:
Adds `Copy Icons` which allows for you to copy the icon URL of guilds & users, works with server specific avatars.


Vendicated:
exists already https://vencord.dev/plugins/ViewIcons

Scyye:
not sure how i forgot about that- i literally use it lmfao

PR #2555, opened 2024-06-06
[Plugin] Drag favorite emotes

pwall2222:


Sqaaakoi:
i will test this later; this could be very useful

pwall2222:
Okay so right now it's very very very WIP, (although 90% of functionality is there)

But in the `update` function, it basically is hard coded to make 1 *real* drag (by setting a global flag). 
Right now that's with the purpose of debugging, because if you modify the emoji order, for some reason the drag data doesn't update, so when you try to do a second drag it references the old data.

(You can test this by putting a log-point/console.log/breakpoint in the `drop` function, and seeing that `source` and `target` are the ones that were there before the change)

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+    },
+    drop(emoji: EmojiData) {
+        return useDrop(() => ({
+            accept: "emoji",
+            collect: monitor => ({
+                canDrop: !!monitor.canDrop(),
+                isOver: !!monitor.isOver(),
+            }),
+            drop(item: { id: string; }) {
+                const source = item.id;
```
pwall2222, line 112: Here is broken and I still don't know why

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+type EmojiDescriptor = {
+    category: string;
+    emoji: EmojiData;
+};
+
+export default definePlugin({
+    name: "DraggableElements",
+    authors: [],
```
MrDiamondDog, line 44: Add yourself to Devs and put yourself in this array

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+    description: "",
```
MrDiamondDog, line 45: Add a description

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+            },
+        },
+        {
+            find: "EmojiTypes.GUILD?S",
+            replacement: {
+                match: /\[(.{1,2}\.emojiItemSelected)\]/,
+                replace: "[e.collected.isDragging?'':$1]",
+            },
+        },
+    ],
```
MrDiamondDog, line 91: rather than having a bunch of patches with the same find, you can have 1 that has an array of replacements instead

also, your find contains a variable name i believe and that will break, so use something else

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+export default definePlugin({
+    name: "DraggableElements",
+    authors: [],
+    description: "",
+    startAt: StartAt.WebpackReady,
+    patches: [
+        {
+            find: "EmojiTypes.GUILD?S",
+            replacement: {
+                match: /(\("li",{\.\.\..{1,2},key:.{1,2},ref:.{1,2})}/,
```
MrDiamondDog, line 51: use \i to match variable names, please do this for all matches

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+                replace: "$1,style:{position:'relative'}}",
+            },
+        },
+        {
+            find: "EmojiTypes.GUILD?S",
+            replacement: {
+                match: /(\(0,.{1,2}\.jsx\)\(.{1,2},{ref:.*?inNitroLockedSection:.{1,2}}\))/,
+                replace: "[$1,emoji.category==='FAVORITES'?$self.wrapper(Y):undefined]",
```
MrDiamondDog, line 59: you can not hardcode variable names into your patches as they will break with almost every discord update. make your patch find those variable names so you can use capturing groups or use the `arguments` variable. this applies to allot of your replacements

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+                    if (!window.AL) {
+                        const sourceIndex = e.emojis.findIndex(emoji => emoji === this.source);
+                        const targetIndex = e.emojis.findIndex(emoji => emoji === this.target);
+                        const finalIndex = targetIndex === 0 ? targetIndex : targetIndex - 1;
+                        if (sourceIndex === finalIndex) {
+                            return false;
+                        }
+                        e.emojis.splice(sourceIndex, 1);
+                        e.emojis.splice(finalIndex, 0, this.source);
+                        window.AL = true;
```
MrDiamondDog, line 124: dont pollute the `window` object like this. just make this a variable in your plugin definition and use `this`

MrDiamondDog (changes requested):

MrDiamondDog:
Please also make sure to run eslint in your project so your formatting matches the rest of the repo

pwall2222:
All the functionality works now, purged  all the bugs (maybe I actually need to return something to `FrecencyUserSettingsActionCreators` but unsure).

The thing is I am not familiar how to create patches and I know I have a few hardcoded variables,
like the `Y` to get target emote, I'll try to fix it as best I can.

<details><summary>Editable PR</summary>
<p>
(Also I can't make the PR editable, because I have an organization for my forks [github damn](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
</p>
</details>

pwall2222:
Okay, I have made the best effort to get a better `find` and to replace all hard coded variables.

byeoon:
this will be the best plugin of all time

pwall2222:
In my eyes, all is done, the only improvements you could make is make the drag item, be only the emote, but for native emojis, it's a pain. Although it is fairly easy for custom emotes with `getEmojiURL`.

Sqaaakoi:
force pushing kinda horror
it'll get squashed into 1 commit later

src/plugins/draggables/style.css
```diff
@@ -0,0 +1,17 @@
+.dragging-indicator::before {
+    content: "";
+    position: absolute;
+    z-index: 1;
+    top: 4px;
+    right: 0;
+    left: 2px;
+    height: 40px;
```
Sqaaakoi, line 8: Not tall enough. Set this to 48px

src/plugins/draggables/style.css
```diff
@@ -0,0 +1,17 @@
...
+    max-width: 4px;
```
Sqaaakoi, line 9: Why is this max-width? Set this to width

Sqaaakoi: Also add `transform: translateX(-4px);` to position it correctly

src/utils/constants.ts
```diff
@@ -529,6 +529,10 @@ export const Devs = /* #__PURE__*/ Object.freeze({
     nekohaxx: {
         name: "nekohaxx",
         id: 1176270221628153886n
+    },
+    PWall: {
+        name: "PWall",
+        id: 0n
```
Sqaaakoi, line 535: Fill this in with your ID if you forgot to

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    ref.current.classList.add(dndCls.autoPointerEvents);
+                    return;
+                }
+                ref.current.classList.remove(dndCls.autoPointerEvents);
+            }
+            );
+            return () => cancelAnimationFrame(frame);
+        }, [collected, ref]);
+        return (
+            <div className={`${dndCls.wrapper} dragging-wrapper`} aria-hidden="true">
```
Sqaaakoi, line 142: Prefix your classes with `vc-`

src/plugins/draggables/style.css
Sqaaakoi: actually this is because the emojis resize :trollface:

pwall2222:
> force pushing kinda horror it'll get squashed into 1 commit later

Half the forfe pushes, are rebases.

src/plugins/draggables/style.css
pwall2222: > Why is this max-width? Set this to width

It used not to work with width, but now it does work with width magic I guess.

src/plugins/draggables/style.css
pwall2222: I'm gonna put 90% so it fits resizes (tested)

src/plugins/draggables/style.css
pwall2222: > Also add `transform: translateX(-4px);` to position it correctly

you think it's too on the right?

src/plugins/draggables/style.css
Sqaaakoi: Actually I think you should be doing `calc(100% - (var(--custom-emoji-picker-constants-emoji-container-padding-vertical) * 2))` instead

pwall2222: > Actually I think you should be doing `calc(100% - (var(--custom-emoji-picker-constants-emoji-container-padding-vertical) * 2))` instead

Why? /gen (I say this because 90% sorta works)

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    match: /(\(0,\i\.jsx\)\(\i,{ref:.*?isFavoriteEmojiWithoutFetchingLatest\((\i)\).*?inNitroLockedSection:\i}\))/,
+                    replace: "[$1,emoji.category==='FAVORITES'?$self.wrapper($2):undefined]",
+                },
+                {
+                    match: /(\[\i,\i\]=\i\.useState\(""\))/,
+                    replace: "$1,[collected,drag]=$self.drag(arguments[0]),emoji=arguments[0].descriptor",
+                },
+                {
+                    match: /(\(\i,{ref:)(\i),/,
+                    replace: "$1emoji.category==='FAVORITES'?drag:$2,collected:collected,",
```
Sqaaakoi, line 66: note this pollutes the DOM element attributes in a way that it looks like `button[collected="[object Object]"]`

pwall2222: I did notice, I just don't know how to pass the props another way (not too acquainted with React)

Sqaaakoi: sounds like the component you are shoving "collected" into does `...props` somewhere

pwall2222: Yes, it deconstructs (`let {whatever,...rest} = props;`) props at the beginning of `button` and whatever is not deconstructed it passes along.

pwall2222: 926243

Sqaaakoi:
Heads up; over the past week I've noticed that this plugin causes Discord to crash whenever you type text into the "add reaction" emoji picker; the React error message is "Rendered fewer hooks than expected. This may be caused by an accidental early return statement."

sadan4:
hi, are you still interested in continuing this pr? if not, do you mind if I finish it? I really like the idea and want to see it merged.

pwall2222:
> hi, are you still interested in continuing this pr? if not, do you mind if I finish it? I really like the idea and want to see it merged.

Is there something to finish? Did it break?
When I left it was pretty much done as far as I remember.

sadan4:
it broke a bit

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+export default definePlugin({
+    name: "DraggableEmotes",
+    authors: [Devs.PWall],
+    description: "Adds the ability to change the order of your favourite emotes",
+    startAt: StartAt.WebpackReady,
+    patches: [
+        {
+            find: 'allowAnimatedEmoji","selectedItemClassName',
+            //     ^ Probably should be replaced
+            replacement: [
```
programminglaboratorys: consider checking out regex positive lookbehind, it allows you to match while not consuming the substring as it seems like some of those just inserts, also work on making them more readable. maybe add comments to propose the purpose of each patch.

src/plugins/draggables/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    replace: "$1collected.isDragging?$self.dragItem():$2",
+                },
+                {
+                    match: /\[(\i\.emojiItemSelected)\]/,
+                    replace: "[collected.isDragging?'':$1]",
+                },
+            ],
+        },
+    ],
+    drag(e: { descriptor: EmojiDescriptor; }) {
```
programminglaboratorys: consider using object destructuring instead of E

programminglaboratorys:
add a readme

pwall2222:
Added the readme

PR #2557, opened 2024-06-07
fix USRBG & ViewIcons in new profiles

AutumnVN:


src/plugins/usrbg/index.tsx
```diff
@@ -75,16 +71,18 @@ export default definePlugin({
         },
         {
             find: /overrideBannerSrc:\i,overrideBannerWidth:/,
```
Vendicated, line 73: is this patch even still needed? according to report this find doesnt find any module

AutumnVN: fixed, i forgot that

src/plugins/usrbg/index.tsx
```diff
@@ -63,28 +63,26 @@ export default definePlugin({
...
+            replacement: {
+                match: /(\i)\.premiumType/,
+                replace: "$self.premiumHook($1)||$&"
+            }
+        },
+        {
+            find: "BannerLoadingStatus:function",
+            replacement: {
+                match: /(?<=void 0:)\i.getPreviewBanner\(\i,\i,\i\)/,
+                replace: "$self.useBannerHook(arguments[0])||$&"
```
Vendicated, line 83: this breaks the rules of hooks. hooks must not be called conditionally

image

Vendicated: oh... this is not a react hook

Vendicated:
thanks :3

PR #2561, opened 2024-06-07
fix badges for simplified profiles

Masterjoona:
makes badges support the new simplified profiles

### DIFF
--- a/src/plugins/_api/badges/index.tsx
+++ b/src/plugins/_api/badges/index.tsx
@@ -87,7 +87,7 @@ export default definePlugin({
             ]
         },
         {
-            find: "_.QuestContent.QUEST_BADGE",
+            find: 'action:"HOVER_BADGE"}),(0,',
             replacement: [
                 {
                     match: /(?<=href:(\i)\.link.{20,60})src:/,
@@ -98,16 +98,12 @@ export default definePlugin({
                     replace: "children:$1.component ? () => $self.renderBadgeComponent($1) :"
                 },
                 {
-                    match: /}=\i,{/,
-                    replace: ",VcUser:VcUser$&"
+                    match: /(?<=badges:(\i).{50,180}Context\)\(\);)/,
+                    replace: "let VcUser=$1.pop();"
                 },
                 {
                     match: /href:(\i)\.link/,
                     replace: "...($1.onClick && { onClick: vcE => $1.onClick(vcE, VcUser) }),$&"
-                },
-                {
-                    match: /(?=;return.{40,140}(.{1})\.map)/,
-                    replace: ";$1.unshift(...Vencord.Api.Badges._getBadges(VcUser))"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


src/plugins/_api/badges/index.tsx
```diff
@@ -85,7 +85,49 @@ export default definePlugin({
                     replace: "...($1.onClick && { onClick: vcE => $1.onClick(vcE, arguments[0]) }),$&"
                 }
             ]
-        }
+        },
+        {
+            find: "_.QuestContent.QUEST_BADGE",
```
MrDiamondDog, line 90: The _ in this find is a variable so this find would break very often

MrDiamondDog (changes requested):

src/plugins/_api/badges/index.tsx
```diff
@@ -85,7 +85,49 @@ export default definePlugin({
...
+                    match: /=\(0,\i\.default\)\((\i)\);/,
+                    // this $1 user has userId instead of id
+                    // and i didnt feel like adding more types idk
+                    replace: "$&let VcUser=Vencord.Webpack.Common.UserStore.getUser($1.userId);"
+                },
+                {
+                    match: /badges:\i,/,
+                    replace: "$&VcUser:{user:VcUser},"
+                }
+            ]
```
Vendicated, line 129: why do these two patches need to be separate? they can most likely be merged into one

src/plugins/_api/badges/index.tsx
```diff
@@ -85,7 +85,49 @@ export default definePlugin({
...
+                },
+                {
+                    match: /href:(\i)\.link/,
+                    replace: "...($1.onClick && { onClick: vcE => $1.onClick(vcE, VcUser) }),$&"
+                },
+                {
+                    match: /(?=;return.{40,140}(.{1})\.map)/,
+                    replace: ";$1.unshift(...Vencord.Api.Badges._getBadges(VcUser))"
+                }
+            ]
```
Vendicated, line 112: it is very dangerous to have one replacement declare a variable and another replacement reference it. if the first replacement breaks but the second one still works, this essentially guarantees a crash since you're referencing an undeclared variable

try to avoid doing that at all cost. if you absolutely must, you can use `group: true` on the patch to mark its replacements to be one group, which makes Vencord only apply them if all work!

src/plugins/_api/badges/index.tsx
Masterjoona: Oops, i was testing around, where i had the unshift here too, what need that variable also

Masterjoona:
From what I gathered, you need to pass it somehow to the another component.
I guess i could throw the user object to badges array and then get it and remove it, which might be more safe? Idk

Vendicated:
thanks a lot, but superseded by https://github.com/Vendicated/Vencord/commit/65970618d89d4c7383fc88dde8380423e67cd5e5, which does this cleaner, without needing to pass any props / variables around patches

PR #2569, opened 2024-06-09
Summaries: Configurable max stored summaries

Sqaaakoi:
The default expiry time of 3 days is often not the reason why old summaries are discarded, in very busy chats I have experienced the limit of 50 saved summaries being reached after around 7 hours of messages.

TL;DR: This will allow the summary log in large servers to be stored for a more reasonable period of time


Vendicated:
we added the limit for a reason. 50 is very reasonable, i don't really see demand for a higher limit than that

with infinite limit it will eventually lead to lag due to the sheer amount of summaries stored

mantikafasi:
I think it being configurable wouldnt be bad, but yea making it very high would cause lags.

zejjnt:
Maybe I should start a new PR for this but summaries have not worked for me for at least 2 months; is this a known issue?

PR #2571, opened 2024-06-09
feat(CopyEmojiMarkdown): Add feature to copy from messages

vishnyanetchereshnya:
Feature to copy emoji from message content and message reaction


Vendicated:
honestly don't really get the point, just use the Copy Text button or ViewRaw

and why would you even need it from a message

PR #2573, opened 2024-06-11
[Plugin] UserNotes

vishnyanetchereshnya:
Allows you to write unlimited notes for users. Unlike Discord, which restricts note saving to a maximum of 500 users and removes older notes when this limit is exceeded.

### DIFF
--- a/src/plugins/userNotes/components/UserNotesModal.tsx
+++ b/src/plugins/userNotes/components/UserNotesModal.tsx
@@ -1,0 +1,66 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { classNameFactory } from "@api/Styles";
+import {
+    closeModal, ModalCloseButton, ModalFooter, ModalHeader, ModalProps, ModalRoot, openModal
+} from "@utils/modal";
+import { Button, Text, TextArea, useState } from "@webpack/common";
+import { User } from "discord-types/general";
+
+import { getUserNotes, saveUserNotes } from "../data";
+
+const cl = classNameFactory("vc-user-notes-modal-");
+
+export function UserNotesModal({ modalProps, close, user, userNotes }: {
+	modalProps: ModalProps;
+	close(): void;
+	user: User;
+	userNotes: string;
+}) {
+    const [value, setValue] = useState(userNotes);
+
+    console.log(user);
+
+    return (
+        <ModalRoot className={cl("root")} {...modalProps}>
+            <ModalHeader className={cl("header")}>
+                <Text variant="heading-lg/semibold" style={{ flexGrow: 1 }}>Notes: {user.username}</Text>
+                <ModalCloseButton onClick={close} />
+            </ModalHeader>
+            <div className={cl("content")}>
+                <TextArea
+                    className={cl("text-area")}
+                    placeholder="Click to add a note"
+                    value={value}
+                    onChange={setValue}
+                />
+            </div>
+            <ModalFooter className={cl("footer")}>
+                <Button
+                    color={Button.Colors.GREEN}
+                    onClick={() => {
+                        saveUserNotes(user.id, value);
+                        close();
+                    }}
+                >
+					Save
+                </Button>
+            </ModalFooter>
+        </ModalRoot>
+    );
+}
+
+export const openUserNotesModal = async (user: User) => {
+    const key = openModal(modalProps => (
+        <UserNotesModal
+            modalProps={modalProps}
+            close={() => closeModal(key)}
+            user={user}
+            userNotes={getUserNotes(user.id) ?? ""}
+        />
+    ));
+};
--- a/src/plugins/userNotes/data.ts
+++ b/src/plugins/userNotes/data.ts
@@ -1,0 +1,57 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import {
+    clear,
+    createStore,
+    del,
+    entries,
+    set,
+    setMany,
+} from "@api/DataStore";
+
+const NotesStore = createStore("UserNotesData", "UserNotesStore");
+
+const usersNotes = new Map<string, string>();
+
+const cacheUsersNotes = async () => {
+    entries(NotesStore).then(usersNotesDB => {
+        usersNotesDB.forEach(([userId, userNotes]) => {
+            usersNotes.set(userId as string, userNotes);
+        });
+    });
+};
+
+cacheUsersNotes();
+
+export const getUserNotes = (userId: string): string | undefined => {
+    return usersNotes.get(userId);
+};
+
+export const saveUserNotes = async (userId: string, userNotes: string) => {
+    if (userNotes.trim() === "") {
+        usersNotes.delete(userId);
+        await del(userId, NotesStore);
+    } else {
+        usersNotes.set(userId, userNotes);
+        await set(userId, userNotes, NotesStore);
+    }
+};
+
+export const clearUserNotes = async () => {
+    usersNotes.clear();
+    await clear(NotesStore);
+};
+
+export const transferUserNotes = async (regularUsersNotes: { [userId: string]: string; }) => {
+    await setMany(Object.keys(regularUsersNotes).reduce((usersNotesDB, userId) => {
+        const userNotes = regularUsersNotes[userId];
+        usersNotesDB.push([userId, userNotes]);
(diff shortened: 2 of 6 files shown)


src/plugins/userNotes/components/UserNotesModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function UserNotesModal({ modalProps, close, user, userNotes }: {
+	modalProps: ModalProps;
+	close(): void;
+	user: User;
+	userNotes: string;
+}) {
+    const [value, setValue] = useState(userNotes);
+
+    console.log(user);
```
MrDiamondDog, line 26: dont forget to remove any `console.log`s in your plugin

src/plugins/userNotes/data.ts
```diff
@@ -0,0 +1,57 @@
...
+import {
+    clear,
+    createStore,
+    del,
+    entries,
+    set,
+    setMany,
+} from "@api/DataStore";
+
+const NotesStore = createStore("UserNotesData", "UserNotesStore");
```
MrDiamondDog, line 16: You don't need to do this, just use `DataStore.set("UserNotesData", ...)` or whatever function you need

src/plugins/userNotes/settings.tsx
```diff
@@ -0,0 +1,70 @@
...
+        type: OptionType.COMPONENT,
+        description:
+			"Transfer existing Discord notes into this plugin's data",
+        component: () =>
+            <Button onClick={() => {
+                // didn't find better way to get current user notes :))
+                fetch("https://discord.com/api/v9/users/@me/notes", {
+                    method: "GET",
+                    headers: {
+                        Authorization: `${token.getToken()}`
```
MrDiamondDog, line 33:
```suggestion
                        Authorization: token.getToken()
```

MrDiamondDog (commented):
I would keep the Discord notes system and this plugin tied together so there aren't two different places for notes

Basically just put the notes from the plugin in the normal Discord notes box, and add the notes to the plugin when you type in the normal notes box

src/plugins/userNotes/data.ts
vishnyanetchereshnya: > You don't need to do this, just use `DataStore.set("UserNotesData", ...)` or whatever function you need

with it then I need to save data as object in store instead just using userId as key

vishnyanetchereshnya:
> I would keep the Discord notes system and this plugin tied together so there aren't two different places for notes
> 
> Basically just put the notes from the plugin in the normal Discord notes box, and add the notes to the plugin when you type in the normal notes box

yea, I thought about it and decided to do it that way cos I feel like the default notes box isn't really convenient due to it being so small

src/plugins/userNotes/data.ts
MrDiamondDog: > with it then I need to save data as object in store instead just using userId as key

what? look at how other plugins do it. You don't need to create a custom store

vishnyanetchereshnya: > what? look at how other plugins do it. You don't need to create a custom store

I mean with the custom store I can update user note just by `set(userId, userNotes, NotesStore)` (just set single value for single key) instead each time set a whole `usersNotes` map into the default store (slower)

Vendicated: yep if you're storing many rows, maintaining your own store is better, it's like a sql database

vishnyanetchereshnya: > yep if you're storing many rows, maintaining your own store is better, it's like a sql database
> 
> although you might even want to create the store manually so you can create an index on the column you query

yea, I tried to do that but unfortunately have no experience with `indexedDB` so I'll leave it in it's current state fow now. mb in future I'll have more knowledge haha

Scyye:
at that point id just hide the user notes section, and turn it into a button that opens the popup, unless thats already whats being done.

vishnyanetchereshnya:
> at that point id just hide the user notes section, and turn it into a button that opens the popup, unless thats already whats being done.

yup, good idea, just added it

Scyye:
Maybe make it like a kind of preview thing? like have it still render the notes box, but clicking it opens the popup?

just an idea, and i think itll make it easier to read the first few lines of the notes that are written for a user.

vishnyanetchereshnya:
> Maybe make it like a kind of preview thing? like have it still render the notes box, but clicking it opens the popup?
> 
> just an idea, and i think itll make it easier to read the first few lines of the notes that are written for a user.

your ideas will always be great ha-ha
will do

Scyye:
image
In the popup menu, id add tooltips to the icons (like when you hover over them, it tells you what they actually do)

src/plugins/userNotes/components/NotesDataModal.tsx
```diff
@@ -0,0 +1,294 @@
...
+                            label="Copy ID"
+                            action={() => copyWithToast(userInfo!.id)}
+                        />
+                        {
+                            !pending &&
+                            (
+                                <>
+                                    <Menu.MenuItem
+                                        id={cl("copy-user-name")}
+                                        label="Copy UserName"
```
Scyye, line 191: Refer to it as `Username`, its one word.

src/plugins/userNotes/components/NotesDataModal.tsx
```diff
@@ -0,0 +1,294 @@
...
+    } as const;
+
+    const [userNotes, setUserNotes] = useState(userNotesArg);
+
+    return (
+        <div
+            className={cl("user")}
+            onContextMenu={event => {
+                ContextMenuApi.openContextMenu(event, () =>
+                    <Menu.Menu
```
Scyye, line 175: Also, to the context menu with all the copies, I'd add a "View User Profile" button, that opens their profile (this can be done with `discord.com/users/USER_ID`

src/plugins/userNotes/components/NotesDataModal.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+    const visibleNotes = filteredNotes.slice(0, visibleNotesNum);
+
+    const canLoadMore = visibleNotesNum < filteredNotes.length;
+
+    return (
+        <ModalRoot className={cl("root")} {...modalProps}>
+            <ModalHeader className={cl("header")}>
+                <Text className={cl("header-text")} variant="heading-lg/semibold">Notes Data</Text>
+                <TextInput className={cl("header-input")} value={searchQuery} onChange={e => setSearchQuery(e)} placeholder="Filter Notes (ID/Notes)" />
```
Scyye, line 59: Make this function with usernames as well rather than just IDs.

Scyye, line 59: Add maybe a drop down, or 2 buttons here that let you filter by friends/people you have blocked

src/plugins/userNotes/components/NotesDataModal.tsx
```diff
@@ -0,0 +1,294 @@
...
+            {
+                pending ? <LoaderSpinner /> :
+                    <Avatar
+                        className={cl("user-avatar")}
+                        size="SIZE_56"
+                        src={userInfo.avatar}
+                    />
+            }
+            <div className={cl("user-info")}>
+                <Text className={cl("user-info-name")} variant="text-lg/bold">{userInfo.name}</Text>
```
Scyye, line 220: maybe add an option in settings to use friend nicknames/global names rather than usernames. I dont know about anyone else, but I recognize those names more than the usernames.

src/plugins/userNotes/components/NotesDataModal.tsx
vishnyanetchereshnya: > Make this function with usernames as well rather than just IDs.

I tried to do that but for it I need to all users was cached into UserStore but it can cause rate limited and I'm not sure where and how I have to request all of them. mb u can help me with that

vishnyanetchereshnya:
> In the popup menu, id add tooltips to the icons (like when you hover over them, it tells you what they actually do)

added too

vishnyanetchereshnya:
@Scyye btw finally added users cache menu to fetch all users and be able to filter by global name or username

Vendicated:
the limit is 1500 not 500. why would anyone ever need more notes than that?

vishnyanetchereshnya:
> the limit is 1500 not 500. why would anyone ever need more notes than that?

is it? before I'm always tried to hold my notes less than 500 lol
I remember when I used betterdiscord and there I had 2k+ notes in userNotes so I find it useful
most useful thing here is searching by notes as I think

vishnyanetchereshnya:
@Vendicated if they increased notes amount from 500 to 1500, what do u thing if I remade this plugin into `notesSearcher` to just search user by notes? it feels really useful. about `userNotes` I can leave it as userplugin and drop into 3rd party plugins channel

Vendicated:
that works

PR #2580, opened 2024-06-15
Add NoAutoMute plugin

HAHALOSAH:
This plugin stops Discord from automatically unmuting you when you join a VC

### DIFF
--- a/src/plugins/noAutoUnmute/index.ts
+++ b/src/plugins/noAutoUnmute/index.ts
@@ -1,0 +1,35 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2024 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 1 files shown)


src/plugins/noAutoUnmute/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NoAutoUnmute",
+    description: "Prevents you from being automatically unmuted by Discord (such as by joining a VC call)",
```
Vendicated, line 24:
```suggestion
    description: "Prevents you from being automatically unmuted / undeafened when joining a DM call",
```

src/plugins/noAutoUnmute/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    authors: [Devs.HAHALOSAH],
+    patches: [
+        {
+            find: "isSelfMute(){",
+            replacement: {
+                match: /video:\i}=\i;/,
+                replace: "$&return;"
```
Vendicated, line 31: the function you're fully nooping here is also responsible for other stuff, eg the no input detected notice
image
image

so a more precise patch would be better

```suggestion
                match: /\(\i.mute||\i.deaf\)&&/,
                replace: "false&&"
```

Scyye:
*please* use actual commit messages.
image

Nuckyz:
> _please_ use actual commit messages. image

why does it matter?

Scyye:
> why does it matter? It's a pull request

its a lot easier to follow whats goin on in the pr if the messages arent just "Updated X file", thats just me tho.

HAHALOSAH:
> _please_ use actual commit messages. image

I clicked the button to merge suggested changes

PR #2581, opened 2024-06-16
Add Hide Groups plugin

KingRainbow44:
This plugin allows users to hide group DMs via the context menu.


src/plugins/hideGroups/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+const hiddenGroupsStore = createStore("HiddenGroups", "HiddenGroupsList");
+
+const hiddenGroups: string[] = [];
+const elements: { [key: string]: HTMLElement } = {};
+
+async function saveToStorage(): Promise<void> {
+    await set("list", JSON.stringify(hiddenGroups), hiddenGroupsStore);
+}
+
+function onChange(): void {
```
MrDiamondDog, line 37: Avoid DOM manipulation when patches are available.

MrDiamondDog (commented):
I feel like this should be able to hide normal dms as well, seems weird to restrict it to only group dms

KingRainbow44:
> I feel like this should be able to hide normal dms as well

I felt that the native Discord feature of closing DMs was comparable to this, which is why I only added support for group DMs.

MrDiamondDog:
> > I feel like this should be able to hide normal dms as well
> 
> 
> 
> I felt that the native Discord feature of closing DMs was comparable to this, which is why I only added support for group DMs.

Ngl I forgot you could do that, ignore that then

src/plugins/hideGroups/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    description: "Allows hiding group DMs.",
+    settings,
+
+    startAt: StartAt.DOMContentLoaded,
+    patches: [
+        {
+            find: 'location:"private_channel"',
+            replacement: {
+                match: /return .\.isMultiUserDM\(\)/,
+                replace: (match) => {
```
MrDiamondDog, line 111: You can use capturing groups and \i to avoid finding the variable in the match string

programminglaboratorys:
it would be cool if it automatically hides recently joined groups. it's annoying to suddenly see 300 notifications from a group you didn't ask to be joined in

programminglaboratorys:
> it would be cool if it automatically hides recently joined groups. it's annoying to suddenly see 300 notifications from a group you didn't ask to be joined in

I understood the purpose of the plugin, quite confused of how this could be useful when it only hides dm groups in the dms list

KingRainbow44:
> quite confused of how this could be useful when it only hides dm groups in the dms list

This plugin is meant to "add" the functionality of closing DMs for groups, not necessarily "hide" group DMs outright. It's useful for people (like me) who are in many group DMs and want to remove the clutter.

programminglaboratorys:
> This plugin is meant to "add" the functionality of closing DMs for groups

fair point

KingRainbow44:
never do commit shenanigans ever again

PR #2594, opened 2024-06-19
doomsday fix: ChatInputButtonAPI

ryan-0324:


Vendicated:
thanks but your patches are always really overengineered... try keeping them simple

PR #2599, opened 2024-06-19
doomsday fix: MoreUserTags

ryan-0324:


Vendicated:
looks good, but I deliberately didn't update this plugin because it honestly  needs a rewrite badly. the patches are very explosive since they rely on variables defined by other patches

Vendicated:
thanks ryan!

Scyye:
any chance it could be made into an API? and let other plugins add other tags; and have the base ones in moreUserTags?

> looks good, but I deliberately didn't update this plugin because it honestly needs a rewrite badly. the patches are very explosive since they rely on variables defined by other patches

Vendicated:
that shouldn't be necessary. plugins can use the decorations api to add stuff to messages

PR #2600, opened 2024-06-19
Add NoAccountLimit plugin

HAHALOSAH:
This plugin removes the 5 account limit for the account switcher


lewisakura:
Duplicate of #1353, see justification there for why we don't want it in the mod.

Issue #2477, opened 2024-05-18
[Bug] [Vencord's VoiceMessages plugin] using the default microphone from windows instead the one chosen by the discord user under the [Voice & Video] section


z2vvqx:
### Discord Account

m9jd

### What happens when the bug or crash occurs?

The title pretty much covers it all, the issue is that the plugin is not using the microphone I want it to use, and there's plenty of microphones connected to my PC so that's a problem.

### What is the expected behaviour?

I expect it to use my settings in the [Voice & Video] section instead of providing it's on settings such as [Noise suppression, Echo cancellation], or provide a new 'own' setting that let's me choose a microphone, but I think it's better to use the user's [Voice & Video] settings if possible.

### How do you recreate this bug or crash?

1. Plug a microphone to your PC
2. Plug another microphone to your PC
3. Choose the microphone you want to use in [Voice & Video] section of discord settings
4. Create a voice message using the VoiceMessages vencord plugin and it'll use [most likely the first plugged microphone to your PC]

### Errors

```
NONE
```

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [X] I have read the requirements for opening an issue above

PR #2602, opened 2024-06-19
VoiceMessages: properly respect user's microphone choice

Masterjoona:
Use currently selected input device instead of the default one.
#2477

### DIFF
--- a/src/plugins/voiceMessages/WebRecorder.tsx
+++ b/src/plugins/voiceMessages/WebRecorder.tsx
@@ -20,7 +20,7 @@ import { Button, useState } from "@webpack/common";
 
 import type { VoiceRecorder } from ".";
 import { settings } from "./settings";
-import { getInputDeviceId } from "./utils";
+import { MediaEngineStore } from "./utils";
 
 export const VoiceRecorderWeb: VoiceRecorder = ({ setAudioBlob, onRecordingChange }) => {
     const [recording, setRecording] = useState(false);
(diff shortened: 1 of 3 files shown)


src/plugins/voiceMessages/utils.ts
```diff
@@ -17,5 +17,7 @@
 */
 
 import { classNameFactory } from "@api/Styles";
+import { findByPropsLazy } from "@webpack";
 
+export const { getInputDeviceId } = findByPropsLazy("getInputDeviceId");
```
Vendicated, line 22:
```suggestion
import { findStoreLazy } from "@webpack";

export const MediaEngineStore = findStoreLazy("MediaEngineStore");
```

src/plugins/voiceMessages/WebRecorder.tsx
```diff
@@ -40,6 +41,7 @@ export const VoiceRecorderWeb: VoiceRecorder = ({ setAudioBlob, onRecordingChang
                 audio: {
                     echoCancellation: settings.store.echoCancellation,
                     noiseSuppression: settings.store.noiseSuppression,
+                    deviceId: getInputDeviceId()
```
Vendicated, line 44: this might return "default". are you sure that'll work?

Masterjoona: I tested it like a week ago and it worked fine

Vendicated: yep seems to work fine, at least for web

Vendicated:
kiitos joona

PR #2603, opened 2024-06-20
Google it

AbstractMelon:
# Let Me Google That For You Command

This command generates a "Let Me Google That For You" (LMGTFY) URL. By default, it uses `https://letmegooglethat.vercel.app/`, but you can customize it to use any other LMGTFY-like service.

### Features
- Generates a LMGTFY URL based on a command
- Customizable URL

## Example
When a user types the command with a query, the bot responds with a LMGTFY URL.

example


Vendicated:
textreplace can do this

AbstractMelon:
image

PR #2606, opened 2024-06-20
fix: canonicalize find in PatchHelper

j4k0xb:
Before:
image

After the fix `\i` works properly:
image

### DIFF
--- a/src/components/VencordSettings/PatchHelperTab.tsx
+++ b/src/components/VencordSettings/PatchHelperTab.tsx
@@ -311,12 +311,13 @@ function PatchHelper() {
         try {
             let parsedFind = v as string | RegExp;
             if (/^\/.+?\/$/.test(v)) parsedFind = new RegExp(v.slice(1, -1));
+            const canonicalFind = canonicalizeMatch(parsedFind);
 
             setFindError(void 0);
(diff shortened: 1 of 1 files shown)


src/components/VencordSettings/PatchHelperTab.tsx
```diff
@@ -311,12 +311,13 @@ function PatchHelper() {
...
             let parsedFind = v as string | RegExp;
             if (/^\/.+?\/$/.test(v)) parsedFind = new RegExp(v.slice(1, -1));
+            const canonicalFind = canonicalizeMatch(parsedFind);
 
             setFindError(void 0);
             setParsedFind(parsedFind);
 
             if (v.length) {
-                findCandidates({ find: parsedFind, setModule, setError: setFindError });
+                findCandidates({ find: canonicalFind, setModule, setError: setFindError });
```
Scyye, line 320: just ooc, is there a reason for this to be a variable rather than just passing in `canonicalizeMatch(parsedFind);`

j4k0xb: not really except for consistency sake, there's also a `canonicalMatch` and `canonicalReplace` variable

PR #2608, opened 2024-06-21
feat(themes): Online Themes Redesign

MrDiamondDog:
Better than the previous pr and doesn't require any migration, and unlike the last one, uses the same validation as before so no theme links break.

image


Vendicated:
i tried rebasing but pring from `main` branch makes git super weird. please **use feature branches**. i can't reopen cause i only have write access while PR is open, reopened via #2610

PR #2614, opened 2024-06-23
add ability to add ignored words to message logger 

imZuxi:
i think that this is useful :)

### DIFF
--- a/src/api/Notifications/notificationLog.tsx
+++ b/src/api/Notifications/notificationLog.tsx
@@ -19,6 +19,8 @@
 import * as DataStore from "@api/DataStore";
 import { Settings } from "@api/Settings";
 import { classNameFactory } from "@api/Styles";
+import { Flex } from "@components/Flex";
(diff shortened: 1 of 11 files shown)


src/plugins/messageLogger/index.tsx
```diff
@@ -267,6 +272,7 @@ export default definePlugin({
             ignoreChannels.includes(ChannelStore.getChannel(message.channel_id)?.parent_id) ||
             (isEdit ? !logEdits : !logDeletes) ||
             ignoreGuilds.includes(ChannelStore.getChannel(message.channel_id)?.guild_id) ||
+            ignoreWords.split(",").some(word => message.content.toLowerCase().includes(word.trim().toLowerCase())) ||
```
Scyye, line 275: what if you want to include a `,` in an ignored word?

imZuxi: i do not believe there is a behavior that i can specifically check for if the word in the list contains , however i would just blacklist x ie discord.gg in the list or some outher word also to my knowledge no engilsh word contains a ,

PR #2615, opened 2024-06-24
feat(plugin): KeywordNotify

camila314:
Simulates a ping whenever a certain regex-defined keyword is in a message. Custom inbox for messages with keywords. Blacklist and whitelist for servers and channels.

😁


NepNep21:
It would be nice to have a toggle to make a regex case insensitive

Scyye:
> It would be nice to have a toggle to make a regex case insensitive, since that's not possible to do just by changing the string passed to the Regex constructor afaik

unless i dont understand what you mean, you can literally just have it match lowercase or uppercase for each of the things.

src/plugins/keywordNotify/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+            const whitelistMode = entry.listType === ListType.Whitelist;
+            if (!whitelistMode && listed) {
+                return;
+            }
+            if (whitelistMode && !listed) {
+                return;
+            }
+
+            if (settings.store.ignoreBots && m.author.bot) {
+                if (!whitelistMode || !entry.listIds.includes(m.author.id)) {
```
Scyye, line 329: why nest this? 
```ts
if (settings.store.ignoreBots && m.author.bot && (!whitelistMode || !entry.listIds.includes(m.author.id)))
```
imo this is fine, although idk how others like it, or if this is against some standard.

src/plugins/keywordNotify/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+        return false;
+    }
+}
+
+enum ListType {
+    BlackList = "BlackList",
+    Whitelist = "Whitelist"
+}
+
+function highlightKeywords(s: string, r: Array<string>) {
```
Scyye, line 60: use better variable names to explain what these actually are.

src/plugins/keywordNotify/index.tsx
```diff
@@ -0,0 +1,447 @@
...
+    forceUpdate();
+}
+
+async function removeKeywordEntry(idx: number, forceUpdate: () => void) {
+    keywordEntries.splice(idx, 1);
+    await DataStore.set(KEYWORD_ENTRIES_KEY, keywordEntries);
+    forceUpdate();
+}
+
+function safeMatchesRegex(s: string, r: string) {
```
Scyye, line 47: use better variable names to explain what these actually are.

NepNep21:
> > It would be nice to have a toggle to make a regex case insensitive, since that's not possible to do just by changing the string passed to the Regex constructor afaik
> 
> unless i dont understand what you mean, you can literally just have it match lowercase or uppercase for each of the things.

Strings aren't necessarily exclusively lower or uppercase, so no. To emulate case insensitivity without adding it in the second parameter of the constructor the user would have to add every possible upper/lowercase permutation, which gets extremely large extremely quickly.

NepNep21:
The easy solution is to conditionally add `i` to the flags string in https://github.com/Vendicated/Vencord/blob/31b18b221ddabff500404acf340d42403495c4ff/src/plugins/keywordNotify/index.tsx#L63 depending on the config.

camila314:
@NepNep21 got good news for you

bdotsamir:
Hello! Lucky I followed the chain all the way here from potentially suggesting a duplicate extension. I have just *one* suggestion for this: my name is a pretty common english word. I would love it if there was a guild blacklist (or whitelist for that matter) that would filter out whenever the word is mentioned in chat. A cursory look through the diff doesn't seem to show that? Unless I'm blind?

programminglaboratorys:
did you put in mind that the message op is blocked or a channel is muted?

PR #2620, opened 2024-06-24
ReplaceGoogleSearch: improve search engine selection

OtterCodes101:


KrstlSkll69:
Isn't this what the custom search engine is for?
(_also adding your self to devs for a two line change is wild_)
Capture

KrstlSkll69:
Ok, I'm going to assume you are a NTTS kid, also just as the main dev of this plugin Moxxie said "Things like AOL, Kagi, Banidu where added as a joke and a funny haha"

KrstlSkll69:
Thats actually not a bad idea, and it actually wouldn't be that much of a pain

Kodarru:
Couldn't we turn the const in a var so we can edit it, rename it and allow the user to delete and add search engines?

Kodarru:
Then the client could fetch the image url to use via the favicon

PR #2623, opened 2024-06-25
friendsSince: add support for new profiles

Chicken:
I'm actually not sure if this should be merged *yet*, I don't think the new user popups and profiles are pushed out for everyone yet (?)
Added matching modules with a string array because I couldn't figure out a good single string match and using RegExp felt too overkill.
Added myself to the authors because felt this was a big enough contribution.


src/plugins/friendsSince/index.tsx
```diff
@@ -63,36 +53,57 @@ export default definePlugin({
...
+        return (
+            <section className={section.section}>
+                <Heading variant="text-xs/semibold">
+                    Friends Since
+                </Heading>
+
+                <Text variant="text-sm/normal">
+                    {getCreatedAtDate(friendsSince, locale.getLocale())}
+                </Text>
+            </section>
```
Vendicated, line 106: seems to be mostly duplicate. it would likely be better to instead take a boolean inside props and do conditional rendering

src/plugins/friendsSince/index.tsx
```diff
@@ -4,50 +4,40 @@
...
+const containerWrapper = findByPropsLazy("memberSinceWrapper");
+const container = findByPropsLazy("memberSince");
 
-const container = findByPropsLazy("memberSinceWrapper");
 const getCreatedAtDate = findByCodeLazy('month:"short",day:"numeric"');
 const locale = findByPropsLazy("getLocale");
-const lastSection = findByPropsLazy("lastSection");
-
-const cl = classNameFactory("vc-friendssince-");
+const section = findLazy((m: any) => Object.values(m).length === 1 && m.section !== void 0);
```
Vendicated, line 20: flipping the condition will be much more performant since it avoids thousands of unnecessary Object.values() calls
```suggestion
const section = findLazy((m: any) => m.section && Object.values(m).length === 1);
```

Vendicated (commented):
thanks for the contribution!

I don't think the find array addition is a good idea. it would be better to stick to just using a regex

As for rollout or not, why not keep the old patch too alongside the new one?

Chicken:
Is there some reason to not support an array of string matches? I feel like it allows for easy matching when a simple string match is hard to figure out and probably performs better compared to a RegExp. It also doesn't take much code to support.

Keeping the old patches is quite brilliant, no idea why I didn't think of that. Though, would probably need an almost duplicate component as the new layout required quite different styles and writing that as conditionals would be way too messy to read.

Vendicated:
> Though, would probably need an almost duplicate component as the new layout required quite different styles and writing that as conditionals would be way too messy to read.

if its just a different class, you can use our classes utility:
```js
className={classes({
    [styles.old]: !isNewProfile,
    [styles.new]: isProfile
})}
```

Vendicated:
as for performance of array vs regex, as long as you keep the regex simple, performance won't be a big deal. if anything, using an array find might be slower because it has to walk the same string multiple times

it's best to keep the find as simple as possible and regex should cover everything you need

Chicken:
>if its just a different class

Well it did change from a div to a section, different class name, different heading variant and also a different *amount* of wrapper divs.

Chicken:
For the regex I would just use something like `/onOpenProfile.+subsection.+USER_PROFILE_MEMBER_SINCE/`

Masterjoona:
Hei Antti!
You can do either `bio)!==""&&` or `.PANEL}),nicknameIcons` as a find :)

Chicken:
Removed all the array stuff, used suggested find above, used the more performant find function, kept old patches, put the sidebar stuff in a boolean prop but kept new and old layouts as separate components. Removed the old layout css file and classes though as the css file was never imported and used (?)

src/plugins/friendsSince/index.tsx
```diff
@@ -88,11 +102,55 @@ export default definePlugin({
                             <path d="M3 5v-.75C3 3.56 3.56 3 4.25 3s1.24.56 1.33 1.25C6.12 8.65 9.46 12 13 12h1a8 8 0 0 1 8 8 2 2 0 0 1-2 2 .21.21 0 0 1-.2-.15 7.65 7.65 0 0 0-1.32-2.3c-.15-.2-.42-.06-.39.17l.25 2c.02.15-.1.28-.25.28H9a2 2 0 0 1-2-2v-2.22c0-1.57-.67-3.05-1.53-4.37A15.85 15.85 0 0 1 3 5Z" />
                         </svg>
                     )}
-                    <Text variant="text-sm/normal" className={classes(cl("body"), textClassName)}>
```
Vendicated, line 91: why did you remove these classes?

Chicken: Explained in the above comment, as far as I'm aware they never did anything as the css file wasn't even included.

src/plugins/friendsSince/index.tsx
```diff
@@ -4,50 +4,64 @@
...
                 match: /(\.Messages\.USER_PROFILE_MEMBER_SINCE.+?userId:(.+?),textClassName:)(\i\.userInfoText)}\)/,
-                replace: (_, rest, userId, textClassName) => `${rest}!$self.getFriendSince(${userId}) ? ${textClassName} : void 0 }), $self.friendsSince({ userId: ${userId}, textClassName: ${textClassName} })`
+                replace: (_, rest, userId, textClassName) => `${rest}!$self.getFriendSince(${userId}) ? ${textClassName} : void 0 }), $self.friendsSinceOld({ userId: ${userId}, textClassName: ${textClassName} })`
+            }
+        },
+        // DM User Sidebar - new layout
+        {
+            find: ".PANEL}),nicknameIcons",
+            replacement: {
+                match: /USER_PROFILE_MEMBER_SINCE.+?userId:(.+?)}\)}\)/,
```
Vendicated, line 55: try to use .+? sparingly

it could happen that this string appears way earlier in the file and you end up matching way too much or the wrong thing

either remove `.+?` or replace it with a fixed length match, eg `.{0,200}?` to match up to 200 chars

src/plugins/friendsSince/index.tsx
Vendicated: wow you're right hahaha

src/plugins/friendsSince/index.tsx
Vendicated: same below

Chicken: Fixed to use fixed length matches

src/plugins/friendsSince/index.tsx
```diff
@@ -4,50 +4,64 @@
...
+            }
+        },
+        // User Profile Modal - new layout
+        {
+            find: "action:\"PRESS_APP_CONNECTION\"",
+            replacement: {
+                match: /USER_PROFILE_MEMBER_SINCE,.{0,100}userId:(.{0,3}\.id),.{0,100}}\)}\),/,
+                replace: "$&,$self.friendsSinceNew({userId:$1,isSidebar:false}),"
             }
         }
```
Vendicated, line 66: ah, so these were actually meant to match variable names. you can achieve that using our special `\i` escape
```suggestion
        {
            find: ".PANEL}),nicknameIcons",
            replacement: {
                match: /USER_PROFILE_MEMBER_SINCE,.{0,100}userId:(\i\.id)}\)}\)/,
                replace: "$&,$self.friendsSinceNew({userId:$1,isSidebar:true})"
            }
        },
        // User Profile Modal - new layout
        {
            find: "action:\"PRESS_APP_CONNECTION\"",
            replacement: {
                match: /USER_PROFILE_MEMBER_SINCE,.{0,100}userId:(\i\.id),.{0,100}}\)}\),/,
                replace: "$&,$self.friendsSinceNew({userId:$1,isSidebar:false}),"
            }
        }
```

Chicken: oh, cool

Vendicated:
maybe it would look better in one line? wastes a lot of space being separate

image

or even like this
image

Chicken:
Well, I agree. But the point of this PR is to just make it work like it used to. That can be looked into later.

Vendicated:
thanks for the fix!

PR #2627, opened 2024-06-25
Migrate to eslint flat config; update dependencies

SomeAspy:
This PR migrates the ESLint config away from `.eslintrc.json` to `eslint.config.mjs`

- Updated ESLint related packages (including the patch for `no-useless-escape`)

I tried to keep changes minimal, but some things probably should be checked.

- Unsure if `path-alias/no-relative` rule is working. I think the plugin needs to be updated. (Upstream is as ~v2) However, this was another patched package, and I do not have an adequate understanding of patches to re-patch it (Upstream has been rewritten to TS, but its a simple patch so I don't think it should be too hard)
- Some weird indentation issues in `src/utils/types.ts`, `src/shared/SettingsStore.ts`, and `src/api/Settings.ts` I am leaving these up to you to review, as I'm not sure what your preference is.
- A few other warnings (6 unused `eslint-disable`s. Leaving this up to you to review.)


SomeAspy:
Hi @Vendicated I was able to patch the package in c9982b5, and this resolved the issues related to `no-relative`

Vendicated:
oh god wtf is that diff

Vendicated:
thanks for your work!

Issue #2619, opened 2024-06-24
[F.Rq.] Smarter ShowMeYourName -- skip if case-insensitive match of nick


YoshiRulz:
### Content

Title. The double username is redundant in this case. For example:
![screencap](https://github.com/Vendicated/Vencord/assets/13409956/acdf72ca-315b-4fe4-921d-94a9ca85a4e8)

### Request Agreement

- [X] I have read the requirements for opening an issue above

PR #2630, opened 2024-06-26
ShowMeYourName: compare username & nick case insensitive

Bloofield:
Update to ShowMeYourName which makes the matching nickname/username check case insensitive

As requested in #2619


Vendicated:
thanks!

PR #2632, opened 2024-06-26
feat(plugin update): ShowHiddenThings: Always show View Server As Role button

hauntii:
Always show the option to "view a server as role" regardless of whether you're an admin (manage server + manage roles)
image


Scyye:
image

KrstlSkll69:
> image

Its a does something that quite literally fits the plugins description 
Capture

Vendicated:
this is already part of permissionviewer. but it's hidden because it may leave your client in an invalid state and send a lot of 401 api requests

thanks anyway

PR #2634, opened 2024-06-28
new API: MarkdownAPI

programminglaboratorys:
I have notice some pr plugins, patch the markdown rules or react components to add their own markdown, with this API there won't be need to do so and won't create conflicts between plugins. easily have the ability to add and modify discord markdown rules. 
this adds a new dependency [simple-markdown](https://www.npmjs.com/package/simple-markdown/) for typing purposes as this is the same library discord uses for their markdown


Scyye:
id draft this and go through and replace the plugins that use md to use this API instead.

if you grab a list of them, i could make a pr to the feature branch and youd just have ta merge it.

programminglaboratorys:
> id draft this and go through and replace the plugins that use md to use this API instead.
> 
> if you grab a list of them, i could make a pr to the feature branch and youd just have ta merge it.

I am unsure of this idea, I see that the API should be added first than the authors could integrate to use the API; I could assist them

src/api/Markdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function insertSlateRules(slate: any) {
+    return Object.assign(
+        slate,
+        Object.fromEntries(
+            Array.from(
+                Object.entries(Rules), ([_, rules]: [string, Rules]) => Object.entries(rules).map(([name, rule]: [string, Rule]) => [name, (rule.Slate ?? slate[name]) ?? { type: "skip" }])).flat()
+        )
+    );
```
MrDiamondDog, line 78: this is completely unreadable and i'm sure there is a better way to do this.

what is this function even trying to do anyway?

Keenelge:
```suggestion
    return Object.assign(
        slate,
        Object.values(Rules).reduce((acc, rule) => {
            Object.keys(rule).forEach((subRuleKey) => {
                const subRule = rule[subRuleKey];

                acc[subRuleKey] = subRule.Slate ?? slate[subRuleKey] ?? { type: "skip" };
            });

            return acc;
        }, {})
    );
```

MrDiamondDog: That's better but you should definitely add a comment that says what this does. 

Also you're two different variables named Rules and rules are not good, there is no way to differ between them. Choose a different name for one of them

Keenelge: agree with MrDiamondDog. u can just rename
`interface MarkDownRules` > `interface MarkDownRulesType`
`const Rules: MarkDownRules = {} as MarkDownRules` > `const MarkDownRules = {} as MarkDownRuleTypes` (u no need to put `:MarkDownRules` here cos u already declared type of object)

PR #2636, opened 2024-06-28
feat(plugin): DefaultStatusForever

ImLvna:
# DefaultStatusForever

Makes the custom status screen default to "Don't Clear" as the duration

The custom status screen, with "Don't Clear" selected and as the first option

### DIFF
--- a/src/plugins/defaultStatusForever/README.md
+++ b/src/plugins/defaultStatusForever/README.md
@@ -1,0 +1,5 @@
+# DefaultStatusForever
+
+Makes the custom status screen default to "Don't Clear" as the duration
+
+![The custom status screen, with "Don't Clear" selected and as the first option](https://github.com/Vendicated/Vencord/assets/44179559/e8cf0ca1-589c-43ef-9a6d-4fa3ed03a1f9)
--- a/src/plugins/defaultStatusForever/index.tsx
+++ b/src/plugins/defaultStatusForever/index.tsx
@@ -1,0 +1,48 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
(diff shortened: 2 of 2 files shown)


src/plugins/defaultStatusForever/index.ts
```diff
@@ -0,0 +1,48 @@
...
+        }
+    ],
+
+    patchChoices(choices: Choice[]) {
+        const nullChoice = choices.find(choice => choice.value === null);
+        if (nullChoice) {
+            choices.splice(choices.indexOf(nullChoice), 1);
+            choices.unshift(nullChoice);
+        }
+        return choices;
```
Vendicated, line 46:
```suggestion
        return choices.sort((a, b) =>
        	a.value === null
        		? -1
        		: b.value === null ? 1 : 0;
        )
```

src/plugins/defaultStatusForever/index.ts
```diff
@@ -0,0 +1,48 @@
...
+            }
+        },
+        {
+            // reorder the list to put "Dont't Clear" at the top
+            find: "get clearAfterOptions",
+            replacement: {
+                match: /(?<=get clearAfterOptions\(\){return).*?}]/,
+                replace: " $self.patchChoices($&)"
+            }
+        }
```
Vendicated, line 37: this is the same module, no?
```suggestion
        {
            find: "this.clearAfterOptions",
            replacement: [
            	// change default status duration to null
            	{
                	match: /(?<=value:)\i(?=,options:this.clearAfterOptions)/,
                	replace: "null"
            	},
            	// reorder the list to put "Dont't Clear" at the top
				{
                	match: /(?<=get clearAfterOptions\(\){return).*?}]/,
                	replace: " $self.patchChoices($&)"
            	}
            ]
        }
```

D3SOX:
This no longer works after the recent Discord changes

Vendicated:
Discord seems to now remember your last choice, so this plugin is obsolete

PR #2641, opened 2024-06-30
feat(MessageLogger): Ignored Word

imZuxi:
Add in ability to add ignored words (accidentally closed my pr when fixing the merge conflict)


src/plugins/messageLogger/index.tsx
```diff
@@ -221,6 +221,11 @@ export default definePlugin({
             description: "Comma-separated list of guild IDs to ignore",
             default: ""
         },
+        ignoreWords: {
+            type: OptionType.STRING,
+            description: "Comma-separated list of words that should be ignored if found (case insensitive)",
```
Scyye, line 226: I know i mentioned this on the other PR, and i did mean to response, but it ended up getting closed;

I feel like there should be a separate setting to define a delimiter, it would allow for the use of things that contain `,`, and you could always just default it to a comma.

imZuxi: yes sorry been one busy week i would add however could cause confusion though

Vendicated:
could you describe a use case for this feature that wouldn't be solved by blacklisting specific users? the only use case I can imagine are things like automated messages that regularely get deleted and have no significance after deletion but you can just blacklist the bot in question

imZuxi:
> could you describe a use case for this feature that wouldn't be solved by blacklisting specific users? the only use case I can imagine are things like automated messages that regularely get deleted and have no significance after deletion but you can just blacklist the bot in question

The use case primarily intended is for when a server uses a bot to delete spam messages. By blocking commonly posted items from being logged, such as links, I can prevent spam links from being logged. For instance, blocking the beginning of a URL like "[http" would ensure that any spam links are deleted of course this would also lead to normal links getting fully deleted however this is a compromise thats willing to be made

PR #2446, opened 2024-05-12
feat(package): @vencord/discord-types

ryan-0324:
Resolves #2431

- Added package @vencord/discord-types
- Migrated from discord-types to @vencord/discord-types
  However, discord-types is still needed for `WebpackInstance`
- Added `FormattedMessage`, `useStateFromStoresArray`, and `useStateFromStoresObject` to `@webpack/common`
- Changed the names of some exports from `@webpack/common/utils` to the names that Discord uses for them
- TSConfig changes:
  - Enabled `noUncheckedIndexedAccess` and `noImplicitAny`
  - Disabled `skipLibCheck` since it disables type checking for all `.d.ts` files, not just the ones in `node_modules`
  - Changed `moduleResolution` from `Node` to `Bundler`
- Updated eslint to 9.x.x and stylelint to 16.x.x
- Enabled type-aware linting and some rules from typescript-eslint
- Updated some dev dependencies and corrected/updated the version specifiers for `@types/lodash`, `@types/node`, `@types/react`, `@types/react-dom`, `highlight.js`, and `moment`
- Updated esbuild to 0.23.0 and made the build scripts compatible
- Enabled type checking for JS files in `scripts`
- Added missing license to `packages/vencord-types`
- Miscellaneous refactoring

PR #2660, opened 2024-07-03
chore: esbuild: ^0.15.18 → ^0.23.0

bb010g:
> [!NOTE]  
> This PR is a patch stack. Do not squash these commits. This patch stack depends on:
> + <https://github.com/Vendicated/Vencord/pull/1661>

I wanted to use RegExp literals using the [`v` flag](<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicodeSets>), but esbuild was too old. Whee.

The build scripts are first moved over to TypeScript (`.mts`) so upgrading with esbuild is sane. VS Code wouldn't show errors for the previous `.mjs` files for me, at least. `tsc` will catch future type errors now. (I can split this out to a separate PR if you *really* want.)

esbuild is then moved forward one breaking release at a time, to make both authoring and reviewing easier.

[esbuild v0.16](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG-2022.md#0160>) doesn't break us.

[esbuild v0.17](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG-2023.md#0170>) breaks our use of `esbuild.build({ watch: true })`, so that's all refactored to support the new `context` design and recommended logging setup. We now log our own summaries & errors, but esbuild can still do so with `pnpm run build --summary`. Care has been taken to prevent logs from separate builds from overlapping. (This delays builds slightly, esp. when `--summary` is passed, but it's worth it for useful output.) I'd really like to instruct esbuild to call a JavaScript callback whenever it logs. Related issues:
+ <https://github.com/evanw/esbuild/issues/2886>

[esbuild v0.18](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG-2023.md#0180>) breaks our type signatures slightly.

[esbuild v0.19](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG-2023.md#0190>) doesn't break us.

[esbuild v0.20](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG.md#0200>) doesn't break us.

[esbuild v0.21](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG.md#0200>) doesn't break us.

[esbuild v0.22](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG.md#0200>) breaks our use of `--platform=node` defaults, switching from `--packages=bundle` to `--packages=external`.

[esbuild v0.23](<https://github.com/evanw/esbuild/blob/e3f4e2d36dd9d777a0676de8aff3f4764ee7079f/CHANGELOG.md#0200>) switches `--platform=node` defaults back to `--packages=bundle`, but I kept our explicit declaration.

Each commit passed `pnpm install && pnpm run lint && pnpm run lint-styles && pnpm run testTsc && pnpm run build --dev` on my machine.

<https://github.com/Vendicated/Vencord/pull/2446> also upgrades esbuild to `^0.21.5`, but the output handling is less robust and I generally think the approach here is cleaner.


scripts/build/build.mts
```diff
@@ -21,43 +21,37 @@ import esbuild from "esbuild";
...
     // If this is a local build (not standalone), optimize
     // for the specific platform we're on
     defines["process.platform"] = JSON.stringify(process.platform);
 
-/**
- * @type {esbuild.BuildOptions}
- */
-const nodeCommonOpts = {
-    ...commonOpts,
+const nodeBuildOpts: esbuild.BuildOptions = {
```
Vendicated, line 42: if you use
```ts
const nodeBuildOpts = {} satisfies BuildOptions;
```
instead, it will have the exact type and you won't have to assert nullability everywhere

scripts/build/build.mts
```diff
@@ -131,46 +125,46 @@ await Promise.all([
...
         entryPoints: ["src/main/index.ts"],
         outfile: "dist/vencordDesktopMain.js",
         footer: { js: "//# sourceURL=VencordDesktopMain\n" + sourceMapFooter("vencordDesktopMain") },
         sourcemap,
         define: {
             ...defines,
-            IS_DISCORD_DESKTOP: false,
-            IS_VESKTOP: true
+            IS_DISCORD_DESKTOP: JSON.stringify(false),
+            IS_VESKTOP: JSON.stringify(true)
```
Vendicated, line 159:
```suggestion
            IS_DISCORD_DESKTOP: "false",
            IS_VESKTOP: "true"
```

scripts/build/buildWeb.mts
```diff
@@ -135,19 +148,14 @@ async function globDir(dir) {
...
 }
 
-/**
- * @type {(dir: string, basePath?: string) => Promise<Record<string, string>>}
- */
-async function loadDir(dir, basePath = "") {
+async function loadDir(dir: string, basePath: string = ""): Promise<Record<string, string>> {
     const files = await globDir(dir);
-    return Object.fromEntries(await Promise.all(files.map(async f => [f.slice(basePath.length), await readFile(f)])));
+    const dirContentEntries = await Promise.all(files.map(async (f): Promise<[string, string]> => [f.slice(basePath.length), await readFile(f, { encoding: "utf-8" })]));
```
Vendicated, line 153:
```suggestion
    const dirContentEntries = await Promise.all(files.map(async f => [f.slice(basePath.length), await readFile(f, { encoding: "utf-8" })] as const));
```

scripts/build/buildWeb.mts
```diff
@@ -159,7 +167,7 @@ async function buildExtension(target, files) {
             if (f.startsWith("manifest")) {
                 const json = JSON.parse(content.toString("utf-8"));
                 json.version = VERSION;
-                content = new TextEncoder().encode(JSON.stringify(json));
```
Vendicated, line 162: why?

scripts/build/common.mts
```diff
@@ -54,11 +54,7 @@ export const banner = {
 };
 
 const PluginDefinitionNameMatcher = /definePlugin\(\{\s*(["'])?name\1:\s*(["'`])(.+?)\2/;
-/**
- * @param {string} base
- * @param {import("fs").Dirent} dirent
- */
-export async function resolvePluginName(base, dirent) {
+export async function resolvePluginName(base: string, dirent: Dirent): Promise<string> {
```
Vendicated, line 57: return types should not be specified unless they can't properly be inferred

```suggestion
export async function resolvePluginName(base: string, dirent: Dirent) {
```

scripts/build/common.mts
```diff
@@ -292,22 +273,177 @@ export const stylePlugin = {
...
+        return buildResult;
+    } catch (error) {
+        context.dispose();
+        if (watch) {
+            throw error;
+        } else {
+            process.exitCode = 1;
+        }
+    }
+}
```
Vendicated, line 425: why is all this needed? seems overengineered

scripts/utils.mts
```diff
@@ -16,15 +16,25 @@
...
+        resolve = res;
+        reject = rej;
+    });
+    return { promise, resolve, reject };
+}
+
+export function getPluginTarget(filePath: string): string | undefined {
     const pathParts = filePath.split(/[/\\]/);
-    if (/^index\.tsx?$/.test(pathParts.at(-1))) pathParts.pop();
+    if (/^index\.tsx?$/.test(pathParts.at(-1) ?? "")) pathParts.pop();
```
Vendicated, line 35:
```suggestion
    if (/^index\.tsx?$/.test(pathParts.at(-1)!)) pathParts.pop();
```

scripts/build/build.mts
bb010g: The type was specified to provide a stable interface to other build scripts. Does that not matter?

scripts/build/common.mts
bb010g: All the logging code is necessary to prevent logs from running over each other, since esbuild doesn't provide sufficient locking. `makeBuildPromise` handles esbuild's new build API without changing how the script is used. I wish all that code wasn't necessary.

jamesbt365:
Superseded by #3218

PR #2662, opened 2024-07-03
Add StyleListener to ClientTheme to fix lightModeFixes

CodeF53:
Light mode fixes of client theme replaces all `--white-500` backgrounds with `--primary-100`, but since discord started lazy loading css chunks, the old approach of generating all these fixes at `DOMContentLoaded` is no longer functional.

This PR adds a mutationObserver to the head of the DOM to generate new fixes for lightMode as the app is used.


src/plugins/clientTheme/cssUtil.ts
```diff
@@ -0,0 +1,44 @@
...
+async function parseStyleLinkNode(node: HTMLLinkElement): Promise<string> {
+    const link = node.getAttribute("href");
+    if (!link) return "";
+
+    try {
+        const res = await fetch(link);
+        return await res.text();
+    } catch (error) {
+        return "";
+    }
```
Vendicated, line 42:
```suggestion
    return fetch(link)
    	.then(res => res.text())
        .catch(() => "");
```

src/plugins/clientTheme/cssUtil.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+// returns all of discord's native styles in a single string
+// ! discord lazy loads css chunks, make sure to use newStyleListener as well
+export async function getStyles(): Promise<string> {
+    let out = "";
+    const styleLinkNodes: NodeListOf<HTMLLinkElement> = document.querySelectorAll('link[rel="stylesheet"]');
+    for (const styleLinkNode of styleLinkNodes)
+        out += await parseStyleLinkNode(styleLinkNode);
+    return out;
```
Vendicated, line 14:
```suggestion
    return Promise.all(styleLinkNodes.map(n => parseStyleLinkNode(n)))
    	.then(contents => contents.join("\n"));
```

src/plugins/clientTheme/cssUtil.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// runs callback with styles of new style nodes added to the head
+export function newStyleListener(callback: (styles: string) => void) {
+    const observer = new MutationObserver(mutations => {
```
Vendicated, line 19: i'm unsure if using a MutationObserver is a good idea. css chunks are loaded via webpack (`wreq.f.css`).

it'd be better to hook into that instead

this func is used for all except the initial chunk
image
image
https://discord.com/assets/12633.bb59c3b8d0e17096489a.css

src/plugins/clientTheme/cssUtil.ts
CodeF53, line 1: @Vendicated using tabs in suggested change, evil

src/plugins/clientTheme/cssUtil.ts
CodeF53: I will ask for advice on how to patch `wreq.f.css` when I have time to fix pr

Vendicated:
lint explosion

src/plugins/clientTheme/cssUtil.ts
CodeF53: you can't regex patch the webpack loader, how do I hook into wreq.f.css then?

PR #2635, opened 2024-06-28
feat(plugin): SimplifiedProfileNotes

Sqaaakoi:
Adds notes back to simplified profile popouts
example showing note at the bottom of a popout

It also includes an option to only show the note section when a note is set. The "add note" button has been reprogrammed to show and focus the note textarea

Also works in the DM sidebar
Note section inserted under the member since text in the DM user profile sidebar

Types haven't been added yet, however the plugin is functional

PR #2665, opened 2024-07-04
feat(UserVoiceShow): update for simplified profiles

vishnyanetchereshnya:
patches update for simplified profiles
Discord_0MPZMH7xoG
Discord_lBw7b2Ee5e


nin0-dev (commented):
you shouldn't override old patches, in case
- by some modern day miracle some people manage to turn off simplified profiles
- bots still use old profiles

vishnyanetchereshnya:
> you shouldn't override old patches, in case
> 
> * by some modern day miracle some people manage to turn off simplified profiles
> * bots still use old profiles

oh yea, u right

vishnyanetchereshnya:
btw I'm not sure to where place voice status in user popup. originally this component located between send message section and roles section. in simplified profile already exists some "voice status"(activity), I just override it by section from this plugin and now this one located ABOVE roles section. would it be better to move it as in original plugin (under roles)?

Vendicated:
could you show a screenshot please?

vishnyanetchereshnya:
> could you show a screenshot please?

yup, added into main pr message

Vendicated:
it takes up too much space. the header should probably be removed in the small profile and replaced with a vc icon

as for the large profile, the header style is inconsistent with discord

vishnyanetchereshnya:
> it takes up too much space. the header should probably be removed in the small profile and replaced with a vc icon
> 
> as for the large profile, the header style is inconsistent with discord

popout: will replace header by vc icon on left side of button
modal: will change text style

btw as nino-dev said bots still uses old profiles. does plugin have to have patches for them?

Sqaaakoi:
> as for the large profile, the header style is inconsistent with discord

I know I've been talking about this a lot (and even opened a technically redundant PR related to it #2690 ), but you can use the Section component I found (see the previously linked PR and #2635 for examples)

> btw as nino-dev said bots still uses old profiles. does plugin have to have patches for them?

enable the experiment for simplified bot profiles. i would ideally make sure it works there

also, not entirely related to this patch, a user can be in multiple voice channels at the same time, as long as they are in different servers. this might be more common with bots so maybe check if that is considered

Nuckyz:
Superseded by https://github.com/Vendicated/Vencord/commit/bae9f8cbe696b3f4ddfd41f80e0bfdb5ccc3a755, thanks regardless!

PR #2666, opened 2024-07-04
Fix: ViewIcons blocking opening profile in DMs

Obsidianninja11:
The ViewIcons plugin was blocking users from opening the bigger profile by clicking the pfp in the right side panel in dms

<details>
<summary>Image of where I'm talking about</summary>

image

</details>

Before fix: Clicking on this pfp would open the pfp as an image preview
After fix: Clicking on this pfp opens the big profile view as expected


Scyye:
Looking at it, this doesnt seem like a fix, more of just a.. feature removal..

i'd add a setting for it, and add a predicate to those 2 patches.

Obsidianninja11:
> Looking at it, this doesnt seem like a fix, more of just a.. feature removal..
> 
> i'd add a setting for it, and add a predicate to those 2 patches.

This is a bug. It used to not block opening from the profile menu from clicking this pfp. The feature is meant to make the normally non-clickable pfps clickable to view the avatar (and also add it to the right click menu)

PR #2668, opened 2024-07-05
Fix messagelogger unblurring removed spoilers

k26pl:
#1815


Vendicated:
this is not correct. you're just entirely disabling the rule `.foo:not(.foo)`

properly fixed via https://github.com/Vendicated/Vencord/commit/b7ac92106ab517e3b68656b62f14cbf117fecd0d

PR #2671, opened 2024-07-06
feat(plugin): Plex Rich Presence

DevJhoan:
Plugin adds a custom rpc with your current playing from Plex or Emby

msedge_aayqcu2MSh
msedge_oOslFECX6S
msedge_xSkpStUuu6

### DIFF
--- a/src/plugins/plexRichPresence/clients/EmbyClient.ts
+++ b/src/plugins/plexRichPresence/clients/EmbyClient.ts
@@ -1,0 +1,127 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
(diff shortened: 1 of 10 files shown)


programminglaboratorys:
too niche and very complex for a plugin. have you put in mind the user perspective?

DevJhoan:
> too niche and very complex for a plugin. have you put in mind the user perspective?

wdym with "user perspective"?

MrDiamondDog:
> wdym with "user perspective"?

In other words: User Experience. This plugin doesn't make much sense to the average user, and it is really niche as those services aren't used by many people.

I think this would do well as a third party plugin though.

happyendermangit:
Nice plugin, some people use plex and emby so it'll be useful for them...

Vendicated:
sorry but this is too niche.. and quite a lot of code for that

why can't you just use https://github.com/phin05/discord-rich-presence-plex ?

i'll be rejecting this, but thanks for your efforts!

PR #2674, opened 2024-07-08
Added an option to not anounce when the user joins

E3L1Z:
I added an option to allow you to mute announcements when you are joining/leaving a call or always when the target of the action is you.

-E3L1Z


Vendicated:
duplicate. also please follow the code style

PR #2680, opened 2024-07-10
feat(plugin:) disableSubtexts

happyendermangit:
- This plugin disables subtexts as they can be used to scam you/trick you
Plugin enabled preview


ImLvna:
why not a plugin to disable all features of markdown, and have them behind settings so the user can choose what to see

also your checks are failing

Sqaaakoi:
i feel like users that would fall for such scams likely wouldn't even enable a plugin like this, let alone know it exists

Vendicated:
le css

PR #2684, opened 2024-07-10
Do not crop non-square donor badges

maddie480:
It seems the donor badge I provided is not actually square, and the `object-fit` is set to `cover`, giving this:

attachment

And here is how it looks with `object-fit: contain`:

attachment

Now, maybe the solution is to replace my badge and make it actually square 😅 but I'm probably not the only one in this situation.

### DIFF
--- a/src/plugins/_api/badges/index.tsx
+++ b/src/plugins/_api/badges/index.tsx
@@ -163,7 +163,7 @@ export default definePlugin({
                 style: {
                     borderRadius: "50%",
                     transform: "scale(0.9)", // The image is a bit too big compared to default badges
-                    objectFit: "contain" // Do not crop non-square images
+                    ["objectFit" as any]: "contain" // Do not crop non-square images
                 }
             },
             onClick() {
--- a/src/plugins/_api/badges/index.tsx
+++ b/src/plugins/_api/badges/index.tsx
@@ -162,7 +162,8 @@ export default definePlugin({
             props: {
                 style: {
                     borderRadius: "50%",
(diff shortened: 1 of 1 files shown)


Vendicated:
i'm unsure if this is a good addition. it might look better for your badge, but some people have badges where the crop is beneficial. it would be useful to compare a good portion of existing badges with and without this filter and see if it's really better overall

but it would be best to simply ensure badge images are proper squares

maddie480:
Then I guess we should just replace my badge image and close this? 😅

Vendicated:
I'm not sure! like I said, it would be beneficial to look at a bunch of badges and see if it's overall better

maybe it is, i have no idea :3

maddie480:
there we go :thinking: [badge-simulator.html.zip](https://github.com/user-attachments/files/16240928/badge-simulator.html.zip) (`cover` on the left, `contain` on the right)

maddie480:
Some conversation happened on Discord after I posted the Badge Simulator :tm: above (which is just a big HTML page that gives a preview of all non-square donor badges with `object-fit: cover` and `object-fit: contain`), and the takeaway is: most badges aren't affected that much, a few look a lot worse, a few look a lot better. 

I think the right move would be to close this PR and let people reach out to fix their badges if they don't like how it fits, to replace it with an image that is actually square (in which case I'm going to reach out :wave:). But I'll let you be the judge and close it if you don't want it ^^

For example:

## Those look worse

image
image

## Those look better

image
image

Vendicated:
thanks for looking into this :3

PR #2686, opened 2024-07-11
Update Translate languages

korbosoft:
I used the table at https://cloud.google.com/translate/docs/languages because the method in `languages.ts` gives languages seemingly not yet supported by the API. AFAIK, this PR will add one language.


Vendicated:
this is the same list

PR #2689, opened 2024-07-12
CtrlEnterSend: fix for new Discord update

UlyssesZh:
Recently CtrlEnterSend has broken twice. This time we need to revert 77492061f to fix the problem. Actually Discord has been switching back and forth between those two places to handle message submission logic for multiple times (4 times as I personally experienced). In every version, both patches can be applied without a problem, so it would benefit if we simply include both patches although only one of them is at effect. This way the plugin should break less often.


ImLvna:
if only one of the patches is to be applied, why did we not set noWarn on them both? to prevent the error logs of patch not applying

UlyssesZh:
The best behavior is having error only when neither of them applies, but I don't know how to do this in Vencord.

PR #2690, opened 2024-07-12
FriendsSince: Migrate to Section component for new profiles

Sqaaakoi:
~~Some of the finds for the old className modules don't work anymore, so~~
I have ported the plugin to use the Section component I found while developing #2635

~~I'm not sure if this breaks the "old layout", I need someone to test that for me~~ Reverted that


Sqaaakoi:
I might be stupid and forgot to see if this was even fixed

Sqaaakoi:
turns out this wasn't entirely needed, at this point idk if this actually improves consistency or not

PR #2692, opened 2024-07-13
add more stuff to vencord debug

sunnniee:


src/plugins/_core/supportHelper.tsx
```diff
@@ -97,14 +98,18 @@ async function generateDebugInfoMessage() {
...
     const commonIssues = {
-        "NoRPC enabled": Vencord.Plugins.isPluginEnabled("NoRPC"),
         "Activity Sharing disabled": tryOrElse(() => !ShowCurrentGame.getSetting(), false),
+        "Link embeds disabled": tryOrElse(() => !ShowEmbeds.getSetting(), false),
         "Vencord DevBuild": !IS_STANDALONE,
         "Has UserPlugins": Object.values(PluginMeta).some(m => m.userPlugin),
         "More than two weeks out of date": BUILD_TIMESTAMP < Date.now() - 12096e5,
     };
 
+    const possiblyUnwantedPlugins = ["NoRPC", "NoProfileThemes", "NoMosaic", "Moyai"].filter(Vencord.Plugins.isPluginEnabled);
```
Vendicated, line 108:
```suggestion
    const possiblyUnwantedPlugins = ["NoRPC", "NoProfileThemes"].filter(Vencord.Plugins.isPluginEnabled);
```
i dont think these two are a real issue and even if they were it would be really obvious once the person describes the symtpoms. As in... the other two could actually cause confusion by breaking unrelated plugins / things and then as to where the issue originates from, while these two only have direct "diagnosable" consequences and don't negatively affect other stuff

sunnniee: i get what you mean, but they're more of a "why not". it's a lot quicker to tell someone to /vencord-debug than asking questions, and at least with moyai you can def make a vague support request like "why does my discord make random sounds". maybe replace the ⚠️ with a ℹ️ to indicate its not necessarily a bad thing?

Vendicated:
i don't really think we need these but thank you

PR #2696, opened 2024-07-15
ThemeAttributes: Add classes to root html element for app state

Sqaaakoi:
Currently, it is hard (impossible?) to reliably detect if Streamer Mode is enabled or not, because the warning notice bar can be dismissed. This injects a `streamer-mode` class into the root HTML element to make this detection reliable from within themes.


Sqaaakoi:
> @[Sqaaakoi](https://github.com/Sqaaakoi) Sqaaakoi closed this [2 minutes ago](https://github.com/Vendicated/Vencord/pull/2696#event-13505735565)
hiiiiii!!!! did you know I love accidentally committing git submodules

PR #2706, opened 2024-07-18
NoTrack: Disable native Sentry (+ core module native support)

Sqaaakoi:
This also resolves the issue when using the Discord deb package where you'd get spammed with Sentry errors


scripts/build/build.mjs
```diff
@@ -93,6 +93,17 @@ const globNativesPlugin = {
...
+            const coreNatives = await readdir(coreNativesPath, { withFileTypes: true });
+            for (const file of coreNatives) {
+                const fileName = file.name;
+                const pluginName = await resolvePluginName(join(coreNativesPath, "../"), file);
+
+                const mod = `p${i}`;
+                code += `import * as ${mod} from "./plugins/_core/native/${fileName}";\n`;
+                natives += `${JSON.stringify(pluginName)}:${mod},\n`;
+                i++;
+            }
```
Vendicated, line 106: why add so much code lol, just add `plugins/_core` to the `pluginDirs` array

src/plugins/_core/native/noTrack.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import Module from "module";
+
+if (IS_DISCORD_DESKTOP) {
+    // Disable Sentry in native by monkeypatching over require
+    // `crashReporterSetup.init(sentry, buildInfo)` will only call `initializeSentrySdk` (wrapper for `sentry.init`) if `sentry != null`
+    Module.prototype.require = new Proxy(Module.prototype.require, {
```
Vendicated, line 12: i don't like monkey patching this. you can proabbly just patch the require cache for it

scripts/build/build.mjs
Sqaaakoi: do you want me to switch it to be more like that?

src/plugins/_core/native/noTrack.ts
Sqaaakoi: I actually tried, however when I tried to, I was unable to require `@sentry/electron` as it was located inside of Discord's asar file

scripts/build/build.mjs
Sqaaakoi: honestly it would probably just be simpler anyway

Sqaaakoi: i was insane

Vendicated:
in its current state this isn't really acceptable... monkey patching global require is really hacky. it would be better to find some way to disable it without the need for that

PR #2712, opened 2024-07-21
IgnoreActivities: Add option for blacklist filter

Covkie:


KannaDev:
Aren't you referring to this, which already exists? 
image
Also, all you have done is rename the variables it seems, and add yourself as a contributor. Making this a pointless PR. 

Mind showing what changes this does?

Nuckyz:
Thank you!

PR #1975, opened 2023-11-11
[Plugin] ReplyPingControl

ant0n-0x0000:
Expanding on PR [#1973](https://github.com/Vendicated/Vencord/pull/1973) to allow the user to toggle between disabling all replies or enabling them.

PR #2713, opened 2024-07-21
feat(plugins): AlwaysPing

Z1xus:
hi! this plugin will always ping the user when they are replied to, regardless of the ping toggle

please provide your feedback and suggestions!

### DIFF
--- a/src/plugins/alwaysPing/index.ts
+++ b/src/plugins/alwaysPing/index.ts
@@ -1,0 +1,41 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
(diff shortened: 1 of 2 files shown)


cheesesamwich:
Looks cool actually, the description could use some work

Feathers8:
Wait, sorry if I'm wrong, but, doesn't #1975 supersede this?

cheesesamwich:
> Wait, sorry if I'm wrong, but, doesn't #1975 supersede this?

From what i can tell, that plugin modifies the actual message component to look like it mentioned you (i might be wrong)

KannaDev:
> dless of the ping toggle

if always ping. why ping toggle..
image

MrDiamondDog:
> > Wait, sorry if I'm wrong, but, doesn't #1975 supersede this?
> 
> From what i can tell, that plugin modifies the actual message component to look like it mentioned you (i might be wrong)

it actually mentions you

PR #2714, opened 2024-07-21
FakeNitroThemes: fix crash when encountering invalid colours

ingobeans:
…o to crash your client when viewing the profile


ingobeans:
Example: putting this: "󠁛󠀣󠁦󠁦󠁦󠁦󠁦󠁦󠁦󠁦󠁦󠀬󠀣󠁥󠁥󠁝" in your bio will make anyone using the plugins client crash when viewing it. Decoding it looks like "[#fffffffff,#ee]". It crashes because it doesn't validate the hex codes to be only 6 chars long, which then crashes the client.

ingobeans:
The copy 3y3 button always gives a 6 char hex code, so I don't necessarily see the need of including support for it.

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -57,7 +57,7 @@ function decode(bio: string): Array<number> | null {
     if (bio == null) return null;
 
     const colorString = bio.match(
-        /\u{e005b}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]+?)\u{e002c}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]+?)\u{e005d}/u,
+        /\u{e005b}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]{6})\u{e002c}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]{6})\u{e005d}/u,
```
ryan-0324, line 60: It has to be `{0,6}` to maintain compatibility with similar plugins for other client mods.
```suggestion
        /\u{e005b}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]{0,6})\u{e002c}\u{e0023}([\u{e0061}-\u{e0066}\u{e0041}-\u{e0046}\u{e0030}-\u{e0039}]{0,6})\u{e005d}/u,
```

ryan-0324: My bad, I meant `{1,6}`.

ingobeans: But the current code doesn't support non 6 char hex codes. #fff will not be interpreted as white by the current code, but rather as #000fff, which is blue. Therefore I don't see the need for allowing shorter codes

ryan-0324: It does not deal with CSS colors. The colors always get passed to encode as 24-bit colors, so that is the intended behaviour. If you force it to only accept 6 characters, you will break compatibility with other plugins for no reason.

ryan-0324: Sorry, that's not how I intended to come across.

Vendicated:
thanks for the fix

PR #2719, opened 2024-07-24
feat(plugin) FakeNitro: option to hide emote URLs via ghost message

arbitrator1:
## Description
Create option for link hiding via ghost messages (client parser bug that has been around for a while with evidently no action from discord to fix it). Adding for feature parity with the BD plugin.

It's off by default because it doesn't work particularly well for multiple emotes and editing (you have to delete the links yourself), although it will append new emotes upon edit.

This more for people like me who are bothered by the little changes when switching over from BD; I personally think the hyperlinks are cleaner when they are allowed but will continue using the ghost message feature until it is patched since it cannot not limited by permissions.

## Preview of what it looks like

message as seen by official clients (stable, PTB, Canary), vencord, BD, etc.

actual message contents


Vendicated:
i don't think this is a good idea because it could be considered spam and get automodded

also you can just do this with textreplace plugin if you really want

PR #2720, opened 2024-07-24
Fix broken find in ClientTheme

maddie480:
This affects that red button in the plugin settings, that errors out instead of switching to dark theme:

Capture d’écran 2024-07-24 à 09 42 18


maddie480:
Fixed by https://github.com/Vendicated/Vencord/pull/2723

Vendicated:
thank you anyway maddie

Issue #1786, opened 2023-10-10
[Feature Request] Use DeepL in translate plugin


Sanceilaks:
### Discord Account 
voidptr_t 
### What is it that you'd like to see? 
It would be convenient to be able to translate messages via DeepL.

### Request Agreement

- [x] I have read the requirements for opening an issue above

FirstWiseman:
I have added an Engine Selector my self, tried to implement DeepL and Bypass the apiKey thing and it worked for a sec but now i cant translate at all not even in the browser anymore xD.

That would be the only problem with DeepL, you have to use an Api Key and to receive the free api key you have to give your payment informations :/

![image](https://github.com/Vendicated/Vencord/assets/54641477/5450ad02-eb16-41a0-a6da-63af6ce553c3)

AshtonMemer:
DeepL API has a CORS policy that blocks requests from the frontend. A way to deal with this would be to use a proxy server, but then you would be sending your API key and potentially private message content to an individual's proxy server before DeepL.

Sqaaakoi:
> DeepL API has a CORS policy that blocks requests from the frontend. A way to deal with this would be to use a proxy server, but then you would be sending your API key and potentially private message content to an individual's proxy server before DeepL.

we can just delete the CORS header in desktop clients (electron based) in the same way that CSP is modified https://github.com/Vendicated/Vencord/blob/main/src/main/index.ts#L90-L124

AshtonMemer:
> we can just delete the CORS header in desktop clients (electron based) in the same way that CSP is modified [https://github.com/Vendicated/Vencord/blob/main/src/main/index.ts#L90-L124|](https://github.com/Vendicated/Vencord/blob/main/src/main/index.ts#L90-L124%7C)
> 
> it's also possible to make a "native" script that fetches that data for you

has this been done by an existing plugin yet?

Sqaaakoi:
> > we can just delete the CORS header in desktop clients (electron based) in the same way that CSP is modified [https://github.com/Vendicated/Vencord/blob/main/src/main/index.ts#L90-L124|](https://github.com/Vendicated/Vencord/blob/main/src/main/index.ts#L90-L124%7C)
> > it's also possible to make a "native" script that fetches that data for you
> 
> has this been done by an existing plugin yet?

https://github.com/Vendicated/Vencord/blob/main/src/plugins/openInApp/native.ts

PR #2721, opened 2024-07-24
Translate: Add DeepL support

AshtonMemer:
Adds an option to the Translate plugin to use DeepL Free or Pro with an API key.
- Addresses feature request #1786.
- Google Translate remains the default translation service.


src/plugins/translate/native.ts
```diff
@@ -0,0 +1,36 @@
...
+    data: string;
+}
+
+export function makeRequest(_: IpcMainInvokeEvent, pro: boolean, apiKey: string, payload: string): Promise<DeepLResponse> {
+    const url = pro
+        ? "https://api.deepl.com/v2/translate"
+        : "https://api-free.deepl.com/v2/translate";
+
+    return new Promise<DeepLResponse>((resolve, reject) => {
+        const req = request(new URL(url), {
```
Vendicated, line 21: i think it's safe for us to use fetch instead of request now since even discord stable has electron 30

src/plugins/translate/utils.ts
```diff
@@ -38,6 +49,46 @@ export interface TranslationValue {
 export async function translate(kind: "received" | "sent", text: string): Promise<TranslationValue> {
     const sourceLang = settings.store[kind + "Input"];
     const targetLang = settings.store[kind + "Output"];
+    const { service } = settings.store;
+
+    // DeepL not supported on web due to CORS policy
+    if (IS_DISCORD_DESKTOP && service !== "google") {
```
Vendicated, line 55: this is a bit messy. would be cleaner to define two functions `googleTranslate()` and `deeplTranslate()` and then use those based on if else

src/plugins/translate/settings.ts
```diff
@@ -53,7 +53,25 @@ export const settings = definePluginSettings({
...
+            ]
+        },
+        deeplApiKey: {
+            description: "DeepL API key",
+            type: OptionType.STRING,
+            default: "",
+            placeholder: "Get your API key from https://deepl.com/your-account",
+            disabled: () => settings.store.service === "google"
+        }
+    } : {})
```
Vendicated, line 74: this is kinda confusing for users cause they will have no clue why it's missing
```suggestion
    service: {
        description: IS_WEB ? "Translation service (Not supported on Web!)" : "Translation service",
        type: OptionType.SELECT,
        disabled: () => IS_WEB,
        options: [
            { label: "Google Translate", value: "google", default: true },
            { label: "DeepL Free", value: "deepl" },
            { label: "DeepL Pro", value: "deepl-pro" }
        ]
    },
    deeplApiKey: {
        description: "DeepL API key",
        type: OptionType.STRING,
        default: "",
        placeholder: "Get your API key from https://deepl.com/your-account",
        disabled: () => settings.store.service === "google"
    }
```

Vendicated:
looks good other than that!

AshtonMemer:
Resolved all of the reviews 👍

src/plugins/translate/utils.ts
```diff
@@ -73,3 +84,43 @@ export async function translate(kind: "received" | "sent", text: string): Promis
...
+    if (!settings.store.deeplApiKey) {
+        showToast("DeepL API key is not set", Toasts.Type.FAILURE);
+        throw new Error("DeepL API key is not set");
+    }
+
+    try {
+        // CORS jumpscare
+        const { status, data } = await Native.makeRequest(settings.store.service === "deepl-pro", settings.store.deeplApiKey, JSON.stringify({
+            text: [text],
+            target_lang: targetLang
```
Vendicated, line 101: are you sure deepl supports the same languages as google? we likely need a separate list of supported languages

src/plugins/translate/utils.ts
```diff
@@ -73,3 +84,43 @@ export async function translate(kind: "received" | "sent", text: string): Promis
...
+                throw new Error(
+                    `Failed to translate "${text}" (${sourceLang} -> ${targetLang})`
+                    + `\n${status} ${data}`
+                );
+        }
+
+        const { translations }: DeepLData = JSON.parse(data);
+        return { src: translations[0].detected_source_language, text: translations[0].text };
+    } catch (e) {
+        console.error(e);
```
Vendicated, line 123:
```suggestion
        new Logger("Translate").error("Failed to translate:", e);
```

src/plugins/translate/utils.ts
Vendicated: https://developers.deepl.com/docs/resources/supported-languages

src/plugins/translate/index.tsx
```diff
@@ -87,7 +92,9 @@ export default definePlugin({
             if (!settings.store.autoTranslate) return;
             if (!message.content) return;
 
-            message.content = (await translate("sent", message.content)).text;
+            message.content = settings.store.service === "google"
+                ? (await googleTranslate("sent", message.content)).text
+                : (await deeplTranslate("sent", message.content)).text;
```
Vendicated, line 97: i didn't mean to inline this if / else everywhery. i meant like
```ts
export function translate(text: string) {
	return IS_WEB || settings.store.service === "google"
		? googleTranslate(text)
		: deeplTranslate(text);
}

function googleTranslate(text: string) {}
function deeplTranslate(text: string) {}
```
and then use that translate func everywhere

Vendicated:
thanks ash!!

PR #2724, opened 2024-07-25
feat(AnonymizeFilenames): add file extension exceptions

leumasme:
New setting in AnonymizeFilenames:
A list of file extensions for which the opposite of `anonymiseByDefault` will be used.
So if anonymize by default is True, this is a list for file extensions which shouldn't be anonymized by default, and the other way around.

List setting could probably use a more useful description than the current one? Tried to keep it short.
`File extensions for which to use the oppposite Anonymize by Default behavior`

`invertedExts` could be cached but eh, its a tiny split-map-trim-filter


leumasme:
see also https://github.com/Vendicated/Vencord/pull/2797

PR #2729, opened 2024-07-27
feat(plugin): NoReactOverLimit

MrDiamondDog:
Makes the "add reaction" buttons disappear when the limit for reactions (20) has already been reached


Sqaaakoi (commented):
I'm uncertain if removing quick react is a good idea; I am a chronic quick react user (to the point I use 2 plugins to improve that experience) and as such I would appreciate settings for each patch?

Also how about making the add reaction labels say something like "Add Reaction (reactions/20)"

ProjectCrytos:
> I'm uncertain if removing quick react is a good idea; I am a chronic quick react user (to the point I use 2 plugins to improve that experience) and as such I would appreciate settings for each patch?
> 
> Also how about making the add reaction labels say something like "Add Reaction (reactions/20)"

Yeah i honestly think this is better

Vendicated:
this doesn't seem very useful. why would anyone need this?

PR #2732, opened 2024-07-27
Add plugin toggling to venbot embeds

Masterjoona:
https://github.com/user-attachments/assets/a2156ef4-89ca-4fbf-b508-2080e920a083


src/components/PluginSettings/togglePluginEnabled.ts
Sqaaakoi, line 1: Is this the right place for this code? I think that most of this should just be in `src/plugins/index.ts` with everything else, as this would be generally useful

src/components/PluginSettings/utils.ts
Sqaaakoi, line 1: Oops I reviewed the older code; yet I still feel the same way, I think that most of this should just be in src/plugins/index.ts with everything else, as this would be generally useful

Masterjoona: ah thanks, i was pondering where i should place this

Sqaaakoi:
after a tiny bit of extra thinking, why not just delete the original embed and replace it with a PluginCard using MessageAccessoriesAPI

Masterjoona:
looking into this

Masterjoona:
the excluded looks kinda ugly,  idk what would be a good component for it

lewisakura:
this is really cool, never thought about doing something like this when i wrote the original feature

for the exclusion stuff, i think it would be best to just render a plugin card but then replace the switch with a yellow exclamation icon (like what discord uses) with a tooltip that explains why, it'd require modification of the plugin card to support this though

src/plugins/_core/supportHelper.tsx
```diff
@@ -146,13 +148,22 @@ export default definePlugin({
...
+            replacement: {
+                match: /BEGINNING_DM\.format\(\{.+?\}\),(?=.{0,300}(\i)\.isMultiUserDM)/,
+                replace: "$& $self.renderContributorDmWarningCard({ channel: $1 }),"
+            },
+        },
+        {
+            find: "this.renderInlineMediaEmbed",
+            replacement: {
+                match: /return null(?<=var \i;let{embed:.{1,80})/,
+                replace: "if(this.props?.embed?.url.includes('https://vencord.dev/plugins/')){return $self.renderPluginCard({embed:this.props.embed})}$&"
```
Sqaaakoi, line 163: I would change this to `.startsWith` instead of `includes`

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +261,35 @@ export default definePlugin({
         );
     }, { noop: true }),
 
+    renderPluginCard: ErrorBoundary.wrap(({ embed }: { embed: Embed; }) => {
+        const pluginName = embed.rawTitle;
```
Sqaaakoi, line 265: This is dangerous / bad design considering how there is no filtering done on the author of the message in this implementation

Use the plugin name in the URL as an embed with a different title is possible. I know my regex hack could be improved though, so maybe don't use that considering you disregard the http/https compatibility anyway.

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +261,35 @@ export default definePlugin({
...
+        if (excludedPlugin) {
+            return (
+                <Button
+                    key="vc-plugin-notice"
+                    color={Button.Colors.RED}
+                >
+                    Only available on {ExcludedReasons[excludedPlugin]}
+                </Button>
+            );
+            // button really wide, good ideas?
```
Sqaaakoi, line 276: Don't use a button here. See @lewisakura's comment, maybe use a tooltip like required plugins do in the plugins page

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +261,35 @@ export default definePlugin({
...
+        } else {
+            const plugin = plugins[pluginName];
+            return (
+                <div style={{ width: "40%" }}>
+                    <PluginCard
+                        onRestartNeeded={() => showToast("Restart to apply changes!")}
```
Sqaaakoi, line 282: I don't like that this is being oversimplified. Make the code in PluginSettings' first useEffect function generic and call it here if you really don't want to duplicate trivial code

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +261,35 @@ export default definePlugin({
...
+                >
+                    Only available on {ExcludedReasons[excludedPlugin]}
+                </Button>
+            );
+            // button really wide, good ideas?
+        } else {
+            const plugin = plugins[pluginName];
+            return (
+                <div style={{ width: "40%" }}>
+                    <PluginCard
```
Sqaaakoi, line 281: No required tooltip?

src/plugins/_core/supportHelper.tsx
Sqaaakoi: I still think checking the protocol being http **could** be useful **but** is not required (again, rare case)

Sqaaakoi (commented):
I feel like you've oversimplified what code I gave you earlier in order to not duplicate things and to make it easier to understand, rather than keeping consistency with how PluginCard is wrapped in PluginSettings
Additionally I have some concerns about how URLs aren't being used to get the plugin name

src/plugins/_core/supportHelper.tsx
Masterjoona: I changed the patch to only check for venbot's messages and fixed the name stuff

src/plugins/_core/supportHelper.tsx
Masterjoona: resolved i guess?  i didnt really get what you meant but did something

Sqaaakoi: > resolved i guess? i didnt really get what you meant but did something

I need to try it for myself to see how natural your implementation of it is

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +262,55 @@ export default definePlugin({
...
+                    <ExcludedPluginCard
+                        name={pluginName}
+                        description={embed.rawDescription}
+                        key={pluginName}
+                    />
+                </div>
+            );
+        }
+
+        const onRestartNeeded = () => showRestartAlert(<p>You have to restart Vencord to apply the changes!</p>);
```
Sqaaakoi, line 282: I feel like this message could be reworded; the language usage is unnatural

Sqaaakoi (commented):
I love it so far! I still feel there are a few more details that need to be added/tweaked

src/plugins/_core/supportHelper.tsx
```diff
@@ -146,13 +149,22 @@ export default definePlugin({
...
+            replacement: {
+                match: /BEGINNING_DM\.format\(\{.+?\}\),(?=.{0,300}(\i)\.isMultiUserDM)/,
+                replace: "$& $self.renderContributorDmWarningCard({ channel: $1 }),"
+            },
+        },
+        {
+            find: "this.renderInlineMediaEmbed",
+            replacement: {
+                match: /return null(?<=var \i;let{embed:.{1,80})/,
+                replace: `if(this.props?.message?.author?.id==='${VENBOT_USER_ID}'&&this.props?.embed?.url?.startsWith('https://vencord.dev/plugins/')){return $self.renderPluginCard({embed:this.props.embed})}$&`
```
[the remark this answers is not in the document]
Masterjoona: this is kinda fire

src/plugins/_core/supportHelper.tsx
```diff
@@ -250,6 +262,55 @@ export default definePlugin({
         );
     }, { noop: true }),
 
+    renderPluginCard: ErrorBoundary.wrap(({ embed }: { embed: Embed; }) => {
+        const pluginName = embed.url!.slice(28);
+        const plugin = plugins[pluginName];
```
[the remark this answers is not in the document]
Masterjoona: yop, good idea

Vendicated:
maybe instead of completely replacing the original message, we could instead add the plugin card as Accessory? So the message will have both

please also resolve conflicts 🙏🏻

Masterjoona:
fixeddd

Vendicated:
I'm gonna rework how plugin cards look soon with components v2

let's wait for that and make these buttons look less crap then (not the ugly blue buttons)

src/components/settings/tabs/plugins/PluginCard.tsx
```diff
@@ -22,14 +24,23 @@ const { startDependenciesRecursive, startPlugin, stopPlugin, isPluginEnabled } =
 
 export const ButtonClasses = findByPropsLazy("button", "disabled", "enabled");
 
+const ExcludedReasons: Record<"web" | "discordDesktop" | "vencordDesktop" | "desktop" | "dev", string> = {
```
Vendicated: is this just copy pasted from settings? best to just export it from somewhere to avoid fragmentation

Vendicated: oh this is settings dont mind me

Vendicated:
I'm not really a fan of all these changes to the settings. the code should for the most part be only in the plugin file, settings shouldn't have to know about any of this

Masterjoona:
reverted those changes

src/components/settings/tabs/plugins/index.tsx
```diff
@@ -104,6 +97,18 @@ function ExcludedPluginsList({ search }: { search: string; }) {
     );
 }
 
+export const depMap: Record<string, string[]> = {};
```
Vendicated: this shouldn't be here. if you really need this exported, make it a function (and call it from the component and anywhere else)

src/components/settings/tabs/plugins/PluginCard.tsx
```diff
@@ -108,3 +120,35 @@ export function PluginCard({ plugin, disabled, onRestartNeeded, onMouseEnter, on
             } />
     );
 }
+
+export function UnavailablePluginCard({ name, description, isMissing }: { name: string; description: string, isMissing: boolean; }) {
```
Vendicated: why here? the settings will never use this component

src/plugins/_core/supportHelper.tsx
```diff
@@ -236,6 +239,23 @@ export default definePlugin({
     renderMessageAccessory(props) {
         const buttons = [] as JSX.Element[];
 
+        const createButton = (key: string, onClick: () => Promise<void>, label: string, color?: string) => {
+            const [disabled, setDisabled] = useState(false);
```
Vendicated: you can't call hooks in if statements so this is not correct

Vendicated: just remove this function and use normal react

src/plugins/_core/supportHelper.tsx
```diff
@@ -245,71 +265,47 @@ export default definePlugin({
...
                 }
             }
         }
 
-        return buttons.length
-            ? <Flex>{buttons}</Flex>
-            : null;
+        return (
+            <>
+                <Flex>{buttons}</Flex>
```
Vendicated:
```suggestion
                {buttons.length > 0 && <Flex>{buttons}</Flex>}
```

src/plugins/_core/supportHelper.tsx
```diff
@@ -327,3 +323,94 @@ export default definePlugin({
         );
     }, { noop: true }),
 });
+
+function UrlToPluginCard(url: string, description: string) {
```
Vendicated: use react correctly. this should be a proper component

src/plugins/_core/supportHelper.tsx
```diff
@@ -327,3 +323,94 @@ export default definePlugin({
...
+    const settings = useSettings();
+    const update = useForceUpdater();
+
+    const pluginName = new URL(url).pathname.split("/")[2];
+    const p = plugins[pluginName];
+    const excludedPlugin = ExcludedPlugins[pluginName];
+
+    const onRestartNeeded = () => Alerts.show({
```
Vendicated: this should be below the return

src/plugins/_core/supportHelper.tsx
```diff
@@ -327,3 +323,94 @@ export default definePlugin({
...
+            <UnavailablePluginCard
+                name={pluginName}
+                description={description}
+                isMissing={!p}
+                key={pluginName}
+            />
+        );
+    }
+
+    const isRequired = p.required || p.isDependency || depMap[p.name]?.some(d => settings.plugins[d].enabled);
```
Vendicated: this shouldn't be here to avoid fragmentation

src/plugins/_core/supportHelper.tsx
```diff
@@ -327,3 +323,94 @@ export default definePlugin({
...
+            onRestartNeeded={onRestartNeeded}
+            disabled={false}
+            plugin={p}
+            update={update}
+            key={p.name}
+        />
+    );
+}
+
+function PluginCards({ message }: { message: Message; }) {
```
Vendicated:
```suggestion
const PluginCards = ErrorBoundary.wrap(function PluginCards({ message }: { message: Message; }) {
```
instead of using the component manually

src/plugins/_core/supportHelper.tsx
```diff
@@ -327,3 +323,94 @@ export default definePlugin({
...
+    const embedPlugins = message.embeds?.map(embed => {
+        if (!embed.url?.startsWith("https://vencord.dev/plugins/")) return null;
+        return UrlToPluginCard(embed.url!, embed.rawDescription);
+    });
+
+    const pluginCards = embedPlugins.filter(Boolean);
```
Vendicated:
```suggestion
    const embedPlugins = message.embeds
    	?.map(embed => {
        	if (!embed.url?.startsWith("https://vencord.dev/plugins/")) return null;
        	return UrlToPluginCard(embed.url!, embed.rawDescription);
    	});
		.filter(isNonNullish);
```

src/components/settings/tabs/plugins/index.tsx
Masterjoona: Why should it be a function? If it was a function it would have to loop over plugins n times for the same result, why not just loop once

src/plugins/_core/supportHelper.tsx
Masterjoona: neither should the plugincard code below be there but you werent fan of the changes to settings...

src/components/settings/tabs/plugins/PluginCard.tsx
```diff
@@ -50,15 +59,15 @@ export function PluginCard({ plugin, disabled, onRestartNeeded, onMouseEnter, on
             if (restartNeeded) {
                 // If any dependencies have patches, don't start the plugin yet.
                 settings.enabled = true;
-                onRestartNeeded(plugin.name, "enabled");
+                onRestartNeeded(plugin.name, wasEnabled ? "disabled" : "enabled");
```
Vendicated: this is always false here because of the if (!wasEnabled)
```suggestion
                onRestartNeeded(plugin.name, "enabled");
```

src/components/settings/tabs/plugins/index.tsx
Vendicated: because circular imports are nasty.

PR #2733, opened 2024-07-27
Fix showConnections and adds back dropdowns for permissionsViewer and…

Masterjoona:
Seemingly they added scrolling back to the popout so this also adds back dropdowns for permissionsviewer and reviewdb. Toggleable between the current buttons and these dropdowns


Vendicated:
this is wayy too much and doesn't fit at all into the small profiles

Vendicated:
Cherry-picked your ShowConnections fix via https://github.com/Vendicated/Vencord/commit/3013c669c035a9e6447aaa87d16c0f6a985c8f5f, thanks for the fix!

But the rest is not mergeable. Maybe in some compacted way but definitely not like this

PR #2707, opened 2024-07-18
[WIP] package vencord as asar

Vendicated:
TODO:
- make git updater work
- test
- update installer

PR #2735, opened 2024-07-27
Fix all hotlinks

TheCrazyInsanity:
Dynamically grabbing files from all these random github repos is incredibly insecure.

This isn't the best way to accomplish this goal by far, but it would be better fixed sooner rather than later.

My changes that stop hotlinking from cdns is controversial, it can safely be undone.

My editor changed some indentation in the areas with html, I can go back and fix that if needed


TheCrazyInsanity:
Oh, the test will fail because none of these uploaded files will be on main yet, if you want we can make a separate PR to push the uploaded files first.

Nuckyz:
grabbing pictures from other repos truly is so insecure

TheCrazyInsanity:
> grabbing pictures from other repos truly is so insecure

there are 4 javascript files, 1 wasm file, and 1 zip file loaded as a chrome extension
and even then, images from single maintainer repos outside of the main vencord repo, could easily get changed to illegal content, which i consider insecure.

Nuckyz:
everything remotely fetched from Vencord is either from a trusted author or commit/version pinned which means impossible to modify

also you do realize you are calling assets from the owner of Vencord and other file types such as png and css insecure

TheCrazyInsanity:
> everything remotely fetched from Vencord is either from a trusted author or commit/version pinned which means impossible to modify
> 
> also you do realize you are calling assets from the owner of Vencord and other file types such as png and css insecure

yes, pngs and css files can easily be changed to illegal content.

being a trusted author doesn't mean anything if it takes no authorization to change it, they could get hacked, go rogue, or get bribed.
being version pinned only means immutable if you 100% trust the cdn.
and commit pinned files can still be modified by abusing git rebase

TheCrazyInsanity:
for example, the polyfill attack applied to version fixed versions of polyfill, and affected thousands of sites.
https://sansec.io/research/polyfill-supply-chain-attack
the mentality of "well I'm too small to get targeted with something like that" isn't good for a program with thousands of users, and it doesn't even have to be a direct target.

Vendicated:
Unpkg and cdnjs are both trusted cdns and the versions are pinned meaning the library authors can't change the code. Fyi, cdnjs is owned by cloudflare, so if you don't trust cdnjs, you shouldn't be using Discord at all, because cloudflare can already run arbitrary code in Discord due to Discord using them as a cdn.

Images and Audios are inherently safe since they can't execute code.

All code files linked to github are pinned to long commit hash which means it's impossible for the author to change the file.

In conclusion, there is not a single unsafe source here and this PR makes little sense. Nevertheless, it is already planned to eventually remove these external dependencies by packaging them with Vencord itself, but that depends on #2707

Vendicated:
Correction: Cdnjs is not actually owned by cloudflare, but it is made by mostly cloudflare engineers so the same thing still applies.

src/main/utils/extensions.ts
```diff
@@ -71,7 +71,7 @@ export async function installExt(id: string) {
             // React Devtools v4.25
             // v4.27 is broken in Electron, see https://github.com/facebook/react/issues/25843
             // Unfortunately, Google does not serve old versions, so this is the only way
-            ? "https://raw.githubusercontent.com/Vendicated/random-files/f6f550e4c58ac5f2012095a130406c2ab25b984d/fmkadmapgofadopljbjfkapdkoienihi.zip"
+            ? "https://raw.githubusercontent.com/Vendicated/Vencord/main/assets/random-files/f6f550e4c58ac5f2012095a130406c2ab25b984d/fmkadmapgofadopljbjfkapdkoienihi.zip" //hotlink
```
Vendicated, line 74: Actually, you changed this from a long commit hash link (immutable) to a main branch link (can be replaced by us at any time)

Thus you changed this from fully safe to remote code execution...

TheCrazyInsanity: it can only be rce if you have no main branch protection, which is also bad. I can't make it a commit hash link without it already being in the repo, this was intended to be modified before merged.

TheCrazyInsanity:
> Unpkg and cdnjs are both trusted cdns and the versions are pinned meaning the library authors can't change the code. Fyi, cdnjs is owned by cloudflare, so if you don't trust cdnjs, you shouldn't be using Discord at all, because cloudflare can already run arbitrary code in Discord due to Discord using them as a cdn.
> 
> Images and Audios are inherently safe since they can't execute code.
> 
> All code files linked to github are pinned to long commit hash which means it's impossible for the author to change the file.
> 
> In conclusion, there is not a single unsafe source here and this PR makes little sense. Nevertheless, it is already planned to eventually remove these external dependencies by packaging them with Vencord itself, but that depends on #2707

"Images and Audios are inherently safe since they can't execute code." They can't execute code, but they can easily be changed to illegal content (i have said this several times already)
"All code files linked to github are pinned to long commit hash which means it's impossible for the author to change the file." isn't necessarily true, while they can't just be arbitrarily changed to whatever, they can still intentionally break things.

happyendermangit:
why bro's pr is closed :frowning_face:

Sqaaakoi:
> why bro's pr is closed ☹️

Because this solves a non-issue and makes an ugly mess of files. Even as someone who thinks hotlinking is a bad idea in Vencord, it doesn't even fix certain hotlinks that bothered me, and of those that it does fix, it does them in such a messy way (you do not need commit hashes in the filesystem like that) that I'm not surprised that this was closed

lewisakura:
Okay, since I can already foresee this going on for ages, I'm just going to lock it with the final statements:
- This PR solves nothing, really, and would be immediately superseded by #2707 anyway.
- The image assets that Vencord fetches are, once again, immutable and cannot be replaced, simply removed.
- As Vee already said, cdnjs is used by Discord themselves so if a supply chain attacked occurred there would be much bigger issues. Same with unpkg.
- Loads of your hotlinks now link to `main`, which is WORSE. I'm aware that you're trying to get us to merge files first but this now causes a security issue for this one commit. No bueno.
- You're changing immutable Discord CDN links (like the emoji links) because you believe the content can be replaced when it can't. There's no advantage to this.
- You have for some reason flagged up that InvisibleChat violates the plugin rules. It doesn't, otherwise it wouldn't have been merged (unless you mean the "untrusted APIs" rule, in which case this doesn't apply because it is trusted by us).
Anyhow, thank you for the contribution, but a lot of it is nonsense security fearmongering and won't be a problem when the ASAR PR is merged.

PR #2736, opened 2024-07-28
XSOverlay: Update to new API

### LINKED ISSUE #2734: [XSOverlay] API Changes
### Content

As of build 667, the WebSocket API has gone public, and is now the recommended way for sending notifications.

WebSocket docs: https://xsoverlay.vercel.app/Developer/API/websockets/websockets

Reference: https://store.steampowered.com/news/app/1173510/view/7792991106417394332?l=german

### Request Agreement

nyakowint:
this makes the plugin a "limited" client because of the origin not being localhost anymore, but it doesn't matter as limited clients 
can still send notifications just fine

fixes #2734


src/plugins/xsOverlay/index.tsx
```diff
@@ -248,7 +278,33 @@ export default definePlugin({
             if (shouldIgnoreForChannelType(channel)) return;
             sendMsgNotif(titleString, finalMsg, message);
         }
-    }
+    },
+
+    async start() {
```
Vendicated, line 283: you should make this a top level function. that way you don't have to define a variable for your plugin and can instead just call the method directly

src/plugins/xsOverlay/index.tsx
```diff
@@ -248,7 +278,33 @@ export default definePlugin({
...
+        if (socket) socket.close();
+        socket = new WebSocket(`ws://127.0.0.1:${settings.store.webSocketPort ?? 42070}/?client=Vencord`);
+
+        const connected = await new Promise(res => setTimeout(() => res(socket.readyState === WebSocket.OPEN), 3000));
+        if (!connected) {
+            logger.error("WebSocket connection failed, retrying after 3000ms.");
+            setTimeout(() => {
```
Vendicated, line 290: won't this spam reconnect attempts if the user doesn't have XSOverlay open? I don't think a retry is a good idea.

Why not just check websocket state & try to start if necessary on demand, whenever a notification is supposed to be sent?

src/plugins/xsOverlay/index.tsx
```diff
@@ -289,10 +344,29 @@ function sendOtherNotif(content: string, titleString: string) {
...
+        icon: "default",
         sourceApp: "Vencord"
     };
-    Native.sendToOverlay(msgData);
+    sendToOverlay(msgData);
+}
+
+function sendToOverlay(notif: NotificationObject) {
+    if (socket && socket.readyState === WebSocket.OPEN) {
+        try {
```
Vendicated, line 355: what exactly is being caught here? it doesn't seem like any error could occur in the code contained in it

src/plugins/xsOverlay/index.tsx
```diff
@@ -289,10 +344,29 @@ function sendOtherNotif(content: string, titleString: string) {
...
+                jsonData: JSON.stringify(notif),
+                rawData: null
+            };
+            socket.send(JSON.stringify(apiObject));
+        } catch (error) {
+            logger.error("Error sending to overlay:", error);
+        }
+    } else {
+        XSOverlay.start();
+    }
```
Vendicated, line 369: won't this just discard this event?

src/plugins/xsOverlay/index.tsx
```diff
@@ -136,7 +166,12 @@ const settings = definePluginSettings({
     },
 });
 
-const Native = VencordNative.pluginHelpers.XSOverlay as PluginNative<typeof import("./native")>;
+let socket: WebSocket;
+
+async function start() {
+    if (socket) socket.close();
+    socket = new WebSocket(`ws://127.0.0.1:${settings.store.webSocketPort ?? 42070}/?client=Vencord`);
```
Vendicated, line 173: it might take some time for the websocket connection to open, which will cause this to fail:
```js
start();
socket.send(...)
```	

You can solve that by making this function return a Promise that resolves once a connection has been opened:
```suggestion
    socket = new WebSocket(`ws://127.0.0.1:${settings.store.webSocketPort ?? 42070}/?client=Vencord`);
    return new Promise<void>((resolve, reject) => {
    	socket.onopen = resolve;
    	socket.onerror = reject;
    	setTimeout(reject, 3000);
    });
```

Now just
```js
await start();
socket.send(...);
```

Vendicated:
thanks!

Natsumi-sama:
Would it be possible to keep support for the old UDP protocol? there are quite a few of us still using the last Unity UI build because it's significantly more stable.

nyakowint:
sure

PR #2738, opened 2024-07-29
improve replacegooglesearch

ashley0143:


src/utils/constants.ts
```diff
@@ -538,6 +538,10 @@ export const Devs = /* #__PURE__*/ Object.freeze({
         name: "Joona",
         id: 297410829589020673n
     },
+    Ashley: {
+        name: "Ashley",
+        id: 396571938081865741n
+    },
```
nin0-dev, line 544:
```suggestion
```

don't add yourself to devs for minor changes

src/plugins/replaceGoogleSearch/index.tsx
```diff
@@ -96,7 +100,7 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, _props)
 export default definePlugin({
     name: "ReplaceGoogleSearch",
     description: "Replaces the Google search with different Engines",
-    authors: [Devs.Moxxie, Devs.Ethan],
+    authors: [Devs.Moxxie, Devs.Ethan, Devs.Ashley],
```
nin0-dev, line 103:
```suggestion
    authors: [Devs.Moxxie, Devs.Ethan],
```

same as below

src/plugins/replaceGoogleSearch/index.tsx
```diff
@@ -12,12 +12,16 @@ import { Flex, Menu } from "@webpack/common";
 
 const DefaultEngines = {
     Google: "https://www.google.com/search?q=",
-    DuckDuckGo: "https://duckduckgo.com/",
+    YouTube: "https://youtube.com/results?search_query=",
+    DuckDuckGo: "https://duckduckgo.com/?q=",
```
KrstlSkll69, line 16: `/?q=` isnt actually needed to make a search on duckduckgo its unneeded filler
*(ignore the `?ai=web` duckduckgo added that after the fact)*
Capture

KrstlSkll69: its also just something that duckduckgo will add themselves if/when its need

PR #2739, opened 2024-07-29
Base64 Links Decoder Plugin

nawelasg:
This Plugins decodes the base64 links in the client itself, It has a dedicated UI button on the ChatButtons and Open a separate Modal window to display the decoded base64 string.

### DIFF
--- a/src/plugins/decodeBase64Links/index.tsx
+++ b/src/plugins/decodeBase64Links/index.tsx
@@ -82,7 +82,7 @@ const openDecodedStrings = decodedContent => {
 export default definePlugin({
     name: "DecodeBase64Links",
     description: "Decode base64 content of any selected text in a message.",
-    authors: [{ name: "Nawelasg", id: 0n }],
+    authors: [Devs.Nawelasg],
     dependencies: ["MessagePopoverAPI"],
 
     start() {
--- a/src/plugins/decodeBase64Links/index.tsx
+++ b/src/plugins/decodeBase64Links/index.tsx
@@ -1,0 +1,120 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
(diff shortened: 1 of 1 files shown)


src/plugins/decodeBase64Links/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                </ModalFooter>
+            </ModalRoot>
+        </ErrorBoundary>
+    ));
+};
+
+export default definePlugin({
+    name: "DecodeBase64Links",
+    description: "Decode base64 content of any selected text in a message.",
+    authors: [{ name: "Nawelasg", id: 0n }],
```
nin0-dev, line 85: for your next plugins, add yourself to the Devs object instead of putting the author inline

nin0-dev (commented):
this can easily be done with a user app, i doubt this will be merged

src/plugins/decodeBase64Links/index.tsx
nawelasg: I have resolved the Devs object issue in the new commit.

nawelasg:
> this can easily be done with a user app, i doubt this will be merged

But it will more useful as a plugin in vencord, because this operates only on the client side and has nothing to do with discord. 
Whereas if its an user app or a bot, Discord will be able to see what is being shared and decoded. And the reason why someone encode such stuff, There is a reason they dont want discord to be able to see that stuff.

Koishukaze:
it's not private either way, anyone can easily decode it

nawelasg:
But its not directly under discord surveillance. Thats what it achieves.

nin0-dev:
i don't think that you have understood that for you to even get the b64 message to decrypt

you...

need to make it go through discord servers

this plugin isn't achieving anything

Vendicated:
this is
a) not useful
b) a duplicate
c) better done via a bot
d) base64 achieves nothing for privacy. use an end to end encrypted messenger like signal, matrix, xmpp, etc

PR #2740, opened 2024-07-29
feat(plugin): UnreadCountBadge

Masterjoona:


Masterjoona:
Resolved issues

Vendicated:
seems like it doesn't work for channels without loaded messages
image

Vendicated:
not really sure how one would fix that but it decreases usefulness

MBear227:
I noticed right clicking or click-and-holding a channel in the channel list will add the unread count even if you never loaded the channel, maybe this could be used to make the count work for unloaded messages?

MBear227:
the badge isnt showing up anymore, probably a complication with css, idk if you know how to fix but maybe i can help?

PR #2741, opened 2024-07-29
Improvement to the maskedLinkPaste plugin

NeoNyaa:
Added a setting which allows the user to disable link embedding for their masked pastes by wrapping the link inbetween the `<>` symbols


Vendicated:
hmm i feel like it would be more useful if it worked via modifier key?

normal paste -> embed
shift + paste -> suppress embed

NeoNyaa:
> > hmm i feel like it would be more useful if it worked via modifier key?
> > normal paste -> embed shift + paste -> suppress embed
> 
> I'd argue otherwise purely because the point of masked links is to keep messages with links neat and tidy, another issue with embeds is they strip URL parameters so things like game invits break often.
> 
> Having it be a toggle without the need to restart vencord worked well and I recon most people would opt to have links not embed when masking them.

To expand on this, from a UX standpoint, it would be better to have the user choose either embedding masked links or non embedding masked links by default. If their usual default doesn't work for a random use case, they can manually remove the `<>` wrapping the link. Using keyboard shortcuts can become a little iffy when many different operating systems have different defaults. While it may work on Windows to use `CTRL + V` or `CTRL + SHIFT + V` who's to say that it'll work on Linux for example which can have some very random shortcuts.

I'd much prefer setting a default and manually removing the `<>` wrapper than fighting with system shortcuts for whatever reason, especially if the shortcut is used often for the OS and some random Vencord plugin is the only thing causing issues.

src/plugins/maskedLinkPaste/index.ts
```diff
@@ -28,10 +39,17 @@ export default definePlugin({
...
+                SlateTransforms.insertText(
+                    editor,
+                    `[${editor.operations[0].text}](<${content}>)`
+                );
+            } else {
+                SlateTransforms.insertText(
+                    editor,
+                    `[${editor.operations[0].text}](${content})`
+                );
+            }
```
Vendicated, line 52:
```suggestion
			const link = settings.store.disableEmbed ? `<${content}>` : content;
            SlateTransforms.insertText(
                editor,
                `[${editor.operations[0].text}](${link})`
            );
```

src/plugins/maskedLinkPaste/index.ts
```diff
@@ -4,18 +4,29 @@
...
+        default: true,
+        description: "Wrap the link inbetween < > to prevent it from embedding.",
+        restartNeeded: false
+    }
+});
+
 export default definePlugin({
     name: "MaskedLinkPaste",
-    authors: [Devs.TheSun],
+    authors: [Devs.TheSun, Devs.Neo],
```
Vendicated, line 27: please don't add yourself to authors for such a minor change

NeoNyaa: I was following the documentation for what to do :/

Vendicated: that's okay, i wasn't blaming you or anything!

Nuckyz:
This plugin is a stock Discord feature now.

PR #2743, opened 2024-07-30
feat(plugins/openInApp) Add Apple Music support 

surgedevs:
Adds opening Apple Music links (https://music.apple.com) in iTunes. Technically this works with any Apple URL, though a lot of them crash the app, so I decided its best to only support Apple Music for now.

The same URL schema does work on Cider aswell, but due to upcoming (breaking) changes I've decided to wait on supporting Cider:
image


surgedevs:
It was decided that the plugin needs a rework so this isn't needed anymore.

PR #2744, opened 2024-07-30
feat(plugins/openInApp) Refactor code and add Apple Music support

surgedevs:
Refactors the code of the plugin to make it easier to add new apps.
Adds Apple Music via iTunes support.

### DIFF
--- a/src/plugins/openInApp/index.ts
+++ b/src/plugins/openInApp/index.ts
@@ -18,52 +18,70 @@
 
 import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin, { OptionType, PluginNative } from "@utils/types";
+import definePlugin, { OptionType, PluginNative, SettingsDefinition } from "@utils/types";
 import { showToast, Toasts } from "@webpack/common";
 import type { MouseEvent } from "react";
 
-const ShortUrlMatcher = /^https:\/\/(spotify\.link|s\.team)\/.+$/;
-const SpotifyMatcher = /^https:\/\/open\.spotify\.com\/(track|album|artist|playlist|user|episode)\/(.+)(?:\?.+?)?$/;
-const SteamMatcher = /^https:\/\/(steamcommunity\.com|(?:help|store)\.steampowered\.com)\/.+$/;
-const EpicMatcher = /^https:\/\/store\.epicgames\.com\/(.+)$/;
-const TidalMatcher = /^https:\/\/tidal\.com\/browse\/(track|album|artist|playlist|user|video|mix)\/(.+)(?:\?.+?)?$/;
-const ItunesMatcher = /^https:\/\/music\.apple\.com\/[a-z]{2}\/?(album|artist|playlist|song|curator)\/([^/?#]+)\/?([^/?#]+)?(?:\?.*)?(?:#.*)?$/;
+type URLReplacementRule = {
+    match: RegExp;
+    replace: (...matches: string[]) => string;
+    description: string;
+    shortlinkMatch?: RegExp;
+    accountViewReplace?: (userId: string) => string;
+};
 
-const settings = definePluginSettings({
+// Do not forget to add protocols to the ALLOWED_PROTOCOLS constant
+const urlReplacementRules: Record<string, URLReplacementRule> = {
     spotify: {
-        type: OptionType.BOOLEAN,
+        match: /^https:\/\/open\.spotify\.com\/(track|album|artist|playlist|user|episode)\/(.+)(?:\?.+?)?$/,
+        replace: (_, type, id) => `spotify://${type}/${id}`,
         description: "Open Spotify links in the Spotify app",
-        default: true,
+        shortlinkMatch: /^https:\/\/spotify\.link\/.+$/,
+        accountViewReplace: userId => `spotify:user:${userId}`,
     },
     steam: {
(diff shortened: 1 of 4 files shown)


src/plugins/openInApp/index.ts
```diff
@@ -18,46 +18,70 @@
...
+import definePlugin, { OptionType, PluginNative, SettingsDefinition } from "@utils/types";
 import { showToast, Toasts } from "@webpack/common";
 import type { MouseEvent } from "react";
 
-const ShortUrlMatcher = /^https:\/\/(spotify\.link|s\.team)\/.+$/;
-const SpotifyMatcher = /^https:\/\/open\.spotify\.com\/(track|album|artist|playlist|user|episode)\/(.+)(?:\?.+?)?$/;
-const SteamMatcher = /^https:\/\/(steamcommunity\.com|(?:help|store)\.steampowered\.com)\/.+$/;
-const EpicMatcher = /^https:\/\/store\.epicgames\.com\/(.+)$/;
-const TidalMatcher = /^https:\/\/tidal\.com\/browse\/(track|album|artist|playlist|user|video|mix)\/(.+)(?:\?.+?)?$/;
+type URLReplacementRule = {
```
Vendicated, line 25:
```suggestion
interface URLReplacementRule {
```

src/plugins/openInApp/index.ts
```diff
@@ -18,46 +18,70 @@
...
-        type: OptionType.BOOLEAN,
+        match: /^https:\/\/tidal\.com\/browse\/(track|album|artist|playlist|user|video|mix)\/(.+)(?:\?.+?)?$/,
+        replace: (_, type, id) => `tidal://${type}/${id}`,
         description: "Open Tidal links in the Tidal app",
-        default: true,
-    }
-});
+    },
+    itunes: {
+        match: /^https:\/\/music\.apple\.com\/[a-z]{2}\/?(album|artist|playlist|song|curator)\/([^/?#]+)\/?([^/?#]+)?(?:\?.*)?(?:#.*)?$/,
```
Vendicated, line 60: is this [a-z]{2} for locale matching? is it guaranteed that part is always present?

src/plugins/openInApp/index.ts
```diff
@@ -89,61 +113,27 @@ export default definePlugin({
...
+                showToast("Opened link in native app", Toasts.Type.SUCCESS);
 
-            const match = TidalMatcher.exec(url);
-            if (!match) break tidal;
+                const newUrl = url.replace(rule.match, rule.replace);
+                VencordNative.native.openExternal(newUrl);
 
-            const [, type, id] = match;
-            VencordNative.native.openExternal(`tidal://${type}/${id}`);
+                console.log(`Opening ${url} (Resolved to ${newUrl}) in ${key} app`);
```
Vendicated, line 132: Either remove this console.log or use a dedicated Logger (from `@vencord/utils`)

src/plugins/openInApp/index.ts
```diff
@@ -155,14 +145,12 @@ export default definePlugin({
...
-        if (platformType === "spotify" && settings.store.spotify) {
-            VencordNative.native.openExternal(`spotify:user:${userId}`);
-            event.preventDefault();
-        } else if (platformType === "steam" && settings.store.steam) {
-            VencordNative.native.openExternal(`steam://openurl/https://steamcommunity.com/profiles/${userId}`);
-            showToast("Opened link in Steam", Toasts.Type.SUCCESS);
-            event.preventDefault();
+    handleAccountView(_, platformType: string, userId: string) {
+        for (const [key, rule] of Object.entries(urlReplacementRules)) {
+            if (key === platformType && rule.accountViewReplace && pluginSettings.store[key]) {
```
Vendicated, line 150: no need for a loop here :P 
```suggestion
        const rule = urlReplacementRules[platformType];
        if (rule?.accountViewReplace && pluginSettings.store[platformType]) {
```

src/plugins/openInApp/index.ts
```diff
@@ -18,46 +18,70 @@
...
+    match: RegExp;
+    replace: (...matches: string[]) => string;
+    description: string;
+    shortlinkMatch?: RegExp;
+    accountViewReplace?: (userId: string) => string;
+};
 
-const settings = definePluginSettings({
+// Do not forget to add protocols to the ALLOWED_PROTOCOLS constant
+const urlReplacementRules: Record<string, URLReplacementRule> = {
```
Vendicated, line 34: I would use PascalCase / SCREAMING_SNAKE_CASE for constant stuff like this so it's clearer it's not a local variable

```suggestion
const UrlReplacementRules: Record<string, URLReplacementRule> = {
```

src/plugins/openInApp/index.ts
surgedevs: Forgot to remove the console.log.... its so over

src/plugins/openInApp/index.ts
surgedevs: Im not sure to be honest, from all the links I've gotten from iTunes/Apple Music and tested they've included the locale. But then again, without passing locale iTunes either crashes or errors. There are also no locale problems as iTunes translates the page again (i.e. using `de` as the locale on an English device also displays the page in English) so I see no issue using `us` in the replace here. Anyway just to be safe I'll update the match to have the locale optional.

PR #2748, opened 2024-08-01
fix(BetterFolders): Close folders when switching accounts

### LINKED ISSUE #1644: [Bug] BetterFolders doesn't close an open folder sidebar when switching accounts
### Discord Account

alosh

The server folder sidebar stays open, and empties itself so it's basically just another sidebar that you can't close

If you open a folder with the blank sidebar the servers will fulfill but the sidebar but still won't close/go away when closing the server folder you just opened

If you go back to the account that opened the sidebar you can close it just fine and the servers fulfill (unintended fix?)

**I'm pretty sure this is intended feature** but the empty sidebar persists upon restarts

I expect the plugin to close the open sidebar when the account switches

1. Have 2 accounts in the Account Switcher
2. Stay on 1 account, open a server folder
3. Switch to another account while keeping that server folder open
4. Profit

### Errors

```
N/A
```

### Request Agreement

n1ckoates:
Closes #1644

### DIFF
--- a/src/plugins/betterFolders/index.tsx
+++ b/src/plugins/betterFolders/index.tsx
@@ -249,6 +249,10 @@ export default definePlugin({
                     dispatchingFoldersClose = false;
                 });
             }
+        },
+
+        LOGOUT() {
+            closeFolders();
         }
     },
 


Vendicated:
thanks for the fix, nick

PR #2749, opened 2024-08-01
Introduce StyleListener API

CodeF53:
Adds an API for listening to styles as they are added to the app via webpack

Replaces #2662 

Will be used by future plugins of mine

### DIFF
--- a/src/plugins/_api/styleListener.ts
+++ b/src/plugins/_api/styleListener.ts
@@ -1,0 +1,59 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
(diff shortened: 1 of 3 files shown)


src/plugins/_api/styleListener.ts
CodeF53, line 1: @Nuckyz @Vendicated How can I make `StyleListenerAPI` require restart when enabled

Enabling it hot doesn't work because it doesn't have a way to retroactively fetch old styles

Nuckyz: you already fetch the existing styles on start, why does enabling it hot doesn't work?

ZoamIl:
cool idea but did your idea die

Scyye:
> cool idea but did your idea die

what?

PR #2753, opened 2024-08-02
XSOverlay: Return old API for compatibility 

nyakowint:
beats the allegations i mean resolves this oversight https://github.com/Vendicated/Vencord/pull/2736#issuecomment-2261782256 (?)


src/plugins/xsOverlay/index.tsx
```diff
@@ -349,6 +356,10 @@ function sendOtherNotif(content: string, titleString: string) {
 }
 
 async function sendToOverlay(notif: NotificationObject) {
+    if (!IS_WEB && settings.store.preferUDP) {
```
Sqaaakoi, line 359: Vesktop can use plugin natives iirc

src/plugins/xsOverlay/index.tsx
```diff
@@ -102,6 +102,11 @@ const settings = definePluginSettings({
             await start();
         }
     },
+    preferUDP: {
+        type: OptionType.BOOLEAN,
+        description: "Enable if you use an older build of XSOverlay unable to connect through websockets. This setting is ignored on web.",
+        default: false
```
Vendicated, line 108:
```suggestion
        default: false,
        disabled: () => IS_WEB
```

src/plugins/xsOverlay/index.tsx
```diff
@@ -178,6 +183,8 @@ async function start() {
     });
 }
 
+const Native = VencordNative.pluginHelpers.XsOverlay as PluginNative<typeof import("./native")>;
```
Vendicated, line 186: do u even test 😭 
```suggestion
const Native = VencordNative.pluginHelpers.XSOverlay as PluginNative<typeof import("./native")>;
```

src/plugins/xsOverlay/index.tsx
nyakowint: I didnt know you could do that lol

src/plugins/xsOverlay/index.tsx
nyakowint: actual web cant lol

PR #2754, opened 2024-08-02
Extend command argument value types

Commandtechno:
They can be a bunch of different things, could also do any if that makes it easier but string is annoying because I have to keep writing @ts-ignore

example boolean option


Vendicated:
lint fails due to the change

as for your own code, I generally recommend using Vencord's `findOption` helper. now it'll be typed:
```ts
const filesOption = findOption<boolean>(options, "files");
```

PR #2762, opened 2024-08-05
fix(ServerInfo): don't display avatar in modal when MentionAvatars is enabled

ryanccn:
| Before | After |
| - | - |
| before | after |


Vendicated:
why is it so squished :o 

maybe it would be better to instead make the MentionAvatars plugin not apply to that location

Vendicated:
thanks but superseded by https://github.com/Vendicated/Vencord/commit/e6994e1946ea2b34c0b6b842d3a323beb9cfa3ef

PR #2763, opened 2024-08-05
Vc notifier plugin

konek101:
this plugin adds notifications to the voice channels for now only the notification for when the voice channel has a slot available for the user but i want to expand it to allow for overall custom notifications if anyone has any ideas i started with this notification as it was an issue of mine that this is a solution for 

its just an addition to the channel context menu that pops up this

image

i understand if it is to little for a plugin to be added but if you have ideas for notifications please let me know :)


Scyye:
What?

What's the use case for this?
Why would you need a notification when the channel youre in is one person away from filling?

sadan4:
> What?
> 
> What's the use case for this? Why would you need a notification when the channel youre in is one person away from filling?

while it seems niche, some servers cap their voice chats at low member counts
this seems like it would notify you when a slot opens up so you can join

konek101:
well yes it would notify you when a slot for you would open up and the notification when clicked will join you to that channel but as i said i want to make it have more than that notification its just that i dont have ideas for it

konek101:
if you want to know i started it as a thing for me cuz i played games in servers that capped their vcs to the max amount of players allow in a game lobby and you had to wait for slot to open up and join before someone else

Vendicated:
way too niche

PR #2767, opened 2024-08-07
fix(SpotifyControls): sync with external shuffle toggles

Masterjoona:
https://github.com/Vencord/plugin-requests/issues/1006


src/plugins/spotifyControls/index.tsx
```diff
@@ -70,21 +70,20 @@ export default definePlugin({
...
         {
             find: 'repeat:"off"!==',
-            replacement: {
-                match: /repeat:"off"!==(.{1,3}),/,
-                replace: "actual_repeat:$1,$&"
-            }
+            replacement: [
+                {
+                    // Discord doesn't give you shuffle state and the repeat kind, only a boolean
+                    match: /repeat:"off"!==(.{1,3}),/,
```
Vendicated, line 78: i know this is old but always good to refactor old stuff while you're already editing it
```suggestion
                    match: /repeat:"off"!==(\i),/,
```

src/plugins/spotifyControls/index.tsx
```diff
@@ -70,21 +70,20 @@ export default definePlugin({
...
+                    replace: "shuffle:arguments[2].shuffle_state,actual_repeat:$1,$&"
```
Vendicated, line 79:
```suggestion
                    replace: "shuffle:arguments[2]?.shuffle_state??false,actual_repeat:$1,$&"
```

Vendicated:
thanks!

PR #2768, opened 2024-08-08
Expose "translate to" settings publicly

Metadorius:
Reason: I recommended this plugin to my non-English-speaking friends, and apparently you can't edit the target language without manually exporting/importing the settings. I realize it's probably not the best impl. since it's unconstrained to the list of languages unlike the select option type, but I am not sure if you can supplement a dynamicly changing list of languages based on translation provider since I am not familiar with the "framework" you people are working in. If anyone can nudge me in the right direction and/or implement the option list it would be great.

### DIFF
--- a/src/plugins/translate/settings.ts
+++ b/src/plugins/translate/settings.ts
@@ -30,7 +30,7 @@ export const settings = definePluginSettings({
         type: OptionType.STRING,
         description: "Language that received messages should be translated to",
         default: "en",
-        hidden: true
+        hidden: false
     },
     sentInput: {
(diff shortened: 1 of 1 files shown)


Vendicated:
you're supposed to use the translate button in the chatbar

Metadorius:
Lol, I thought it's just a button that translates the message. There's nothing in the UI indicating that it's not simply a translation button, in fact even the settings reference it like so.

image

Is there a way we could make it a bit more obvious so such cases won't happen and I don't need to explain to people how to use it?

PR #2770, opened 2024-08-09
feat(plugin): Custom background plugin

NexWan:
Some basic implementation which targets the custom-theme-background class of Discord and overrides it with an image, it doesn't affect any other style (as far as I've tested it)

Some images with the plugin active:
image
image


Vendicated:
just use css

PR #2773, opened 2024-08-10
fix PictureInPicture

lumap:
discord added an extra parameter the children array and broke this 😕 here's a fix for that. readded myself as the dev as well, for good this time i swear


Vendicated:
thanks

PR #2776, opened 2024-08-11
OpenInApp: Add support for localization in Spotify URL regex

vxray:
The previous implementation only supported English links because only non-English Spotify URLs include the /intl-xx/ prefix


Nuckyz:
Thank you!

PR #2782, opened 2024-08-14
new plugin NoMaskedLinkPaste

SupertigerDev:
This plugin disables the feature where when you highlight text and paste text to convert it to a masked link.
I personally find this feature SOO annoying when i just want to quickly edit a message. 

For example, I send a wrong link by accident.
1. Press the up arrow key
2. Ctrl + A, CTRL + V + Enter

And now my day is ruined, I have added extra work for myself ☹️ 

I know that I should've created an issue first, But I don't mind if this PR gets declined. It was all a fun and learning process.


src/plugins/noMaskUrlPaste/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default definePlugin({
+    name: "NoMaskUrlPaste",
+    authors: [Devs.CatNoir],
+    description: "Pasting a link while having text selected will not paste a hyperlink.",
+    patches: [
+        {
+            find: ".selection,preventEmojiSurrogates:",
+            replacement: {
+                match: "if(null!=e.selection&&s.M8.isExpanded(e.selection))",
```
Sqaaakoi, line 18: You cannot copy paste raw Discord code here as the variable names will change when Discord updates this module. Use regex combined with Vencord's special identifier selector `\i`. You should capture the `e` variable, and you will also need to find a new match that can capture the `n` variable used in the replacement.

src/plugins/noMaskUrlPaste/index.ts
```diff
@@ -0,0 +1,23 @@
...
+                replace: "return e.insertText(n);$&"
```
Sqaaakoi, line 19: `e` and `n` should be replaced with regex captures as described in the previous comment

Sqaaakoi (commented):
Good plugin, but the patch needs to be redesigned to use regex capture groups for variable names instead of just an exact string find

src/plugins/noMaskUrlPaste/index.ts
Masterjoona: This is supposed to return a boolean with it as well, though seemingly it doesnt matter I feel like to have same logic as discord. Personally I would just make the if statement be always false

Sqaaakoi: I have taken a look at this code and you could just replace this with `if(false)` and forget the capture groups

Sqaaakoi: > This is supposed to return a boolean with it as well, though seemingly it doesnt matter I feel like to have same logic as discord. Personally I would just make the if statement be always false

That boolean is likely if the paste was successful or not

SupertigerDev: I think I managed to do it, seems to be working aswell!

SupertigerDev: oh wait! i missed the last e

SupertigerDev: Fixed, sorry 😅

Masterjoona: you forgot change the M8 :sob: personally i would just do 
```
{
    match: "return!1;if\(",
    replace: "$&false&&"
}
```

Kyuuhachi: Regexes only match inside the module matched by the `find`, and last I checked that module was pretty small, maybe a dozen or two lines. So little risk of false positives I would think.

Vendicated: the current match is fine and better than something overly generic. Just need to make sure not to use minified variable names

SupertigerDev:
Just want to confirm that it's all ready to merge :weary:

src/plugins/noMaskUrlPaste/index.ts
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants.js";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "NoMaskUrlPaste",
+    authors: [Devs.CatNoir],
+    description: "Pasting a link while having text selected will not paste a hyperlink.",
```
Vendicated, line 13:
```suggestion
    description: "Pasting a link while having text selected will not paste as masked URL",
```

Vendicated:
thanks for the contribution! 🐇

PR #2786, opened 2024-08-16
fix Online Themes tab

HAHALOSAH:
The CSS class for textarea is no longer in the main webpack bundle, so this PR just copies over the styles


Nuckyz:
Thanks for the fix!

Issue #2787, opened 2024-08-16
[Bug] XSOverlay Notifications Don't show PFP


TypicalZedF:
### Discord Account

typicalzedf

### What happens when the bug or crash occurs?

When the plugin is enabled and a notification is displayed, it fails to show a profile picture correctly and instead shows a gray icon.
![image](https://github.com/user-attachments/assets/4d9da8e3-8365-4b9c-87ef-bd27e83c702e)

### What is the expected behaviour?

The message should correctly show a profile picture as shown below
![image](https://github.com/user-attachments/assets/aaaf52c4-33ae-438f-8f2d-254aa737d0c4)

### How do you recreate this bug or crash?

1. You must have SteamVR with XSOverlay installed of course. The XSOverlay Plugin should also be enabled!
2. Have a buddy send you a message while you're in VR! You'll get no profile picture in the notification!

### Errors

There seem to be no errors that I can gather, this is more of a visual bug of sorts or the plugin is failing to grab users profile pictures or is failing to display it properly.

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [X] I have read the requirements for opening an issue above

TypicalZedF:
Another bug I'd like to report on the side of this is the XSOverlay seems to ignore the size of the message and it breaks the pop-up showing the text breaking out of the notification bubble.

nyakowint:
Any faults in the notification itself you can blame xiexe for... but I'll fix up the profile picture thing right now

nyakowint:
> Another bug I'd like to report on the side of this is the XSOverlay seems to ignore the size of the message and it breaks the pop-up showing the text breaking out of the notification bubble. Perhaps if the notification is very long.. only show the first few words followed with "..." to show there is more in the message?

I cannot reproduce this, and regardless is something you should create a bug report with XSOverlay's dev instead.

TypicalZedF:
Is this fixed? It was closed and marked as complete but I didn't notice a fix.

nyakowint:
i didnt fix your second comment, cause that is not really an issue with the vencord plugin

the linked pr fixes what was originally reported in this issue, and it works from my testing (just now).. if you are having a different issue you'll need to elaborate, or send some logs or something

TypicalZedF:
The images are still gray as per the original intent of the reported issue. After checking for updates, there is not update available for the XSOverlay Plugin and images still are not loading for the notifications

nyakowint:
[...]

PR #2788, opened 2024-08-16
XSOverlay: fix profile images

nyakowint:
This fixes the root cause of #2787
since this issue only happened from lack of thorough testing this will be a draft until i'm able to test it


Vendicated:
hop on
```js
fetch(url)
	.then(res => res.blob())
	.then(blob => new Promise<string>(resolve => {
		const r = new FileReader();
		r.onload = () => resolve(r.result.split(",")[1]);
		r.readAsDataURL(blob);
	}));
```

PR #2790, opened 2024-08-18
feat(Plugin): CopyFileContents

Obsidianninja11:
Adds a button to copy the contents of a text file (even if the text isn't in the tiny preview).
After clicking, the copy icon will turn into a checkmark (like in large codeblocks), and the text will be copied to your clipboard.
For files over 50kb (50 thousand characters) that don't fully show in the "View whole file" preview, it shows "File too large to copy". I made a version that fetches files that are large, but I don't think that's necessary for 99% of users, and it could lead to problems if a user tries to copy a huge file.
This pr also updates the copy icon in `components\icons.tsx` (because the icon is used in this plugin), as the icon that was in there is long outdated.

<details><summary>Images</summary>

image
image

A file containing 51kb that is unable to be copied (the first 683 numbers in the fibonacci sequence in this example).
image
</details>


src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+export default definePlugin({
+    name: "CopyFileContents",
+    description: "Adds a button to copy text file attachment contents.",
+    authors: [Devs.Obsidian, Devs.Nuckyz],
+    patches: [
+        {
+            find: ".Messages.PREVIEW_BYTES_LEFT.format(",
+            replacement: {
+                match: /\.footerGap.+?url:\i,fileName:\i,fileSize:\i}\),(?<=fileContents:(\i),bytesLeft:(\i).+?)/g,
+                replace: "$&$self.AddCopyButton({ fileContents: $1, bytesLeft: $2 }),"
```
Scyye, line 30: I dont think youre supposed to put spaces in replaces
```suggestion
                replace: "$&self.AddCopyButton({$1,$2}),"
```
verify that works, im not sure, and im tired

src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+    ],
+    AddCopyButton: ErrorBoundary.wrap(({ fileContents, bytesLeft }: { fileContents: string, bytesLeft: number; }) => {
```
Scyye, line 34: shouldnt this be `addCopyButton`? 
not certain.

src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+        const [recentlyCopied, setRecentlyCopied] = useState(false);
+        return (
+            <Tooltip text={recentlyCopied ? "Copied!" : bytesLeft > 0 ? "File too large to copy" : "Copy File Contents"}>
+                {tooltipProps => (
+                    <div
+                        {...tooltipProps}
+                        className="vc-cfc-button"
```
Scyye, line 41: i think youre meant to use a class name factory for this.

Scyye (commented):
Im tired and youre asking for a review, so i gave it to ya lol
just verify everything is accurate before changing stuff.

src/plugins/CopyFileContents/index.tsx
Obsidianninja11: Class name factory seems to be for when there's multiple class names all prefixed by `vc-cfc-` (cfc in this case). So I don't think I am supposed to use it for this.

Scyye: ahh, okay.

src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+        }
+    ],
+    addCopyButton: ErrorBoundary.wrap(({ fileContents, bytesLeft }: { fileContents: string, bytesLeft: number; }) => {
+        const [recentlyCopied, setRecentlyCopied] = useState(false);
+        return (
+            <Tooltip text={recentlyCopied ? "Copied!" : bytesLeft > 0 ? "File too large to copy" : "Copy File Contents"}>
+                {tooltipProps => (
+                    <div
+                        {...tooltipProps}
+                        className="vc-cfc-button"
```
Nuckyz, line 41: Since the plugin only has two classes I would hardcode with the style attribute and delete the css file

src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    authors: [Devs.Obsidian, Devs.Nuckyz],
+    patches: [
+        {
+            find: ".Messages.PREVIEW_BYTES_LEFT.format(",
+            replacement: {
+                match: /\.footerGap.+?url:\i,fileName:\i,fileSize:\i}\),(?<=fileContents:(\i),bytesLeft:(\i).+?)/g,
+                replace: "$&$self.addCopyButton({fileContents:$1,bytesLeft:$2}),"
+            }
+        }
+    ],
```
Nuckyz, line 33:
```suggestion
    ],
    
```

Nuckyz (changes requested):

src/plugins/CopyFileContents/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import "./style.css";
+
+import ErrorBoundary from "@components/ErrorBoundary";
+import { CopyIcon, NoEntrySignIcon } from "@components/Icons";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { Clipboard, Tooltip, useState } from "@webpack/common";
+
+const CheckMarkIcon = () => {
+    return <svg aria-hidden="true" width="24" height="24" fill="none" viewBox="0 0 24 24">
```
Nuckyz, line 16:
```suggestion
    return <svg width="24" height="24" fill="currentColor" viewBox="0 0 24 24">
```

Nuckyz (changes requested):

Nuckyz:
You never change setRecentlyCopied back to false, is that intended?

src/plugins/CopyFileContents/index.tsx
Obsidianninja11: :hover doesn't work in style=

Nuckyz: right

Obsidianninja11:
> You never change setRecentlyCopied back to false, is that intended?

Yes. That's how it works in code blocks, so I recreated it here.

Scyye:
I still feel like it should

Obsidianninja11:
I added a toast when copying. It's currently stay as a checkmark with a toast (the first video).
I don't think the checkmark changing back is necessary because it can create some buggy behaviour, and I don't think it looks as good (second video).

https://github.com/user-attachments/assets/1acdfc0e-078c-4a00-ab57-90dfbc52b504

https://github.com/user-attachments/assets/0c488b46-5231-4f2f-af2c-c53d7c38bf0e

Nuckyz:
I fixed the buggy behavior and made the checkmark change back, it makes way more sense

Nuckyz:
Thank you for the new plugin!

PR #2792, opened 2024-08-18
[Webkeybinds] Don't override browser keybinds

rniii:
Ctrl+Tab already didn't work, but for some reason it seems you can actually override Ctrl+(Shift)+T (open new tab/reopen last tab)

### DIFF
--- a/src/plugins/webKeybinds.web/index.ts
+++ b/src/plugins/webKeybinds.web/index.ts
@@ -35,6 +35,7 @@ export default definePlugin({
         if (hasCtrl) switch (e.key) {
             case "t":
             case "T":
+                if (!IS_VESKTOP) return;
                 e.preventDefault();
                 if (e.shiftKey) {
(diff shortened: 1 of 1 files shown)


src/plugins/webKeybinds.web/index.ts
```diff
@@ -47,14 +48,15 @@ export default definePlugin({
...
-                break;
             case "Tab":
+                if (!IS_VESKTOP) return;
                 const handler = e.shiftKey ? KeyBinds.SERVER_PREV : KeyBinds.SERVER_NEXT;
                 handler.action(e);
                 break;
+            case ",":
+                e.preventDefault();
+                SettingsRouter.open("My Account");
+                break;
```
Scyye, line 59: Why move this?

Vendicated: more logical

PR #2796, opened 2024-08-19
Updated Anonymize File Names

TheMasterKitty:
Added default file extension selection for users to select (like ".png") so that specific file types can be anonymized automatically whereas files not in that list stay under the same name by default.


TheMasterKitty:
o shoot forgot to undo prev changes one sec

PR #2798, opened 2024-08-19
Remake of User Voice Show plugin

llytz:
uservoiceshow plugin preview

no more big buttons that take too much space in users profile, small icons in the top right corner of profile instead

now icons are also shown in direct messages list, server members list and friends list


Vendicated (commented):
Thanks for the PR, it looks a lot better than before!

Code still needs some work though

fdkchx:
I get it, but moments like this don't look good
image - Instead of broadcasting in the profile banner, I'd leave the voice channel icon, just with a different tooltip
image
image

PR #2799, opened 2024-08-22
Ignore blocked users in whoReacted avatars

lannoene:
yeah that's it. i took the relationship store thing from noblockedmessages and there they used a try catch when getting the blocked status for users. i haven't had any issues without it though. if i should add it, feel free to tell me.

### DIFF
--- a/src/plugins/whoReacted/index.tsx
+++ b/src/plugins/whoReacted/index.tsx
@@ -21,18 +21,26 @@ import { Devs } from "@utils/constants";
 import { sleep } from "@utils/misc";
 import { Queue } from "@utils/Queue";
 import { useForceUpdater } from "@utils/react";
-import definePlugin from "@utils/types";
+import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy, findComponentByCodeLazy } from "@webpack";
-import { ChannelStore, Constants, FluxDispatcher, React, RestAPI, Tooltip } from "@webpack/common";
+import { ChannelStore, Constants, FluxDispatcher, React, RestAPI, Tooltip, RelationshipStore } from "@webpack/common";
 import { CustomEmoji } from "@webpack/types";
 import { Message, ReactionEmoji, User } from "discord-types/general";
+import { definePluginSettings } from "@api/Settings";
 
 const UserSummaryItem = findComponentByCodeLazy("defaultRenderUser", "showDefaultAvatarsForNullUsers");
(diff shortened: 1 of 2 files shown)


src/plugins/whoReacted/index.tsx
```diff
@@ -155,7 +156,7 @@ export default definePlugin({
         }, [message.id]);
 
         const reactions = getReactionsWithQueue(message, emoji, type);
-        const users = Object.values(reactions).filter(Boolean) as User[];
+        const users = Object.values(reactions).filter((v) => {return v && !RelationshipStore.isBlocked(v.id)}) as User[];
```
Vendicated, line 159:
```suggestion
        const users = Object.values(reactions).filter((v) => v && !RelationshipStore.isBlocked(v.id)) as User[];
```

src/plugins/whoReacted/index.tsx
```diff
@@ -100,7 +101,7 @@ function handleClickAvatar(event: React.MouseEvent<HTMLElement, MouseEvent>) {
 export default definePlugin({
     name: "WhoReacted",
     description: "Renders the avatars of users who reacted to a message",
-    authors: [Devs.Ven, Devs.KannaDev, Devs.newwares],
+    authors: [Devs.Ven, Devs.KannaDev, Devs.newwares, Devs.lannoene],
```
Vendicated, line 104: don't add yourself as author for such minor changes

src/plugins/whoReacted/index.tsx
```diff
@@ -29,6 +29,7 @@ import { Message, ReactionEmoji, User } from "discord-types/general";
 
 const UserSummaryItem = findComponentByCodeLazy("defaultRenderUser", "showDefaultAvatarsForNullUsers");
 const AvatarStyles = findByPropsLazy("moreUsers", "emptyUser", "avatarContainer", "clickableAvatar");
+const RelationshipStore = findByPropsLazy("getRelationships", "isBlocked");
```
Vendicated, line 32: would be a good idea to move this to webpack commons since it's used in multiple places

Vendicated (commented):
This behaviour should definitely be toggleable or only apply if Hide Blocked users is enabled

sadan4:
this should really be an option, not hardcoded

Vendicated:
i think it would be way more intuitive to just depend on whether HideBlockedUsers is enabled or not instead of adding a setting

lannoene:
@Vendicated Could you link me to where HideBlockedUsers is located? I'm having trouble finding it. Thanks!

PR #2803, opened 2024-08-24
feat: hideRepliesToBlockedMessages

CodeF53:
Adds new option to NoBlockedMessages, default off:
Discord_bIarBcg95U

Without on
ShareX_rhH4I0OWJQ

With on
ShareX_wKHcRm2lWx

### DIFF
--- a/src/plugins/noBlockedMessages/index.ts
+++ b/src/plugins/noBlockedMessages/index.ts
@@ -80,15 +80,15 @@ export default definePlugin({
         if (!Settings.plugins.NoBlockedMessages.hideRepliesToBlockedMessages || !message)
             return false;
 
-		try {
-        	const { messageReference } = message;
-	        if (!messageReference) return false;
-	        
-	        const replyMessage = MessageStore.getMessage(messageReference.channel_id, messageReference.message_id);
-	        return this.isBlocked(replyMessage);
-	    } catch (e) {
+        try {
+            const { messageReference } = message;
+            if (!messageReference) return false;
+
+            const replyMessage = MessageStore.getMessage(messageReference.channel_id, messageReference.message_id);
+            return this.isBlocked(replyMessage);
+        } catch (e) {
             new Logger("NoBlockedMessages").error("Failed to check if user is blocked:", e);
-		}
+        }
     },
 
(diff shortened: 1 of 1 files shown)


src/plugins/noBlockedMessages/index.ts
```diff
@@ -60,9 +68,26 @@ export default definePlugin({
...
+    },
+
+    isReplyToBlocked(message: Message) {
+        if (!Settings.plugins.NoBlockedMessages.hideRepliesToBlockedMessages)
+            return false;
+
+        const { messageReference } = message;
+        if (!messageReference) return false;
+        const replyMessage = MessageStore.getMessage(messageReference.channel_id, messageReference.message_id);
+        return this.isBlocked(replyMessage);
```
Vendicated, line 86:
```suggestion
        if (!Settings.plugins.NoBlockedMessages.hideRepliesToBlockedMessages || !message)
            return false;

		try {
        	const { messageReference } = message;
	        if (!messageReference) return false;
	        
	        const replyMessage = MessageStore.getMessage(messageReference.channel_id, messageReference.message_id);
	        return this.isBlocked(replyMessage);
	    } catch (e) {
            new Logger("NoBlockedMessages").error("Failed to check if user is blocked:", e);
		}
```

src/plugins/noBlockedMessages/index.ts
```diff
@@ -48,10 +49,17 @@ export default definePlugin({
...
-                    replace: (_, props) => `if($self.isBlocked(${props}.message))return;`
+                    replace: (_, props) => `if($self.isBlocked(${props}.message)||$self.isReplyToBlocked(${props}.message))return;`
                 }
             ]
-        }))
+        })),
+        {
+            find: ".messageListItem",
+            replacement: {
+                match: /(?<=\i=)(?=\(0,(\i)\.jsx)/,
```
Vendicated, line 59: way too generic, please make this more specific, by adding some kind of anchor

Oroborius:
If possible add option to hide the user from userlist as well?

Feathers8:
There is a different PR for that: #2170

PR #2805, opened 2024-08-25
feat(superReactionTweaks): Allow disabling super reactions

n1ckoates:
Added a `0` option to the slider, and a check that returns `false` if the limit is `0`.


src/plugins/superReactionTweaks/index.ts
```diff
@@ -25,7 +25,7 @@ export const settings = definePluginSettings({
         description: "Max Super Reactions to play at once",
```
Nuckyz, line 25:
```suggestion
        description: "Max Super Reactions to play at once. 0 to disable playing Super Reactions",
```

Nuckyz (changes requested):

Nuckyz:
Thank you!

PR #2806, opened 2024-08-25
Remove dependency on virtual-merge, only required by fakeProfileThemes plugin

wearrrrr:
Removes dependency on [virtual-merge](https://www.npmjs.com/package/virtual-merge), this was only "required" by fakeProfileThemes, however you can just merge objects normally in this case, without worrying about reflecting the merge back onto the passed in `user` parameter. Will add myself as a plugin author if needed/wanted, but this feels like a simple enough change to not need it :3

### DIFF
--- a/src/plugins/fakeProfileThemes/index.tsx
+++ b/src/plugins/fakeProfileThemes/index.tsx
@@ -29,7 +29,6 @@ import definePlugin, { OptionType } from "@utils/types";
 import { extractAndLoadChunksLazy, findComponentByCodeLazy } from "@webpack";
 import { Button, Flex, Forms, React, Text, UserProfileStore, UserStore, useState } from "@webpack/common";
 import { User } from "discord-types/general";
-import virtualMerge from "virtual-merge";
 
 interface UserProfile extends User {
     themeColors?: Array<number>;
@@ -232,10 +231,7 @@ export default definePlugin({
             if (settings.store.nitroFirst && user.themeColors) return user;
             const colors = decode(user.bio);
             if (colors) {
-                return virtualMerge(user, {
-                    premiumType: 2,
-                    themeColors: colors
-                });
+                return { ...user, premiumType: 2, themeColors: colors };
             }
         }
         return user;
--- a/package.json
+++ b/package.json
@@ -41,8 +41,7 @@
         "fflate": "^0.8.2",
         "gifenc": "github:mattdesl/gifenc#64842fca317b112a8590f8fef2bf3825da8f6fe3",
         "monaco-editor": "^0.50.0",
-        "nanoid": "^5.0.7",
-        "virtual-merge": "^1.0.1"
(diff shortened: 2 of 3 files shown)


wearrrrr:
I also feel like nanoid is a pretty somewhat dependency here, unless it's planned to be used in more places than just the notificationLog (maybe a vencord api to generate a random ID?)

src/plugins/fakeProfileThemes/index.tsx
```diff
@@ -232,10 +231,7 @@ export default definePlugin({
             if (settings.store.nitroFirst && user.themeColors) return user;
             const colors = decode(user.bio);
             if (colors) {
-                return virtualMerge(user, {
-                    premiumType: 2,
-                    themeColors: colors
-                });
+                return { ...user, premiumType: 2, themeColors: colors };
```
Vap0r1ze, line 234:
```suggestion
                return Object.assign(user, { premiumType: 2, themeColors: colors });
```
you'd wanna return the same user object still, this is being passed to discord code so not many assumptions can be made

Vap0r1ze: this even still is probably not a great idea, i think `virtual-merge` is a fine solution

Vendicated:
this was deliberately done to keep proper expected behaviour. namely, spreading turns the object "read-only" (since it's now a copy) and breaks class instance fields / methods entirely. but we also don't want to pollute Discord's objects with fake properties because it might break things

it's also such a thin dependency that it has very little impact

ryan-0324:
> this was deliberately done to keep proper expected behaviour. namely, spreading turns the object "read-only" (since it's now a copy) and breaks class instance fields / methods entirely. but we also don't want to pollute Discord's objects with fake properties because it might break things, so we can't just write these properties on the original object
> 
> it's also such a thin dependency that it has very little impact

I don't see why the properties can't be set on the [UserProfile](https://github.com/ryan-0324/Vencord/blob/398cba4a6f4e3eb5871fb7e4eaca525140ab8753/packages/discord-types/src/general/UserProfile.ts#L11) object normally, since doing so is [type-safe](https://github.com/ryan-0324/Vencord/blob/398cba4a6f4e3eb5871fb7e4eaca525140ab8753/packages/discord-types/src/general/UserProfile.ts#L29-L34), even when considering TypeScript's soundness holes. I've been doing so in [FakeProfileThemes](https://github.com/ryan-0324/Vencord/blob/398cba4a6f4e3eb5871fb7e4eaca525140ab8753/src/plugins/fakeProfileThemes/index.tsx#L232-L233) for a while and in [my own plugin](https://github.com/ryan-0324/Vencord/blob/64707ed2d7d2825d9acf1e4182f55c503b1086c2/src/plugins/fakeProfileThemesAndEffects/lib/userProfile.ts#L17-L32) for almost a year without any issues.

PR #2809, opened 2024-08-27
feat(Plugin): Add AlwaysExpandRoles

surgedevs:
Adds a plugin to always expand the role list in profile popouts.
This was highly requested and sparked drama in the Discord server, very small plugin, not sure what else to say.

### DIFF
--- a/src/plugins/alwaysExpandRoles/index.ts
+++ b/src/plugins/alwaysExpandRoles/index.ts
@@ -1,0 +1,17 @@
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "AlwaysExpandRoles",
+    description: "Always expands the role list in profile popouts",
+    authors: [Devs.surgedevs],
+    patches: [
+        {
+            find: "action:\"EXPAND_ROLES\"",
+            replacement: {
+                match: /(let\s+\i=\d+===\i\?\i:)\d+;/,
+                replace: "$1Number.MAX_VALUE;"
+            }
+        }
+    ]
+});


Nuckyz:
We should probably migrate the settings from the  old ShowAllRoles

people who had it enabled will likely already want this enabled

surgedevs:
> We should probably migrate the settings from the old ShowAllRoles
> 
> people who had it enabled will likely already want this enabled

Added, also added the missing README.md

Sqaaakoi:
How about adding an option to hide the collapse button?

sadan4:
> How about adding an option to hide the collapse button?

could you just use CSS for that?

src/plugins/alwaysExpandRoles/README.md
```diff
@@ -0,0 +1,3 @@
+# Always Expand Roles
+
+Always expands the role list in profile popouts, like the now deleted ShowAllRoles used to do
```
Nuckyz, line 3:
```suggestion
Always expands the role list in profile popouts
```

src/plugins/alwaysExpandRoles/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import definePlugin from "@utils/types";
+
+migratePluginSettings("AlwaysExpandRoles", "ShowAllRoles");
+export default definePlugin({
+    name: "AlwaysExpandRoles",
+    description: "Always expands the role list in profile popouts",
+    authors: [Devs.surgedevs],
+    patches: [
+        {
+            find: "action:\"EXPAND_ROLES\"",
```
Nuckyz, line 30:
```suggestion
            find: 'action:"EXPAND_ROLES"',
```

src/plugins/alwaysExpandRoles/index.ts
```diff
@@ -0,0 +1,37 @@
...
+            replacement: {
+                match: /(\[\i,\i\]=\i\.useState\()!1\)/,
+                replace: "$1!0)"
```
Nuckyz, line 33:
```suggestion
                match: /(roles:\i(?=.+?(\i)\(!0\)[,;]\i\({action:"EXPAND_ROLES"}\)).+?\[\i,\2\]=\i\.useState\()!1\)/,
                replace: (_, rest) => `${rest}!0)`
```

Sorry if this regex looks like hell, it's just abusing lookaheads to make sure we are changing the correct useState

Nuckyz (changes requested):

Nuckyz:
> How about adding an option to hide the collapse button?

not worth that's so minimal

src/plugins/alwaysExpandRoles/index.ts
Sqaaakoi: Why use a replacement function?

Nuckyz: naming the capturing groups

Sqaaakoi: Also that `\2` is almost certainly a typo

Nuckyz (approved):

Nuckyz:
Thank you!

surgedevs:
> Hello, thanks for the plugin!
> 
> I think adding an option to hide the collapse button would be good as part of the plugin. Since the roles are always expanded, there's not really a reason you would want to collapse them (since you expressed your preference not to collapse).

This was already discussed above.

```css
div:has(> div[class^=collapseButton_]) {
    display: none;
}
```

Nuckyz:
> Hello, thanks for the plugin!
> 
> I think adding an option to hide the collapse button would be good as part of the plugin. Since the roles are always expanded, there's not really a reason you would want to collapse them (since you expressed your preference not to collapse).

The button is so small it makes no difference to have it there or not. Also useful when you want to collapse for some reason

PR #2812, opened 2024-08-28
feat(CustomRPC): Adds a button to disable CustomRPC directly from the account panel 

LooosaZ:
Using the GameToggleActivity button code I was able to add a similar feature to CustomRPC, where before this, you would need to enter the plugins page and disable it from there, now, all you need to do is click on the "Disable CustomRPC" button added to the account panel.
Button


surgedevs:
This change is way to niche, no one really needs a shortcut to enable/disable CustomRPC, additionally this clutters up the profile space and even breaks without a theme:
image

PR #2814, opened 2024-08-29
Some permissions changes

gorciu-official:
I think this will be better


gorciu-official:
some fucking changes included, will be reposted

PR #2816, opened 2024-08-30
feat(Plugin): EnhancedUserTags

vishnyanetchereshnya:
Replaces and extends default tags (Official, Original Poster, etc.) with a crown icon, the type of which depends on the user's permissions
Discord_uJGkaKB9qd
Discord_TOmGSpgzUy
Discord_Q6UweE6QHk
Discord_E3JuwBSavR
Discord_QR6pVKOejF


surgedevs:
This would be much better as a separate option for MoreUserTags and not as its own plugin.
https://github.com/Vendicated/Vencord/tree/main/src/plugins/moreUserTags

vishnyanetchereshnya:
> This would be much better as a separate option for MoreUserTags and not as its own plugin. https://github.com/Vendicated/Vencord/tree/main/src/plugins/moreUserTags

wdym "separate option"? what's exactly

PR #2818, opened 2024-08-31
Add invert option to dearrow plugin

### LINKED ISSUE #2783: [FEATURE REQUEST - Dearrow] Setting for the default state
### Content

A setting could be added to either enable dearrow on youtube embeds by default (current behavior) or disable it by default (and clicking on the button per embed)

### Request Agreement

electricsteve:
Fix #2783 
Just need feedback if the option should be marked restart needed. (line 174)


Vendicated:
why do you need feedback for if a restart is needed? it simply depends on if the option can apply immediately or needs a restart to work

EDM115:
@Vendicated from what I understand the option will have an immediate effect but will not work retroactively on already dearrowed embeds, unless the user reload their client

electricsteve:
Yes, its just idk if it is worth requiring a restart.

Vendicated:
ah i see. that's fine

electricsteve:
Yea ok but my question is still not answered, @EDM115 what do you think?

EDM115:
I believe it's alright, if users want to see changes on their old embeds that are already dearrowed they just have to refresh discord or close & reopen, which isn't a big problem  
I don't think that this setting needs a restart to work out of the box

electricsteve:
Ok, will remove comment in a few min, currently eating.

Nuckyz:
Instead of being an option to invert we should do something like how ShowHiddenChannels does:
image

Every time you click the toggle button it also updates the setting, so it remembers what you last chose

electricsteve:
Wtf is that option name

electricsteve:
> Instead of being an option to invert we should do something like how ShowHiddenChannels does: image
> 
> Every time you click the toggle button it also updates the setting, so it remembers what you last chose

OK will do that

electricsteve:
But wait the name of that setting contradicts the function, because it isn't the default, right?

electricsteve:
(Spelling mistake)

Nuckyz:
> But wait the name of that setting contradicts the function, because it isn't the default, right?

It's the default value which is applied the next time you view the button where it's used

src/plugins/dearrow/index.tsx
```diff
@@ -95,7 +108,9 @@ function DearrowButton({ component }: { component: Component<Props>; }) {
                     onMouseLeave={onMouseLeave}
                     className={"vc-dearrow-toggle-" + (embed.dearrow.enabled ? "on" : "off")}
                     onClick={() => {
+                        const { invert } = settings.store;
```
Nuckyz, line 111:
```suggestion
```

src/plugins/dearrow/index.tsx
```diff
@@ -95,7 +108,9 @@ function DearrowButton({ component }: { component: Component<Props>; }) {
...
                         const { enabled, oldThumb, oldTitle } = embed.dearrow;
+                        settings.store.invert = !invert;
```
Nuckyz, line 113:
```suggestion
                        settings.store.invert = !enabled;
```

Nuckyz (changes requested):

src/plugins/dearrow/index.tsx
```diff
@@ -77,6 +77,19 @@ async function embedDidMount(this: Component<Props>) {
...
+            embed.dearrow.enabled = !enabled;
+            if (oldTitle) {
+                embed.dearrow.oldTitle = embed.rawTitle;
+                embed.rawTitle = oldTitle;
+            }
+            if (oldThumb) {
+                embed.dearrow.oldThumb = embed.thumbnail.proxyURL;
+                embed.thumbnail.proxyURL = oldThumb;
+            }
+        }
```
Vendicated, line 91: this is just undoing the previous work. why not just use a condition above?

src/plugins/dearrow/index.tsx
electricsteve: Ah ok

src/plugins/dearrow/index.tsx
electricsteve: Will do!

Nuckyz:
Thank you for the contribution!

Nuckyz:
> NP! also I didn't notice enabled didn't need to be inverted. LOL

no worries, my suggestion was wrong too, I only noticed after testing haha

PR #2821, opened 2024-09-01
LastFM: Added dropdown for activity type

asleepa:
Activity type now uses a dropdown rather than 1 toggle between playing and listening.

Note: Streaming will show as Playing in the member list, however, it will show as streaming on profiles. The purple status indicator also only appears in the bottom left user card. Watching won't show the 3rd label (album name) but it seems to show how long it has been active instead.

image
image
image

### DIFF
--- a/src/plugins/lastfm/index.tsx
+++ b/src/plugins/lastfm/index.tsx
@@ -65,7 +65,9 @@ interface TrackData {
 // only relevant enum values
 const enum ActivityType {
     PLAYING = 0,
+    STREAMING = 1,
     LISTENING = 2,
+    WATCHING = 3
 }
 
 const enum ActivityFlag {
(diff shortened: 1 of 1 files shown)


Vendicated:
this seems pretty pointless and confusing for users. why would anyone ever want to use anything except listening or playing

asleepa:
Streaming can be used as a replacement word for "listening", such as Streaming <artist>, and some might also want the purple status indicator because it's different from the default online, idle, DND, and invisible.  Watching could be removed, I just added it for the sole intent of more options.

moyaimoment:
> such as Streaming (artist), and some might also want the purple status indicator because it's different from the default online, idle, DND, and invisible

Doesn't work unless you have a Twitch or YouTube link with it

PR #2824, opened 2024-09-01
PermissionsViewer: UI Cleanup, Role Icons Support, Granted By Tooltip

vishnyanetchereshnya:
Single scroller replaced by separate one for each panel
Added icon display for roles
Discord_HmjAd3fEBZ


src/plugins/permissionsViewer/styles.css
```diff
@@ -104,6 +107,12 @@
     flex-shrink: 0;
 }
 
+.vc-permviewer-perms-role-image {
```
Sqaaakoi, line 110: role icons aren't always 1:1 aspect ratio, this will stretch those icons

vishnyanetchereshnya: > role icons aren't always 1:1 aspect ratio, this will stretch those icons

uh, didn't notice, fixed

Nuckyz:
Thanks for the contribution! I went ahead and did more cleanups related to the UI and code, and also added a new tooltip to the permissions in the popoup to show what role is granting it

PR #2825, opened 2024-09-01
feat(plugins): LocalBlock, block annoying people without them knowing

junepark678:
There's times when there's an annoying user you want to block but just can't because you know they'd retaliate. This is for those times. This plugin allows you to block users discretely.


junepark678:
How this is different from #2492: this directly modifies the relationshipstore (which surprisingly doesn't push the changes to discord, only pulls) so stuff like NoBlockedMessages work as expected. This really is a true "Fake Block"

Vendicated:
you can now ignore users so this is obsolete

PR #2827, opened 2024-09-01
ShowMeYourName: Attempt to fix all inconsistencies; DM Support

hideki2k02:
I've noticed the code have some minor inconsistencies that makes it so that certain features don't work as intended (Mainly casing not working on certain modes), this attempts to fix that whilst trying to make the code be future-proof.

I've also added support for Direct Messages and Direct Groups with their respective toggles.

### DIFF
--- a/src/plugins/showMeYourName/index.tsx
+++ b/src/plugins/showMeYourName/index.tsx
@@ -4,20 +4,26 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 rini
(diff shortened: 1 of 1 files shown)


src/plugins/showMeYourName/index.tsx
```diff
@@ -4,20 +4,26 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+/*
```
k26pl, line 7: The license header is duplicated

hideki2k02: ty for pointing that out!

Vendicated:
i don't really get the point of this pr, i don't think it adds any objective improvements

PR #2730, opened 2024-07-27
Port volume booster to web/vesktop

sadan4:
resolves: #1944 
ports volume booster to web/vesktop

TODO: 

- [x] don't error on screenshare
- [x] adjust volume?
- [x] fix deafening

PR #2828, opened 2024-09-01
Fix VolumeBooster on Vesktop & IgnoreActivities tweaks

sadan4:


src/plugins/volumeBooster/index.ts
```diff
@@ -69,7 +69,7 @@ export default definePlugin({
         // Patches needed for web/vesktop
         {
             find: "streamSourceNode",
-            predicate: () => IS_WEB,
+            predicate: () => !IS_DISCORD_DESKTOP,
```
rozbrajaczpoziomow, line 72:
```suggestion
            predicate: () => IS_WEB || IS_VESKTOP,
```

sadan4: https://github.com/Vendicated/Vencord/pull/2730#discussion_r1701700464

PR #2832, opened 2024-09-02
PronounDB: fix pronoun tooltip

maddie480:
The text in the tooltip (after patching) is:
```js
d.Z.Messages.USER_PROFILE_PRONOUNS + vcHasPendingPronouns ? "" : ` (${vcPronounSource})`
```
... which ends up checking if `'Pronomsfalse'` is truthy, so the whole thing ends up evaluating to `''`, and we end up with an empty tooltip on the pronouns field in the profile.

Adding parentheses makes sure it's checking `vcHasPendingPronouns` only:
```js
d.Z.Messages.USER_PROFILE_PRONOUNS + (vcHasPendingPronouns ? "" : ` (${vcPronounSource})`)
```
This evaluates to `'Pronoms (Discord)'` as intended.

### DIFF
--- a/src/plugins/pronoundb/index.ts
+++ b/src/plugins/pronoundb/index.ts
@@ -57,7 +57,7 @@ export default definePlugin({
                 },
                 {
                     match: /text:\i\.\i.Messages.USER_PROFILE_PRONOUNS/,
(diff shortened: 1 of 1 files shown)


Nuckyz:
thanks haha

PR #2834, opened 2024-09-02
fix: CopyFileContents

smileyhead:
Changed the copy button's tooltip to sentence case, as the default buttons already present in Discord (‘Expand {number} lines’, ‘View whole file’, ‘Change language’) do not use title case, either.


Nuckyz:
Even though it's inconsistent, having it the current ways looks better

PR #2835, opened 2024-09-02
NoServerEmojis: fix plugin checking for non-existing property

maddie480:
The plugin seems to have completely stopped working. ~~Looks like there's no `type` property in the emojis anymore:~~ _oops I'm blind it's just a number instead_

image

so, `emoji.type !== "GUILD_EMOJI"` is always `true`, and no emoji end up being skipped because they're all considered unicode emojis :sweat_smile:


Vendicated:
it's just an enum now. look at fakenitro
image

maddie480:
I didn't even see the property, whoops

maddie480:
okay I updated it so that I just replace the string enum value `"GUILD_EMOJI"` with the corresponding int value `1` :sweat_smile:

PR #2840, opened 2024-09-04
Fix: Volume booster on non-desktop not respecting sinkId

sadan4:


src/plugins/volumeBooster/index.ts
```diff
@@ -128,6 +128,10 @@ export default definePlugin({
             gain.connect(data.audioContext.destination);
         }
 
+        if(data.sinkId && "setSinkId" in AudioContext.prototype)
```
Vendicated, line 131:
```suggestion
        if (data.sinkId && "setSinkId" in AudioContext.prototype)
```

PR #2842, opened 2024-09-04
Better string specificity for props: PopoutPosition

fres621:
Adds the type PopoutPosition to replace "string" in multiple component's props as any other string would result in a crash
Also specifies the prop align in Popout for the same reason (Any string other than left, right or center would make it crash)


Vendicated:
merged manually via https://github.com/Vendicated/Vencord/commit/be02baffaa8c46d499b90c6e2f638e4fe7aa28bd, thanks!

PR #2843, opened 2024-09-04
Prevent preview message button from appearing on chat input commands

NotPiny:
Change the `previewMessage` plugin to hide the preview button on messages where the first word matches `/^\/[-_\p{L}\p{N}]{1,32}$/gmu` as the preview will simply be `/ping option:option value` which you can get by copying the command with ctrl + c

### DIFF
--- a/src/plugins/previewMessage/index.tsx
+++ b/src/plugins/previewMessage/index.tsx
@@ -81,8 +81,9 @@ const PreviewButton: ChatBarButton = ({ isMainChat, isEmpty, type: { attachments
 
     const hasAttachments = attachments && UploadStore.getUploads(channelId, DraftType.ChannelMessage).length > 0;
     const hasContent = !isEmpty && draft?.length > 0;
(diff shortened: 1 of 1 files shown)


src/plugins/previewMessage/index.tsx
```diff
@@ -81,8 +81,9 @@ const PreviewButton: ChatBarButton = ({ isMainChat, isEmpty, type: { attachments
 
     const hasAttachments = attachments && UploadStore.getUploads(channelId, DraftType.ChannelMessage).length > 0;
     const hasContent = !isEmpty && draft?.length > 0;
+    const isCommand = draft?.split(" ")[0]?.match(/^\/[-_\p{L}\p{N}]{1,32}$/gmu) !== null;
```
Sqaaakoi, line 84: This is hacky. There is definitely a better way of checking if there is a command being typed. You'll likely find luck using DraftStore

PR #2844, opened 2024-09-04
fix crashing on canary when searching slash commands

nakoyasha:
this fixes a crash caused by vencord commands lacking the `untranslatedName` and `untranslatedDescription` fields by.. adding them

(ty aaamia for finding this bug)


Vendicated:
thanks!

PR #2845, opened 2024-09-05
RoleColorEverywhere: Add role color to the message content

sadan4:
adds role color to messages 
closes: https://github.com/Vencord/plugin-requests/issues/63#issuecomment-2332235459


sadan4:
https://github.com/Vendicated/Vencord/pull/2893 is just better

PR #2847, opened 2024-09-06
TimeBarAllActivities: Fix patch

hauntii:
Discord moved where they keep the activity message text, so here's a fix for the patch.

### DIFF
--- a/src/plugins/timeBarAllActivities/index.tsx
+++ b/src/plugins/timeBarAllActivities/index.tsx
@@ -40,7 +40,7 @@ export default definePlugin({
     settings,
     patches: [
         {
-            find: 'location:"UserProfileActivityCard"',
+            find: ".gameState,children:",
             replacement: [
                 // Insert Spotify time bar component
                 {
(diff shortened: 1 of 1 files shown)


src/plugins/timeBarAllActivities/index.tsx
```diff
@@ -40,7 +40,7 @@ export default definePlugin({
     settings,
     patches: [
         {
-            find: ".Messages.USER_ACTIVITY_PLAYING",
+            find: 'location:"UserProfileActivityCard"',
```
Nuckyz, line 43:
```suggestion
            find: ".gameState,children",
```
Location stuff is related to analytics and more likely to be removed in the future

Nuckyz (changes requested):

src/plugins/timeBarAllActivities/index.tsx
hauntii: Didn't know that, will keep that in mind

Nuckyz:
Thank you for the quick fix!

PR #2848, opened 2024-09-06
CopyFileContents: Add padding

DokterKaj:
The "download" and "change language" buttons both have 4px left padding, so this should too.


Nuckyz:
Thank you!

PR #2850, opened 2024-09-08
feat(plugin): Add BetterGifSend

ryanamay:
This PR adds a plugin that provides options regarding the configuration of the GIF picker visibility/send behavior.

These behaviors are: (configurable via plugin options)

### Gif Visibility Behavior

- **Close after sending**: The picker will close immediately after sending a GIF. (Default)
- **Always keep open**: The picker will always remain open after selecting a GIF.
- **Keep open when SHIFT is held**: Keeps the GIF picker open if the SHIFT key is held down.

### GIF Send Behavior

- **Send immediately**: Sends the GIF directly to the chat. (Default)
- **Insert into chatbox**: Inserts the GIF URL into the chatbox, allowing you to add additional text before sending.
- **Insert link if SHIFT is held**: Inserts the GIF URL into the chatbox if the SHIFT key is held down.

This plugin also replaces **GifPaste**. (I asked permission from Vee)


DokterKaj:
Would it be possible to add the same visibility functionality for the emoji and sticker pickers?

ryanamay:
> Would it be possible to add the same visibility functionality for the emoji and sticker pickers?

Yes, theoretically you can bring it over to the Sticker/Emoji pickers. Thooo it would probably warrant its own plugin.

rynmy:
Upon checking active PRs (which i shouldve done sooner tbh), I noticed an implementation has been made regarding to the shift functionality of this plugin in https://github.com/Vendicated/Vencord/pull/2745.

Since our implementation was more or less the same, just me being more explicit. I've decided to include their dev constant here to give credit.

iminlikewithyou:
i wouldn't say that the implementation is really the same - because i gave the user between two default options: Insert the link into the chatbox, or send immediately. then, on top of that, i gave them the option to perform the opposite behavior by holding down shift.

you might think that this is equivalent to providing the user with three options (which you did): Insert the link into the chatbox, send immediately, or insert link if shift is held, but it isn't. with your implementation, you can't have inserting the link be the default behavior BUT hold shift to instantly send, because it isn't an option. (this is the option i like to use)

i do really like the new closing behavior here, but i think you should do something more similar to what i did, where you present two options (Close immediately, Keep open) (Insert link, Immediately send), and provide them with an option to use the alternate behavior (a shift override) in order to perform the other action instead. or you could go nuclear and present the user with four options, but imo i think that's less understandable than choosing your defaults and then selecting what you want shift to do.

PR #2851, opened 2024-09-08
messageClickActions: Prevent deletion of non-existent messages

imjustprism:
For those who delete messages with 30 cps


Vendicated:
this is not correct

PR #2854, opened 2024-09-08
enhancement: Add filter for userplugins under plugins if running from source build

schmayterling:
SCR-20240909-flbj

note: this is already working, i just dont have any user plugins added rn


Vendicated:
this has been added at some point

PR #2855, opened 2024-09-09
feat(shc): Add feature parity, more styles & show modes

JustOptimize:
- Separate show modes and channel styles, you can now choose between Classic (normal channel), Muted, Show Unreads and "Muted and Show Unreads"
- Add Lock icon on the right
- Remove "hideUnreads" setting since it is now merged with the channelStyle


JustOptimize:
@Vendicated any edits/suggestions?

PR #2860, opened 2024-09-11
[Plugin] Hide a message clientside

garlic-os:
A very simple plugin that lets you right click and hide a message without blocking its author. Hides it just for you.

Oftentimes I see a message and think, "I don't want to block this person, but I just don't want to see this particular thing on my screen right now". This plugin lets you do that, no more, no less.

I am aware that this PR may look like a duplicate, but other closed/rejected versions of this idea had larger scopes and/or complex code. Mine is intentionally small in scope, and the core implementation is just [one line](https://github.com/garlic-os/Vencord/blob/main/src/plugins/hideMessage/index.tsx#L24). So please consider reviewing this and accepting it into the project.


zt64:
@Vendicated you will review

korbosoft:
my buddy is here...

garlic-os:
Closing; HideMedia serves the purpose I was going for with this plugin.

PR #2864, opened 2024-09-11
AppleMusicRichPresence: fix metadata fetching

ryanccn:
Apple got rid of the old `tools.applemediaservices.com` search endpoint, so now we have to fetch from the API that `music.apple.com` uses with an extracted hardcoded token. Cheers!

### DIFF
--- a/src/plugins/appleMusic.desktop/native.ts
+++ b/src/plugins/appleMusic.desktop/native.ts
@@ -16,19 +16,6 @@ async function applescript(cmds: string[]) {
     return stdout;
 }
 
-function makeiTunesUrl(query: string) {
-    const url = new URL("https://itunes.apple.com/search");
-    url.searchParams.set("media", "music");
-    url.searchParams.set("entity", "musicTrack");
-    url.searchParams.set("limit", "1");
-    url.searchParams.set("term", query);
-    return url;
-}
-
-const requestOptions: RequestInit = {
-    headers: { "user-agent": "Mozilla/5.0 (Windows NT 10.0; rv:125.0) Gecko/20100101 Firefox/125.0" },
-};
-
 interface RemoteData {
     appleMusicLink?: string,
     songLink?: string,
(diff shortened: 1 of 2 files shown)


src/plugins/appleMusic.desktop/native.ts
```diff
@@ -58,21 +45,23 @@ async function fetchRemoteData({ id, name, artist, album }: { id: string, name:
...
+            makeiTunesUrl(name + " " + artist + " " + album), requestOptions
+        ).then(r => r.json());
 
-        const appleMusicLink = songData?.songs?.data[0]?.attributes.url;
-        const songLink = songData?.songs?.data[0]?.id ? `https://song.link/i/${songData?.songs?.data[0]?.id}` : undefined;
-
-        const albumArtwork = songData?.songs?.data[0]?.attributes.artwork.url.replace("{w}", "512").replace("{h}", "512");
-        const artistArtwork = artistData?.artists?.data[0]?.attributes.artwork.url.replace("{w}", "512").replace("{h}", "512");
+        const artistHtml = await fetch(artistViewUrl, requestOptions).then(r => r.text());
+        const extractedArtistArtwork = artistHtml.match(/,"image":"(https:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)\.png)"/);
```
Vendicated, line 53: nit: due to high complexity, i recommend storing this regex in a top level variable instead of having it inline. not just makes it more maintainable but also slightly improves performance

src/plugins/appleMusic.desktop/native.ts
```diff
@@ -58,21 +45,23 @@ async function fetchRemoteData({ id, name, artist, album }: { id: string, name:
...
 
         cachedRemoteData = {
             id,
-            data: { appleMusicLink, songLink, albumArtwork, artistArtwork }
+            data: {
+                appleMusicLink: trackViewUrl,
+                songLink: `https://song.link/i/${trackId}`,
+                albumArtwork: artworkUrl100.replace("100x100", "512x512"),
```
Vendicated, line 60: probably would be more robust to use the `URL` api and use urlSearchParams.set("whatever", "512x512"), same one line below

or is the size a path instead of search param?

ryanccn: size is part of the path, yeah

src/plugins/appleMusic.desktop/native.ts
ryanccn: done!

n1ckoates:
For what it's worth, the `https://amp-api-edge.music.apple.com/v1/catalog/us/search` endpoint is very similar and is still active. It *does* require an API key, but you could just use the one hardcoded into https://music.apple.com.

You just have to add the search params `platform=web&with=serverBubbles`, and the response format is slightly different.

ryanccn:
@n1ckoates I did look into that API, but the request body is much more complex and hardcoding a Bearer token is not such a good idea IMHO.

Vendicated:
thank u ryan!

PR #2867, opened 2024-09-12
Fix: BetterFolders rendering pending guild applications

sadan4:
when rendering pending guild applications better folders breaks and will not let you see them, this fixes that


Nuckyz:
Thanks for the fix!

PR #2872, opened 2024-09-13
new plugin: betterInvites

programminglaboratorys:
this plugin upgrades the invite embed allowing you to see the inviter, preview the server before joining and view the invite relative expiration date

inviter
receiver

### DIFF
--- a/src/plugins/betterInvites/README.md
+++ b/src/plugins/betterInvites/README.md
@@ -2,4 +2,4 @@
 
 -   be able to see the inviter and inviter profile
 -   preview discoverable servers by clicking the name
--   be able to see invite expiration date
+-   be able to see the invite relative expiration date
--- a/src/plugins/betterInvites/README.md
+++ b/src/plugins/betterInvites/README.md
@@ -1,0 +1,5 @@
+# features
+
+-   be able to see the inviter and inviter profile
+-   preview discoverable servers by clicking the name
+-   be able to see invite expiration date
--- a/src/plugins/betterInvites/README.md
+++ b/src/plugins/betterInvites/README.md
@@ -1,5 +1,11 @@
+# BetterInvites
+
+![inviter preview](https://raw.githubusercontent.com/programminglaboratorys/resources/main/Vencord.BetterInvites/inviter.png)
+
+![receiver preview](https://raw.githubusercontent.com/programminglaboratorys/resources/main/Vencord.BetterInvites/receiver.png)
+
 # features
 
 -   be able to see the inviter and inviter profile
--   preview discoverable servers by clicking the name
+-   preview discoverable servers by clicking the server name
 -   be able to see the invite relative expiration date
--- a/src/plugins/betterInvites/index.tsx
+++ b/src/plugins/betterInvites/index.tsx
@@ -34,7 +34,7 @@ function lurk(id: string) {
 export default definePlugin({
     name: "BetterInvites",
     description: "bomby",
-    authors: [Devs.Samwich],
+    authors: [Devs.iamme],
     patches: [
         {
             find: ".Messages.HUB_INVITE_ANOTHER_SCHOOL_LINK",
--- a/src/plugins/betterInvites/index.tsx
+++ b/src/plugins/betterInvites/index.tsx
@@ -49,7 +49,7 @@ export default definePlugin({
                 },
                 {
                     match: /(0,\i\.jsx\)\(\i\.\i\.Header,\{)text:(\i)/,
-                    replace: "$1text: $self.Header(arguments[0].invite.inviter, $2)"
+                    replace: "$1text: $self.Header(arguments[0].invite.inviter, $2),style:{display:\"flex\"}"
                 }
             ]
         }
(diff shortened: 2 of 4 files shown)


src/plugins/betterInvites/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+                },
+                {
+                    match: /(\.jsx\)\(\i\.\i\.Header,\{)text:(\i)/,
+                    replace: "$1text: $self.Header(arguments[0].currentUserId, arguments[0].invite.inviter, $2)"
+                }
+            ]
+        }
+    ],
+    handleTip(isGuest: boolean, message: string, expires_at: string) {
+        return <>this invite will expire {Parser.parse(`<t:${Math.round(new Date(expires_at).getTime() / 1000)}:R>`)}{isGuest ? ". " + message : ""}</>;
```
Sqaaakoi, line 58: fix caps here

src/plugins/betterInvites/style.css
```diff
@@ -0,0 +1,23 @@
+h3[class^="h5"]:has([class="vc-bi-header-inner"]) {
```
Sqaaakoi, line 1: these cursed header classes always scare me, you are already patching near this so you should just make a .vc-bi-header class instead

src/plugins/betterInvites/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    Header(currentUserId: string, inviter: User | undefined, defaultMessage: string) {
+        return <div className="vc-bi-header-inner">
+            {(inviter && (currentUserId !== inviter.id)) ? <>
+                <img
+                    className={classes(AvatarStyles.avatar, AvatarStyles.clickable) + " vc-bi-inviter-avatar"}
+                    onClick={() => openUserProfile(inviter.id)}
+                    src={inviter.avatar ? `https://cdn.discordapp.com/avatars/${inviter.id}/${inviter.avatar}.webp?size=80` : "/assets/1f0bfc0865d324c2587920a7d80c609b.png?size=128"}
+                />
+                <p className="vc-bi-invite-title"> {inviter.global_name ? inviter.global_name.toUpperCase() : inviter.username.toUpperCase()} HAS INVITED YOU TO JOIN</p>
```
Sqaaakoi, line 68: just let the CSS text-transform do its job, this will probably look cursed to screen readers and anyone reading the code or DOM in devtools

src/plugins/betterInvites/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+            </> : <p className="vc-bi-invite-title">{defaultMessage}</p>}</div>;
+    },
+    Lurkable: (id: string, features: Guild["features"] | Array<string>) => {
+        let discoverable = false;
+
+        if (features instanceof Set) discoverable = features.has("DISCOVERABLE");
```
Sqaaakoi, line 74: this is ugly and you should just pass the features arg to the Set constructor instead by doing `new Set(feature).has("DISCOVERABLE")` directly in the if condition below

src/plugins/betterInvites/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+interface User {
+    id: string;
+    avatar: string;
+    global_name: string;
+    username: string;
+}
+
+function lurk(id: string) {
+    GuildManager.joinGuild(id, { lurker: true })
+        .then(() => { setTimeout(() => GuildManager.transitionToGuildSync(id), 100); })
```
Sqaaakoi, line 30: why the timeout?

src/plugins/betterInvites/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+    name: "BetterInvites",
+    description: "See invites expiration date, view inviter profile and preview discoverable servers before joining by clicking their name",
+    authors: [Devs.iamme],
+    patches: [
+        {
+            find: ".Messages.HUB_INVITE_ANOTHER_SCHOOL_LINK",
+            replacement: [
+                {
+                    match: /,(\i)&&(\(\i=\(0,\i\.jsx\)\(\i\.TooltipContainer.+)className:(\i.tooltipContainer),text:(\i\.\i\.Messages.GUEST_MEMBERSHIP_EXPLANATION)/,
+                    replace: (_, isGuest, rest, className, message) => `,(${isGuest}||((!${isGuest})&&arguments[0].invite.expires_at)) && ${rest}text:$self.handleTip(${isGuest}, ${message}, arguments[0].invite.expires_at),className:${className}+" vc-bi-tool-tip-container"`
```
Sqaaakoi, line 44: i am unsure about how this patch works and i need to look at it a bit better but it is very hard to read and could be refactored in a few ways

src/plugins/betterInvites/README.md
```diff
@@ -0,0 +1,11 @@
+# BetterInvites
+
+![inviter preview](https://github.com/user-attachments/assets/9f040e27-5540-433f-86e6-d4bd745e36a3)
```
Sqaaakoi, line 3: update this after fixing the caps in "this"

Sqaaakoi (commented):
I like this plugin. I haven't loaded it into my client and tested it yet, but it is looking a lot better than the code you showed me recently. I'll try it out fully soon

src/plugins/betterInvites/index.tsx
programminglaboratorys: > i am unsure about how this patch works and i need to look at it a bit better but it is very hard to read and could be refactored in a few ways

I am all into hearing those few ways I thought if I made the replace a function it would be much readable but ig I just made it longer

src/plugins/betterInvites/style.css
programminglaboratorys: that would probably make the patchs more cursed

Sqaaakoi: it will be fine :trollface:

src/plugins/betterInvites/index.tsx
programminglaboratorys: > If discord pushes an update, wouldn't `/assets/1f0bfc0865d324c2587920a7d80c609b.png?size=128` also cause problems?
that won't be a probably any soon

programminglaboratorys: > just let the CSS text-transform do its job, this will probably look cursed to screen readers and anyone reading the code or DOM in devtools

so ignore the potential of quickcss?

Sqaaakoi: > > just let the CSS text-transform do its job, this will probably look cursed to screen readers and anyone reading the code or DOM in devtools
> 
> so ignore the potential of quickcss?

if someone disables text transform there it would look out of place, so i would just set it to regular sentence case in the code

src/plugins/betterInvites/index.tsx
Sqaaakoi: the type for features could also just be set to Iterable<string>

Snowy-Fluffy:
Can you update this nice plugin?

PR #2876, opened 2024-09-14
New Plugin: FileViewer for PDFs

TheGreenPig:
# New Plugin: File Viewer
Hello, I came across this plugin request https://github.com/Vencord/plugin-requests/issues/3 related to a BD plugin for which I am the original author of. Since I wanted to check out Vencord at some point anyway and this plugin was quite popular on the BD site, I thought I could port it to Vencord...

Notice this is my **first ever Vencord plugin** and although I have experience with BD plugins, I haven't been active in the Discord modding scene for a while so my bad if something isn't exactly up to date. I'm open for comments and suggestions on how to improve the plugin of course. 

The plugin isn't an exact port and has less functionality, unfortunately, since Discord changed the way you can embed and link to Discord attachments. 

## Main Features
image

- Preview **PDF** files (no other file types unfortunately)
- Preview is the normal embed element, so no other service/package/website etc. necessary
- Hide/Show the pdf preview
- Resize the preview
- Cache the files with some settings/customization 

## Drawbacks
Viewing files used to be pretty easy with the old plugin, I was able to just append the discord attachment url to a google or microsoft embed url and it would work. That's why I was able to support excel, word and even some 3d files. After some experimentation, I don't think simply adding the attachment url to such an embedding website works anymore... I had to resort to downloading the file in `native.ts` (otherwise it would just refuse, or open the download window instead of just getting the data), then turning the blob into an objecturl and then embedding that. 

I also had to change some csp rules `"frame-src", "object-src"`, hope that's ok like that.


src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+    name: "FileViewer",
+    description: "Preview PDF Files without having to download them",
+    authors: [Devs.AGreenPig],
+    dependencies: ["MessageAccessoriesAPI", "MessageUpdaterAPI",],
+    settings,
+    patches: [
+        {
+            find: "Messages.IMG_ALT_ATTACHMENT_FILE_TYPE.format",
+            replacement: {
+                match: /function\((.+?)\)(.+Fragment.{0,500}newMosaicStyle.{0,500}?children:\[)(.+?\])/,
```
Masterjoona, line 154: I assume you are new to this type of patching but you don't need to match all of the module and instead only match, the important part, the `children`  here (and something around it to differentiate it from other `children`s) and you don't need capture groups for variables since you can use `arguments[0]` here!

TheGreenPig: Thanks for the help!

sadan4:
idk if this *needs* the native.ts, but if it does you should append the folder name with .desktop

TheGreenPig:
So if I just put the blob fetching code into the normal `index.tsx` I get a `Access-Control-Allow-Origin` error since the origin is `discord.com` and the files live at `cdn.discordapp.com`. I believe I was able to get around the CORS policy with a few changes at some point, but then it would just open the "save file to" window instead of just getting the blob data. 

So I may be missing something obvious, but I'd rather not mess around/disable CORS policies more than I have to and the `native.ts` implementation seems to work so I thought it would be ok?

As to the 
> you should append the folder name with .desktop

I wasn't quite sure what that suffix did, since there are some plugins with a `native.ts` file that didn't have it (like voiceMessages or translate) and others that did have the `.desktop`, but no `native.ts` file like invisibleChat for example. So I have no idea what that folder suffix actually does😅.

src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        default: false
+    },
+    persistPreviewState: {
+        type: OptionType.BOOLEAN,
+        description: "Persist the state of opened/closed File Previews across channel switches and reloads.",
+        default: false
+    },
+    cacheSize: {
+        type: OptionType.SLIDER,
+        description: "Maximum number of PDF files to cache (after that, the least recently used file will be removed). Lower this value if you're running out of memory.",
```
Vendicated, line 37: i don't think this setting is a very good idea. a plugin should never use excessive memory and should just use a sensible default that won't use too much

it's also fairly rare to even encounter a pdf so i doubt having such a high limit would be useful

src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+    url: string;
+    proxy_url: string;
+    content_type: string;
+    content_scan_version: number;
+    title: string;
+    spoiler: boolean;
+    previewBlobUrl?: string;
+}
+
+const stripLink = (url: string) => url.replace("https://cdn.discordapp.com/attachments/", "").split("/").slice(0, 2).join("-");
```
Vendicated, line 63: what exactly does this do? and why? would aid maintainability to add a short doc comment

it might also be a good idea to depend on the more robust `URL` api instead of doing manual url manipulation

src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+async function buildCss() {
+    const visiblePreviews: Set<string> | undefined = await get(STORE_KEY);
+    const elements = [...(visiblePreviews || [])].map(url => `#file-viewer-${stripLink(url)}`).join(",");
+    style.textContent = `
+    :is(${elements})  {
+        display: flex !important;
+    }
+    `;
+}
```
Vendicated, line 80: wouldn't this be better solved via react state?

src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+            find: "Messages.IMG_ALT_ATTACHMENT_FILE_TYPE.format",
+            replacement: {
+                match: /newMosaicStyle,\i\),children:\[(?<=}=(\i);.+?)/,
+                replace: "$&$self.renderPreviewButton($1),"
+            }
+        }
+    ],
+    start() {
+        objectUrlsCache.setMaxSize(Math.round(settings.store.cacheSize));
+        new Logger("FileViewer").info(`Initialized LRU Cache with size ${objectUrlsCache.getMaxSize()}`);
```
Vendicated, line 161: this is just noise that isn't very relevant
```suggestion
```

src/plugins/fileViewer/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+                    ))}
+                </ErrorBoundary>
+            );
+        }, -1);
+
+        style = document.createElement("style");
+        style.id = "VencordFileViewer";
+        document.head.appendChild(style);
+    },
+    renderPreviewButton(e) {
```
Vendicated, line 179:
```suggestion
    renderPreviewButton: ErrorBoundary.wrap(e => {
```
...
```js
}, { noop: true })
```

src/plugins/fileViewer/native.ts
```diff
@@ -0,0 +1,28 @@
...
+            });
+
+            res.on("end", function () {
+                const body = Buffer.concat(chunks);
+                resolve(body);
+            });
+        });
+        req.on("error", reject);
+        req.end();
+    });
```
Vendicated, line 26: 1. this is a full cors / csp bypass, which is very bad. please make sure to validate the `url` to ensure only discord cdn urls can be fetched with this
2. you can just use the `fetch` api instead of using the https module, since we're finally on modern electron

src/plugins/fileViewer/fileViewer.css
```diff
@@ -0,0 +1,29 @@
+.file-viewer.container {
```
Vendicated, line 1: classnames should be prefixed with `vc-` to differentiate them from discord classes and avoid conflicts

```suggestion
.vc-fileViewer.container {
```

src/plugins/fileViewer/cache.ts
```diff
@@ -0,0 +1,62 @@
...
+        for (const key of this.cache.keys()) {
+            URL.revokeObjectURL(this.cache.get(key)!);
+        }
+        this.cache.clear();
+    }
+
+    setMaxSize(maxSize: number) {
+        if (maxSize < 1) {
+            new Logger("FileViewer").error("Cache size must be at least 1");
+            return;
```
Vendicated, line 54:
```suggestion
            throw new Error("Cache size must be at least 1");
```

src/plugins/fileViewer/README.md
```diff
@@ -0,0 +1,3 @@
+# FileViewer
+
+Allows you to Preview PDF Files without having to download them
```
Vendicated, line 3: an image would be nice! you can upload it to #1749 for a permalink

src/components/Icons.tsx
```diff
@@ -407,6 +407,36 @@ export function PencilIcon(props: IconProps) {
...
+export function PreviewVisible(props: IconProps) {
+    return (
+        <Icon
+            {...props}
+            className={classes(props.className, "vc-preview-visible")}
+            viewBox="0 0 226 226"
+        >
+            <path
+                fill="currentColor"
+                d="M113,37.66667c-75.33333,0 -103.58333,75.33333 -103.58333,75.33333c0,0 28.25,75.33333 103.58333,75.33333c75.33333,0 103.58333,-75.33333 103.58333,-75.33333c0,0 -28.25,-75.33333 -103.58333,-75.33333zM113,65.91667c25.99942,0 47.08333,21.08392 47.08333,47.08333c0,25.99942 -21.08392,47.08333 -47.08333,47.08333c-25.99942,0 -47.08333,-21.08392 -47.08333,-47.08333c0,-25.99942 21.08392,-47.08333 47.08333,-47.08333zM113,84.75c-15.60204,0 -28.25,12.64796 -28.25,28.25c0,15.60204 12.64796,28.25 28.25,28.25c15.60204,0 28.25,-12.64796 28.25,-28.25c0,-15.60204 -12.64796,-28.25 -28.25,-28.25z"
```
Vendicated, line 419: what's the origin of these svgs? i can't find them in discord's sources

Vendicated (commented):
thank you for the contribution! like I said on discord, I'm impressed that you so quickly figured out all apis and native capabilities

since this plugin seems to only support pdfs, perhaps a rename to PdfViewer would be good? it can still be renamed later if support for more filetypes would be added

the native file is needed due to CORS, right? if the native file is absolutely needed for the plugin to function, the plugin should be marked as desktop only by naming its folder `plugin.desktop`. this will exclude it from web builds where native scripts won't be available

src/plugins/fileViewer/index.tsx
TheGreenPig: Sure I can remove it I guess. I just thought it might make sense to add some customizability... I guess the memory usage has less inherently to do with the _amount_ of files, more the size of all the objectUrls. I was thinking of a worst-case scenario where someone might have a channel with a ton of pdf files they want to review quickly, whilst keeping them in cache, but that's probably just unnecessary and overkill.

src/plugins/fileViewer/index.tsx
TheGreenPig: That might be a somewhat hacky solution yes... I toggle the visibility of file previews via css, so I need to give each preview element a unique id which I can then use to control their display value. As a means of giving the preview an id, I grabbed the channel and message id from the the attachment url.

So for example  `https://cdn.discordapp.com/attachments/123456789/987654321/filename.pdf?ex=66eaae57&is=66e95cd7&hm=388f&...` would simply turn into `123456789-987654321` which I then assign as an id to the preview element. Technically it doesn't really matter what exactly I use as an id, but it should be unique and related to the attachment. I also store if an attachment is open or not in the `@api/DataStore` and I use the attachment url there as well, so that's why I used that.

I can rework this method to use the URL api, but it's really just about grabbing unique and reasonable data for an id from an url.

src/plugins/fileViewer/index.tsx
TheGreenPig: So the PreviewButton is the main driving force behind toggling the visibility of the preview with is a different component. The PreviewButton uses a state on it's own, but since the actual preview isn't a child of the preview button (since it's sort of in a different place) I can't really prop drill the state? I also tried adding visibility information to the message object it'self (like I do after adding the previewBlobUrl), then I would have to update the Message every time I change the visibility state I think?  

If you think that's a better idea or there is some other react store I can use to set state across components I can, then the `stripLink` thing is also irrelevant since I don't need any css toggling at all.

(This implementation was sort of based on the hideAttachments plugin btw)

src/components/Icons.tsx
TheGreenPig: I don't think it's from discord, probably just some random svg I found some time ago. If you have an idea for a different/more fitting Icon that's built into discord, let me know and I can add that instead.

Huderon: Discord has their own `const { EyeIcon, EyeSlashIcon } = Vencord.Webpack.Common.Forms`

src/plugins/pdfViewer.desktop/native.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function getBufferResponse(_: IpcMainInvokeEvent, url: string): Promise<Buffer> {
+    const urlObj = new URL(url);
+    if (!urlChecks.every(check => check(urlObj))) {
+        throw new Error("Invalid URL");
+    }
+
+    const response = await fetch(url);
+    if (!response.ok) {
+        throw new Error(`Failed to fetch: ${response.statusText}`);
```
Vendicated, line 23: fetch itself might throw if the request fails (usually due to no internet / broken dns)

```suggestion
    const response = await fetch(url).catch(() => null); 
    if (!response?.ok) {
        throw new Error(`Failed to fetch: ${response?.statusText ?? "Failed to connect"}`);
```

src/plugins/pdfViewer.desktop/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function FilePreview({ attachment }: { attachment: Attachment; }) {
+    const { previewBlobUrl, previewVisible } = attachment;
+
+    if (!previewVisible) return null;
+
+    return (
+        <div className={"vc-pdf-viewer-container"}>
+            {previewBlobUrl ? <embed src={previewBlobUrl} className="vc-pdf-viewer-preview" title={attachment.filename} /> : <div style={{ display: "flex" }}><Spinner /></div>}
```
Vendicated, line 59: this just puts the raw pdf blob inside the embed tag, right? i don't know much about the embed tag, is this safe if the pdf is malicious? it absolutely must not be able to load external resources or execute code

src/plugins/pdfViewer.desktop/native.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    const arrayBuffer = await response.arrayBuffer();
+    return Buffer.from(arrayBuffer);
```
Vendicated, line 27: nit (ipc converts Buffer -> U8A anyway so this is likely more efficient):
```suggestion
    return new Uint8Array(arrayBuffer);
```

Vendicated (commented):
good changes!

honestly is there even a point in saving the embed open / closed state across reloads? i don't think this is really something people will care about and just overcomplicates the code. the could could be a loot simpler if you removed the persistance, discord doesn't have anything like that either for text file previews

src/plugins/pdfViewer.desktop/index.tsx
TheGreenPig: I'm not a security expert so I can't really say. I'm basically using the built-in chrome pdf viewer and I've read that it sandboxes it's content automatically. The `iframe` and `object` element should work here too though, not just the `embed` and I'm honestly not 100% which one is definitively better (since browser support isn't exactly an issue in this context). I'm also not sure how I would test this. 
There should be other ways to display the pdfs like using pdf.js or converting the pdfs into images, but these would all make the code a lot more complex, add overhead and I'm not sure if it's necessary in the first place. Maybe someone more knowledgeable can chime in, but I honestly doubt that you can simply execute JS code inside a pdf with this...

SippinOnJuiceBox:
Hey, Whats the progress on this plugin?

DroidZed:
I'd love to have this plugin added !! It's a much needed feature for a lot of us.

PR #2880, opened 2024-09-16
New Plugin: Bedtimes

gBasil:
This plugin allows you to configure bedtimes for users (stored locally):

image

If you send a message to a user during their bedtime, the message is silenced (`@silent ` is added to it).

### DIFF
--- a/src/plugins/bedtimes/README.md
+++ b/src/plugins/bedtimes/README.md
@@ -1,0 +1,3 @@
+# Bedtimes
+
+A plugin that lets you configure locally configure bedtimes for other users and prepends `@silent` to any messages you send during that bedtime.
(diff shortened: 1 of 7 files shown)


indexxing:
this would be cool, also would be nice just to track what timezone your friends are in (most of my friends are in a different timezone 😭 )

gBasil:
The plugin currently doesn't have a visual indicator anywhere, it just adds a `@silent ` if the bedtime is active.

It also doesn't handle timezones (times are stored relative to your current timezone, and compared against your current timezone whenever you send a message), so it isn't the best if you travel across timezones a lot.

indexxing:
still cool

PR #2881, opened 2024-09-16
docs: make `noDevtoolsWarning`  description less ambigous

scarf005:
[as discussed in discord](https://discord.com/channels/1015060230222131221/1026515880080842772/1283952481759662111), current description was ambiguous.

EDIT:

> Disables the 'HOLD UP' banner in the console. As a side effect, also prevents Discord from hiding your token, which prevents random logouts. 

This can be interpreted in two ways:

- `noDevtoolsWarning` prevents discord from hiding your tokens.
- `noDevtoolsWarning` prevents random logouts.

OR:

+ `noDevtoolsWarning` prevents discord from hiding your tokens.
+ `Discord hiding your token` prevents random logouts.

### DIFF
--- a/src/plugins/noDevtoolsWarning/index.ts
+++ b/src/plugins/noDevtoolsWarning/index.ts
@@ -21,7 +21,7 @@ import definePlugin from "@utils/types";
 
 export default definePlugin({
     name: "NoDevtoolsWarning",
-    description: "Disables the 'HOLD UP' banner in the console. As a side effect, also prevents Discord from hiding your token, which prevents random logouts.",
+    description: "Disables the 'HOLD UP' banner in the console. Also stops random logouts when using DevTools.",
     authors: [Devs.Ven],
     patches: [{
         find: "setDevtoolsCallbacks",
--- a/src/plugins/noDevtoolsWarning/index.ts
+++ b/src/plugins/noDevtoolsWarning/index.ts
@@ -21,7 +21,7 @@ import definePlugin from "@utils/types";
 
 export default definePlugin({
     name: "NoDevtoolsWarning",
-    description: "Disables the 'HOLD UP' banner in the console. Also stops random logouts when using DevTools.",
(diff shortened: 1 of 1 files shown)


RobinRMC:
This change would make the description worse in my opinion. The current description perfectly states what the plugin does and what you can expect.

I personally find it very annoying that Discord hides my token when I have DevTools open. If the description of this plugin doesn't state that it prevents this feature, I wouldn't have known that this plugin causes my token to stay visible and I wouldn't have known that I should enable this plugin if I want to prevent Discord from hiding my token.

src/plugins/noDevtoolsWarning/index.ts
```diff
@@ -21,7 +21,7 @@ import definePlugin from "@utils/types";
 
 export default definePlugin({
     name: "NoDevtoolsWarning",
-    description: "Disables the 'HOLD UP' banner in the console. As a side effect, also prevents Discord from hiding your token, which prevents random logouts.",
+    description: "Disables the 'HOLD UP' banner in the console. Also stops random logouts when using DevTools.",
```
Vendicated, line 24: maybe something like this
```suggestion
    description: "Disables the 'HOLD UP' banner in the console. Also stops random logouts & token hiding when using DevTools.",
```

RobinRMC: I still think the original description is perfect as-is. No reason to be juggling words around.

RobinRMC: Actually.. this might be better:

"Disables the 'HOLD UP' banner in the DevTools console. As a side effect, this also prevents Discord from hiding your token and logging you out of your account when you close Discord while DevTools is open."

RobinRMC (commented):
Replace the description with "Disables the 'HOLD UP' banner in the DevTools console. As a side effect, this also prevents Discord from hiding your token and logging you out of your account when you close Discord while DevTools is open."

This describes all of the effects the plugin has on your client and is easy to understand.

sadan4:
tbh just give the plugin a readme.md instead

PR #2884, opened 2024-09-16
Improve plugin settings formatting (Inconsistency title&description)

TheGreenPig:
# Description
This PR addresses the inconsistencies in how plugin settings are displayed across different OptionTypes.

## Problem:
1. For the boolean OptionType (Switch), both the setting title (i.e., the formatted setting ID) and the description are displayed. However, for all other OptionTypes, only the description is shown, leading to inconsistency. This can be confusing for plugin devs.
2. The use of "Description" implies that plugin developers should explain the setting's purpose and functionality in more detail. As a result, descriptions tend to be longer than typical setting titles, which can make settings with lengthy descriptions appear overwhelming when used as a title.
## Solution:
This small PR standardizes the formatting by ensuring that both the setting title and description are displayed across all setting types, not just for the boolean OptionType.

### Before (current)
Example-before
### After
Example-after


src/components/PluginSettings/components/SettingSelectComponent.tsx
```diff
@@ -44,7 +46,8 @@ export function SettingSelectComponent({ option, pluginSettings, definedSettings
 
     return (
         <Forms.FormSection>
-            <Forms.FormTitle>{option.description}</Forms.FormTitle>
+            <Forms.FormTitle>{wordsToTitle(wordsFromCamel(id))}</Forms.FormTitle>
+            <Forms.FormText className={Margins.bottom20} type="description">{option.description}</Forms.FormText>
```
Vendicated, line 50: 20 seems too much, discord seems to use 16! also perhaps we could match their font size

image
image


```suggestion
            <Forms.FormText className={Margins.bottom16} type="description">{option.description}</Forms.FormText>
```

Vendicated (commented):
i definitely agree that it's confusing that the description is used as title and the name itself isn't displayed anywhere and the fact that this is inconsistent with boolean settings

but i think in its current state it also looks a bit odd

Nuckyz:
This might make some plugin settings confusion as the author did not use a proper setting name due to it not being part of the component before. I think we should just wait for a proper settings rewrite in the future

Nuckyz:
> This might make some plugin settings confusion as the author did not use a proper setting name due to it not being part of the component before. I think we should just wait for a proper settings rewrite in the future

I just changed my mind and found a case where the missing setting name made me confused, will merge this :)

Nuckyz:
Thanks!!

PR #2885, opened 2024-09-18
moreUserTags: Add an option to change the [APP] tag back to [BOT]

cootshk:
This was requested [here](https://github.com/Vencord/plugin-requests/issues/430#issuecomment-2081255001).

It requires a restart because of some [weird caching issues](https://github.com/user-attachments/assets/e7d542ff-7372-4b34-aa91-22bf1a95167c)

Disabled:
![[APP] venbot](https://github.com/user-attachments/assets/aaf64bff-81b3-4560-a514-0df354136dd4)
Enabled:
![[BOT] venbot](https://github.com/user-attachments/assets/8347f358-53a0-42d9-8799-29d5f167f330)

### DIFF
--- a/src/plugins/moreUserTags/index.tsx
+++ b/src/plugins/moreUserTags/index.tsx
@@ -175,7 +175,7 @@ const settings = definePluginSettings({
 export default definePlugin({
     name: "MoreUserTags",
     description: "Adds tags for webhooks and moderative roles (owner, admin, etc.)",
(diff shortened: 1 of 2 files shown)


cootshk:
image
It also works in the member list and with verified bots as well

Vendicated:
this is so irrelevant, people should just get used to the new name

PR #2887, opened 2024-09-18
fix(typingtweaks): don't eat the third user in the replaced typing text

lumap:
The typingTweaks plugin has an option to show how many users are typing by replacing the "Several users are typing..." text (appears after 4+ users are typing). Due to the current implementation, only the "first" two users are shown in the plugin's typing text. This PR adds the 3rd user.


Vendicated:
nothing is being eaten. this is intended behaviour to keep the text a sane length and is reflected in the "n others" count

lumap:
if you say so

PR #2889, opened 2024-09-19
feat(plugin): CopyStatusUrls

sadan4:
When you right-click one of these buttons, it copies the URL to your clipboard.

this is useful because Discord provides no other way to see the link you are going to before you click it(if you have always trust enabled), and with always trust enabled, you can see what you are clicking before you click it
image


src/plugins/copyStatusUrls/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            replacement: {
+                match: /(?=onClick)(?=.*index:(\i))/,
+                replace: "onContextMenu: $self.makeContextMenu(arguments[0], $1),"
+            }
+        }
+    ],
+
+    makeContextMenu(props: MakeContextMenuProps, index: number) {
+        return async () => {
+        try {
```
Vendicated, line 38: please fix formatting

src/plugins/copyStatusUrls/index.ts
```diff
@@ -0,0 +1,66 @@
...
+                const { button_urls } = await getMetadataFromApi(props.activity, props.user.id);
+                if (!button_urls[index]) {
+                    throw new Error("button_urls does not contain index");
+                }
+                Clipboard.copy(button_urls[index]);
+                Toasts.show({
```
Vendicated, line 44: you can use the copyWithToast util instead

src/plugins/copyStatusUrls/index.ts
Vendicated, line 38: just get rid of this try catch it's not needed

src/plugins/copyStatusUrls/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export default definePlugin({
+    name: "CopyStatusUrls",
+    description: "Copy the users status url when you right-click it",
+    authors: [Devs.sadan],
+    patches: [
+        {
+            find: "?\"PRESS_WATCH_ON_CRUNCHYROLL_BUTTON\"",
+            replacement: {
+                match: /(?=onClick)(?=.*index:(\i))/,
```
Vendicated, line 30: this is not very robust. it will match the first onClick in the file and then find the next index which can very easily be wrong. you should add some anchor or limit length to ensure you match the right onClick and index

sadan4: the patch should be better now (also addressed other feedback)

rniii:
idk if something like this makes more sense

image

PR #2891, opened 2024-09-20
Update reply icon

Obsidianninja11:
Updates vencord's reply icon to discords new one

Before: 
image
After: 
image


Nuckyz:
This icon is currently not used by any plugins, so we can leave it to be updated in the future if needed. Thanks regardless

PR #2895, opened 2024-09-21
fix(ReplaceGoogleSearch): DuckDuckGo URL

MrDiamondDog:
Without the `?q=`, some searches will go to the wrong place such as the download for DuckDuckGo if you put anything that starts with `x`.


Nuckyz:
Thanks!

PR #2902, opened 2024-09-23
feat(plugin): ShowMeTheMember

Masterjoona:


src/plugins/showMeTheMember/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+                replace: "return $self.renderGuild({Original: $1, position: 'left', ...$2})"
+            }
+        }
+    ],
+    settings,
+    async start() {
+        document.addEventListener("keydown", keydown);
+        document.addEventListener("keyup", keyup);
+        await requireUserContextMenu();
+        await requireGuildContextMenu();
```
Nuckyz, line 86:
```suggestion
        await requireUserContextMenu();
        await requireGuildContextMenu();
```

Only require those when u need to use them in the components. you can just call it everytime inside the components since their execution is cached

Nuckyz (changes requested):

src/plugins/showMeTheMember/index.tsx
Nuckyz: I dont understand where that error is from

Vendicated:
okay this definitely needs a better description and or name because the current ones are really vague and I had no clue what this does based on them

This also needs a README! and I think I fixed your conflicts lol

Masterjoona:
fixedd

(thank you about the conflicts i have no idea how i did those)

Vendicated:
just forcepush shenanigans

PR #2904, opened 2024-09-24
BetterFolders: fix patch

sadan4:
the old patch matched two things, the new one only matches one


Derpious (commented):
broke compatibility with horizontal server bar

sadan4:
> broke compatibility with horizontal server bar

discord issue, nothing to do with this patch

PR #2907, opened 2024-09-25
Using `displayName` properties for sub command display names instead of `name` properties

MaybeAshleyIdk:
Before, the `displayName` properties were ignored for sub commands.  
Fixing it so that registered sub commands actually use the specified display names instead of the regular name properties in the command picker.

I don't know if the change to `subCommandPath` actually does anything, but it seems the correct thing to do.

<details>
<summary><strong>Example</strong></summary>

Given the command object:
```ts
{
    name: "name",
    displayName: "displayName",
    options: [
        {
            name: "subcommand1_name",
            displayName: "subcommand1_displayName",
            type: ApplicationCommandOptionType.SUB_COMMAND,
        },
        {
            name: "subcommand2_name",
            displayName: "subcommand2_displayName",
            type: ApplicationCommandOptionType.SUB_COMMAND,
        },
    ],
}
```

Before patch:  
close-up screenshot of the Discord command picker not showing the display names

After patch:  
close-up screenshot of the Discord command picker, this time showing the display names correctly
</details>

### DIFF
--- a/src/api/Commands/index.ts
+++ b/src/api/Commands/index.ts
@@ -113,11 +113,11 @@ function registerSubCommands(cmd: Command, plugin: string) {
             type: ApplicationCommandType.CHAT_INPUT,
             name: `${cmd.name} ${o.name}`,
             id: `${o.name}-${cmd.id}`,
-            displayName: `${cmd.name} ${o.name}`,
+            displayName: `${cmd.displayName ?? cmd.name} ${o.displayName ?? o.name}`,
             subCommandPath: [{
                 name: o.name,
                 type: o.type,
-                displayName: o.name
+                displayName: o.displayName
             }],
             rootCommand: cmd
         };


Vendicated:
why are you using the displayName property in the first place? just use name

MaybeAshleyIdk:
Mainly just to prevent name clashing.  
I saw that `name` must be unique, else it throws, so if you have a possibly common name for a command, then having two plugins with that same command name would cause problems.

lewisakura:
> then having two plugins with that same command name would cause problems.
plugins shouldn't have the same command name in the first place, that's bound to cause confusion considering they don't have icons

MaybeAshleyIdk:
> plugins shouldn't have the same command name in the first place

Definitely yeah, but on the off chance that it happen, it'd be still nice to have a way to disambiguate.

Plus, since `displayName` exists in the first place, I don't think it's wrong to assume that it should work as expected.
If no one uses and it doesn't seem to be recommended to use it at all, then why not remove it completely?
I'd say either fix it or remove it, but leaving it in a half-broken state isn't really a good choice.

lewisakura:
then i think we should remove support for it entirely, there is no reason for the name and display name to be any different, and you should be using `name` to set the actual name for your command. if you are conflicting with a different plugin, then you need to rename your command or rename that plugin's command to be unique

that's a change that can be made at a later date though, and will probably be rolled up into a different PR

PR #2908, opened 2024-09-25
LastFMRichPresence: add listenbrainz support

cgytrus:
adds support for listenbrainz to the LastFMRichPresence plugin
fulfills https://github.com/Vencord/plugin-requests/issues/634

might want to reconsider the name of the plugin now, if that's feasible
or, alternatively, move listenbrainz support to a completely separate plugin, but currently theyd be basically copy-pasted as theres very little difference from the last.fm plugin unless additional listenbrainz-specific features are added, like the one described below

also, if the listenbrainz functionality will be kept in this plugin, would be nice if the creator of the application could add listenbrainz assets to the like the last.fm ones (and, perhaps, some for popular music players/services as well)


TimovVeen:
@Nuckyz could this be looked at? I'd appreciate this plugin a lot, thanks!

yellowsink:
I've got some code here that can use lb's websocket to listen for listens instead of polling if that's of any interest
https://github.com/yellowsink/shelter-plugins/blob/master/plugins/lastfm/listenbrainz.ts

cgytrus:
@yellowsink would be cool if this was integrated with my current thing, maybe as a toggle, i would accept a PR into this branch if you're up to integrating this here!

yellowsink:
respectfully, I don't intend to contribute to this, I just wanted to let you know that listenbrainz has a websocket, and I've got some tested open source code to use it that you're welcome to if you want :)

PR #2914, opened 2024-09-26
Add Hide Groups plugin

KingRainbow44:
This plugin allows users to hide group DMs via the context menu.

(adjusted to merge into `dev` rather than `main`)


Vendicated:
sorry but this is not very useful. there are little reasons one would want to hide groups instead of just leaving them

plus, you can use pindms to organise dms instead

thanks regardless

PR #2916, opened 2024-09-26
VolumeBooster: fix stream on web based clients

sadan4:
fixes https://github.com/Vencord/Vesktop/issues/867 for web, userscript, and vesktop

### DIFF
--- a/src/plugins/volumeBooster/index.ts
+++ b/src/plugins/volumeBooster/index.ts
@@ -77,6 +77,10 @@ export default definePlugin({
                     match: /Math\.max.{0,30}\)\)/,
                     replace: "arguments[0]"
                 },
+                {
+                    match: /(?=}return"video")/,
+                    replace: "this.updateAudioElement();"
+                },
                 // Patch the volume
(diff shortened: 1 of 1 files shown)


src/plugins/volumeBooster/index.ts
```diff
@@ -77,6 +77,11 @@ export default definePlugin({
                     match: /Math\.max.{0,30}\)\)/,
                     replace: "arguments[0]"
                 },
+                // Fix streams not playing audio untill you update them
```
Vendicated, line 80:
```suggestion
                // Fix streams not playing audio until you update them
```

src/plugins/volumeBooster/index.ts
```diff
@@ -77,6 +77,11 @@ export default definePlugin({
...
+                {
+                    match: /(?=}return"video")/,
```
Vendicated, line 82: matching nothing is confusing and likely worse for performance
```suggestion
                    match: /\}return"video"/,
```

PR #2917, opened 2024-09-27
feat(plugin): Youtube Music Rich Presence Support

teop23:
Screenshot 2024-09-27 052657
Screenshot 2024-09-27 052930
Screenshot 2024-09-27 060342
image

Adds YT Music Rich Presence support by connecting to chrome's(and other chromium based browsers) remote debugger. 
This should satisfy Vencord/plugin-requests#952, Vencord/plugin-requests#689 and Vencord/plugin-requests#241

Credit to [@manucabral](https://github.com/manucabral) for their idea of using chrome's remote debugger. If instead you'd like a standalone app, check out their implementation here: [YoutubeMusicRPC](https://github.com/manucabral/YoutubeMusicRPC)


Vendicated:
The way this works is really cursed and YTM is kinda niche, people can just use LastFMRichPresence instead

PR #2918, opened 2024-09-27
Added a Remove Extra Indent Plugin

AA1999:
A new plugin was added to remove all extra indents from pasted code in the code blocks. Added a README.md file for the plugin. Added the author name to constants.ts

This plugin is supposed to remove all the extra indents from code when you paste it inside a `\`\` code block and patch the indents upon sending/editing

An example gif is included in the README but here's that for easy access:

indent-remover


Vendicated:
already a plugin

PR #2919, opened 2024-09-27
ViewIcons QoL changes

Tiagoquix:
- Default to PNG (transparency support). WebP is... WebP.
- Default to 4096 size (max. allowed).
- Changes descriptions just for fun.


Vendicated:
the defaults are good

Issue #2920, opened 2024-09-29
[Bug] Conflict with collapsed MessageLogger messages and NoBlockedMessages


jamesbt365:
### Discord Account

jamesbt365

### What happens when the bug or crash occurs?

Deleted messages when "Collapse dDeleted" is enabled in MessageLogger are hidden from the chat.

### What is the expected behaviour?

Deleted messages when "Collapse Deleted" is enabled in MessageLogger should not be removed by the NoBlockedMessages plugin.

### How do you recreate this bug or crash?

1. Enable MessageLogger
2. Enable "Collapsed Deleted" in the settings of MessageLogger
3. Enable NoBlockedMessages
4. Restart Vencord.

Some proof by 2 users today:
https://discord.com/channels/1015060230222131221/1026515880080842772/1289567636023414897
https://discord.com/channels/1015060230222131221/1026515880080842772/1289569871381794816

Heres a couple videos demonstrating it working without and not working with:
https://discord.com/channels/1015060230222131221/1026515880080842772/1289571412985643159
https://discord.com/channels/1015060230222131221/1026515880080842772/1289571505289826324

Apparently this is a recent problem and (from the reports) this did not happen before, I'm guessing there was a regression after fixing patches post Discord update.

Also worth mentioning that in addition to this, changing the setting in MessageLogger does not prompt the user to restart, but restarting is required for the settings to apply.

### Errors

N/A.

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [X] I have read the requirements for opening an issue above

PR #2921, opened 2024-09-30
NoBlockedMessage: fix conflicts with message logger

sadan4:
closes: https://github.com/Vendicated/Vencord/issues/2920

### DIFF
--- a/src/plugins/noBlockedMessages/index.ts
+++ b/src/plugins/noBlockedMessages/index.ts
@@ -24,6 +24,11 @@ import { findByPropsLazy } from "@webpack";
 import { Message } from "discord-types/general";
 
 const RelationshipStore = findByPropsLazy("getRelationships", "isBlocked");
+interface MessageDeleteProps {
+    collapsedReason: {
+        message: string;
+    };
+}
 
 export default definePlugin({
     name: "NoBlockedMessages",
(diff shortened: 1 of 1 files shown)


src/plugins/noBlockedMessages/index.ts
```diff
@@ -24,6 +24,11 @@ import { findByPropsLazy } from "@webpack";
 import { Message } from "discord-types/general";
 
 const RelationshipStore = findByPropsLazy("getRelationships", "isBlocked");
+interface MessageDeleteProps {
+    collapsedReason: {
+        message: string;
+    };
+}
```
Nuckyz, line 31:
```suggestion

interface MessageDeleteProps {
    collapsedReason: {
        message: string;
    };
}
```

src/plugins/noBlockedMessages/index.ts
```diff
@@ -35,9 +40,8 @@ export default definePlugin({
             replacement: [
                 {
                     match: /let\{[^}]*collapsedReason[^}]*\}/,
-                    replace: "return null;$&"
-                }
-            ]
+                    replace: "if($self.shouldHide(arguments[0])) return null;$&"
+                }]
```
Nuckyz, line 44:
```suggestion
                    replace: "if($self.shouldHide(arguments[0]))return null;$&"
                }
            ]
```

Nuckyz (changes requested):

PR #2862, opened 2024-09-11
Fix: BlurNsfw not blurring embeds

rynmy:
Closes #2837 

demo:
image

PR #2922, opened 2024-09-30
BlurNSFW: fix

sadan4:
fixes blur nsfw

the old patch was broken, and matched something that wasn't the channel


rynmy:
wont this conflict with #2862? (additionally fixes embeds blurring)

sadan4:
didnt notice that
whoops

PR #2924, opened 2024-09-30
new plugin: StatusPresets

programminglaboratorys:
plugin Allows you to remember your statuses and set them later.

closes [#835](https://github.com/Vencord/plugin-requests/issues/835). any assist to fix those issues would be appreciated
## Preview
modal
status-presets-menu

## TODO

-   [x] When the clear button is clicked, it should only clear the user's custom status and not open the custom status modal
-   [x] When the remove button is clicked, it should not trigger the menu item at the same time
-   [x] when a user does not have a Nitro, it should filter custom emojis from the status selection
-   [x] ~~Module responsible for setting the user custom status does not get executed sometimes~~ (fixed by re-implementing setStatus)
-   [x] Add README
~~Maybe add a warning when switching statuses too fast~~
-   [x] Settings doesn't update on preset status removal


src/plugins/statusPresets/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+    text: string;
+    clearAfter: string | number | null;
+    status: "online" | "dnd" | "idle" | "invisible";
+}
+
+const StatusStyles = findByPropsLazy("statusItem");
+const setStatus = findByCodeLazy(/default\.track\(\i.\i.CUSTOM_STATUS_UPDATED/);
+
+const PMenu = findComponentByCodeLazy(/{id:\i,label:\i,icon:\i,hint:\i,renderSubmenu:\i,...\i}/);
+const EmojiComponent = findComponentByCodeLazy(/\i.translateSurrogatesToInlineEmoji\(\i.name\)/);
```
Nuckyz, line 53:
```suggestion
const setStatus = findByCodeLazy(/default\.track\(\i.\i.CUSTOM_STATUS_UPDATED/);

const PMenu = findComponentByCodeLazy(/{id:\i,label:\i,icon:\i,hint:\i,renderSubmenu:\i,...\i}/);
const EmojiComponent = findComponentByCodeLazy(/\i.translateSurrogatesToInlineEmoji\(\i.name\)/);
```

Prefer to use strings as the filter instead of regex

Nuckyz (changes requested):

src/plugins/statusPresets/index.tsx
programminglaboratorys: the best I could do is
```suggestion
const setStatus = findByCodeLazy(".CUSTOM_STATUS_UPDATED,{");

const PMenu = findComponentByCodeLazy(/{id:\i,label:\i,icon:\i,hint:\i,renderSubmenu:\i,...\i}/);
const EmojiComponent = findComponentByCodeLazy(/\i.translateSurrogatesToInlineEmoji\(\i.name\)/);
```
the other components seem to be so mangled

Nuckyz:
```suggestion
const setStatus = findByCodeLazy(/default\.track\(\i.\i.CUSTOM_STATUS_UPDATED/);

const PMenu = findComponentByCodeLazy(".menuItemLabel", ".menuItemInner");
const EmojiComponent = findComponentByCodeLazy(".translateSurrogatesToInlineEmoji(");
```

PR #2925, opened 2024-10-03
fix permission viewer

sadan4:


src/plugins/permissionsViewer/index.tsx
```diff
@@ -172,7 +172,7 @@ export default definePlugin({
         {
             find: ".VIEW_ALL_ROLES,",
             replacement: {
-                match: /\.collapseButton,.+?}\)}\),/,
+                match: /\.collapseButton,.+?null,/,
```
Nuckyz, line 175:
```suggestion
                match: /\.expandButton,.+?null,/,
```

Nuckyz (changes requested):

src/plugins/permissionsViewer/index.tsx
sadan4: i think this doesnt change the resulting patch, but idrc

Nuckyz: They swapped the order so it's more right

sadan4: ah

PR #2927, opened 2024-10-03
Plugin to loop videos

Terevenen2:
I made a plugin videoLooper to add a button on videos to be able to loop videos

I have no experience in React whatsoever I based most of the plugin on Lumap PIP plugin : https://github.com/Vendicated/Vencord/tree/main/src/plugins/pictureInPicture

I ran some tests and it looks and works fine

Here's a preview
videoLooper

The SVG I used come from here : https://iconduck.com/icons/49438/loop

### DIFF
--- a/src/plugins/videoLooper/index.tsx
+++ b/src/plugins/videoLooper/index.tsx
@@ -1,0 +1,94 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors*
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { Tooltip } from "@webpack/common";
+
+const settings = definePluginSettings({
+    loop: {
+        description: "Whether to make the video loop or not",
+        type: OptionType.BOOLEAN,
+        default: false,
+        restartNeeded: false
+    }
+});
+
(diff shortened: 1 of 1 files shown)


programminglaboratorys:
Add a README.md

src/plugins/videoLooper/README.md
NobreHD, line 1: You probably should remove this README since it's duplicated

NobreHD:
Works on latest version (4391fcc2)

src/plugins/mediaLooper/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import "./styles.css";
+
+import { classNameFactory } from "@api/Styles";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { Tooltip, useEffect, useState } from "@webpack/common";
+import { RefObject } from "react";
+
+import LoopIcon from "./components/LoopIcon"; // The LoopIcon component you provided
```
EepyElvyra, line 16: Ai slop moment

src/plugins/videoLooper/README.md
Terevenen2: Done, I didn't noticed the old folder was still here thank you

src/plugins/mediaLooper/index.tsx
EepyElvyra: Look, I'm not necessarily against using ai as assistant for coding but at least make it non obvious 😭 you can at least take the comments it generated out

Terevenen2: Yes you are right, I removed the comment like you asked

PR #2932, opened 2024-10-06
fix(showHiddenChannels): Disable hidden channel name obfuscation

Davr1:
This patch aims to revert the channel name obfuscation experiment, which might get added to the default gateway flags in the future.

image


Vendicated:
Discord most likely will stop this capability from working once they are finished experimenting, so I doubt this will be effective. I don't think this is relevant yet anyway, is it? Since it's experiment controlled and you can just not enable it

once they roll out this change for everyone, we can reinvestigate

PR #2939, opened 2024-10-08
Fix HideAttachments when message is forwarded

jamesbt365:
Relevant docs:
https://discord.com/developers/docs/resources/message#messages-resource
https://discord.com/developers/docs/resources/message#message-snapshot-object


jamesbt365:
Going to merge this into #2928 (which is good to go i think) as there technically are conflicting changes if you aren't lazy and do things properly.

PR #2940, opened 2024-10-08
Support poll results in roleColorEverywhere

jamesbt365:
First time messing around with patches, mine is probably suboptimal and is mainly a rip of the `.reactorDefault` patch.

Also not that well versed with js/ts, so I don't expect my code to be the most optimal/fit convention of this project, but hey, I tried!

image


src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -131,6 +137,14 @@ export default definePlugin({
             },
             predicate: () => settings.store.reactorsList,
         },
+        {
+            find: ",reactionVoteCounts",
+            replacement: {
+                match: /,onContextMenu:e=>.{0,15}\((\i),(\i),(\i)\).{0,300}nickname/,
```
Vendicated, line 143: no hardcoding!
```suggestion
                match: /,onContextMenu:\i=>.{0,15}\((\i),(\i),(\i)\).{0,300}nickname/,
```

src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -142,6 +156,12 @@ export default definePlugin({
     ],
     settings,
 
+    getPollColor(args: { channel: { guild_id: string | null; }; user: { id: string; }; }) {
```
Vendicated, line 159:
```suggestion
    getPollColor(props: { channel: Channel; user: User; }) {
```

src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -142,6 +156,12 @@ export default definePlugin({
...
+        return args.channel.guild_id
```
Vendicated, line 160:
```suggestion
        return props.channel?.guild_id
```

src/plugins/roleColorEverywhere/index.tsx
jamesbt365: the Channel type's guild_id is a `string` whereas the actual type discord can return here can be null?

src/plugins/roleColorEverywhere/index.tsx
jamesbt365: The patch above this one hardcodes it too, should that also be changed?

Nuckyz (changes requested):

src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -126,11 +131,19 @@ export default definePlugin({
...
+                match: /,onContextMenu:\i=>.{0,15}\((\i),(\i),(\i)\).{0,250}tag:"strong"/,
                 replace: "$&,style:{color:$self.getColor($2?.id,$1)}"
             },
             predicate: () => settings.store.reactorsList,
         },
+        {
+            find: ",reactionVoteCounts",
+            replacement: {
+                match: /\.nickname,(?=children:)/,
+                replace: "$&style:{color:$self.getColor(arguments[0]?.user?.id,{channelId:arguments[0]?.channel.id})},"
```
Vendicated, line 143:
```suggestion
                replace: "$&style:{color:$self.getColor(arguments[0]?.user?.id,{channelId:arguments[0]?.channel?.id})},"
```

or better yet, pass the raw arguments and handle it inside the function with try catch

Nuckyz:
Supersed by https://github.com/Vendicated/Vencord/commit/ea2772476dc6d042b1e3769ad24727b01b02878d

I've added your feature and credits

Issue #2639, opened 2024-06-29
[Bug] Can't move categories above other categories with ShowHiddenChannels enabled


Apacpua:
### What happens when the bug or crash occurs?

Whenever I enable ShowHiddenChannels plugin, I can't move categories above other categories in my server and I also can't move categories blow or above certain channels or move those channels below or above the category. There's always conflicts.

### What is the expected behaviour?

The expected behavior is that I can freely move channels and categories in my own server.

### How do you recreate this bug or crash?

1. Enable ShowHiddenChannels plugin;
2. In your discord server, move a category over another category;
3. Notice that the category does not move.

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [X] I have read the requirements for opening an issue above

LyubomirT:
Can confirm, this is happening to me as well.

littlegiqnt:
Still an issue, although for me only the top category does not move

Vendicated:
should be fixed

PR #2942, opened 2024-10-09
Fix: Moving channels (read desc.)

sadan4:
This fixes a bug in show hidden channels where you could not move things down if they **were not** in a category.
You also could not move uncategorized channels up or down.

This bug was caused by the `Uncategorized` category being filtered out, breaking discords code.


sadan4:
not sure if this fixes https://github.com/Vendicated/Vencord/issues/2639 or if this is another bug I found while looking into it

PR #2946, opened 2024-10-12
feat(VoiceDownload): add native file picker

ssobkowski:
This PR adds a native file picker dialog from electron to VoiceDownload, removing the need to open a browser window.


Vendicated:
this is iincredibly unsafe. but you can use `DiscordNative.fileManager.saveWithDialog(data, filename);` to download safely.

ssssobek:
Sorry about that. Fixed now.

PR #2953, opened 2024-10-15
Add "To GIF" plugin

ofvv:
Adds a /gif slash command which will convert any image or url to .gif format


Vendicated:
better done via a bot https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md#plugin-rules

PR #2955, opened 2024-10-16
feat(plugin) AutoVX Links

ssnob:
useful for sharing links that don't automatically embed such as tiktok and twitter


jamesbt365:
image

PR #2958, opened 2024-10-19
Fix FakeNitro patch for FakeStickers

fres621:
Fix FakeNitro patch for rendering FakeStickers
This patch might be a little sensitive but it doesn't currently conflict with any other code in the module and I don't think it will
An alternative less sensitive patch would be a lookbehind for the function's name "renderAttachments"

To test this patch enable these 3 options and send a FakeNitro's FakeSticker
image
If the patch applied correctly it should look like a regular FakeSticker in the message


Nuckyz:
Done by https://github.com/Vendicated/Vencord/commit/4e893527583a8ec2cdfb3d815b9d49851e7b6614, thanks regardless!

PR #2963, opened 2024-10-22
Fix: ImageModals and ImageZoom

sadan4:
the ImageModal still needs to be typed

not sure if that is the best/correct place to put the patch needed to fix it, but image modal is used by a few plugins


Vendicated:
superseded by https://github.com/Vendicated/Vencord/commit/a11ccde40fdba6157da006542d6709c1488b5c4c

PR #2964, opened 2024-10-23
fix MutualGroupDMs

sadan4:
*making a seprate pr because the other one might take a while and this was an easy fix*


Vendicated:
thanks!

PR #2968, opened 2024-10-23
LastFMRichPresence: Adds a toggle to ignore Last.fm Now Playing tag...

d-rez:
... to work around scrobblers like Plex.

For whatever reason, Plex scrobbler doesn't broadcast "Now Playing" / "Now scrobbling" actions, and only broadcasts that a song was played, halfway through the song. This means that currently this Vencord plugin is unable to show songs being scrobbled by i.e. Plexamp.

This CR adds a new toggle under plugin settings that allows the user to override the default plugin behaviour and simply use the most recent last.fm song for "listening to" activity, as long as it was broadcast less than 5 mins ago, which (the value) seemed reasonable to me.

Tested locally by building and patching my local Discord install with Vencord as per docs - it seems to work as intended :)

Cheers

### DIFF
--- a/src/plugins/lastfm/index.tsx
+++ b/src/plugins/lastfm/index.tsx
@@ -184,6 +184,11 @@ const settings = definePluginSettings({
         description: "show the Last.fm logo by the album cover",
         type: OptionType.BOOLEAN,
         default: true,
+    },
+    ignoreLastFmNowPlayingTag: {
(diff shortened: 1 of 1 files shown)


d-rez:
Hold up, need to revisit this as there's a logic error somewhere causing the status to be displayed past the expected 5 minutes

Vendicated:
sorry but I don't want to merge this. we shouldn't try to work around broken scrobblers by also being broken ourselves

PR #2969, opened 2024-10-23
Add BiggerImagePreview plugin

sheetau:
lets you to enlarge images and GIFs to fill the window in preview screen displayed when clicked. You can adjust the size between the original size and the window size via right-click menu or the plugin settings menu.

### DIFF
--- a/src/plugins/biggerImagePreview/BiggerImagePreview.tsx
+++ b/src/plugins/biggerImagePreview/BiggerImagePreview.tsx
@@ -1,0 +1,171 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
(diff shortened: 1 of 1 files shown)


Vendicated:
Thanks for your contribution! Unfortunately, this plugin doesn't follow our [guidelines](https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md). Specifically:
> No raw DOM manipulation. Use proper patches and React


Additionally, wouldn't it be possible to achieve something like this with pure css?

sheetau:
Thank you for your feedback.
At first, I was trying to achieve this with CSS, but while changing the size of the image is fine, it creates extra space in the parent element. This results in areas where the image isn't displayed, causing the preview not to close when clicked. Therefore, it was necessary to use JavaScript to make both elements the same size as the image. Additionally, using a plugin allowed for adding finer settings.
If I fix the code and there are no issues, I plan to submit another pull request.

PR #2974, opened 2024-10-24
Update source code options for dev build

Wissididom:
I have seen `Github doesn't let you change the source code zip so it is outdated, please refer to the main branch for the latest source code.` being in the description of the DevBuild, so I thought I make a change that also updates the source code options. I searched for `source code` in both pull requests and issues but didn't find something there.

It basically just updates and force-pushes the devbuild tag.

I'm not sure if commits are squashed, so it would be nice to know if I should squash the commits into one or similar.

I have read the contributing guidelines, but I didn't find something regarding CI changes, so I hope I didn't miss anything.

This is my test PR, where you should be able to see the workflow runs: https://github.com/Wissididom/Vencord/pull/1

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,7 @@ on:
     push:
         branches:
             - main
+            - Update-source-code-options-for-dev-build
         paths:
             - .github/workflows/build.yml
             - src/**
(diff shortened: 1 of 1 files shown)


Vendicated:
this is really pointless, just download source from main

PR #2977, opened 2024-10-25
new plugin: BlurMediasToggle

Walkoud:
**Plugin: BlurMediasToggle**

**Description:**
The BlurMediasToggle plugin blurs all media content (images, videos) within channels, allowing users to keep sensitive content private from onlookers. The plugin offers a toggle key for quick enable/disable functionality. Additionally, media will unblur temporarily when hovered over, ensuring convenient access without compromising privacy.

**Key Features:**

- Blurs all images and videos in channels.
- Quickly toggle the blur effect with a hotkey.
- Hover to temporarily unblur media.
- Ideal for ensuring on-screen privacy in shared or public spaces.

https://github.com/Walkoud/VC-BlurMediasToggle
qeAOy72nXV
Discord_rg4d71Q76G

image

### DIFF
--- a/src/plugins/BlurMediasToggle/index.tsx
+++ b/src/plugins/BlurMediasToggle/index.tsx
@@ -1,0 +1,158 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2023 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { addChatBarButton, ChatBarButton, removeChatBarButton } from "@api/ChatButtons";
+import { addPreSendListener, removePreSendListener, SendListener } from "@api/MessageEvents";
+import { definePluginSettings } from "@api/Settings";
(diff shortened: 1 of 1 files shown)


nin0-dev (commented):
don’t use dom manipulation. use proper patches and React

Vendicated:
thanks for your contribution! however, this seems too niche / not very useful. one can just disable/enable embeds in discord settings for a similar effect

+ BlurNSFW exists

angrybirdice:
bandicam 2025-03-24 11-08-32-210
the video still shows as continued playing even though the video ended which its an issue

nakoyasha:
> bandicam 2025-03-24 11-08-32-210 the video still shows as continued playing even though the video ended which its an issue

This is a Discord issue and is not related to the plugin at all (it's not even in vencord...)

Walkoud:
> thanks for your contribution! however, this seems too niche / not very useful. one can just disable/enable embeds in discord settings for a similar effect
> 
> * BlurNSFW exists

I think BlurMediasToggle is actually quite useful, even if Discord has options to disable media previews globally.

First, Discord’s built-in settings are not flexible. If you disable media previews, it applies to everything across all servers and DMs, and if you want to see previews again, you have to manually toggle the setting back on. With BlurMediasToggle, you get a much more granular and instant control, allowing you to quickly decide when you want to see media and when you don’t, without digging through settings.

Second, privacy matters. Many people use Discord in different environments—at home, in public, at work, while streaming, etc. Sometimes, you just don’t want random images popping up on your screen for everyone nearby to see. Even if the media isn’t explicitly NSFW, it can still be awkward, borderline, or just out of context. BlurMediasToggle helps prevent unwanted surprises when you’re in a situation where you need to be mindful of what appears on your screen.

Lastly, BlurNSFW is not the same thing. It specifically hides NSFW-classified content, but it doesn’t cover suggestive, weird, or unexpected media that might not be flagged as NSFW but could still be inappropriate in certain situations.

In short, BlurMediasToggle gives users instant, situational control that neither Discord’s native options nor BlurNSFW provide. It’s a great addition for anyone who wants a quick, reliable way to manage media visibility without messing with settings or relying on automatic detection.

PR #2980, opened 2024-10-27
Add button to enable plugin on veebot plugin embeds

jamesbt365:
Maybe this is too niche, but I've seen an outstanding amount of users not know what to do when prompted with these embeds.

image
image

Could force a restart after the plugin is enabled too (or maybe add another button to do that?) but I don't even know if this will be accepted, let alone that.


Masterjoona:
#2732

jamesbt365:
Aww man.. Will reopen if I actually have something worth superseding yours.

PR #2982, opened 2024-10-28
New Plugin: messageCountdown

Kewwie:
A countdown starts when you send a message, you need to wait for it to finish before it can start again

===

Helps people who likes to grind xp in guilds

### DIFF
--- a/src/plugins/messageCountdown/index.tsx
+++ b/src/plugins/messageCountdown/index.tsx
@@ -1,0 +1,129 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addChatBarButton, ChatBarButton, removeChatBarButton } from "@api/ChatButtons";
+import { addPreSendListener, removePreSendListener } from "@api/MessageEvents";
+import { definePluginSettings } from "@api/Settings";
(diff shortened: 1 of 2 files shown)


Baget88:
AHHH YESS PLEASE ADD THIS, ITS ACTUALLY RLLY GOOD FOR XP 🙏

tester5454:
It's a good plugin why hasn't it been added yet :(
it helped with grinding in guilds and timing rank ups and i think it deserves to be in the plugin store

nin0-dev:
not only is this plugin extremely niche and useless to 99% of people that have a life beyond XP grinding, it also needs to be reviewed before it even is merged

be patient

tester5454:
This plugin's existence is 99% for the people that don't have a life and it's been 2 months how much longer are they gonna take

Baget88:
speaking on behalf of 100people in this server i am in, they want it and so do i
some servers give prizes for getting xp which is another reason that it is useful

lewisakura:
Like nin0 has said, this plugin has zero use outside of grinding bot XP in guilds, and we recently updated our plugin guidelines to ban these sorts of plugins:
> No plugins that interact with specific Discord bots (official Discord apps like Youtube WatchTogether are okay)

Since there was no actual given use case beyond this in the PR description (and the whole plugin is built around the concept of grinding XP based on the settings), I'm going to close and lock this.

PR #2983, opened 2024-10-28
Added the normalEmojis plugin

7ijme:
I have always wanted Apple Emojis on my Linux installation. With this plugin, you get that. 

I remove the source attribute of the `img` tag, and create a `::before` element with the actual emoji character as content. This way, the emoji gets rendered as the font you have installed for emojis.

This is what it looks like:
image

### DIFF
--- a/src/plugins/normalEmojis/index.ts
+++ b/src/plugins/normalEmojis/index.ts
@@ -1,0 +1,35 @@
+import definePlugin from "@utils/types";
+import style from "./style.css";
+
+let interval: NodeJS.Timeout;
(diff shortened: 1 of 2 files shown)


jamesbt365:
Can be done with CSS, no plugin needed.
https://betterdiscord.app/theme/EmojiReplace

jedenastka:
As far as I understand, this is a duplicate of #2061, though this implements jumbo emojis, while my thing purposefully doesn't (though I wanted to maybe make a toggle for that). It also does this very differently.

7ijme:
Closing due to not being added for 3 months

PR #2984, opened 2024-10-28
fix(plugin): make NoBlockedMessages not hide likely spammers

### LINKED ISSUE #2466: [Bug] NoBlockedUsers hides likely spammers
### Discord Account

_No response_

Messages from likely spammers are completely hidden despite not being blocked. iLoveSpam fixes this as it completely overrides the flag I'm guessing.

I would expect them to render as usual, with an option to show messages

1. Find a message from a likely spammer
2. Enable NoBlockedMessages
3. Press Ctrl+R
4. it has gone :o

### Errors

_No response_

### Request Agreement

notTamion:
fixes #2466

### DIFF
--- a/src/plugins/noBlockedMessages/index.ts
+++ b/src/plugins/noBlockedMessages/index.ts
@@ -22,6 +22,7 @@ import { Logger } from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy } from "@webpack";
 import { Message } from "discord-types/general";
+import {i18n} from "@webpack/common";
 
 const RelationshipStore = findByPropsLazy("getRelationships", "isBlocked");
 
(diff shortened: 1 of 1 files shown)


src/plugins/noBlockedMessages/index.ts
```diff
@@ -77,6 +77,7 @@ export default definePlugin({
     },
 
     shouldHide(props: MessageDeleteProps) {
-        return !props?.collapsedReason?.message.includes("deleted");
+        const message = props?.collapsedReason?.message;
+        return !message.includes("deleted") && !message.includes("spammer");
```
jamesbt365, line 81: Are we sure this isn't a problem when translated? the "deleted" part isn't because its generated by Vencord but if the reason changes from spammer if translated this could be a problem.

sadan4: > Are we sure this isn't a problem when translated? the "deleted" part isn't because its generated by Vencord but if the reason changes from spammer if translated this could be a problem.

that could probably be tested with a similar fix I made a bit ago
https://github.com/Vendicated/Vencord/pull/2921

Nuckyz:
Closing as this was fixed in another commit

PR #2987, opened 2024-10-30
This plugin allows users to type /vcBtop to receive a message display…

Mousica-dev:
a small plugin that shows people their username, CPU consumption, RAM usage, and network data.

like this:
Screenshot 2024-10-30 at 03 30 44


Vendicated:
sorry but this is not useful / too niche

~~also how in the world do you only have 260mb ram~~

PR #2989, opened 2024-11-01
Update plugin description: FixImagesQuality due to confusing behavior for less technically inclined users, and general ineffectiveness. 

emkab:
Delete plugin fixImagesQuality.
There is no reason for WEBP to cause quality loss when it isn't called for by the connection speed of the user. While it is both a lossy and lossless format, it's very effective compression algorithm enables it to decrease file sizes enough to often load it at acceptably high qualities on weaker connections. 
All this plugin does is replace any image which would be loaded a WEBP with PNG, which while it is a lossless format, it might take significantly longer to load due to file sizes (PNG's lossless compression still keeps file sizes quite big). It also means that this plugin completely disables discord's ability to match users internet speed and image quality. Which might lead to unexpected behavior for less technically inclined users. This should at least be mentioned in the plugin description.

Signed, a WEBP defender.


Sqaaakoi:
The reason the plugin exists because Discord **had** an issue where their media service would send a lossy image when a client requests a lossless image
Do note that the plugin has been removed in the past because Discord has fixed issues and then readded due to that issue being discovered

emkab:
> The reason the plugin exists because Discord **had** an issue where their media service would send a lossy image when a client requests a lossless image Do note that the plugin has been removed in the past because Discord has fixed issues and then readded due to that issue being discovered

This is Interesting, and I see why it would dismiss my pull request. But why then not just add this information, together with a warning that this plugin may slow loading times, to the plugin description?

Vendicated:
?

PR #2993, opened 2024-11-03
Fix all plugins for new i18n lib

Nuckyz:


src/plugins/_core/settings.tsx
```diff
@@ -148,17 +149,13 @@ export default definePlugin({
...
-        } catch {
-            return firstChild === "PREMIUM";
-        }
+        const names = {
+            top: getIntlMessage("USER_SETTINGS"),
+            aboveNitro: getIntlMessage("BILLING_SETTINGS"),
+            belowNitro: getIntlMessage("APP_SETTINGS"),
+            aboveActivity: getIntlMessage("ACTIVITY_SETTINGS")
+        };
+        return header === names[settingsLocation];
```
Vendicated, line 158: why remove this try catch? please add it back

src/plugins/betterFolders/index.tsx
```diff
@@ -276,7 +277,7 @@ export default definePlugin({
     makeGuildsBarGuildListFilter(isBetterFolders: boolean) {
         return child => {
             if (isBetterFolders) {
-                return child?.props?.["aria-label"] === i18n.Messages.SERVERS;
+                return child?.props?.["aria-label"] === getIntlMessage("SERVERS");
```
Vendicated, line 280: this definitely needs a try catch (was also unsafe before)

src/plugins/moreUserTags/index.tsx
```diff
@@ -295,19 +296,19 @@ export default definePlugin({
 
     isOPTag: (tag: number) => tag === Tag.Types.ORIGINAL_POSTER || tags.some(t => tag === Tag.Types[`${t.name}-OP`]),
 
-    getTagText(passedTagName: string, strings: Record<string, string>) {
-        if (!passedTagName) return strings.APP_TAG;
+    getTagText(passedTagName: string) {
```
Vendicated, line 299: this definitely needs a try catch (was also unsafe before)

src/plugins/typingIndicator/index.tsx
```diff
@@ -72,24 +73,26 @@ function TypingIndicator({ channelId }: { channelId: string; }) {
     const typingUsersArray = Object.keys(typingUsers).filter(id => id !== myId && !(RelationshipStore.isBlocked(id) && !settings.store.includeBlockedUsers));
     let tooltipText: string;
 
+    // the new syntax is getIntlMessage("ONE_USER_TYPING", { a: getDisplayName(guildId, typingUsersArray[0]) });
+
```
Vendicated, line 77:
```suggestion
```

src/utils/patches.ts
```diff
@@ -16,12 +16,31 @@
...
+import { runtimeHashMessageKey } from "./intlHash";
 import { Patch, PatchReplacement, ReplaceFn } from "./types";
 
 export function canonicalizeMatch<T extends RegExp | string>(match: T): T {
-    if (typeof match === "string") return match;
-    const canonSource = match.source
-        .replaceAll("\\i", "[A-Za-z_$][\\w$]*");
+    let partialCanon = typeof match === "string" ? match : match.source;
+    partialCanon = partialCanon.replaceAll(/#{intl::([A-Za-z_$][\w$]*)}/g, (_, key) => {
+        const hashed = runtimeHashMessageKey(key);
```
Vendicated, line 25: please verify this won't cause crashes in case this throws (just make runtimeHashMessageKey throw and see). if it does, we need to add error handling somewhere - if it doesn't it's better not to

src/webpack/common/utils.ts
```diff
@@ -56,7 +57,10 @@ export const { match, P }: Pick<typeof import("ts-pattern"), "match" | "P"> = ma
 
 export const lodash: typeof import("lodash") = findByPropsLazy("debounce", "cloneDeep");
 
-export const i18n: t.i18n = findLazy(m => m.Messages?.["en-US"]);
+export const i18n = mapMangledModuleLazy('defaultLocale:"en-US"', {
+    intl: filters.byProps("string", "format"),
+    t: filters.byProps(runtimeHashMessageKey("DISCORD"))
```
Vendicated, line 62: if this throws an error, entirety of vencord will explode so this absolutely needs a try catch

src/utils/patches.ts
```diff
@@ -16,12 +16,31 @@
...
+
+        const isString = typeof match === "string";
+        const hasSpecialChars = /^[\d]/.test(hashed) || !/^[\w$]+$/.test(hashed);
+
+        if (hasSpecialChars) {
+            return isString
+                ? `["${hashed}`
+                : String.raw`(?:\["${hashed})`.replaceAll("+", "\\+");
```
Sqaaakoi, line 33: is there any reason why the `"]` isn't checked for here? it makes patches look ugly

I have just done `#{intl::ADD_REACTION_NAMED}\"],{emojiName` because of this (previously it was `.ADD_REACTION_NAMED.format`)

Nuckyz: I guess you are right, it can be checked

PR #2994, opened 2024-11-03
feat(plugin): WebPWA

ThaUnknown:
This plugin makes discord installable as a PWA when used in browser, it's effectively a *more than slightly* worse version of Vesktop, but since it can re-use a browser that's already in use, does wonders for reducing RAM usage even further on devices which have little of it.

This enables vencord on web to:
- makes vencord installable as a PWA
  - lets discord open as a standalone window just like an app
  - lets discord be launched at startup
- adds support fo discord's custom title bar
  - supports dynamic sizing, zoom, display pixel density etc
- shows a notification badge
- added global keybinds
  - updated keybinds notice to say how to configure it

Most of this code is re-used from vesktop with slight alterations

there are still a few things to consider:
- what to do on firefox? firefox doesnt support ANY of this, but vencord also doesn't ship on firefox anymore?
- what to do on vesktop, this plugin is a web target, and there's no "extension only" target
- discord FORCES 720p30fps screenshare on web, for no good reason, idk how to not do that, while on the JS side it's easy since it's just changing track constraints, I have 0 success with trying to patch discord's garbled mess

https://github.com/user-attachments/assets/1f66f423-2ce0-4cc5-a4e2-a45796096b44

about://apps settings for auto-start:

image


src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
...YAOhQX0YMqfGLZJgOHEo0G+fuXCAhHnwqU1FQG3nEzLK4nALkacPPwNUE35igDC3La9t/rFkdfTyh5ggAOXw3EoDxBoG+ZMHRkk05HQOU1UR+VuOF4T6spPEUAX50U1M8Ak0hAepxWoGQ5FyPAqCNWZ1cvza90sZZdquZJAhBITJ25e4ASza4gqD+WTkKvDftD+orlPtFTOdHYnKqqgnovouBZAjhiW8DKfCJ5schTE4DVN6OKMsNLy/2u7Ot5AugApf1Oga9Cnha4mwaY1feZ8PNlwfxV7tZUm3aSADrhJJ4pUxgV8l6BtgHkmFLM/wCp9+4amP+sW8/0k7GFJIAuUBP5Bpr9+UXEyhxJBMkMKxvVkRO/R2NIAugBng4iAHw3yq2BjSa1FlHkxNeCkj1fBtIkucWF4j6CmDJH3i+wGHi93bH6JoAndg0KrJBL/cTgyRVAYoyOKFF0df04KHQ7iP9dHh/qBM+s4vHjPCxTiRYvX5JXa1Y3bmxXEkCSVo3HEbRmlhAwC6QMTLIZWS0VBJh3AKjyoa3yucp+O1Npyqt1JQGkaPmJFZxx/CeRSUyYIVcFKYKppbr42jO9BAVVXwzss1Iu87WA1n0ZSQCp4XdE7eLpkWNUcBHFcKX0FRgIrGiKeS0znsmJqc96NWrPYETjzUkCMAPVQ0lJuI2KiGmKJIMkQWZeC6Zl8Meed1q+/SQ1tryaJAALIG9/zIQmM9MUIj5HOg+7A11tUkF/9jFWtaDt+Rfkvt700SkJwHSIj+xg6szdvZVo9kRF5UkqYTIRjbBYBFt1x8xbFMZKVaFVOdF9q7xwB99OBpAEkGZrtN9KzJrIhAkK0wQwnwwif5rFMqf79mQbH7GCN4hRq/pb3li+qP8uczqTrWpBQBKAFpQsLCNWCL5W/ziwbxwDZ4B4JAEnOY4U4t56bCcf3mWV1zNhHftb33Hi81kWmt/yriQBWA65/g7jpNCcORI+ZTSzOhqsDCEFw8H8LyAq0N+icTWI+QATbQPzdgI2xoDNGaRuzIzWb5bLeeNwNqslSQBmIWtRu+LokTg2KAbfIKgYQAoPAGEAq8j/Z0jsMcTIB1EegGwG9yZCBtAFaYgMuIT2xy2Z6hloIOZGJtQTUA8Fe8HYBfBehrLTh9jOxii2r/BoIg2LzGt6N/8PoMvr9NjXOyAAAAAASUVORK5CYII=",
```
Sqaaakoi: there has to be a better way of doing this. fetch this at runtime from within Discord

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import "./styles.css";
```
Sqaaakoi, line 7: This import is static. You do not want to be doing this with the styles you have. Take a look at GameActivityToggle for a good example as to why you should and how you can use a managed style

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+import { filters, waitFor } from "@webpack";
+import { RelationshipStore } from "@webpack/common";
+
+const manifest = {
+  name: "Discord",
+  short_name: "Discord",
+  start_url: "https://discord.com/channels/@me", // URL when PWA launches
+  display: "fullscreen",
+  display_override: ["window-controls-overlay"],
+  lang: "en-US",
```
Sqaaakoi, line 21: Remove this, it's irrelevant

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { filters, waitFor } from "@webpack";
+import { RelationshipStore } from "@webpack/common";
+
+const manifest = {
+  name: "Discord",
+  short_name: "Discord",
+  start_url: "https://discord.com/channels/@me", // URL when PWA launches
```
Sqaaakoi, line 18: Will break on canary and PTB, and the stock desktop app loads `/app` which redirects here

You might want to look into making this a getter and using variables set in `window.GLOBAL_ENV` if setting it to `/app` doesn't work properly

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  display: "fullscreen",
+  display_override: ["window-controls-overlay"],
+  lang: "en-US",
+  background_color: "#2a2a2f",
+  theme_color: "#2a2a2f",
+  scope: "https://discord.com", // scope of all possible URL"s
```
Sqaaakoi, line 24: Set this to `/`

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+  ]
+};
+
+const isMac = navigator.platform.startsWith("Mac");
+
+let link: HTMLLinkElement | undefined;
+
+let GuildReadStateStore: any;
+let NotificationSettingsStore: any;
```
Sqaaakoi: You should be using findStoreLazy instead of fetching these later

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    toFind--;
+    if (toFind === 0) setBadge();
+  });
+}
+
+let enabled = false;
+
+waitForAndSubscribeToStore("GuildReadStateStore", store => (GuildReadStateStore = store));
+waitForAndSubscribeToStore("NotificationSettingsStore", store => (NotificationSettingsStore = store));
+waitForAndSubscribeToStore("RelationshipStore");
```
Sqaaakoi: Remove this bloat and set it up in the start function. These stores should already be loaded by then.

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  },
+  patches: [
+    {
+      find: "platform-web",
+      replacement: {
+        // eslint-disable-next-line no-useless-escape
+        match: /(?<=" platform-overlay"\):)\i/,
+        replace: "$self.getPlatformClass()"
+      }
+    },
```
Sqaaakoi, line 98: A PWA might be installed but the app might not be running within a PWA. Consider checking if you're in a PWA before overriding the stock behaviour. https://web.dev/learn/pwa/detection might help you

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const isMac = navigator.platform.startsWith("Mac");
+
+let link: HTMLLinkElement | undefined;
+
+let GuildReadStateStore: any;
+let NotificationSettingsStore: any;
+
+function setBadge() {
+  if (!enabled) return;
```
Sqaaakoi: Reminder there is no need to manage your own running state as long as you migrate to the start/stop function

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      sizes: "256x256",
+      type: "image/png"
+    }
+  ]
+};
+
+const isMac = navigator.platform.startsWith("Mac");
```
Sqaaakoi, line 36: Consider inlining this

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+let GuildReadStateStore: FluxStore & { getTotalMentionCount: () => number; hasAnyUnread: () => boolean; };
+let NotificationSettingsStore: FluxStore & { getDisableUnreadBadge: () => boolean; };
```
Sqaaakoi, line 39: You should be using findStoreLazy

src/plugins/webPWA/index.ts
```diff
@@ -0,0 +1,113 @@
...
+        cb?.(store);
+        store.addChangeListener(this.setBadge);
+
+        toFind--;
+        if (toFind === 0) this.setBadge();
+      });
+    };
+    waitForAndSubscribeToStore("GuildReadStateStore", store => (GuildReadStateStore = store));
+    waitForAndSubscribeToStore("NotificationSettingsStore", store => (NotificationSettingsStore = store));
+    waitForAndSubscribeToStore("RelationshipStore");
```
Sqaaakoi, line 81: Those stores should be loaded by then, directly attach the listeners

Sqaaakoi (commented):
Brilliant idea with a few too many assumptions about where the code is running.

Ignore the outdated stuff, you made a lot of changes did many force pushes while I was trying to review this.

src/plugins/webPWA/index.ts
ThaUnknown: quite literaly copied 1:1 from vestkop, with types and not any

src/plugins/webPWA/index.ts
ThaUnknown: vesktop didnt, but oh well

src/plugins/webPWA/index.ts
ThaUnknown: uuuuuh??? english pls?

src/plugins/webPWA/index.ts
ThaUnknown: IIRC it was required on some browsers to register a PWA

src/plugins/webPWA/index.ts
ThaUnknown: there's no PTB on web? or am I missing something?

Sqaaakoi: > there's no PTB on web? or am I missing something?

https://ptb.discord.com/app
https://canary.discord.com/app

src/plugins/webPWA/index.ts
ThaUnknown: I'm against this. It requires basic competence from the user, since this isn't reactive, if a user goes into pwa mode without restarting shit will break

Sqaaakoi: > I'm against this. It requires basic competence from the user, since this isn't reactive, if a user goes into pwa mode without restarting shit will break

True. I'm not actually sure what this would look like. Perhaps some hackery with CSS media queries could be done directly? I say this as I don't know if using a React hook would work here. Is it even worth checking?

ThaUnknown: Assuming other web plugins like web right click and web keybindns dont... no?

src/plugins/webPWA/index.ts
ThaUnknown: I see little value in inlining this, so i wont

src/plugins/webPWA/index.ts
Sqaaakoi: > uuuuuh??? english pls?

You might not need to be loading it all like this, if you lazy load the stores using `findStoreLazy` you *should* be able to directly call addChangeListener on the stores here.

If that doesn't work, I would recommend replacing waitForAndSubscribeToStore with a promise wrapper around waitFor that also attaches the listeners and then using. `Promise.all` like `Promise.all(["Store"].map(waitForAndSubscribeToStore)).then(() => setBadge())`.
Make sure to replace the placeholder variables I used there. Do note this isn't _technically necessary_, it'll just make the code less messy

If you copied this from Vesktop, this idea could be upstreamed as a refactor ~~that will never happen~~

ThaUnknown:
@Sqaaakoi any chance if you know if discord still hardcodes 720p 30fps for screensharing? I remember from GM days the browser code looked something like:
```js
screenShare(videoQuality, audioQuality) {
// just ignore the settings lol
return navigator.mediaDevices.getDisplayMedia({ video: { height: 720, frameRate: 30 } })
}
```
but I'm struggling to navigate the new minified code, I wanted to fix that in this plugin too, but I can't verify if it's an issue

src/plugins/webPWA/index.ts
Sqaaakoi: > Assuming other web plugins like web right click and web keybindns dont... no?

They check a different way, which reminds me, Vencord plugins can be specified as web only by adding a `.web` suffix to the plugin folder name. You should probably do this for your plugin.

ThaUnknown: right, wish this was documented lmao

Sqaaakoi:
> @Sqaaakoi any chance if you know if discord still hardcodes 720p 30fps for screensharing? I remember from GM days the browser code looked something like:
> 
> ```js
> screenShare(videoQuality, audioQuality) {
> // just ignore the settings lol
> return navigator.mediaDevices.getDisplayMedia({ video: { height: 720, frameRate: 30 } })
> }
> ```
> 
> but I'm struggling to navigate the new minified code, I wanted to fix that in this plugin too, but I can't verify if it's an issue

I'm pretty sure it does, but do not count on my word.

I am not sure if this plugin is the right place to be doing it, 2 reasons

1. Large PRs often get ignored by maintainers.
2. This would be better suited as a port of the existing Vesktop code to support web as well as Electron. If you want to do this, may I suggest asking in the Discord server (likely #vesktop-development, you may need to ask for permissions through modmail, do state your purpose as they might suggest somewhere else to be better suited) for help?

src/plugins/webPWA.web/index.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import { disableStyle, enableStyle } from "@api/Styles";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { findStoreLazy } from "@webpack";
+import { RelationshipStore } from "@webpack/common";
+import type { FluxStore } from "@webpack/types";
+
+import style from "./style.css?managed";
```
Faf4a, line 14: file name mismatches
```suggestion
import style from "./styles.css?managed";
```

xPaw:
I suspect this isn't going to work when using the userscript version because of Discord's CSP. I tried to manually inject the webmanifest just by copying the code out of here:

> Refused to load manifest from 'blob:https://discord.com/2720f46f-fefa-4280-ac70-2df63a550683' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback.

ThaUnknown:
I'm unsure why? This plugin works fine on my end? I'm running brave with the vencord extension

xPaw:
I mean the userscript version of Vencord, not the extension.

ThaUnknown:
ah yeah, it defo won't work there, nothing you can do about it, but it's a non issue, the extension is fine, and it's not heavy or problematic in any ways, the bigger issue is discord calls randomly dying because chromium suspends audiocontext in the tab for inactivity

xPaw:
I tried the extension build, and it's dying on document.querySelector('link[rel="icon"]') being null. For some reason Discord replaces that icon on load with a base64 version? Don't know why it's not being found by this code during load.

image

xPaw:
> que? this part works fine for me?

open devtools and look at the application > manifest, it reports an invalid scope url.

> that's what this is for

I don't know what that does, if anything.

ThaUnknown:
I'm uncertain why that patch wouldn't work for you, I copied it 1:1 from vesktop

ThaUnknown:
I'm uncertain why that patch wouldn't work for you, I copied it 1:1 from vesktop

xPaw:
Does it for you? Can you make a screenshot.

ThaUnknown:
I'm doing these updates from my lap on my phone

xPaw:
https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API is what that option is for.

ThaUnknown:
yes im aware, but i'm patching discord to use native title bar, it's not for some reason

src/plugins/webPWA.web/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    } catch (e) {
+      console.error(e);
+    }
+  },
+  start() {
+    enableStyle(style);
+    const manifest = {
+      name: "Discord",
+      short_name: "Discord",
+      start_url: 'https' + window.GLOBAL_ENV.WEBAPP_ENDPOINT + "/app", // URL when PWA launches
```
uwx, line 45: This is borked
```suggestion
      start_url: 'https:' + window.GLOBAL_ENV.WEBAPP_ENDPOINT + "/app", // URL when PWA launches
```

ThaUnknown:
@xPaw i fixed this, finally had time to touch it

src/plugins/webPWA.web/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    // installability
+    const manifest = {
+      name: "Discord",
+      short_name: "Discord",
+      start_url: "https:" + window.GLOBAL_ENV.WEBAPP_ENDPOINT + "/app", // URL when PWA launches
+      display: "fullscreen",
+      display_override: ["window-controls-overlay"],
+      lang: "en-US",
+      background_color: "#2b2d31", // var(--background-secondary)
+      theme_color: "#1e1f22", // var(--background-tertiary)
```
splatterxl, line 46: Is there any way to resolve these css vars automatically?

ThaUnknown: I've investigated this. No it is not possible. These variables only accept primitive CSS colors, that is hex and rgb, in semi-rare cases rgba on SOME linux distros, but not all, discord and many themes use cielab, calc and other, which would make this non-functional.

The best I came up with was to maybe introduce a setting, where users can change this to some variable if they use themes, but having themes re-calculate said styling isn't easy either, so I left it as is.

Additionally chrome does NOT recognize manifest updates due to security, as one could for example make a PWA look like a note taking app, then once installed update it to for example act like a phishing attack for email, bank etc, so updating these values post-installation does nothing as well.

Vendicated: you can dynamically resolve them https://github.com/Vencord/Vesktop/blob/main/src/renderer/themedSplash.ts

ThaUnknown: @Vendicated once again, manifest only supports primitive colors, this returns `oklab(0.323409 0.00288205 -0.0101295)` and oklab isn't supported by manifest

Vendicated: that code returns rgb, or at least should. if not, it probably broke in a recent chrome version and we should look into fixing it

ThaUnknown: no, it returns it correctly, it's simply discord that's changing to cielab, it returns whatever colorspace method is used, so if a theme uses cielab, it will return cielab, if it uses lab, it will return lab, if it uses hls it will return hls, getComputedStyle simply computes and normalizes the values of the color space, it doesnt convert the colorspace into another colorspace

PR #2995, opened 2024-11-04
mutual server members tab

Z1xus:
adds a new tab to serverinfo plugin that shows server members that are also in your other servers, sorted by amount of mutual servers
image


misopog:
lgtm

PR #2996, opened 2024-11-05
fix no blocked messages for new i18n system

sadan4:
plus it's not dependent on lang anymore

### DIFF
--- a/src/plugins/noBlockedMessages/index.ts
+++ b/src/plugins/noBlockedMessages/index.ts
@@ -21,14 +21,14 @@ import { Devs } from "@utils/constants";
 import { Logger } from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy } from "@webpack";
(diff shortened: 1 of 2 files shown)


src/plugins/noBlockedMessages/index.ts
```diff
@@ -77,6 +78,6 @@ export default definePlugin({
     },
 
     shouldHide(props: MessageDeleteProps) {
-        return !props?.collapsedReason?.message.includes("deleted");
+        return props.collapsedReason() === i18n.t[runtimeHashMessageKey("BLOCKED_MESSAGE_COUNT")]();
```
Nuckyz, line 81:
```suggestion
        return props.collapsedReason() === getIntlMessage("BLOCKED_MESSAGE_COUNT")
```

silly

Nuckyz (changes requested):

sadan4:
this is cursed, there needs to be a way to generate an arbitrary message

Nuckyz (approved):

PR #2999, opened 2024-11-06
feat(plugin): muteBlockedUsers

Vexify4103:
Plugin Description:
1. Adds a "Mute/Unmute and Block/Unblock" button for all users. Unlike the standard Block/Unblock button, which only restricts certain users from messaging you, this option also mutes them in voice channels for added control.
2. Includes an automatic feature that mutes blocked users in voice channels, enhancing visibility and manageability.
This plugin effectively transforms standard blocks into "real" blocks, creating a smoother and more organized communication environment.
image
image

### DIFF
--- a/src/plugins/muteBlockedUsers/index.tsx
+++ b/src/plugins/muteBlockedUsers/index.tsx
@@ -1,0 +1,216 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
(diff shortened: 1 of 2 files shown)


Vexify4103:
It was brought to my attention that this just adds a "feature" to an existing plugin. 
If so I would love to contribute to that plugin instead of crying about it on Discord. 
Thanks <3

PR #3005, opened 2024-11-07
New Plugin: SortFriendList

Giingu:
**SortFriendList**
----
Adds to the Friend List the date you became friends and sorts them.

image


Sqaaakoi:
Why aren't you modifying the existing SortFriendRequests plugin to support this use case, especially considering the code is almost the same? I know the name would be wrong, but plugin settings can be kept after a rename by statically calling `migratePluginSettings(newName, ...oldNames)` (see how other plugins do it for an example)

<sub>also how do you have 1k friend requests??</sub>

PR #3006, opened 2024-11-07
feat(plugin): Theme Toggler

RedCrafter07:
This is a small plugin I made for toggling themes with a hotkey or if you start a stream. You may also show a popup once you toggle the theme. All of the mentioned options are configurable, you can find a picture below for reference.

I'd also want to give some credit to Kodarru, also credited in the code. I used fragments of the code she wrote for the StreamerModeOnStream plugin for the theme disable functionality.

Hope you like it :)

image


Vendicated:
you can basically already achieve this with the VencordToolbox plugin. Simply import all your themes via QuickCSS and then toggle QuickCSS via the toolbox

as such this plugin is essentially superfluous

thanks for your contribution nevertheless!

RedCrafter07:
Thanks for the feedback! You're totally right there are other approaches - I just found the hotkey/stream method really convenient in my own day to day use. Since some friends mentioned they'd find this useful too, I figured I'd share this plugin via PR.

PR #3009, opened 2024-11-11
feat(plugin): Letter emoji shortcuts (for easier typing of the blue letters)

UnmatchedBracket:
This plugin adds easier ways to type the blue letter emojis (`:regional_indicator_*:`).

----------

## Option 1: Autocomplete
Makes e.g. `a` and `aa` suggest 🇦 first, `b` and `bb` suggest 🇧 first, etc.

https://github.com/user-attachments/assets/8a8dc738-6c55-406c-a57b-078a1354eda3

This will not insert the emoji if e.g. `:bb:` is typed.
## Option 2: Alias
Makes typing e.g. `:cc:` correct to 🇨. Single letters like `:c:` don't work to avoid conflicts with 🅰️ and 🅱️.

https://github.com/user-attachments/assets/4d817803-6c90-47ed-83e4-711e96d6d159

Defaults to off due to the side-effect of making the plaintext `:cc:` render as 🇨.
## Option 3: All Letters Autocomplete
Makes `ltr`, `letter`, and `letters` suggest every letter, most useful for typing with reactions/typing lots of letters as emojis.

https://github.com/user-attachments/assets/68efa72d-de72-4e13-8c69-ca232d441c4e

----------
Each option is individually toggleable without reload:
plugin settings popup

### DIFF
--- a/src/plugins/letterEmojiShortcut/index.tsx
+++ b/src/plugins/letterEmojiShortcut/index.tsx
@@ -1,0 +1,89 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+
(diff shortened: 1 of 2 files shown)


Vendicated:
can be done with the TextReplace plugin already

thanks for your effort though

UnmatchedBracket:
Not entirely, this works in the context of reactions and the emoji picker as well

Vendicated:
too niche then

cursefroge:
partially replicated with this textreplace (regex):
find: `:(?<char>[a-z]):`
replace: `:regional_indicator_$<char>:`

PR #3017, opened 2024-11-15
feat(themes): Online Themes CSS Override

### LINKED ISSUE #1466: [Feature Request] A theme configurator for Vencord/Vesktop
### Discord Account

coolesding (real)

I would like to have a theme configurator when importing a theme through vencord, for example when you paste in the theme link you get a little configurator symbol where you can then edit all the variables easily so you don't have to put them into quickcss or edit it on github(because:). 

for example something like this and when you click the gear you open a little window that lets you scroll through the variables and easily lets you edit them
image
also, maybe add categories so for example I can sort everything that changes the background into "Background" (wow so shiny) because this is just stupid and annoying.
image
thank you, you're great. :shiggy:

### Request Agreement

yuna0x0:
Resolves #1466.

This PR implemented an online theme CSS override feature, which allows user to configure their online theme using CSS override.

## Limitation
- When there's an `@import` CSS at-rule in the remote stylesheet, those imported stylesheets' CSS rules won't be visible.  It will make things more complex if we want to fetch and parse those stylesheets recursively.

## Screenshots
Screenshot 2024-11-16 at 3 23 11 AM
Screenshot 2024-11-16 at 3 23 52 AM


yuna0x0:
> I think it'd be better as a modal rather than an inline, inline (at least, from my experience) has always been a lot more messy and hard to use

I agree with this. Change to modal component.

Screenshot 2024-11-16 at 8 28 26 AM
Screenshot 2024-11-16 at 8 28 41 AM

yuna0x0:
Added handly icon buttons for quick edit and reset.

https://github.com/user-attachments/assets/043c995c-d1f4-49e2-8c8f-3c9cc6c736ce

MiguVT:
Any update of this? look like Vencord is not accepting new features

PR #3026, opened 2024-11-20
feat(silentTyping): make button/icon visual look better

xNasuni:
before:
image
after:
image

:3


Vendicated:
this looks very out of place because it is wider than the other icons. that's why the old icon was smaller

optimally we'd just use a different proper square icon

image

xNasuni:
the reason I had used a bigger size (36 x 28) was because I thought it looked more odd if it WASN'T scaled up to fit the height of other buttons, it also made it bigger in width, which looking back wasn't really the best move

I sized it back down to 24x24 like all the other chatbar buttons and I think now it does look better.

image
image

Nuckyz:
Is it possible to achieve the same result without the use of transform and scale?

xNasuni:
> Is it possible to achieve the same result without the use of transform and scale?

I was trying to do it without using scale and transform because to me it seemed finnicky or bad practice, and I tried for a while and couldn't get the red bar to look the same, so I think it's possible but not worth the time unless someone that is knowledgeable can remake it without transformations, if it is strictly needed.

I used the scale attribute of the SVG element so that the red bar would be the same as the bell's red bar, for consistency, and the icon would stay normal.
image

Nuckyz:
Alright it's fine as it is, thanks!

PR #3028, opened 2024-11-21
Replace API add/remove funcs with methods in plugin definition

Vendicated:
Turns most APIs that have add & remove functions into singular `on*` & `render*` methods you can use in your plugin definition

Your handler will automatically be registered & unregistered for you and the dependency will be added if needed

## Example

Before
image

After:
image


Vendicated:
TODO: test every plugin I migrated to make sure I didn't break anything by accident

Vendicated:
This renamed a lot of the API functions to be more specific (`addDecoration` -> `addMessageDecoration`) so it will likely break many third party plugins

src/plugins/messageLinkEmbeds/index.tsx
```diff
@@ -372,7 +372,7 @@ export default definePlugin({
     settings,
 
     start() {
-        addAccessory("messageLinkEmbed", props => {
+        addMessageAccessory("messageLinkEmbed", props => {
```
Nuckyz, line 375: These shouldnt be using the start method

Nuckyz (changes requested):

src/api/MessageAccessories.tsx
```diff
@@ -16,43 +16,45 @@
...
+            <ErrorBoundary message={`Failed to render ${key} Message Accessory`} key={key}>
+                <accessory.render {...props} />
+            </ErrorBoundary>
+        );
 
-        if (!Array.isArray(accessories))
-            accessories = [accessories];
-        else if (accessories.length === 0)
+        const accessories = Array.isArray(res) ? res : [res];
+        if (!accessories.length)
```
Nuckyz, line 57:
```suggestion
        const accessories = Array.isArray(res) ? res : [res];
        if (!accessories.length)
```

This wont ever be an array because it's wrapped in an ErrorBoundary

src/plugins/clearURLs/index.ts
```diff
@@ -133,17 +136,4 @@ export default definePlugin({
             );
         }
     },
-
-    start() {
-        this.createRules();
```
Nuckyz, line 138: this,createRules(); is gone

Nuckyz (changes requested):

src/api/MessageAccessories.tsx
```diff
@@ -16,44 +16,42 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import ErrorBoundary from "@components/ErrorBoundary";
 import { JSX } from "react";
 
-export type AccessoryCallback = (props: Record<string, any>) => JSX.Element | null | Array<JSX.Element | null>;
-export type Accessory = {
-    callback: AccessoryCallback;
+export type MessageAccessoryFactory = (props: Record<string, any>) => JSX.Element | null | Array<JSX.Element | null>;
```
Nuckyz, line 22:
```suggestion
export type MessageAccessoryFactory = (props: Record<string, any>) => JSX.Element | null | Array<JSX.Element | null>;
```

This should probably just return React.ReactNode

Nuckyz (changes requested):

Nuckyz (approved):

PR #3037, opened 2024-11-22
BetterFolders: Fix dedicated sidebar

sadan4:


sadan4:
discord seems to have reverted

PR #3050, opened 2024-11-26
feat: add linkSummarizer plugin

macdontstop:
Adds a button to the message popover to summarize links in messages using Ollama.


Vendicated:

unfortunately, I will have to reject this for the following reasons:
1. this doesn't need to be a plugin, you can just make a user installable Discord app and use a context menu command on the message with the link
2. as this requires the user to selfhost ollama, it is not very user friendly and doesn't work out of the box

sorry :(

PR #3051, opened 2024-11-26
RoleColorEverywhere: add a min contrast for colors

sadan4:
Draft because the code is bloated and I'm working on other things rn. Feel free to close if not useful / code is too bad.

Adds an option to raise the contrast to a minimum level defined in settings. Useful for role colors that make text really hard to read.


sadan4:
this code is terrible and was only ever meant to be a POC

PR #1865, opened 2023-10-25
Fix PinDMs, SMYN, SilentTyping, ValidUser, PlatformIndicators

rniii:
Fixes: ShowMeYourName, SilentTyping, PinDMs, PlatformIndicators (+ apis), ValidUser

burger pr: https://codeberg.org/Ven/cord/pulls/35

PR #3053, opened 2024-11-26
feat(SilentTyping): Add per-channel/guild toggling + other improvements.

EtorixDev:
Part of the improvements to the command (number 3 below) requires #3047 to be merged as well.

### Changes
1. The tooltip for the chat bar icon now states the current status of the plugin.
    - Previously it would say "Disable Silent Typing" which could be confused with the feature itself, disabling typing from showing.
    - Now it states "Typing Hidden" which clearly states the current status without possible confusion for what clicking it would do.
2. The chat context menu settings were expanded and moved into a drop down group.
    - The chat context menu item now appears above the `Send Message Button` button to place it visually with the only other drop down setting, `Suggestions`.
3. The chat command `/silent-typing` was greatly expanded with more settings.
4. The chat icon will now show up even when a thread/post is opened halfway.
5. Expanded to toggle per DM, channel, guild, and globally instead of a single global toggle.
    - For the purpose of the plugin, threads are considered separate channels from their parent channel.
6. The chat icon can now toggle the current DM, channel, guild, and globally.
    - Left click to toggle the current DM or channel.
    - Middle click to toggle the current guild, if applicable.
    - Right click to toggle globally (disable/enable the plugin).
7. Blacklist / Whitelist
    - By default, every single DM, channel, and guild will hide your typing. This is while the `Default Hidden` setting is enabled. You can whitelist locations from hiding your typing by adding them to the `Disabled Locations` list.
    - If you disable `Default Hidden`, then every single DM, channel, and guild will show your typing. You can blacklist locations to hide your typing by adding them to the `Enabled Locations` list.
    - Both lists are independent and only active when `Default Hidden` is enabled or disabled respectively.
    - Left and middle clicking the chat icon will add the DM, channel, or guild ID to whichever list is currently active.
    - The lists can be modified manually in the settings with a comma separated list of IDs.


src/plugins/silentTyping/index.tsx
```diff
@@ -70,37 +124,71 @@ const SilentTypingToggle: ChatBarButton = ({ isMainChat }) => {
...
+        </Menu.MenuItem>
     );
 };
 
-
 export default definePlugin({
     name: "SilentTyping",
-    authors: [Devs.Ven, Devs.Rini, Devs.ImBanana],
-    description: "Hide that you are typing",
+    authors: [Devs.Ven, Devs.ImBanana, Devs.Etorix],
```
jamesbt365, line 183: Why did you remove someone from the authors?

EtorixDev: Nuckyz said in the core-development channel that you shouldn't add yourself as a developer for minor changes. I originally started modifying the [enhanced](https://github.com/D3SOX/vc-silentTypingEnhanced/tree/master) fork before switching to main and D3SOX only had Rini and not ImBanana listed so I decided to just look at everyone's commit history on the file and the only thing Rini did was a single line for the match in #1865 so I thought it didn't really qualify. I don't mind adding it back if needed though.

jamesbt365: I don't think you should be the one dictating the removal of other users from contributions, let a maintainer make that decision.

Vendicated: she wrote the majority of the plugin. the commit you linked just changes the listed author from her old username to her new one. 

https://github.com/rniii/rindetta/commit/9338b92b1a3b68308950a20f0d694901e4c39e2a

EtorixDev: Got it, my bad. Updated.

EtorixDev:
Updated this for present day. Would appreciate a review <:blobcatcozy:1026533070955872337>

PR #3054, opened 2024-11-27
ImageZoom: Fix zoom level not saving

sadan4:


src/plugins/imageZoom/components/Magnifier.tsx
```diff
@@ -54,6 +54,14 @@ export const Magnifier = ErrorBoundary.wrap<MagnifierProps>(({ instance, size: i
...
     const imageRef = useRef<HTMLImageElement | null>(null);
 
+    useEffect(() => {
+        return () => {
+            if (settings.store.saveZoomValues) {
+                settings.store.zoom = zoom.current;
+                settings.store.size = size.current;
+            }
+        };
+    });
```
Nuckyz, line 64: Shouldnt this effect have an empty dependency array?

Nuckyz (changes requested):

src/plugins/imageZoom/components/Magnifier.tsx
sadan4: isnt nothing the same as an empty dep array?

sadan4: just looked up and [its not the same](https://react.dev/reference/react/useEffect#examples-dependencies)

however, if a dep array is passed, it doesn't work

Nuckyz: that must be why the layout effect doesn't work either

try a layout effect with no deps

PR #3055, opened 2024-11-27
new plugin: stopAutoUnread

lumap:
image
stops discord from automatically setting a channel's unread indicator as "all messages" after a while


Vendicated:
is this an experiment?

lumap:
yeah, rolled out to some people (idk what's the exact %) so thought it might be useful to push here

Vendicated:
plugins should be easy to understand, widely useful and not about experiments

once this feature has finalised and fully rolled out, we can add plugins around it. until then, it's best to keep it as a third party plugin

PR #3056, opened 2024-11-29
AdsEverywhere

Inbestigator:
See ads everywhere, so Discord finally feels like a real website.

image


Sqaaakoi:
Is the code trying to be of the same quality as the idea of the plugin itself?

Inbestigator:
> Is the code trying to be of the same quality as the idea of the plugin itself?

The code **is** of the same quality, it's impossible to do this without dom manip afaik (if you can do it with patches, feel free to lmk)

src/plugins/adsEverywhere/index.ts
```diff
@@ -0,0 +1,103 @@
...
+                row.append(link);
+            }
+        }
+    });
+    setTimeout(addAds, 1000);
+}
+
+export default definePlugin({
+    name: "AdsEverywhere",
+    description: "See ads everywhere, so Discord feels like a real website.",
```
sadan4, line 100:
```suggestion
    description: "See ads everywhere, so Discord feels like a real website.",
    required: true,
```

think you forgot this

Inbestigator: Google moment

PR #3057, opened 2024-11-29
CommandsAPI: Fix spread overwriting omitted subcommand options

EtorixDev:
If you omit options from a subcommand as you are allowed to do, the options argument will be filled in with the parent's options creating a recursive loop ending in `RangeError: Maximum call stack size exceeded`. This resets the options to `undefined` if they were intentionally left out.


Nuckyz:
Thanks for this fix and the one in the other pull request!

PR #3058, opened 2024-11-29
New Plugin: PluralKitIntegration

JohnyTheCarrot:
Uses third party API for pluralkit, without authentication.


JohnyTheCarrot:
I forgot that a proxy tag can have both a prefix and a postfix. Will fix soon

Vendicated:
Thank you for your contribution. However, this is too niche (and has already been rejected before)

hqnna:
> Thank you for your contribution. However, this is too niche (and has already been rejected before)

Respectfully, this is a load of bullshit. PluralKit is in nearly 1,000,000 servers (roughly). The fact you deny it on a matter of "too niche" when it also plays part in accessibility is fucking stupid, but I guess anything not used by majorities is "too niche" for this repository isn't it, despite having plugins that do things as minor as "removes the gap between code blocks and text below it" despite that being able to be fixed with themes and doesn't even need to be a plugin.

lewisakura:
Just locking this now in advance because I know how this is going to go.

The codeblock fix plugin was allowed because it makes a change to the source code which is more performant than using the CSS equivalent. The other plugins you have cited were added to the mod before proper quality control restrictions were put in place. They are kept purely out of the fact that they're funny and for the fact that we don't perform retroactive QC, but would not be acceptable under our rules now (and a couple of plugins from that period have been removed).

I know PluralKit is popular (we had it in our server a while ago), but PluralKit themselves provide all of this functionality through their bot. Adding a plugin that only works in certain servers with this bot is not very useful overall in the mod because the majority of Discord servers that people tend to be in don't have PluralKit (I'm aware of the figures you've cited, but out of the ~150 communities I'm in, only 1 has PK/equiv. in active use, so I don't think going off of the bot figures is as practical as it seems). We also don't have much experience with the PK API and we have to maintain the plugins we add to the mod, so it'll increase our future maintenance cost when it breaks.

Genuinely, I understand how much this plugin means to you, and if you wish to use it you may use it via a DevBuild, but it will not be added to the main mod for the reasons cited by myself and V.

Vendicated:
Prefacing your message with "Respectfully" doesn't make it any less disrespectful and outright hostile.

This hostility isn't welcome here. This is your first and last warning to play it nice or we will have to restrict your access to contribute.

Not to mention that this is our project, so it is completely up to us what to accept and what to reject. You are welcome to create your own fork if you have different ideas

PR #3061, opened 2024-11-30
docs: Add a bunch of READMEs for plugins missing them, and a few (possibly ok) consistency related fixes

korbosoft:
Heavily based off plugin description. It may not be the best, but it's at least a good start.


korbosoft:
the new READMEs are now at a good state.

PR #3064, opened 2024-12-04
BetterFolders: Fix including open folders in main sidebar

sadan4:


sadan4:
only on canary as of now

PR #3065, opened 2024-12-04
Add saveCheckboxSettings plugin

holybananapants:
### **Description**

This plugin allows Discord users to save the states of specific checkboxes while watching screen shares. It ensures that selected preferences (e.g., "Show participants without video") persist across sessions, remaining active or inactive as set by the user, even after restarting Discord.

image

### DIFF
--- a/src/plugins/saveCheckboxSettings/index.ts
+++ b/src/plugins/saveCheckboxSettings/index.ts
@@ -1,0 +1,66 @@
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default {
+    name: "SaveCheckboxSettings",
+    description: "Saves the states of checkboxes in Discord settings and restores them on restart.",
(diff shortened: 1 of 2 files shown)


src/plugins/saveCheckboxSettings/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            messagePreviewCheckbox.addEventListener("change", () => {
+                savedSettings.showMessagePreview = messagePreviewCheckbox.checked;
+                localStorage.setItem("SaveCheckboxSettings", JSON.stringify(savedSettings));
+            });
+
+            console.log("[SaveCheckboxSettings] All checkboxes initialized!");
+
+            // Stop observing once checkboxes are found and initialized
+            observer.disconnect();
+        });
```
sadan4, line 57: don't use dom manipulation, use patches

if you need help with that, use the plugin development in the discord

PR #3069, opened 2024-12-10
Add DisableAnimations plugin

wont-stream:
Applies the property `skipAnimation` to all react-sprint instances Discord has. 

Also adds basic css `* { transition: none !important; animation: none !important; }`


Vendicated:
just enable Reduced Animations in Accessibility settings

wont-stream:
I’d argue reduced motion doesn’t get you all the way there but alright.

PR #3080, opened 2024-12-14
feat: Add Vencord styles to popout windows

Masterjoona:
Previously vencord styles and quick css was only getting added to the main window and not popout windows, this should fix it

example
before

after

### DIFF
--- a/src/utils/quickCss.ts
+++ b/src/utils/quickCss.ts
@@ -17,88 +17,148 @@
 */
 
 import { Settings, SettingsStore } from "@api/Settings";
-import { ThemeStore } from "@webpack/common";
+import { findStoreLazy } from "@webpack";
+import { FluxDispatcher, ThemeStore } from "@webpack/common";
 
+const PopoutWindowStore = findStoreLazy("PopoutWindowStore");
 
-let style: HTMLStyleElement;
-let themesStyle: HTMLStyleElement;
+class StyleManager {
+    style: HTMLStyleElement | null = null;
+    themesStyle: HTMLStyleElement | null = null;
+    systemValuesStyle: HTMLStyleElement | null = null;
+    doc: Document;
 
-function createStyle(id: string) {
-    const style = document.createElement("style");
-    style.id = id;
-    document.documentElement.append(style);
-    return style;
-}
-
-async function initSystemValues() {
-    const values = await VencordNative.themes.getSystemValues();
-    const variables = Object.entries(values)
-        .filter(([, v]) => v !== "#")
-        .map(([k, v]) => `--${k}: ${v};`)
-        .join("");
-
-    createStyle("vencord-os-theme-values").textContent = `:root{${variables}}`;
-}
-
-export async function toggle(isEnabled: boolean) {
-    if (!style) {
-        if (isEnabled) {
-            style = createStyle("vencord-custom-css");
-            VencordNative.quickCss.addChangeListener(css => {
-                style.textContent = css;
-                // At the time of writing this, changing textContent resets the disabled state
-                style.disabled = !Settings.useQuickCss;
-            });
-            style.textContent = await VencordNative.quickCss.get();
-        }
-    } else
-        style.disabled = !isEnabled;
-}
-
-async function initThemes() {
-    themesStyle ??= createStyle("vencord-themes");
-
-    const { themeLinks, enabledThemes } = Settings;
-
-    // "darker" and "midnight" both count as dark
-    const activeTheme = ThemeStore.theme === "light" ? "light" : "dark";
-
-    const links = themeLinks
-        .map(rawLink => {
-            const match = /^@(light|dark) (.*)/.exec(rawLink);
-            if (!match) return rawLink;
-
-            const [, mode, link] = match;
-            return mode === activeTheme ? link : null;
-        })
-        .filter(link => link !== null);
-
-    if (IS_WEB) {
-        for (const theme of enabledThemes) {
-            const themeData = await VencordNative.themes.getThemeData(theme);
-            if (!themeData) continue;
-            const blob = new Blob([themeData], { type: "text/css" });
-            links.push(URL.createObjectURL(blob));
-        }
-    } else {
-        const localThemes = enabledThemes.map(theme => `vencord:///themes/${theme}?v=${Date.now()}`);
-        links.push(...localThemes);
+    constructor(doc: Document) {
+        this.doc = doc;
     }
 
-    themesStyle.textContent = links.map(link => `@import url("${link.trim()}");`).join("\n");
+    createStyle(id: string) {
+        const style = this.doc.createElement("style");
+        style.id = id;
+        this.doc.documentElement.appendChild(style);
+        return style;
+    }
+
+    async initSystemValues() {
+        const values = await VencordNative.themes.getSystemValues();
+        const variables = Object.entries(values)
+            .filter(([, v]) => v !== "#")
+            .map(([k, v]) => `--${k}: ${v};`)
+            .join("");
+
+        this.createStyle("vencord-os-theme-values").textContent = `:root{${variables}}`;
+    }
+
+    async toggleCustomCSS(isEnabled: boolean) {
+        if (!this.style) {
+            if (isEnabled) {
+                this.style = this.createStyle("vencord-custom-css");
+                VencordNative.quickCss.addChangeListener(css => {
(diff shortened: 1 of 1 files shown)


src/utils/quickCss.ts
```diff
@@ -17,88 +17,148 @@
...
-    ThemeStore.addChangeListener(initThemes);
+            if (elapsed >= maxWait) {
+                clearInterval(interval);
+            }
 
-    if (!IS_WEB)
-        VencordNative.quickCss.addThemeChangeListener(initThemes);
+            elapsed += 500;
+        }, 500);
+    });
```
Sqaaakoi, line 163: Why is there a loop waiting for the document to be ready? Is it not possible to use the document readystatechange event?

src/utils/quickCss.ts
```diff
@@ -17,88 +17,148 @@
...
+
+        SettingsStore.removeChangeListener("useQuickCss", this.toggleCustomCSS);
+        SettingsStore.removeChangeListener("themeLinks", this.initThemes);
+        SettingsStore.removeChangeListener("enabledThemes", this.initThemes);
+        ThemeStore.removeChangeListener(this.initThemes);
+    }
 }
 
+const mainWindowManager = new StyleManager(document);
+export const toggle = mainWindowManager.toggleCustomCSS.bind(mainWindowManager);
```
Sqaaakoi, line 127: shouldn't toggling be a static method of StyleManager that does it for all windows?

Sqaaakoi (commented):
I'm glad this finally exists. It's a long overdue feature.

For anyone else trying to make sense of this: each StyleManager operates autonomously and does not get centrally managed.

src/utils/quickCss.ts
Sqaaakoi: update: this isn't called anywhere just get rid of it

src/utils/quickCss.ts
Masterjoona: nop, if it the styles get instantly added when the event is dispatched they get overwritten or smth

Sqaaakoi: > nop, if the styles get instantly added when the event is dispatched they get overwritten by the actual popout html (?) or smth

cursed

is it not possible to inject into the head?
is using a MutationObserver a possibility?

Masterjoona: neither worked from my tests

Vendicated:
don't use a class that's terrible

Masterjoona:
ill try to come up with somthing else then

Masterjoona:
i came up with something better

src/utils/quickCss.ts
Sqaaakoi: After (???) amount of usage, I've noticed this doesn't work. I been researching how PopoutWindowStore works to see how this could be fixed.

Replace the subscription to "POPOUT_WINDOW_OPEN" with the "message" event on the `window`.
Replace finding the most recent windowKey with  `event.data?.discordPopoutEvent?.key`

This is how PopoutWindowStore adds the styles (]through a chain of functions)

Extra note: There is a ternary to switch between recreating all `link[rel="stylesheet"]` elements and importing all `link[rel="stylesheet"]` and `style` elements. This is always set to true (though a variable that is seemingly never changed), so the `style` elements are not imported. If this was set to false, it would copy all style elements, including those added by Vencord.

Masterjoona: ah, thank you!, should've spotted that myself

Sqaaakoi: > ah, thank you!, should've spotted that myself

hold on; since I posted that I have realised that another check needs to be implemented
you should also check `event.data?.discordPopoutEvent?.type === "loaded"` or else an error will occur when closing windows

Vendicated:
could you update this pr please? and it would be good to add all styles, even ones added by plugins and managed styles

Vendicated:
This should be considerably easier to do now that all vencord styles are in one central DOM node

Vendicated:
image

Masterjoona:
Yeah that's much better!

Though I'm just not sure how to handle plugins (blurnsfw or clienttheme) that might need to update their styles

imjustprism:
> Yeah that's much better!
> 
> Though I'm just not sure how to handle plugins (blurnsfw or clienttheme) that might need to update their styles

watch the vencord root element for any changes. when a plugin adds, removes, or modifies any style, the observer detects it and automatically copies all styles to every open popout window. plugins like blurnsfw or clienttheme dont need any special handling cuz when they update their styles, the observer picks it up and syncs to popouts automatically

src/api/Themes.ts
```diff
@@ -89,6 +112,14 @@ document.addEventListener("DOMContentLoaded", () => {
     SettingsStore.addChangeListener("themeLinks", initThemes);
     SettingsStore.addChangeListener("enabledThemes", initThemes);
 
+    window.addEventListener("message", event => {
```
imjustprism, line 113: i think its better to use PopoutWindowStore.addChangeListener() + isWindowFullyInitialized(key) instead

src/api/Themes.ts
```diff
@@ -76,6 +80,25 @@ async function initThemes() {
...
+    if (!popoutWindow?.document) return;
+
+    const doc = popoutWindow.document;
+
+    doc.getElementsByTagName("vencord-root")[0]?.remove();
+
+    doc.documentElement.appendChild(vencordRootNode.cloneNode(true));
+}
+
+function updatePopoutWindows() {
```
imjustprism, line 94: use a set to track which popouts already have styles. skip if already applied, remove from set when window close

src/api/Themes.ts
```diff
@@ -89,6 +112,14 @@ document.addEventListener("DOMContentLoaded", () => {
     SettingsStore.addChangeListener("themeLinks", initThemes);
     SettingsStore.addChangeListener("enabledThemes", initThemes);
 
+    window.addEventListener("message", event => {
+        const { discordPopoutEvent } = event.data || {};
+        if (discordPopoutEvent?.type !== "loaded") return;
+
+        const popoutWindow = PopoutWindowStore.getWindow(discordPopoutEvent.key);
```
imjustprism, line 117: minor guard, but ig better to check window exists before calling applytopopout

src/api/Themes.ts
```diff
@@ -76,6 +80,25 @@ async function initThemes() {
...
     themesStyle.textContent = links.map(link => `@import url("${link.trim()}");`).join("\n");
+    updatePopoutWindows();
+}
+
+function applyToPopout(popoutWindow: Window) {
+    if (!popoutWindow?.document) return;
+
+    const doc = popoutWindow.document;
+
+    doc.getElementsByTagName("vencord-root")[0]?.remove();
```
imjustprism: nitpick but querySelectorAll is a bit safer i think

src/api/Themes.ts
```diff
@@ -19,8 +19,11 @@
 import { Settings, SettingsStore } from "@api/Settings";
 import { createAndAppendStyle } from "@utils/css";
 import { ThemeStore } from "@vencord/discord-types";
+import { findStoreLazy } from "@webpack";
 
-import { userStyleRootNode } from "./Styles";
+import { userStyleRootNode, vencordRootNode } from "./Styles";
+
+const PopoutWindowStore = findStoreLazy("PopoutWindowStore");
```
imjustprism: i typed this, can be helpful or you can add if you wish: https://github.com/imjustprism/Equicord/blob/c025463a5e214c443bd125c79d399285f8482778/packages/discord-types/src/stores/PopoutWindowStore.d.ts

Vendicated:
Idk tbh. It's probably fine to require a re-open of the popout? Users won't change the settings of those plugins too often

src/api/Themes.ts
Vendicated:
```suggestion
    doc.querySelector("vencord-root")?.remove();
```

src/api/Themes.ts
Masterjoona: I feel like this and the set would add unneeded complexity

src/api/Themes.ts
Masterjoona: thanks!

Masterjoona:
> It's probably fine to require a re-open of the popout? 

Fine by me

Vendicated:
epic lint fail

Vendicated:
thanks!

PR #3083, opened 2024-12-15
feat(ViewRaw): add View Role option

vishnyanetchereshnya:


Vendicated:
added it here too 
image

Vendicated:
thank u good idea!

PR #3084, opened 2024-12-17
quickSnip!

Err0r430:
A quick preset message snippet tool for sending those repetitive messages. Great for mods and admins of large servers who have to answer the same questions over and over again with the exact same answer. It has a base level of support for when in dms replacing ${myName} with your username and ${userName} with their name (I'll definitely be adding more of these inline replacement things in the future).

### DIFF
--- a/src/plugins/quickSnip/index.tsx
+++ b/src/plugins/quickSnip/index.tsx
@@ -1,0 +1,408 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addChatBarButton, ChatBarButton, removeChatBarButton } from "@api/ChatButtons";
+import { addPreSendListener, removePreSendListener } from "@api/MessageEvents";
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import { getCurrentChannel, insertTextIntoChatInputBox, sendMessage } from "@utils/discord";
+import { closeModal, ModalCloseButton, ModalContent, ModalFooter, ModalHeader, ModalProps, ModalRoot, openModal } from "@utils/modal";
+import definePlugin, { OptionType } from "@utils/types";
+import { Button, Forms, useEffect, useRef, useState } from "@webpack/common";
+
+import { getMyUsername, getSnippets, saveSnippets, Snippet } from "./localStorageUtils";
+
+const settings = definePluginSettings({
+    autoSendSnippet: {
+        description: "Instantly send a snippet when you click its name.",
+        type: OptionType.BOOLEAN,
+        default: false,
+    },
+});
+
+interface SnippetSettingsModalProps {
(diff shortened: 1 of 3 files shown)


src/plugins/quickSnip/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+    name: "QuickSnip",
+    description: "Send quick customizable preset snippets in DMs and channels.",
+    authors: [Devs.error],
+    dependencies: ["ChatInputButtonAPI"],
+    settings,
+    start() {
+        addChatBarButton("QuickSnip", ChatBarIcon);
+        this.listener = addPreSendListener((context, msg) => {
+            console.log("Message content:", msg.content);
+        });
```
jamesbt365, line 402: Seems you left some unnecessary logging in?

Err0r430: Thanks for catching it, my bad for not removing it.

src/plugins/quickSnip/index.tsx
```diff
@@ -0,0 +1,405 @@
...
+    description: "Send quick customizable preset snippets in DMs and channels.",
+    authors: [Devs.error],
+    dependencies: ["ChatInputButtonAPI"],
+    settings,
+    start() {
+        addChatBarButton("QuickSnip", ChatBarIcon);
+    },
+    stop() {
+        removeChatBarButton("QuickSnip");
+        removePreSendListener(this.listener);
```
jamesbt365, line 403: You no longer have this listener

Err0r430: Updated that, I'm confused why my eslint has decided to not work in the slightest.

sadan4 (commented):
needs a readme.md

sadan4:
i feel like this would be better as an addition to message tags instead of its own plugin

Err0r430:
Added the README as requested. I'm unsure of the format of these (I took a look at a few but wasn't consistent) so I just wrote one that highlighted its features. 

> i feel like this would be better as an addition to message tags instead of its own plugin

I considered this, but I felt as though messageTags is geared towards a more keyboard-based usage experience, whereas QuickSnip is a simple button interface aimed at server admins/mods who repeatedly send the same basic messages to users.

sadan4:
> I considered this, but I felt as though messageTags is geared towards a more keyboard-based usage experience, whereas QuickSnip is a simple button interface aimed at server admins/mods who repeatedly send the same basic messages to users.
i see no reason not to combine the two, sometimes I don't want to type a command / would rather type a command. plus it would be confusing having snippets in two separate places

Err0r430:
So shall I merge this plugin with messageTags? I still feel as though this could be a standalone plugin but am happy to merge if you think so.

Inbestigator:
Duplicate functionality plugins can introduce confusion for users

Vendicated:
two plugins covering the same functionality is not allowed because it's very confusing

Err0r430:
> two plugins covering the same functionality is not allowed because it's very confusing 

Working on merging with message tags tonight.

PR #3086, opened 2024-12-17
Add plugin ForceOwner

RyfterWasTaken:
Makes the client think the user is the owner of all the servers, allows access to see most of the settings of a server, as well as to see all members (through role settings).


lewisakura:
This has a massive likelihood to cause the client to trigger a bunch of requests that result in 403s for no real benefit to the end user besides causing antispam flags. Sorry, but no.

PR #3087, opened 2024-12-18
EmoteCloner: Fix a hang when the sticker/original server is deleted

nakoyasha:
This fixes the modal hanging and infinitely loading in cases where the sticker can't be fetched (e.g when it or the server it's from was deleted, general Discord weirdness, etc). Also allows for the sticker to still be cloned.

### DIFF
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -317,6 +317,20 @@ function isGifUrl(url: string) {
     return u.pathname.endsWith(".gif") || u.searchParams.get("animated") === "true";
 }
 
+function buildFakeSticker(sticker: any) {
+    return {
+        id: sticker?.id,
+        name: sticker?.name,
+        format_type: sticker?.format_type,
+        // Discord has a character limit of at least 1 for tags (aka related emoji)
+        tags: " ",
+        description: "",
+        type: "2",
+        available: true,
+        guild_id: 0
+    };
+}
+
 const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) => {
     const { favoriteableId, itemHref, itemSrc, favoriteableType } = props ?? {};
 
@@ -343,27 +357,7 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
                 // (e.g when using MessageLinkEmkbeds)
                 if (sticker === undefined) return;
 
-                return buildMenuItem("Sticker", async () => {
-                    const fetchedSticker = await fetchSticker(favoriteableId);
-
-                    // Workaround for incase the sticker or the server it's from is deleted.
-                    // Allows the sticker to still be cloned.
-                    if (fetchedSticker === undefined) {
-                        return {
-                            id: sticker?.id,
-                            name: sticker?.name,
-                            format_type: sticker?.format_type,
-                            // Discord has a character limit of at least 1 for tags (aka related emoji)
-                            tags: " ",
-                            description: "",
-                            type: "2",
-                            available: true,
-                            guild_id: 0
-                        };
-                    }
-
-                    return fetchedSticker;
-                });
+                return buildMenuItem("Sticker", () => fetchSticker(favoriteableId).then(s => s ?? buildFakeSticker(sticker)));
         }
     })();
 
--- a/src/plugins/emoteCloner/index.tsx
+++ b/src/plugins/emoteCloner/index.tsx
@@ -342,7 +342,7 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
 
                 // Workaround for cases when it's not available
                 // (e.g when using MessageLinkEmkbeds)
-                if (sticker == undefined) {
+                if (sticker === undefined) {
                     return;
                 }
 
@@ -351,7 +351,7 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
 
                     // Workaround for incase the sticker or the server it's from is deleted.
                     // Allows the sticker to still be cloned, albeit less accurately.
-                    if (fetchedSticker == undefined) {
+                    if (fetchedSticker === undefined) {
                         return {
                             "id": sticker?.id,
                             "name": sticker?.name,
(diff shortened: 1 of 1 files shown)


src/plugins/emoteCloner/index.tsx
```diff
@@ -63,16 +63,20 @@ async function fetchSticker(id: string) {
...
-    });
+        FluxDispatcher.dispatch({
+            type: "STICKER_FETCH_SUCCESS",
+            sticker: body
+        });
 
-    return body as Sticker;
+        return body as Sticker;
+    } catch (err) {
+        return undefined;
```
splatterxl, line 78:
```suggestion
    } catch {
        // do nothing
```

src/plugins/emoteCloner/index.tsx
```diff
@@ -336,7 +340,33 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
                 const sticker = props.message.stickerItems.find(s => s.id === favoriteableId);
                 if (sticker?.format_type === 3 /* LOTTIE */) return;
 
-                return buildMenuItem("Sticker", () => fetchSticker(favoriteableId));
+                // Workaround for cases when it's not available
+                // (e.g when using MessageLinkEmkbeds)
+                if (sticker === undefined) {
+                    return;
+                }
```
splatterxl, line 347:
```suggestion
                if (!sticker) return;
```

src/plugins/emoteCloner/index.tsx
```diff
@@ -336,7 +340,33 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
...
+                            // Discord has a character limit of at least 1 for tags (aka related emoji)
+                            "tags": " ",
+                            "description": "",
+                            "type": "2",
+                            "available": true,
+                            "guild_id": 0
+                        };
+                    } else {
+                        return fetchedSticker;
+                    }
```
splatterxl, line 368:
```suggestion
                    }
                    
                    return fetchedSticker;
```

src/plugins/emoteCloner/index.tsx
```diff
@@ -336,7 +340,33 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
...
+                        return {
+                            "id": sticker?.id,
+                            "name": sticker?.name,
+                            "format_type": sticker?.format_type,
+                            // Discord has a character limit of at least 1 for tags (aka related emoji)
+                            "tags": " ",
+                            "description": "",
+                            "type": "2",
+                            "available": true,
+                            "guild_id": 0
```
splatterxl, line 364: format to have normal key names not in quotes

splatterxl (commented):
cleaner code suggestions but lgtm otherwise

src/plugins/emoteCloner/index.tsx
nakoyasha: woops, that's a side-effect of me using a linter that complains about non-explicit returns xd

src/plugins/emoteCloner/index.tsx
```diff
@@ -63,16 +63,19 @@ async function fetchSticker(id: string) {
...
-    });
+        FluxDispatcher.dispatch({
+            type: "STICKER_FETCH_SUCCESS",
+            sticker: body
+        });
 
-    return body as Sticker;
+        return body as Sticker;
+    } catch (err) {
+    }
```
Vendicated, line 78:
```suggestion
		return null;
    }
```

src/plugins/emoteCloner/index.tsx
```diff
@@ -336,7 +339,31 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
                 const sticker = props.message.stickerItems.find(s => s.id === favoriteableId);
                 if (sticker?.format_type === 3 /* LOTTIE */) return;
 
-                return buildMenuItem("Sticker", () => fetchSticker(favoriteableId));
+                // Workaround for cases when it's not available
+                // (e.g when using MessageLinkEmkbeds)
+                if (sticker === undefined) return;
```
Vendicated, line 344: why is this needed? we already check favoriteableId above, and that's all we need

src/plugins/emoteCloner/index.tsx
```diff
@@ -336,7 +339,31 @@ const messageContextMenuPatch: NavContextMenuPatchCallback = (children, props) =
...
+
+                return buildMenuItem("Sticker", async () => {
+                    const fetchedSticker = await fetchSticker(favoriteableId);
```
Vendicated, line 347: extract the fake sticker into a separate function and call it inline for nicer code

```suggestion
                return buildMenuItem("Sticker", () => fetchSticker(favoriteableId).then(s => s ?? buildFakeSticker(sticker));
```

src/plugins/emoteCloner/index.tsx
nakoyasha: it's needed because of the incompatibility with the plugin mentioned in the comment, it results in this without the check:
Discord_SyxhRNa8sJ
it's a niche case, but i wanna be safe in case the issue exists elsewhere too

src/plugins/emoteCloner/index.tsx
```diff
@@ -314,6 +318,20 @@ function isGifUrl(url: string) {
     return u.pathname.endsWith(".gif") || u.searchParams.get("animated") === "true";
 }
 
+function buildFakeSticker(sticker: any) {
```
splatterxl, line 321: could this be some other type than any? like `Partial<Sticker>` or smth?

src/plugins/emoteCloner/index.tsx
```diff
@@ -314,6 +318,20 @@ function isGifUrl(url: string) {
...
+    return {
+        id: sticker?.id,
+        name: sticker?.name,
+        format_type: sticker?.format_type,
+        // Discord has a character limit of at least 1 for tags (aka related emoji)
+        tags: " ",
+        description: "",
+        type: "2",
```
splatterxl, line 329: is there an enum for this, for clarity?

splatterxl (commented):
nitpicks

src/plugins/emoteCloner/index.tsx
nakoyasha: [there is](https://discord.com/developers/docs/resources/sticker#sticker-object-sticker-types), but im pretty sure emotecloner makes no use of type 1 stickers thus no reason to add it

src/plugins/emoteCloner/index.tsx
nakoyasha: done

PR #3093, opened 2024-12-19
Update the prodversion in the extension download URL

Julienraptor01:
The version 32 is too old and cause the request to fail
The 32bit int limit seems to be the max value and should make the request not fail for a few years hopefully
*This change doesn't really matter for Vencord as its only use is to load react dev tools which are not taken from the chrome web store anyway, but it will fix the feature if a userplugin try to load chrome extensions*


Vendicated:
superseded by https://github.com/Vendicated/Vencord/commit/9ccc74bde37ceb30017d434907b12f0e0f19c9d1

PR #3094, opened 2024-12-19
feat: Disable faulty plugins to prevent unrecoverable loop

million1156:
An issue a lot of people have when Discord updates is their plugins break. Sometimes, the update is so significant that Vencord itself is broken (i.e. cannot manually disable plugins). I experienced this issue myself, and so I created a pretty decent solution.

When the user enters an "unrecoverable loop" (this is triggered when Vencord crashes 2x in <1s), we disable all plugins **EXCEPT** for plugins that are API related (i.e. `CommandsAPI`, plugins that are marked as `required` (such as `NoTrack`), or plugins that are already disabled (why would you disable an already disabled plugin?) & create a notification that lets the user copy the plugins that were disabled (so they can go ahead and re-enable them when Vencord is fixed). 

This PR is currently a draft as I could probably make the copying part a bit better, but let me know! :)

### DIFF
--- a/src/Vencord.ts
+++ b/src/Vencord.ts
@@ -85,6 +85,7 @@ async function syncSettings() {
     }
 }
 
+
 async function init() {
     await onceReady;
     startAllPlugins(StartAt.WebpackReady);
--- a/src/Vencord.ts
+++ b/src/Vencord.ts
@@ -85,7 +85,6 @@ async function syncSettings() {
     }
 }
 
-
 async function init() {
     await onceReady;
     startAllPlugins(StartAt.WebpackReady);
--- a/src/plugins/crashHandler/index.ts
+++ b/src/plugins/crashHandler/index.ts
@@ -89,11 +89,11 @@ export default definePlugin({
                 if (!shouldAttemptRecover) {
                     // Disable all non-core plugins (last-ditch effort to prevent a crash loop)
                     try {
-                        CrashHandlerLogger.log("Disabling all non-core plugins...");
+                        CrashHandlerLogger.debug("Disabling all non-core plugins...");
                         const pluginSettings = Settings.plugins;
-                        const disabledPlugins: string[] = [];
+                        const disabledPlugins: string[] = []; // horror
                         for (const pluginName in pluginSettings) {
-                            if (pluginSettings[pluginName].enabled && !pluginSettings[pluginName].required && !pluginSettings[pluginName].hidden) {
+                            if (pluginSettings[pluginName].enabled && !pluginSettings[pluginName].required && !pluginSettings[pluginName].hidden && !pluginName.endsWith("API")) {
                                 pluginSettings[pluginName].enabled = false;
                                 disabledPlugins.push(pluginName);
                                 CrashHandlerLogger.log(`Disabled plugin: ${pluginName}`);
--- a/src/plugins/crashHandler/index.ts
+++ b/src/plugins/crashHandler/index.ts
@@ -24,7 +24,7 @@ import { closeAllModals } from "@utils/modal";
 import definePlugin, { OptionType } from "@utils/types";
 import { maybePromptToUpdate } from "@utils/updater";
 import { filters, findBulk, proxyLazyWebpack } from "@webpack";
-import { DraftType, ExpressionPickerStore, FluxDispatcher, NavigationRouter, SelectedChannelStore } from "@webpack/common";
+import { Clipboard, DraftType, ExpressionPickerStore, FluxDispatcher, NavigationRouter, SelectedChannelStore } from "@webpack/common";
 
 const CrashHandlerLogger = new Logger("CrashHandler");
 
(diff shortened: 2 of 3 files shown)


million1156:
I change my mind, I think this is the best way :v

million1156:
> in its current state this is kinda awful because it will completely overwrite your settings and force you to reenable everything

it doesn't override settings, it just disables plugins. they can easily be re-enabled and that's why there's a simple button to copy any disabled plugins. and again, this is really only triggered in the worst-case scenario (i.e. when a user cannot disable faulty plugins & they need to access discord). 

for example, in the latest episode of "vencord broke plz help", (almost) all plugins were broken due to a few changes Discord made. people could not open servers (if they had `MoreUserTags`/`ShowHiddenChannels`), i myself was unable to even open Discord settings (although this was a skill issue). the fix i made would theoretically fix this issue entirely, and allow the user to view any support channels/anything they needed.

Vendicated:
sorry but you cannot "easily re-enable" 100 plugins 😭

if you actually want to make this better, you should not do any disabling and instead add a safe mode functionality that skips starting non essential plugins and such without needing to modify any settings

this safe mode could then be used to disable plugins / update / whatever

Vendicated:
doing so would be fairly trivial. you just need to add a new safeMode setting and edit this function:

https://github.com/Vendicated/Vencord/blob/48a9aef2ebafbc72a0245a3725bf3c84bc9d76c6/src/plugins/index.ts#L45-L51

new code would look something like this
```ts
 export function isPluginEnabled(p: string) { 
     return ( 
         Plugins[p]?.required || 
         Plugins[p]?.isDependency || 
         (settings[p]?.enabled && !settings.safeMode) 
     ) ?? false; 
 } 
```

million1156:
> sorry but you cannot "easily re-enable" 100 plugins 😭
> 
> if you actually want to make this better, you should not do any disabling and instead add a safe mode functionality that skips starting non essential plugins and such without needing to modify any settings
> 
> this safe mode could then be used to disable plugins / update / whatever

fair point, i'm coming from the perspective of only having like 4-5 plugins but if you're a power-user then yeah you're right. i'll change this back to a draft and work on it a bit more!

million1156:
> doing so would be fairly trivial. you just need to add a new safeMode setting and edit this function:
> 
> https://github.com/Vendicated/Vencord/blob/48a9aef2ebafbc72a0245a3725bf3c84bc9d76c6/src/plugins/index.ts#L45-L51
> 
> new code would look something like this
> 
> ```ts
>  export function isPluginEnabled(p: string) { 
>      return ( 
>          Plugins[p]?.required || 
>          Plugins[p]?.isDependency || 
>          (settings[p]?.enabled && !settings.safeMode) 
>      ) ?? false; 
>  } 
> ```

i'll take a look, ty!

Vendicated:
what I am thinking of:

when an unrecoverable crash happens, it should show a popup notifying the user of such and offer them the option to restart in safe mode. now vencord will set the safeMode setting to true and restart the app.

once all plugins finished initialising, it should automatically change the setting back to false so the next time the user opens the app it will be normal mode again

a command line toggle `--vencord-safe-mode` (or similar) could also be added

million1156:
> what I am thinking of:
> 
> when an unrecoverable crash happens, it should show a popup notifying the user of such and offer them the option to restart in safe mode. now vencord will set the safeMode setting to true and restart the app.
> 
> once Vencord finished initialising, it should automatically change the setting back to false so the next time the user opens the app it will be normal mode again
> 
> a command line toggle `--vencord-safe-mode` (or similar) could also be added

I just saw this! I love your idea, I'll try it that way instead :)

million1156:
@Vendicated 👋 As requested, I went ahead and did what you suggested instead, which is as follows:

- If the client encounters an unrecoverable crash, a popup will occur similar to this: Screenshot 2024-12-20 at 6 31 44 PM

- If the user clicks Ok, safeMode is enabled on all plugins besides:
 - Those that end in `API`
 - Those that are marked as `required` (such as `NoTrack`)
 - Those that are marked as `hidden` (limited to a subsection)

- When the client re-opens and tries to enable patches or plugins, any that are marked with safeMode are not loaded, then set to false (so it doesn't repeatedly disable patches)


I think this is much better than what I originally had in mind, let me know what you think!

Vendicated:
why are you overengineering this so much? you just need to edit the isPluginEnabled function, that's it

million1156:
> why are you overengineering this so much? you just need to edit the isPluginEnabled function, that's it

in my testing this did not stop it from being loaded, hence why i tapped into addPatch directly

Vendicated:
it is all you have to do

million1156:
> it is all you have to do

i took another look, turns out i was wrong! i fixed the isPluginEnabled function & i only run this after the isPluginEnabled check is ran (rather than modifying addPatch/startPlugin):
```js
    // Disable safe-mode after check is ran
    if (p.safeMode) p.safeMode = false;
```

Vendicated:
I pushed a fix to this PR since you didn't seem to have understood what I was trying to say and were still vastly overengineering this

ghost:
there is this incel, and recently he trooned out and started posting on his community about him pissgasming all over his lap. i think he is an autogynephile and his clitty shatters every second

PR #3096, opened 2024-12-20
Translate: fix links sometimes breaking

nay-cat:
I use the translate plugin very often, but there is something that makes me very nervous.

When I send memes to my friends, THIS HAPPENS

image

This also happens!

image

I have added a small piece of code that takes care of checking the links in the message, avoiding them when translating the message, I have built Vencord with the modified plugin and it seems to work (I have tested it in Italian, French and Spanish).

image
image


Inbestigator:
peak commit message

nay-cat:
eeehhh my bad

src/plugins/translate/index.tsx
```diff
@@ -88,13 +88,28 @@ export default definePlugin({
             if (!settings.store.autoTranslate) return;
             if (!message.content) return;
 
+            const urlRegex = /(https?:\/\/[^\s]+)/g;
```
Vendicated, line 91:
```suggestion
            const urlRegex = /(https?:\/\/\S+)/g;
```

Vendicated:
image

Vendicated:
using a string like `[[LINK]]` is bad not only because it might be translated but also because some languages might reorder stuff (maybe RTL) and what if the user included that text in their message? xD 

i would instead generate an id (maybe with discords snowflake generator or uuid or string hash) for each link, then replace that again

nay-cat:
You're right about that (I'm an idiot)

Excuse me!! 

I have uploaded a version that I have tested quickly from Russian to English and English to Russian, it should not give any problems now, I have used xxhash64 as I have seen it already used in the project.

Sorry for the inconvenience and I hope that now it works for all cases (probably I have forgotten something).
image

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,37 @@ export default definePlugin({
...
             tooltipTimeout = setTimeout(() => setShouldShowTranslateEnabledTooltip?.(false), 2000);
-
-            const trans = await translate("sent", message.content);
-            message.content = trans.text;
-
+        
+            const tempContent = message.content.replace(urlRegex, (url) => {
+                const hash = h64(url);  // xxhash64 from @intrnl/xxhash64
+                // hash to string
+                const hashString = hash.toString(16);
```
Vendicated, line 107:
```suggestion
                const hash = h64(url).toString(16); 
```

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,37 @@ export default definePlugin({
...
+                return hashString;  
+            });
+
+            const trans = await translate("sent", tempContent);
+        
+            let translatedContent = trans.text;
+
+            // hash to original
+            urlMap.forEach((url, hash) => {
+                translatedContent = translatedContent.replace(hash, url);
```
Vendicated, line 118:
```suggestion
                translatedContent = translatedContent.replaceAll(hash, url);
```

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,37 @@ export default definePlugin({
...
             if (!message.content) return;
-
+        
+            const urlRegex = /(https?:\/\/\S+)/g;
+        
+            const matches = message.content.match(urlRegex);
+        
+            if (matches && matches[0] === message.content.trim()) return;
+        
+            const urlMap = new Map();
```
Vendicated, line 98: this map honestly isn't needed, just hash again

nay-cat: I just tested a version by recalculating the hash seems to work well. I will upload the commit now, anyway, isn't it better to use a map than to recalculate the hashes all the time in terms of performance?

Vendicated: if you're calculating a lot of hashes, sure

but messages shouldn't contain more than like 10 links. and Maps are based on hashes so it literally hashes the string for you under the hood too to generate the map key haha

either way with so little links the performance difference is irrelevant but using a Map just makes it more complicated

nay-cat: Have you managed to test if everything is working properly?

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,31 @@ export default definePlugin({
         this.preSend = addPreSendListener(async (_, message) => {
             if (!settings.store.autoTranslate) return;
             if (!message.content) return;
-
+        
+            const urlRegex = /(https?:\/\/\S+)/g;
```
Vendicated, line 92:
```suggestion
            const urlRegex = /https?:\/\/\S+/g;
```

Vendicated: additionally this regex should be moved to the top level!

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,31 @@ export default definePlugin({
...
+        
+            const matches = message.content.match(urlRegex);
+        
+            if (matches && matches[0] === message.content.trim()) return;
```
Vendicated, line 96:
```suggestion
            if (matches?.[0] === message.content.trim()) return;
```

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,31 @@ export default definePlugin({
...
+        
             setShouldShowTranslateEnabledTooltip?.(true);
             clearTimeout(tooltipTimeout);
             tooltipTimeout = setTimeout(() => setShouldShowTranslateEnabledTooltip?.(false), 2000);
-
-            const trans = await translate("sent", message.content);
-            message.content = trans.text;
-
+        
+            const tempContent = message.content.replace(urlRegex, (url) => h64(url).toString(16));
```
Vendicated, line 102: if we do hex here, there is technically a very small chance to generate english words like beef and have Google translate it xD. making this purely numeric would be much better (e.g. go the discord snowflake route like I suggested)

src/plugins/translate/index.tsx
```diff
@@ -87,14 +88,31 @@ export default definePlugin({
...
+            const trans = await translate("sent", tempContent);
+        
+            let translatedContent = trans.text;
+        
+            if (matches) {
+                matches.forEach((url) => {
+                    const hash = h64(url).toString(16);
+                    translatedContent = translatedContent.replaceAll(hash, url); 
+                });
+            }
```
Vendicated, line 113:
```suggestion
            matches?.forEach(url => {
                const hash = h64(url).toString(16);
                translatedContent = translatedContent.replaceAll(hash, url); 
            });
```

Vendicated (commented):
thanks for the changes, that's already better!

src/plugins/translate/index.tsx
nay-cat: The probability of that happening is extremely low!!! If this bug happens to someone, they should play the lottery

Anyway, you're right, I'll be working on a version using snowflake or an alternative to avoid generating weird words

flawless-run:
as mentioned in #3205, this issue also happens with emojis. would be nice to add emojis to the regex as well.

after some testing, i've noticed that it can mess up mentions as well.

{13B33970-5107-4E6C-BDAF-1FCFEC8E7915}

might be worth adding the regex to mentionables such as users, channels, etc.

---

as mentioned in #3205, it would be great if we could add this to the plugin settings:
```ts
  includeOriginalText: {
      type: OptionType.BOOLEAN,
      description: "Include the original message content in the translated message",
      default: false
  }
 ```
 a lot of Chinese users have requested this in communities i'm in. would simply add the original text at the bottom for easier reading
 ```ts
 if (settings.store.includeOriginalText) {
    message.content = `${translatedMessage}\n-# ${message.content}`;
} else {
    message.content = translatedMessage;
}
 ```

nay-cat:
> as mentioned in #3205, this issue also happens with emojis. would be nice to add emojis to the regex as well.
> 
> after some testing, i've noticed that it can mess up mentions as well.
> 
> {13B33970-5107-4E6C-BDAF-1FCFEC8E7915}
> 
> might be worth adding the regex to mentionables such as users, channels, etc.
> 
> as mentioned in #3205, it would be great if we could add this to the plugin settings:
> 
> ```ts
> includeOriginalText: {
>     type: OptionType.BOOLEAN,
>     description: "Include the original message content in the translated message",
>     default: false
> }
> ```
> 
> a lot of Chinese users have requested this in communities i'm in. would simply add the original text at the bottom for easier reading
> 
> ```ts
> if (settings.store.includeOriginalText) {
>    message.content = `${translatedMessage}\n-# ${message.content}`;
> } else {
>    message.content = translatedMessage;
> }
> ```

I will try to work on that too

nay-cat:
@Vendicated 

I have tested with discord snowflakes and they work fine in almost all languages, but when testing with languages like Chinese, the translator would add random numbers to the snowflake or translate them, for example.

I tried sending 100 messages and on two occasions, the Chinese translator added a 9 to the end of the snowflake making it unrecognisable or edited a number.

image
image

I tried with mathematical characters and many other things, even with html tags but it always gave an error, either in DeepL or in Google Translator

It didn't happen with the hash, but when using the hash, the messages were broken when translating from Chinese to Italian.

I tried so many things, that at one point I thought that some kind of mystical entity was trying to communicate with me through the translator.
image
image

In the end, I simply made the messages be replaced by <$index> according to the number of transformations needed, this seems much simpler, but with languages like Chinese, Japanese and Arabic, the hashes and snowflakes worked badly at times.

I modified the emoji regex as it did not support animated emojis.
image

Vendicated:
lmaoo 😭 
image

Thank you for investigating, sounds good!

PR #3100, opened 2024-12-23
New Plugin: OpenIn

ItzRxg3:
Originally just created this for myself, but i think it may be useful for somebody else :>


Vendicated:

unfortunately, I'll have to reject this for multiple reasons:
- this plugin is a huge security hazard. your native openLink plugin lets the browser execute any code on the host system, which is a full sandbox escape
- it seems way too niche
- it won't work equally on all operating systems

PR #3101, opened 2024-12-23
OpenInApp: Add support for geo.music.apple.com links

faejr:
There's a form of Apple Music link that starts with `geo`, you can commonly get these from link websites from songs and such. 
As a result I noticed that they currently don't open in Apple Music, like the rest of the regular share links. 

All this PR does is add an optional `geo.` part to the regex match.

### DIFF
--- a/src/plugins/openInApp/index.ts
+++ b/src/plugins/openInApp/index.ts
@@ -57,8 +57,8 @@ const UrlReplacementRules: Record<string, URLReplacementRule> = {
         description: "Open Tidal links in the Tidal app",
     },
     itunes: {
-        match: /^https:\/\/(geo\.)?music\.apple\.com\/([a-z]{2}\/)?(album|artist|playlist|song|curator)\/([^/?#]+)\/?([^/?#]+)?(?:\?.*)?(?:#.*)?$/,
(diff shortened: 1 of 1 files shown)


src/plugins/openInApp/index.ts
```diff
@@ -57,8 +57,8 @@ const UrlReplacementRules: Record<string, URLReplacementRule> = {
         description: "Open Tidal links in the Tidal app",
     },
     itunes: {
-        match: /^https:\/\/music\.apple\.com\/([a-z]{2}\/)?(album|artist|playlist|song|curator)\/([^/?#]+)\/?([^/?#]+)?(?:\?.*)?(?:#.*)?$/,
-        replace: (_, lang, type, name, id) => id ? `itunes://music.apple.com/us/${type}/${name}/${id}` : `itunes://music.apple.com/us/${type}/${name}`,
+        match: /^https:\/\/(geo\.)?music\.apple\.com\/([a-z]{2}\/)?(album|artist|playlist|song|curator)\/([^/?#]+)\/?([^/?#]+)?(?:\?.*)?(?:#.*)?$/,
+        replace: (_, __, lang, type, name, id) => id ? `itunes://music.apple.com/us/${type}/${name}/${id}` : `itunes://music.apple.com/us/${type}/${name}`,
```
Sqaaakoi, line 61: you could just put ?: in the capture group to make it a non capturing match

Sqaaakoi (commented):
looks good now

PR #3102, opened 2024-12-23
New Plugin NeverAnimate

rylie-b:
Plugin simply disables all animations from profiles/servers/avatars instead of forcing them all on either on load or on hover (always on is already a plugin); I amended the AlwaysAnimate plugin to suit folks like myself with epilepsy, seizure risks, et al - as Discords response to updating their accessibility features functionally neutered this.

### DIFF
--- a/src/plugins/DisableAnimations/index.ts
+++ b/src/plugins/DisableAnimations/index.ts
@@ -1,0 +1,44 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors*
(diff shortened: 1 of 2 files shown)


KrstlSkll69:
Is this not just this but with patches?

https://github.com/Vendicated/Vencord/pull/3069

rylie-b:
> Is this not just this but with patches?
> 
> #3069

As far as I am aware of their plugin, it does not disable the same things. The one here would disable avatar/pfp animations, even on hover; disable server icons from animating, server and profile banners from animating as well as emoji-animations within the status bar of the "member list".

Issue #3085, opened 2024-12-17
[Bug] QuickReply moves caret to the beginning of the message.


ofluffydev:
### What happens when the bug or crash occurs?

More of a programming logic bug and not a crashing one, but when you use quick reply in the middle of typing a message it will move your caret to the beginning of the message. Not the worst bug but I've done it like 10 times today on accident, it's getting pretty annoying.

### What is the expected behaviour?

When you activate quick reply with ctrl up/down it should preserve the location of the caret in the message.

### How do you recreate this bug or crash?

1. Type something into the message box
2. Attempt to use a quick reply
3. Caret at the beginning of the message...a 

### Errors

n/a

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well
- [x] I have read the requirements for opening an issue above

Sqaaakoi:
I'm currently trying to fix this. I can explain why this happens, and why this is slightly harder to fix than I initially thought.

The listener for the keyboard events is set on the document itself, but this occurs after we have a chance to cancel the browser moving the caret.

I'll still try to find a way to fix this though.

Sqaaakoi:
i forgot preventDefault existed

PR #3104, opened 2024-12-24
QuickReply: Prevent caret from moving when selecting message

Sqaaakoi:
Fixes Vendicated/Vencord#3085
Additionally this refactors flux listeners to new standard and tries to replicate stock reply button behaviour slightly more accurately

### DIFF
--- a/src/plugins/quickReply/index.ts
+++ b/src/plugins/quickReply/index.ts
@@ -68,8 +68,12 @@ export default definePlugin({
     },
 
     flux: {
-        DELETE_PENDING_REPLY: () => { replyIdx = -1; },
-        MESSAGE_END_EDIT: () => { editIdx = -1; },
+        DELETE_PENDING_REPLY() {
+            replyIdx = -1;
+        },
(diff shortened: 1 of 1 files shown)


src/plugins/quickReply/index.ts
```diff
@@ -60,24 +60,20 @@ export default definePlugin({
...
-        Dispatcher.unsubscribe("MESSAGE_START_EDIT", onStartEdit);
-        Dispatcher.unsubscribe("CREATE_PENDING_REPLY", onCreatePendingReply);
         document.removeEventListener("keydown", onKeydown);
     },
-});
 
-const onDeletePendingReply = () => replyIdx = -1;
-const onEndEdit = () => editIdx = -1;
+    flux: {
+        DELETE_PENDING_REPLY: () => { replyIdx = -1; },
```
Vendicated, line 71:
```suggestion
        DELETE_PENDING_REPLY() {
            replyIdx = -1;
        },
```

PR #3106, opened 2024-12-26
break better folders

Zach11111:
This pr breaks better folders because it has been far too long since discord has broken it.


nin0-dev (commented):
lgtm

sadan4:
10/10 LGTM

PR #2938, opened 2024-10-08
ShowMeYourName: add support for user mentions

sadan4:
adds support for user mentions in show me your name

PR #3107, opened 2024-12-26
feat(ShowMeYourName): Nick, Display, Username Permutations; Streamer Mode; Mentions; Affixes; Colors

EtorixDev:
### Changes
1. Added a setting to respect Streamer Mode which is enabled by default.
2. Added all permutations (15) of Nick, Display, and Usernames.
    - Added a setting to exclude duplicates if there is overlap.
    - Nick/Display fill in for each other if either is missing, and if both are then username fills in.
3. Set custom prefixes and suffixes (up to 3 characters long) for each. Example: `Nick [DISPLAY] (@username)`
    - The first in the order does not have the symbols unless the setting to always show the symbols for that specific name is enabled.
4. Set custom colors for each, including separately for the prefixes and suffixes.
   - Accepts any hex(a), rgb(a), or hsl(a) input.
   - Accepts `Role+-#` input (ex: `Role+15`) to set it to the role color brightened/darkened by that percentage. Falls back to Discord's default gray.
   - Doesn't affect the first name listed unless the setting to always have the color for that specific name is enabled.
5. Borrowed and modified @sadan4's pr #2938 to add a setting to display the names in @ mentions.
   - Added an additional setting to hide the default `@` sign from mentions (and replies) due to the possibility of it visually conflicting with the prefixes and suffixes.

~~Ignore the branch name. I originally just planned to add streamer mode support before deciding to add more. (Also wasn't aware there was already a PR for that, oops.)~~

This PR adds a ton of settings due to the customization but it's all quite straightforward and the default settings should serve most people fine in my opinion if they don't want to tinker with anything.

### Examples
Before Message
After Message
Before Mention
After Mention


src/plugins/showMeYourName/index.tsx
```diff
@@ -4,48 +4,236 @@
...
-    message: Message;
-    withMentionPrefix?: boolean;
-    isRepliedMessage: boolean;
-    userOverride?: User;
+import { findStoreLazy } from "@webpack";
+import { ChannelStore, GuildMemberStore, UserStore } from "@webpack/common";
+import { GuildMember, User } from "discord-types/general";
+
+const StreamerModeStore = findStoreLazy("StreamerModeStore");
+const colorPattern = /^#(?:[\da-f]{3}){1,2}$|^#(?:[\da-f]{4}){1,2}$|(rgb|hsl)a?\((\s*-?\d+%?\s*,){2}(\s*-?\d+%?\s*)\)|(rgb|hsl)a?\((\s*-?\d+%?\s*,){3}\s*(0|(0?\.\d+)|1)\)$/iu;
```
sadan4, line 16: it would be eaiser to just only allow an hex code as the color.

sadan4 (commented):
it would probably be eaiser if the user just passed a format string eg`{username} ({nickname})` instead of having six settings for prefixes and suffixes

src/plugins/showMeYourName/index.tsx
```diff
@@ -4,48 +4,236 @@
...
+        description: "Show username color even if the username is the only name displayed, or is the first displayed.",
+        default: false,
+    }
 });
 
 export default definePlugin({
     name: "ShowMeYourName",
-    description: "Display usernames next to nicks, or no nicks at all",
-    authors: [Devs.Rini, Devs.TheKodeToad],
+    description: "Display any permutation of nicknames, display names, and usernames in chat. Nicknames are per-server and display names are global. You will need to hover over old messages to update them after changing any settings.",
```
sadan4, line 235:
```suggestion
    description: "Display any permutation of nicknames, display names, and usernames in chat",
```

describing nicknames and display names doesn't belong in the plugin description

if you don't want to hover over messages to update them after changing settings, you can use settings as react state

EtorixDev:
> it would probably be eaiser if the user just passed a format string eg`{username} ({nickname})` instead of having six settings for prefixes and suffixes

Updated to switch to a string format input instead of a drop down.

src/plugins/showMeYourName/index.tsx
EtorixDev: Perhaps but it's already implemented and as far as I can tell quite robust and allows more freedom of input.

EtorixDev:
Updated with gradient support, but this requires `discord-types` to be updated with the new `GuildMember` field:
```js
colorStrings: {
    primaryColor: string;
    secondaryColor: string | undefined;
    tertiaryColor: string | undefined;
};
```

Edit: Just switched to using (x as any) syntax to get rid of the error.

PR #3108, opened 2024-12-27
feat(plugin): SpoilerReveal

rushy-me:
Allows spoilers to be revealed on hover, with ctrl+click to hide them again.

kziKEwQ6l3fVNwmaZp9q9jJqYLaGmIg


sadan4 (commented):
i feel like it would make more sense for this plugin to be merged with revealAllSpoilers, they have a lot in common

PR #3110, opened 2024-12-28
NoUnblockToJump: Also allow jump for ignored users

sadan4:
why whas that the pr message?


src/plugins/noUnblockToJump/README.md
```diff
@@ -0,0 +1,5 @@
+# No Unblock To Jump
+
+Removes this popup when you try to jump to a message (eg: in search results).
```
Nuckyz, line 3:
```suggestion
Removes the popup preventing you to jump to a message from a blocked/ignored user (eg: in search results)
```

PR #3111, opened 2024-12-28
Create userplugins

alyaseer2:


Vendicated:
there is no tooth fairy, and you can't use BetterDiscord plugins on Vencord :(

PR #3114, opened 2024-12-29
New Plugin : CustomCursor

HaddajiDev:
Replace the default cursor with a custom cursor


nin0-dev:
afaik you can add custom cursors with CSS directly

also in the future, avoid linking to external assets

PR #3119, opened 2025-01-02
New Plugin: fixupLinks

kodenamekrak:
Adds configurable domain replacement in messages, primarily intended for easier use of fixup websites such as fxtwitter that provide prettier discord embeds


Vendicated:
TextReplace can do this

PR #3126, opened 2025-01-06
NoBlockedMessages: Add ignored messages

jamesbt365:
Adds ignored user support to NoBlockedMessages.

Duplicating the patch and adding yet another setting for blocking the messages from the message store seems redundant, so i just changed the function a little bit and used one patch instead.


src/plugins/noBlockedMessages/index.ts
```diff
@@ -67,22 +67,44 @@ export default definePlugin({
...
         try {
-            return RelationshipStore.isBlocked(message.author.id);
+            if (RelationshipStore.isBlocked(message.author.id)) {
+                return true;
+            }
+            if (this.options.ignoreIgnoredMessages && RelationshipStore.isIgnored(message.author.id)) {
+                return true;
+            }
+
+            return false;
```
sadan4, line 90:
```suggestion
	return this.options.ignoreIgnoredMessages && RelationshipStore.isIgnored(message.author.id);
```

BAHO777:
Hi, can you make sure that the blocked messages don't pop up in other users' replies? That would be really cool!

jamesbt365:
> Hi, can you make sure that the blocked messages don't pop up in other users' replies? That would be really cool!

Hey! Could you not necro an old PR and send me notifications and emails for something that has fuck all to what I fixed please? That would be really cool!

PR #3130, opened 2025-01-08
Add Plugin: quickNav

RedlineDevs:
## Description:

Basically this plugin lets you set certain keybinds in Servers, Channels or DMs using a Slash Command to be able to.. as the title says, quickly navigate to them whenever, wherever

### Setting Keybinds
- `/qn-set-kb`: Create a new navigation keybind
  - `name`: Name for the keybind (required)
  - `keys`: Comma-separated list of keys (required)
    - Must include at least one modifier (ctrl, alt, or shift)
    - Example: `ctrl,shift,l` (btw has to include the commas, no spaces)
  - `channel`: Target channel (optional, defaults to current channel)
  - `category`: Group name for organizing keybinds (optional)
  - `description`: Description of the keybind's purpose (optional)

### Managing Keybinds
- `/qn-list-kb`: View all keybinds
  - `category`: Filter by category name (optional)
  - `sort`: Sort by name, created, or lastused (optional)
- `/qn-remove-kb`: Delete a keybind
  - `id`: ID of the keybind to remove (required)

### Valid Keys
- **Modifiers**: ctrl, alt, shift
- **Letters**: a-z
- **Numbers**: 0-9
- **Function**: f1-f12
- **Navigation**: tab, capslock, space, backspace, delete, enter, escape, home, end, pageup, pagedown, arrowup, arrowdown, arrowleft, arrowright

image
*and yes I started coding this at 4am for some reason... I was bored*

### DIFF
--- a/src/plugins/quickNav/index.tsx
+++ b/src/plugins/quickNav/index.tsx
@@ -1,0 +1,464 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { ApplicationCommandInputType, ApplicationCommandOptionType, findOption, sendBotMessage } from "@api/Commands";
+import * as DataStore from "@api/DataStore";
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+import { NavigationRouter } from "@webpack/common";
+import { Channel } from "discord-types/general";
+
+interface QuickNavKeybind {
+    id: number;
+    name: string;
+    keys: string[];
+    guildId?: string;
+    channelId: string;
+    category?: string;
+    description?: string;
+    createdAt: number;
+    lastUsed?: number;
+}
+
+const DATA_KEY = "QuickNav_KEYBINDS";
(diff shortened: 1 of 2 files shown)


Vendicated:
Just use QuickSwitcher: `CTRL` + `k` / `CTRL` + `t`

RedlineDevs:
> Just use QuickSwitcher: `CTRL` + `k` / `CTRL` + `t`

Thought of that too but tbh, but lazy people (Like me) just want the lazy way which is a keybind, this would also help if you are talking in diff servers/dms at the same time and u cba to keep pressing back/forward

Kyuuhachi:
Why is this configured with commands rather than configuration?

RedlineDevs:
> Why is this configured with commands rather than configuration?

Like `registerCommand`?

RedlineDevs:
> > Why is this configured with commands rather than configuration?
> 
> Like `registerCommand`?

Ahhh Why the Keybinds are configured with commands rather than Config, Mainly because it's gonna be easier and faster, instead of needing to open settings > plugins > quickNav and then modify the settings

Plus Embeds look better

If needed I might try and implement support for the Config though!

PR #3131, opened 2025-01-08
sentry-ipc grey screen fix

SK3Artemis:
This fixed the grey screen for me...
The Error was: Refused to connect to 'sentry-ipc://sentry-electron.renderer-start/sentry_key' because it violates the document's Content Security Policy.


Vendicated:
this is not valid so it can't be the fix. the `:` at the end of schemes is required, so this has no effect

Vendicated:
Also this error is irrelevant and doesn't cause any issues. It is caused by OpenAsar, uninstalling OpenAsar fixes it

Vendicated:
https://x.com/discord_support/status/1876990070713987223

PR #3132, opened 2025-01-08
feat: 🎉 `blurAuthor` plugin

TobyBridle:
This adds the ability to set author(s) which have their attachments automatically hidden, similar to the existing `blurNsfw` plugin.

In addition to setting it within the Plugin configuration page, it adds an action to the context menu to add/remove a given user to the blocklist.

Photos:
- Hidden by default
- Shown when the IMAGE (not the message) is hovered

Videos:
- Hidden by default
- Shown when playing, hovered or in fullscreen

*I understand that there may not be an open request for this. I was interested in some plugin development and wanted to do it for personal uses*

### DIFF
--- a/src/plugins/blurAuthor/index.ts
+++ b/src/plugins/blurAuthor/index.ts
@@ -49,7 +49,7 @@ const settings = definePluginSettings({
         description: "User Blacklist (Comma Separated User-IDs)",
         placeholder: "261607958822125568,865521228859706162",
         isValid: (ids: string) => {
-            if (!!!ids || !ids.trim().length) return true;
+            if (!ids.trim().length) return true;
 
             if (/^(?:\d{18})(?:,\s?\d{18})*$/.test(ids)) return true;
             else return "Please ensure the User IDs are valid and separated correctly!";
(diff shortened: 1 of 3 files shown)


src/plugins/blurAuthor/index.ts
```diff
@@ -0,0 +1,108 @@
...
+            }]
+        }
+    ],
+
+    hasUser: (id: string): boolean => {
+        return hasUser(id);
+    },
+
+    start() {
+        style = document.createElement("style");
```
sadan4, line 98: import a css file instead of using dom mainp

sadan4: nevermind, im stupid

didnt notice the template string :sob:

Vendicated:
The BlurNSFW plugin has awful code and nothing new should be based on it. We also don't allow plugins that are just slight modifications of others. It would make much more sense to just add this functionality to the original plugin. Could just add an option to mark users as NSFW

TobyBridle:
> The BlurNSFW plugin has awful code and nothing new should be based on it. We also don't allow plugins that are just slight modifications of others. It would make much more sense to just add this functionality to the original plugin. Could just add an option to mark users as NSFW

If that's welcome, I'm perfectly fine with doing that

PR #3137, opened 2025-01-11
Switch profile button

Sqaaakoi:
Moves the View Main/Server Profile button out of the overflow menu

Screenshot of the View Main Profile button

Some code was copied from ReviewDB

How it works:
Replaces the MenuItem components with the button, props are reused as-is by the button component
Moves the item out of the overflow menu component props and into the row of buttons

Suggestions for a different icon are welcome

### DIFF
--- a/src/plugins/switchProfileButton/index.tsx
+++ b/src/plugins/switchProfileButton/index.tsx
@@ -1,10 +1,9 @@
-import { migratePluginSettings } from "@api/Settings";
 import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants";
 import { classes } from "@utils/misc";
 import definePlugin from "@utils/types";
 import { findByPropsLazy } from "@webpack";
-import { Button, Icons, TooltipContainer, Text, type MenuTypes } from "@webpack/common";
+import { Button, Icons, TooltipContainer, Text } from "@webpack/common";
 import { ReactNode } from "react";
 
 const RoleButtonClasses = findByPropsLazy("button", "buttonInner", "icon", "banner");
--- a/src/plugins/switchProfileButton/index.tsx
+++ b/src/plugins/switchProfileButton/index.tsx
@@ -1,0 +1,65 @@
+import { migratePluginSettings } from "@api/Settings";
+import ErrorBoundary from "@components/ErrorBoundary";
+import { Devs } from "@utils/constants";
+import { classes } from "@utils/misc";
+import definePlugin from "@utils/types";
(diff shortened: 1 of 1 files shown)


RedCrafter07:
Hey there, I love the idea! 
What about a server icon? I think Discord has one, but don't quote me on that

src/plugins/switchProfileButton/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    SwitchProfileButton: ErrorBoundary.wrap((props: SwitchProfileMenuItemProps) => {
+        return <TooltipContainer
+            text={<>
+                {props.label}
+                <Text
+                    color="text-muted"
+                    variant="text-xs/medium"
+                >{props.subtext}</Text>
+            </>}
+            aria-label={false}
```
Vendicated, line 55: wha?

src/plugins/switchProfileButton/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Devs } from "@utils/constants";
+import { classes } from "@utils/misc";
+import definePlugin from "@utils/types";
+import { findByPropsLazy } from "@webpack";
+import { Button, Icons, Text, TooltipContainer } from "@webpack/common";
+import { ReactNode } from "react";
+
+const RoleButtonClasses = findByPropsLazy("button", "buttonInner", "icon", "banner");
+
+type SwitchProfileMenuItemProps = {
```
Vendicated, line 17:
```suggestion
interface SwitchProfileMenuItemProps {
```

src/plugins/switchProfileButton/index.tsx
Sqaaakoi: This is how Discord does it for the other buttons there. I am aware that ReviewDB does not do it this way.

Sqaaakoi: I went digging again to find where I took this code from and it looks like I stole it from this component, and it turns out I didn't steal all of the code

Screenshot of the component code, where the outer aria-label is set to false and the inner aria-label is set to an explicitly defined value
(from module 475413)

Now that I've looked at it, I agree that this can be improved, but I'm unsure as to how I should improve it.
I can't use the actual component for this as the aria-label can't be set separately, which is what I need to do as the tooltip text isn't a string.

Sqaaakoi:
Discord has forced me to fix the code. I still don't know what icon to use

Sqaaakoi:
I've fixed the patch for the button being broken now, though for some reason there is a huge delay when clicking the button?

PR #3142, opened 2025-01-15
feat(Plugin): NoJoinOnClick

Moritzoni:
Restores the behavior of clicking voice channels in the Active Now tab without joining

### DIFF
--- a/src/plugins/noJoinOnClick/index.ts
+++ b/src/plugins/noJoinOnClick/index.ts
@@ -1,0 +1,24 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2025 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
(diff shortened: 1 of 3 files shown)


src/plugins/noJoinOnClick/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    description: "Restores the behavior of clicking voice channels in the Active Now tab without joining",
+    authors: [Devs.Moritzoni],
+
+    patches: [{
+        find: "h.default.selectVoiceChannel(r.id),",
+        replacement: [
+            {
+                match: "h.default.selectVoiceChannel(r.id),",
+                replace: "h.default.selectChannel(r.id),"
+            }
```
sadan4, line 21: dont hardcode variables

sadan4 (commented):
this should be merged with doubleclick to join, not its own plugin

PR #3153, opened 2025-01-19
Unified Styles API

Sqaaakoi:
### The problem
Vencord does not inject styles into popout windows.
Additionally, plugins that create their own styles at runtime contain a lot of (ugly) code to add these styles.

### How this solves the problem
Almost every call to `document.createElement("style")`[^1] is replaced with the unified styles API.
This API should be considered the single source of truth regarding stylesheets for everything from now on.
The underlying data structure for "managed styles" has been changed to not store references to DOM `<style>` elements.
Instead, DOM elements are looked up in all available documents (current document and popouts) when necessary.
When a popout is created, all enabled styles are injected to that document.
Please note that there is no "style updated" event system implemented into this API.

### How to use it
The major functions that should be commonly used are `createStyle`[^2], `deleteStyle`, and `setStyle`. The existing `enableStyle` and `disableStyle` can still be used for managed styles.

I am aware that there's still a decent amount of out-of-date documentation in the code, and some things should be renamed or moved around. Please do provide feedback on this.
Some code was taken from #3080 (thank you @Masterjoona for originally developing code for CSS in popouts)

[^1]: Not used in YoutubeAdblock or ~~ShikiCodeblocks, though the code in ShikiCodeblocks looks redundant.~~ Fixed
[^2]: `createStyle` is also used to update styles


scripts/build/common.mjs
```diff
@@ -276,7 +276,7 @@ export const stylePlugin = {
     name: "style-plugin",
     setup: ({ onResolve, onLoad }) => {
         onResolve({ filter: /\.css\?managed$/, namespace: "file" }, ({ path, resolveDir }) => ({
-            path: relative(process.cwd(), join(resolveDir, path.replace("?managed", ""))),
+            path: relative(process.cwd(), join(resolveDir, path.replace(/\?managed$/, ""))),
```
Vendicated, line 279: why

src/plugins/blurNsfw/index.ts
```diff
@@ -56,19 +43,16 @@ export default definePlugin({
...
         }
     },
 
     start() {
-        style = document.createElement("style");
-        style.id = "VcBlurNsfw";
-        document.head.appendChild(style);
-
-        setCss();
+        style.edit = src => src.replace("blur-amount", Settings.plugins.BlurNSFW.blurAmount);
```
Vendicated, line 51: this is horror

Sqaaakoi: i am aware. what would you suggest doing instead? Keeping it inline? Get rid of the edit function entirely?

scripts/build/common.mjs
Sqaaakoi: in this current state, yes it's unnecessary
i was just working on adding enabled by default styles, would there be ANY usecase for that?

Sqaaakoi: idc about the styles being enabled idea anymore, should this just be left as a string

PR #3154, opened 2025-01-19
Settings API: add support for custom objects / arrays

Nuckyz:
Adds support for using objects/array settings which sync whenever they are modified, however they have no UI implementation and are always hidden.

This should be used for saving data which is handled by its own UI, like PinDMs. The difference of using this from DataStore is that the data gets cloud synced.

### DIFF
--- a/src/plugins/messageTags/index.ts
+++ b/src/plugins/messageTags/index.ts
@@ -37,18 +37,15 @@ function getTags() {
 }
 
 function getTag(name: string) {
-    return settings.store.tagsList.find(tag => tag.name === name) ?? null;
+    return settings.store.tagsList[name] ?? null;
 }
 
 function addTag(tag: Tag) {
-    settings.store.tagsList.push(tag);
+    settings.store.tagsList[tag.name] = tag;
 }
 
 function removeTag(name: string) {
-    const index = settings.store.tagsList.findIndex(tag => tag.name === name);
-    if (index === -1) return;
-
-    settings.store.tagsList.splice(index, 1);
+    delete settings.store.tagsList[name];
 }
 
 function createTagCommand(tag: Tag) {
@@ -81,10 +78,8 @@ const settings = definePluginSettings({
         default: true
     },
     tagsList: {
-        type: OptionType.ARRAY,
-        description: "",
-        hidden: true,
-        default: [] as Tag[],
+        type: OptionType.CUSTOM,
+        default: {} as Record<string, Tag>,
     }
 });
 
@@ -105,7 +100,7 @@ export default definePlugin({
         // TODO: Remove DataStore tags migration once enough time has passed
         const oldTags = await DataStore.get<Tag[]>(DATA_KEY);
         if (oldTags != null) {
-            settings.store.tagsList = oldTags;
+            settings.store.tagsList = Object.fromEntries(oldTags.map(oldTag => [oldTag.name, oldTag]));
             await DataStore.del(DATA_KEY);
         }
 
--- a/src/plugins/messageTags/index.ts
+++ b/src/plugins/messageTags/index.ts
@@ -18,7 +18,7 @@
 
 import { ApplicationCommandInputType, ApplicationCommandOptionType, findOption, registerCommand, sendBotMessage, unregisterCommand } from "@api/Commands";
 import * as DataStore from "@api/DataStore";
-import { Settings } from "@api/Settings";
+import { definePluginSettings, Settings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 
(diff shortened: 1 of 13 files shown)


src/plugins/messageTags/index.ts
```diff
@@ -32,28 +32,29 @@ interface Tag {
...
+function getTag(name: string) {
+    return settings.store.tagsList.find(tag => tag.name === name) ?? null;
+}
+
+function addTag(tag: Tag) {
+    settings.store.tagsList.push(tag);
+}
+
+function removeTag(name: string) {
+    settings.store.tagsList = settings.store.tagsList.filter(tag => tag.name !== name);
```
Vendicated, line 48: better to use splice with findIndex (make sure it's not -1 or it pops last element)

src/plugins/messageTags/index.ts
```diff
@@ -32,28 +32,29 @@ interface Tag {
...
-    tags = await tags.filter((t: Tag) => t.name !== name);
-    DataStore.set(DATA_KEY, tags);
-    return tags;
-};
+function getTags() {
+    return settings.store.tagsList;
+}
+
+function getTag(name: string) {
+    return settings.store.tagsList.find(tag => tag.name === name) ?? null;
```
Vendicated, line 40: kinda unrelated to this pr but it would make much more sense to use an object instead of array for this, since this array is basically used like a KeyVal Record. would be super simple with just a private setting

src/plugins/pinDms/data.ts
```diff
@@ -19,109 +18,127 @@ export interface Category {
...
+    initCategories(UserStore.getCurrentUser()?.id);
+    forceUpdate();
+}
+
+let currentUserCategoriesIndex = -1;
+export let currentUserCategories: Category[] = [];
 
-export let categories: Category[] = [];
+export function initCategories(userId: string) {
+    const categoriesIndex = settings.store.userBasedCategoryList.findIndex(categoryList => categoryList.userId === userId);
```
Vendicated, line 42: also could just be an object using private settings instead of array

src/plugins/pinDms/index.tsx
```diff
@@ -157,9 +161,9 @@ export default definePlugin({
     },
 
     startAt: StartAt.WebpackReady,
-    start: init,
+    start: () => init(),
     flux: {
-        CONNECTION_OPEN: init,
+        CONNECTION_OPEN: () => init(),
```
Vendicated, line 166: why

src/plugins/textReplace/index.tsx
```diff
@@ -120,28 +123,27 @@ function Input({ initialValue, onChange, placeholder }: {
...
-
-        await DataStore.set(rulesKey, rulesArray);
-        update();
     }
 
     async function onChange(e: string, index: number, key: string) {
-        if (index === rulesArray.length - 1)
+        if (index === rulesArray.length - 1) {
             rulesArray.push(makeEmptyRule());
+        }
```
Vendicated, line 137: why

src/shared/SettingsStore.ts
```diff
@@ -34,7 +34,7 @@ export interface SettingsStore<T extends object> extends SettingsStoreOptions {
  */
 export class SettingsStore<T extends object> {
     private pathListeners = new Map<string, Set<(newData: any) => void>>();
-    private globalListeners = new Set<(newData: T, path: string) => void>();
+    private globalListeners = new Set<(newData: any, path: string) => void>();
```
Vendicated, line 37:
```suggestion
    private globalListeners = new Set<(newData: T, path: string) => void>();
```

src/shared/SettingsStore.ts
```diff
@@ -81,6 +88,41 @@ export class SettingsStore<T extends object> {
...
+    private makeArrayProxy(array: any[], arrayPath: string) {
+        const self = this;
+
+        return new Proxy(array, {
+            set(target, key: string, value) {
+                if (target[key] === value) return true;
+
+                Reflect.set(target, key, value);
+
+                self.globalListeners.forEach(cb => cb(target, arrayPath));
```
Vendicated, line 115: this is wrong. globalListeners get the root settings object

src/shared/SettingsStore.ts
```diff
@@ -81,6 +88,41 @@ export class SettingsStore<T extends object> {
...
+
+                self.globalListeners.forEach(cb => cb(undefined, setPath));
+                self.pathListeners.get(setPath)?.forEach(cb => cb(undefined));
+
+                return true;
+            }
+        });
+    }
+
+    private makeArrayProxy(array: any[], arrayPath: string) {
```
Vendicated, line 106: why exactly do we need a separate function? can't the normal proxy also work for arrays? also unless i'm missing something, this doesn't proxy the array elements so if you have 2d arrays or objects inside your array, those won't be dynamic

src/shared/SettingsStore.ts
```diff
@@ -67,20 +78,66 @@ export class SettingsStore<T extends object> {
                     });
                 }
 
-                if (typeof v === "object" && v !== null && !Array.isArray(v))
-                    return self.makeProxy(v, root, `${path}${path && "."}${key}`);
+                if (typeof v === "object" && v != null && !v[SYM_IS_PROXY]) {
```
Vendicated, line 81: nit: this can never be undefined at that point so === is more accurate :P

```suggestion
                if (typeof v === "object" && v !== null && !v[SYM_IS_PROXY]) {
```

src/plugins/pinDms/index.tsx
```diff
@@ -46,21 +44,28 @@ export const settings = definePluginSettings({
...
             { label: "Custom (right click channels to reorder)", value: PinOrder.Custom }
-        ],
-        onChange: () => forceUpdate()
+        ]
     },
-
-    dmSectioncollapsed: {
+    canCollapseDmSection: {
+        type: OptionType.BOOLEAN,
+        description: "Allow DMs section to be collapsable",
```
Vendicated, line 51:
```suggestion
        description: "Allow uncategorised DMs section to be collapsable",
```

src/shared/SettingsStore.ts
```diff
@@ -28,6 +31,93 @@ interface SettingsStoreOptions {
...
 
+interface ProxyContext<T extends object = any> {
+    settingsStore: SettingsStore<T>;
+    root: T;
+    path: string;
+}
+
+const proxyContexts = new WeakMap<any, ProxyContext>();
+
+const proxyHandler: ProxyHandler<any> = {
```
Vendicated, line 42: this and the proxyContents weak map should really be moved into the class so different stores don't interfere with one another (and this will also make accessing privates "illegally" obsolete)

src/shared/SettingsStore.ts
```diff
@@ -6,6 +6,9 @@
 
 import { LiteralUnion } from "type-fest";
 
+const SYM_IS_PROXY = Symbol("SettingsStore.isProxy");
+const SYM_GET_RAW_TARGET = Symbol("SettingsStore.getRawTarget");
```
Vendicated, line 10: eh why not
```suggestion
export const SYM_IS_PROXY = Symbol("SettingsStore.isProxy");
export const SYM_GET_RAW_TARGET = Symbol("SettingsStore.getRawTarget");
```

Vendicated (approved):

PR #3155, opened 2025-01-19
New Plugin: InboxJumpKeybind

arutonee1:
This is a keybind that I've been manually pasting into the dev console for the past year or so. It binds `Ctrl+;` to the topmost inbox jump button.
screenshot

### DIFF
--- a/src/plugins/inboxJumpKeybind/README.md
+++ b/src/plugins/inboxJumpKeybind/README.md
@@ -1,0 +1,3 @@
+# Inbox Jump Keybind
+
+Adds a keybind for that tiny jump button in the inbox
--- a/src/plugins/inboxJumpKeybind/index.ts
+++ b/src/plugins/inboxJumpKeybind/index.ts
@@ -1,0 +1,28 @@
+/*
+* Vencord, a Discord client mod
+* Copyright (c) 2025 Vendicated and contributors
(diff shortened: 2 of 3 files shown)


Scyye:
is it reasonable to make it customizable?

arutonee1:
@Scyye Ah yes, that would be reasonable. I'll update it.

arutonee1:
@Scyye Implemented! (Sorry for taking so long, I'm learning a new keyboard layout)

RyfterWasTaken:
Why not make 1 plugin that does a lot of shortcuts, and allows changing existing bindings, as this seems to have a very specific use case.

arutonee1:
@RyfterWasTaken I personally haven't found any actions without keybinds that are as useful as the jump button- It would be really nice to have a plugin that changes arbitrary keybinds, but I would consider that worthy of a separate plugin altogether (if someone wants to make such a plugin, I would support it the whole way through!).

In addition, I don't necessarily think there's anything wrong with having extremely specific plugins- just look at AlwaysAnimate, AlwaysExpandRoles, BetterGifAltText, BetterNotesBox, ColorSighted, DontRoundMyTimestamps, FixCodeblockGap, ForceOwnerCrown, MoreCommands, MoreKaomoji, or NoF1 to name a few.

PR #3156, opened 2025-01-20
ShowConnections: Cap max number of connections displayed

sadan4:


src/plugins/showConnections/index.tsx
```diff
@@ -57,6 +58,11 @@ const settings = definePluginSettings({
             { label: "Cozy", value: Spacing.COZY }, // US Spelling :/
             { label: "Roomy", value: Spacing.ROOMY }
         ]
+    },
+    maxNumberOfConnections: {
+        type: OptionType.NUMBER,
```
Vendicated, line 63: slider would be better. just do 5-50

src/plugins/validUser/index.tsx
```diff
@@ -101,7 +95,7 @@ async function getUser(id: string) {
         fakeBadges.push(badges.premium);
 
     // Fill in what we can deduce
-    const profile = UserProfileStore.getUserProfile(id);
+    const profile: Partial<UserProfile> = UserProfileStore.getUserProfile(id) ?? {};
```
sadan4, line 98: not sure what the point of this is in the first place, so not sure if this is the best/correct fix

src/plugins/showConnections/index.tsx
```diff
@@ -88,20 +90,51 @@ function ConnectionsComponent({ id, theme }: { id: string, theme: string; }) {
...
+                    <span className="vc-sc-connection-name">
+                        View All Connections
+                    </span>
+                </span>
+            }
+        >
+            {props => (
+                <Icons.MoreHorizontalIcon
+                    {...props}
+                    onClick={onClick}
```
Vendicated, line 130: this is not very accessible because there is nothing indicating this is a button

you should wrap this icon inside a button with the onClick 

(Discord button has ButtonLooks.BLANK or use their Clickable component)

sadan4: ButtonLooks.BLANK messes up the styles, so clickable works best

src/plugins/showConnections/index.tsx
```diff
@@ -88,20 +90,54 @@ function ConnectionsComponent({ id, theme }: { id: string, theme: string; }) {
...
+                    </span>
+                </span>
+            }
+        >
+            {props => (
+                <Clickable
+                    onClick={onClick}
+                >
+                    <Icons.MoreHorizontalIcon
+                        {...props}
```
Vendicated, line 132: spread the props into the Clickable instead, so the aria-label is exposed and the mouse listeners are on the right object

src/plugins/showConnections/index.tsx
```diff
@@ -88,20 +90,54 @@ function ConnectionsComponent({ id, theme }: { id: string, theme: string; }) {
...
+    return (
+        <Tooltip
+            text={
+                <span className="vc-sc-tooltip">
+                    <span className="vc-sc-connection-name">
+                        View All Connections
+                    </span>
+                </span>
+            }
+        >
```
Vendicated, line 126:
```suggestion
        <Tooltip text="View all Connections">
```

src/plugins/showConnections/index.tsx
sadan4: a slider doesn't work due to the fact that the size of the icons is customizable, leading to a multiple of seven  -1 not always being the ideal number for an even row

src/components/PluginSettings/components/SettingSliderComponent.tsx
```diff
@@ -40,16 +40,23 @@ export function SettingSliderComponent({ option, pluginSettings, definedSettings
...
         if (typeof isValid === "string") setError(isValid);
         else if (!isValid) setError("Invalid input provided.");
         else {
             setError(null);
-            onChange(newValue);
+            onChange(getValue(newValue));
         }
     }
-
+    console.log("rendering");
```
Vendicated: erm

Vendicated:
looks good but soooo many conflicts

PR #3157, opened 2025-01-20
Rewrite WebpackPatcher to support new features

Nuckyz:


src/utils/misc.ts
```diff
@@ -100,6 +100,11 @@ export function pluralise(amount: number, singular: string, plural = singular +
     return amount === 1 ? `${amount} ${singular}` : `${amount} ${plural}`;
 }
 
+export function interpolateIfDefined(strings: TemplateStringsArray, ...args: any[]) {
+    if (args.some(arg => arg == null)) return "";
+    return strings.reduce((acc, str, i) => `${acc}${str}${args[i] ?? ""}`, "");
```
Vendicated, line 105: this should work and be better

```suggestion
    return String.raw({ raw: strings }, ...args);
```

src/webpack/webpack.ts
```diff
@@ -86,16 +86,25 @@ export const filters = {
...
 export const moduleListeners = new Set<CallbackFn>();
-export const factoryListeners = new Set<(factory: (module: any, exports: any, require: WebpackInstance) => void) => void>();
-export const beforeInitListeners = new Set<(wreq: WebpackInstance) => void>();
+export const factoryListeners = new Set<FactoryListernFn>();
 
-export function _initWebpack(webpackRequire: WebpackInstance) {
+export function _initWebpack(webpackRequire: WebpackRequire) {
     wreq = webpackRequire;
+
+    if (webpackRequire.c == null) return;
```
Vendicated, line 99: should this be moved above assignment to `wreq`?

Vendicated (commented):
still have to look at patchWebpack

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+    try {
+        shouldPatchFactories = false;
 
-    get: () => webpackChunk,
-    set: v => {
-        if (v?.push) {
-            if (!v.push.$$vencordOriginal) {
-                logger.info(`Patching ${WEBPACK_CHUNK}.push`);
-                patchPush(v);
+        const hardcodedModuleAttempt = wreq(128014)?.b;
```
Vendicated, line 35: this is a bit dangerous since 128014 could potentially become a different module with side effects. maybe check m[128014] first and see if it's the right thing

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+        if (typeof hardcodedModuleAttempt === "function" && typeof hardcodedModuleAttempt() === "number") {
```
Vendicated, line 36: this function call should be wrapped in an individual try catch

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+        return undefined;
+    },
+    // Same thing as get
+    has: (target, p) => {
+        return false;
+    },
+    */
+
+    // The set trap for patching or defining getters for the module factories when new module factories are loaded
+    set: (target, p, newValue, receiver) => {
```
Vendicated, line 155:
```suggestion
    set(target, p, newValue, receiver) {
```

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+        // If the property is not a number, we are not dealing with a module factory
```
Vendicated, line 156: should we really rely on this? there is no guarantee and there were some discord pushes in the past that changed module ids to strings

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+ * @param ignoreExistingInTarget Whether to ignore checking if the factory already exists in the moduleFactoriesTarget
+ * @returns Whether the original factory was updated, or false if it doesn't exist in any Webpack instance
+ */
+function updateExistingFactory(moduleFactoriesTarget: AnyWebpackRequire["m"], id: PropertyKey, newFactory: AnyModuleFactory, ignoreExistingInTarget: boolean = false) {
+    let existingFactory: TypedPropertyDescriptor<AnyModuleFactory> | undefined;
+    let moduleFactoriesWithFactory: AnyWebpackRequire["m"] | undefined;
+    for (const wreq of allWebpackInstances) {
+        if (ignoreExistingInTarget && wreq.m === moduleFactoriesTarget) continue;
+
+        if (Reflect.getOwnPropertyDescriptor(wreq.m, id) != null) {
```
Vendicated, line 187: should be equivalent and better

```suggestion
        if (Object.hasOwn(wreq.m, id)) {
```

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+        }
+
+        // eslint-disable-next-line prefer-const
+        let [module, exports, require] = args;
 
-function patchFactories(factories: Record<string, (module: any, exports: any, require: WebpackInstance) => void>) {
-    for (const id in factories) {
-        let mod = factories[id];
+        if (wreq == null) {
+            if (!wreqFallbackApplied) {
```
Vendicated, line 289: is this something that can actually be reached during normal functionality or is this only if something breaks?

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+            try {
+                callback(exports, id);
+            } catch (err) {
+                logger.error("Error in Webpack module listener:\n", err, callback);
+            }
+        }
+
+        for (const [filter, callback] of waitForSubscriptions) {
+            try {
+
```
Vendicated, line 357:
```suggestion
```

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+/**
+ * Patches a module factory.
+ *
+ * @param id The id of the module
+ * @param factory The original module factory
+ * @returns The patched module factory
+ */
+function patchFactory(id: PropertyKey, factory: AnyModuleFactory) {
+    // 0, prefix to turn it into an expression: 0,function(){} would be invalid syntax without the 0,
```
Vendicated, line 406: are we sure we don't need to remove \n anymore?

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+        const moduleMatches = typeof patch.find === "string"
+            ? code.includes(patch.find)
+            : (patch.find.global && (patch.find.lastIndex = 0), patch.find.test(code));
+
+        if (!moduleMatches) continue;
+
+        if (
+            !Settings.eagerPatches &&
+            (patch.fromBuild != null && getBuildNumber() !== -1 && getBuildNumber() < patch.fromBuild) ||
+            (patch.toBuild != null && getBuildNumber() !== -1 && getBuildNumber() > patch.toBuild)
```
Vendicated, line 424: getBuildNumber should be cached in a variable instead of called 4 times

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
 
-            if (!patch.all) patches.splice(i--, 1);
-        }
+                code = newCode;
+                patchedFactory = (0, eval)(`// Webpack Module ${String(id)} - Patched by ${[...patchedBy].join(", ")}\n${newCode}\n//# sourceURL=WebpackModule${String(id)}`);
+            } catch (err) {
+                logger.error(`Patch by ${patch.plugin} errored (Module id is ${String(id)}): ${replacement.match}\n`, err);
+
+                if (IS_DEV) {
+                    const changeSize = code.length - lastCode.length;
```
Vendicated, line 486: let's extract this into a separate function so it's cleaner

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,536 @@
...
+                    const [titleFmt, ...titleElements] = Logger.makeTitle("white", "Diff");
+                    logger.errorCustomFmt(titleFmt + fmt, ...titleElements, ...elements);
+                }
 
-        if (IS_DEV) {
-            if (mod !== originalMod) {
-                factory.$$vencordPatchedSource = String(mod);
-            } else if (wreq != null) {
-                const existingFactory = wreq.m[id];
+                patchedBy.delete(patch.plugin);
```
Vendicated, line 518: this seems wrong. the plugin might have successfully applied other patches (it was already broken previously I think?). unsure if it's worth to bother to fix this behaviour

src/webpack/patchWebpack.ts
Nuckyz: yeah, I tested a lot a while ago, and this also prevents issues of removing new lines which mess up strings in some parts

scripts/generateReport.ts
```diff
@@ -26,7 +26,7 @@
 import { readFileSync } from "fs";
 import pup, { JSHandle } from "puppeteer-core";
 
-for (const variable of ["DISCORD_TOKEN", "CHROMIUM_BIN"]) {
+for (const variable of ["CHROMIUM_BIN"]) {
```
Vendicated, line 29: should also update the reporter workflow file to remove references to this secret

scripts/generateReport.ts
```diff
@@ -292,7 +292,7 @@ page.on("error", e => console.error("[Error]", e.message));
 page.on("pageerror", e => {
     if (e.message.includes("Sentry successfully disabled")) return;
 
-    if (!e.message.startsWith("Object") && !e.message.includes("Cannot find module")) {
+    if (!e.message.startsWith("Object") && !e.message.includes("Cannot find module") && !/^.{1,2}$/.test(e.message)) {
```
Vendicated, line 295: what's this for? would be good to add a short comment explaining why we ignore this so we can understand & know if we still need this in the future

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,579 @@
...
-                    if (newCode === code) {
-                        if (!patch.noWarn) {
-                            logger.warn(`Patch by ${patch.plugin} had no effect (Module id is ${id}): ${replacement.match}`);
-                            if (IS_DEV) {
-                                logger.debug("Function Source:\n", code);
-                            }
-                        }
+        const buildNumber = getBuildNumber();
+        if (
+            !Settings.eagerPatches &&
```
Vendicated, line 446: we can do a bit less checks by also caching this:
```js
const shouldCheckBuildId = !Settings.eagerPatches && buildNumber !== -1;
if (shouldCheckBuildId && (...))
```

scripts/generateReport.ts
Nuckyz: I don't remember LOL

src/webpack/patchWebpack.ts
```diff
@@ -1,353 +1,584 @@
...
+            if (
+                shouldCheckBuildId &&
+                (replacement.fromBuild != null && buildNumber < replacement.fromBuild) ||
+                (replacement.toBuild != null && buildNumber > replacement.toBuild)
+            ) {
+                continue;
+            }
 
-                    patchedBy.delete(patch.plugin);
+            if (patch.plugin === "Vesktop") {
```
Vendicated, line 478:
```suggestion
			// TODO: remove once Vesktop has been updated to use addPatch
            if (patch.plugin === "Vesktop") {
```

PR #3158, opened 2025-01-22
OverrideForumDefaults: Add sorting by unread

lolsuffocate:
# SortForumsByUnread

Allows sorting forum posts so that unread posts are at the top

Image


Vendicated:
maybe merge with OverrideForumDefaults would be reasonable

lolsuffocate:
Sure, all done

PR #3162, opened 2025-01-24
Use much stricter, whitelist based CSP

Vendicated:


Vendicated:
No sane theme (theme listed on the BD list) should be affected by this

If anyone wants to test this with popular themes, do so

But I say we should just merge this and see if there are many complaints. If there are, we can whitelist some more domains as necessary

Nuckyz (approved):

colleirose:
Could you whitelist the DuckDuckGo image proxy? It's https://external-content.duckduckgo.com and is what I've been using for loading images in themes until recently.

edit: #3481

RequiDev:
Theme that I semi-maintain is using github.io and is now blocked. I would argue that is sane.

SomnusVeritas:
Could it be that local paths also get blocked? When I try to set the image to a file on my local machine, it doesn't work. It worked before this update.

Vendicated:
local files never worked. use our support channel if you need help

PR #3163, opened 2025-01-25
New plugin: ViewMembersWithRole

Ryfters:
Allows to view members with a role, will show up to 100 members per role + the cached members


Scyye:
What's the use case?

Estereg:
> What's the use case?
I would use it to view the Whitelist, in nfts projects for example

Estereg:
> What's the use case?

I would use it to view the Whitelist, in nfts projects for example

Vendicated:
Stock Discord feature now

PR #3169, opened 2025-01-27
ircColors: fix ColorDot/RoleColorsEverywhere/maybe smth else

henmalib:
image

Basically instead of modifying just the username, now it's changing color inside of discord's helper func

And this function is used by RoleColorsEverywhere, so it's a win-win situation
https://github.com/Vendicated/Vencord/blob/main/src/plugins/roleColorEverywhere/index.tsx#L28


jedenastka:
I don't have time to test this right now, but wouldn't this break the member list toggle?

Maybe this is desirable. That said, in that case the toggle has to be removed.

henmalib:
> I don't have time to test this right now, but wouldn't this break the member list toggle?
> 
> Maybe this is desirable. That said, in that case the toggle has to be removed.

Setting on:
image

Setting off:
image

i haven't touched your patch for member list, so yeah... it works as expected

src/plugins/typingTweaks/index.tsx
```diff
@@ -57,6 +57,15 @@ interface Props {
     guildId: string;
 }
 
+const getUserColor = (userId: string, guildId: string) => {
+    if (Vencord.Plugins.isPluginEnabled("IrcColors")) {
+        // @ts-ignore
+        const color = Vencord.Plugins.plugins.IrcColors?.resolveUsedColor?.(userId);
```
Vendicated, line 63: you should just import the plugin instead
```js
import IrcColors from "../ircColors";

IrcColors.resolveUserColor(userId)
```

src/plugins/ircColors/index.ts
```diff
@@ -79,31 +104,28 @@ export default definePlugin({
...
-        }
+        return id ? this.resolveUsedColor(id) : undefined;
 
-        return (!settings.store.applyColorOnlyToUsersWithoutColor || !colorString)
-            ? color
-            : colorString;
+    },
+    calculateNameColorForListContext(context: any) {
+        const id = context?.user?.id;
+        return id ? this.resolveUsedColor(id) : undefined;
```
Vendicated, line 129:
```suggestion
        return id && this.resolveUsedColor(id);
```

src/plugins/ircColors/index.ts
```diff
@@ -79,31 +104,28 @@ export default definePlugin({
             predicate: () => settings.store.memberListColors
         }
     ],
+    // Propped to be used in TypingTweaks/RoleColorsEverywhere
+    calculateHSLforId,
+    resolveUsedColor(userId: string) {
```
Vendicated, line 109:
```suggestion
    resolveUserColor(userId: string) {
```

src/plugins/typingTweaks/index.tsx
```diff
@@ -47,41 +45,49 @@ const settings = definePluginSettings({
...
 
-interface TypingUserProps {
+interface Props {
     user: User;
     guildId: string;
 }
 
-const TypingUser = ErrorBoundary.wrap(function TypingUser({ user, guildId }: TypingUserProps) {
+const getUserColor = (userId: string, guildId: string) => {
+    if (Vencord.Plugins.isPluginEnabled("IrcColors")) {
```
Vendicated:
```suggestion
    if (Vencord.Plugins.isPluginEnabled(IrcColors.name)) {
```

src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -145,23 +148,27 @@ export default definePlugin({
...
             replacement: {
                 match: /(?<=isUnsupported\]:(\i)\.isUnsupported\}\),)(?=children:\[)/,
                 replace: "style:$self.useMessageColorsStyle($1),"
             },
             predicate: () => settings.store.colorChatMessages
         }
     ],
-
     getColorString(userId: string, channelOrGuildId: string) {
+        if (Vencord.Plugins.isPluginEnabled("IrcColors")) {
```
Vendicated:
```suggestion
        if (Vencord.Plugins.isPluginEnabled(IrcColors.name)) {
```

src/plugins/ircColors/index.ts
```diff
@@ -59,83 +76,47 @@ const settings = definePluginSettings({
...
+                match: /(?<=\.MEMBER_LIST}\),\[\]\),)(.+?color:)null!=.{0,50}?(?=,)/,
+                replace: (_, rest) => `ircColor=$self.calculateNameColorForListContext(arguments[0]),${rest}ircColor`
             },
             predicate: () => settings.store.memberListColors
         }
     ],
+    // Propped to be used in TypingTweaks/RoleColorsEverywhere
+    calculateHSLforId,
+    resolveUserColor(userId: string) {
+        const channelId = SelectedChannelStore.getChannelId();
```
Vendicated: this needs a try catch to avoid crashes if things break

PR #2161, opened 2024-02-10
feat(MoreUserTags): Role Colors for User Tag

OIRNOIR:
This feature is inspired by https://github.com/Puyodead1/replugged-stafftags
It took me 3 hours to figure out how to do this in Vencord and I'm so glad I did because it looks just so much better
Basically it allows setting a toggle in settings that makes staff tags have the same background as the member's role, falling back to the default colors it was before.

A screenshot of the concept. The first image was a screenshot of replugged-stafftags and the second was a screenshot of the then-vencord plugin. This is a screenshot of a screenshot, so you can see my modification in action near the top.

Screenshot 2024-02-10 at 1 06 23 AM

Here are some more screenshots, demonstrating that it can support both dark and light role colors. It also works in the member list and on profiles.

Screenshot 2024-02-10 at 1 06 45 AM
Screenshot 2024-02-10 at 1 07 30 AM

This is my first contribution to Vencord, so let me know if I should make any changes!

Note: The patches may seem over-engineered, but as the member object is not accessible within the section where it creates the bot tag, the colors need to be passed from the place it exists to where it needs to go. The extra patches are designed to do that.

PR #3170, opened 2025-01-27
moreUserTags: attempt to rewrite

henmalib:
Reason:
image

While discord hasn't pushed to stable i'm trying to rewrite patches, so they work both on canary and stable

Currently I've got tags working in messages:
image

### DIFF
--- a/src/plugins/moreUserTags/index.tsx
+++ b/src/plugins/moreUserTags/index.tsx
@@ -1,0 +1,147 @@
+import definePlugin from "@utils/types";
+import { Devs } from "@utils/constants";
+import { settings } from './settings';
+import { computePermissions, isWebhook, Tag, tags } from "./consts";
+import { getIntlMessage } from "@utils/discord";
+import { ChannelStore, GuildStore, PermissionsBits } from "@webpack/common";
+import { Channel, Message, User } from "discord-types/general";
+import { tag } from "type-fest/source/opaque";
+
+
+export default definePlugin({
+    name: "MoreUserTags",
+    description: "Adds tags for webhooks and moderative roles (owner, admin, etc.)",
+    authors: [Devs.Cyn, Devs.TheSun, Devs.RyanCaoDev, Devs.LordElias, Devs.AutumnVN],
+    settings,
+    patches: [
+        // Add tags
+        {
+            find: '.ORIGINAL_POSTER=',
+            replacement: {
+                match: /function\(\i\){(?=.{1,30}.BOT)/,
+                replace: "$&$self.genTagTypes(arguments[0]);"
+            }
+        },
+        // Render Tags in messages
+        {
+            find: '.Types.ORIGINAL_POSTER',
+            replacement: {
+                match: /(?<=let (\i).{1,200}children:\i}=\i;.{0,100}.isSystemDM.{0,350}),null==\i\)(?=.{1,30}?null:)/,
+                replace: ',($1=$self.getTag({...arguments[0],location:"chat",origType:$1}))$&'
+            }
+        },
+        {
+            find: '.STAFF_ONLY_DM:',
+            replacement: {
+                match: /(?<=type:(\i).{10,1000}.REMIX.{10,100})default:(\i)=/,
+                replace: "default:$2=$self.tagObj[$1];"
+            }
+        }
+    ],
+    tagObj: {},
+
+    genTagTypes(obj) {
+        let i = 100;
+
+        for (const { name } of tags) {
+            obj[name] = ++i;
+            obj[i] = name;
+            obj[`${name}-BOT`] = ++i;
+            obj[i] = `${name}-BOT`;
+            obj[`${name}-OP`] = ++i;
+            obj[i] = `${name}-OP`;
+        }
+
+        this.tagObj = obj;
+    },
+
+    getTagText(passedTagName: string, originalText: string) {
+        try {
+            if (!passedTagName) return getIntlMessage("APP_TAG");
+            const [tagName, variant] = passedTagName.split("-");
+
+            const tag = tags.find(({ name }) => tagName === name);
+            if (!tag) return getIntlMessage("APP_TAG");
+
+            if (variant === "BOT" && tagName !== "WEBHOOK" && this.settings.store.dontShowForBots) return getIntlMessage("APP_TAG");
+
+            const tagText = settings.store.tagSettings?.[tag.name]?.text || tag.displayName;
+            switch (variant) {
+                case "OP":
+                    return `${getIntlMessage("BOT_TAG_FORUM_ORIGINAL_POSTER")} • ${tagText}`;
+                case "BOT":
+                    return `${getIntlMessage("APP_TAG")} • ${tagText}`;
+                default:
+                    return tagText;
+            }
+        } catch {
+            return originalText;
+        }
+    },
+
+    getTag({
+        message, user, channelId, origType, location, channel
+    }: {
+        message?: Message,
+        user: User & { isClyde(): boolean; },
+        channel?: Channel & { isForumPost(): boolean; isMediaPost(): boolean; },
+        channelId?: string;
+        origType?: number;
+        location: "chat" | "not-chat";
+    }): number | null {
+        if (!user)
+            return null;
+        if (location === "chat" && user.id === "1")
+            return Tag.Types.OFFICIAL;
+        if (user.isClyde())
+            return Tag.Types.AI;
+
+        let type = typeof origType === "number" ? origType : null;
+
+        channel ??= ChannelStore.getChannel(channelId!) as any;
+        if (!channel) return type;
+
+        const settings = this.settings.store;
+        const perms = this.getPermissions(user, channel);
+
+        for (const tag of tags) {
+            if (location === "chat" && !settings.tagSettings[tag.name].showInChat) continue;
+            if (location === "not-chat" && !settings.tagSettings[tag.name].showInNotChat) continue;
+
+            // If the owner tag is disabled, and the user is the owner of the guild,
+            // avoid adding other tags because the owner will always match the condition for them
+            if (
+                tag.name !== "OWNER" &&
+                GuildStore.getGuild(channel?.guild_id)?.ownerId === user.id &&
+                (location === "chat" && !settings.tagSettings.OWNER.showInChat) ||
+                (location === "not-chat" && !settings.tagSettings.OWNER.showInNotChat)
+            ) continue;
+
+            if (
+                tag.permissions?.some(perm => perms.includes(perm)) ||
+                (tag.condition?.(message!, user, channel))
+            ) {
+                if ((channel.isForumPost() || channel.isMediaPost()) && channel.ownerId === user.id)
+                    type = Tag.Types[`${tag.name}-OP`];
+                else if (user.bot && !isWebhook(message!, user) && !settings.dontShowBotTag)
+                    type = Tag.Types[`${tag.name}-BOT`];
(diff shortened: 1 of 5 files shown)


Vendicated:
by rewrite I meant make patches less dependant on one another. for example, implement this without adding new members to Discord's enums

Vendicated:
right now, the plugin targets so many different modules and if even one single patch breaks, everything else stops working and either causes crashes or makes everyone have APP tag or stuff like that

more robust means that the plugin either works or fully breaks (silently), not constantly this in between

henmalib:
> by rewrite I meant make patches less dependant on one another. for example, implement this without adding new members to Discord's enums

Well, I'll look what i can do about it
Currently I'm saving `tagObj` to use it later without dependency, but without a patch for creating tags discord doesn't resolve it as a number ( maybe i just too eepy rn, will look later how to avoid it )


> right now, the plugin targets so many different modules and if even one single patch breaks, everything else stops working and either causes crashes or makes everyone have APP tag or stuff like that
> 
> more robust means that the plugin either works or fully breaks (silently), not constantly this in between

That's how i see it:
* fewer patches
* patches are easier to read/modify
* patches shouldn't depend on one another
* patches shouldn't depend on things that might change easily (like current one for messages, I've already tried to make it better than before, but yeah, it needs improvement )

henmalib:
Got it working without adding/changing discord's tags object, gonna eep now

image

Vendicated:
nini

src/plugins/moreUserTags/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+export default definePlugin({
+    name: "MoreUserTags",
+    description: "Adds tags for webhooks and moderative roles (owner, admin, etc.)",
+    authors: [Devs.Cyn, Devs.TheSun, Devs.RyanCaoDev, Devs.LordElias, Devs.AutumnVN, Devs.hen],
+    settings,
+    patches: [
+        // Render Tags in messages
+        // Maybe there is a better way to catch this horror
+        {
+            find: ".Types.ORIGINAL_POSTER",
```
Nuckyz, line 26: This find uses a non mangled key, it should be better to change it

src/plugins/moreUserTags/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+            replacement: {
+                match: /(?<=type:(\i).{10,1000}.REMIX.{10,100})default:(\i)=/,
+                replace: "default:$2=$self.getTagText($self.tagObj[$1]);"
+            }
+        }
+    ],
+    tagObj: {},
+    startAt: StartAt.Init,
+    start() {
+        this.genTagTypes();
```
Nuckyz, line 44: tagObj should just be a iife

Nuckyz (changes requested):

henmalib:
While reading through the deleted file I've got a question:

Should i support data attributes for theming? If I'm not wrong you can just grab content in css for same results

Tags inside of profile were reliant on 2-3 patches, I'll try to reduce, but if I wouldn't find a way to do this: should i just drop support for it? Seeing tags from member list/message isn't that hard

henmalib:
Okay, so OP tags are working out of the box, i have no idea why original one had a patch with a comment "show OP tags correctly"
image

Also I've added member list
image

I've left a comment how it currently patches discord code and I'm not sure if it's gonna be safe enough ( kinda feel weird because of the assignment inside parenthesis).
If someone can give a feedback about current state of patches it will be appreciated

henmalib:
Tags in profile are also now here
image

It doesn't work with bots and requires 2 patches... I'm not happy about it

short explanation why:
Render DOM looks like:
```
server_profile(user,channelId):
	top_part(user)
``` 
Username, pronouns, tags are rendered inside `top_part`  and for tags we need channelId, so the only way i know is prop it down from `server_profile`

It works and if patch with `server_profile` dies or doesn't exists at all - we will just get 0 tags in profiles ( bot tags will be present ).

Original plugin has 4 patches for profiles, including bot ones, so idk how to proceed with it

2 patches with "safe fail" seems reasonable, but still not perfect
adding 3rd patch just for tags for bots seems a lot
dropping profile support altogether - easier to maintain, lose some functionality ( are many people need tags there? )

src/plugins/moreUserTags/types.ts
```diff
@@ -0,0 +1,34 @@
...
+    showInNotChat: boolean;
+}
+export interface TagSettings {
+    WEBHOOK: TagSetting,
+    OWNER: TagSetting,
+    ADMINISTRATOR: TagSetting,
+    MODERATOR_STAFF: TagSetting,
+    MODERATOR: TagSetting,
+    VOICE_MODERATOR: TagSetting,
+    TRIAL_MODERATOR: TagSetting,
```
PlavorSeol, line 32:
```suggestion
    CHAT_MODERATOR: TagSetting,
```

henmalib: Lmao, it was like this in the original plugin 

Thanks for noticing

henmalib:
Using MessageDecoratior API i can get almost the same results without using any patches ( thanks theCodeToad )

image

But we are losing tags in replies like here
image

And bots are... ugh, should find a way to remove default one
image


ngl, i feel like a Davin doing all these comments

henmalib:
lange rede kurzer sinn

It looks like i can't remove default tag without another patch, so what's the point of decorations if it still requires patching?

Also MessageDecorationProps.props.decorations seems to be undefined

OIRNOIR:
Will there be an option to set tag colors as the guild member color (similar to #2161)? I personally think user tags are informative but very ugly when members' tags use the bot colors because it makes it harder to differentiate between priviliged members and bots at first glance.

henmalib:
> Will there be an option to set tag colors as the guild member color (similar to #2161)? I personally think user tags are informative but very ugly when members' tags use the bot colors because it makes it harder to differentiate between priviliged members and bots at first glance.

Your plugins add a bunch of new patches which caused this plugin to get deleted

I might try to recreate it using decorations, then they will be a lot more useful

henmalib:
I would love to hear answers to the following questions by someone (Vee or Nuckyz):
* Should data attributes be supported like in an original plugin?
* Tags inside of a profile require 3 patches ( for both bots and users ), which are going to be dependent on one another. Should I support it? If one of the patches breaks, default tag will be displayed ( none for users, app for bots )
* If adding a plugin idea from @OIRNOIR, it's probably to use  DecorationsAPI, but will require a patch to remove default tags. What seems more reasonable: changing tags directly or removing them?

TheKodeToad:
> And bots are... ugh, should find a way to remove default one image

My solution was just not to try and merge the bot and custom tag. I always found separating them with • to be somewhat weird - and this works just fine IMO:
image

henmalib:
> My solution was just not to try and merge the bot and custom tag. I always found separating them with • to be somewhat weird - and this works just fine IMO:

That might be a nice solution, but wouldn't it look weird with [APP] [WEBHOOK]?

> Of course you can't add everything with decorators though
Well, the only thing i haven't seen an API for was for user profiles, messages/member list are quite easy

I'll probably change to decorations again tomorrow, try to implement a pr from OIRNIOR and will use splitting like you do

lolsuffocate:
> Well, the only thing i haven't seen an API for was for user profiles, messages/member list are quite easy

I believe Nuckyz just PR'd an API for this https://github.com/Vendicated/Vencord/pull/3197

henmalib:
> > Well, the only thing i haven't seen an API for was for user profiles, messages/member list are quite easy
> 
> I believe Nuckyz just PR'd an API for this #3197

Well, untill it gets merged there's no point of relying on it 
But thanks for pointing this out. I'm pretty sure It will get merged soon, so yeah 
Decorations might be the way to go

TheKodeToad:
> That might be a nice solution, but wouldn't it look weird with [APP] [WEBHOOK]?

True, I just didn't implement webhook tags. That could become a separate plugin again or just have its own handling... because surely just changing a message is unlikely to break that much

Nuckyz:
I've changed the decorators API a bit, so now if you want to use them you no longer need to manually add margin or align vertically in messages. If you want those changes just merge dev into your branch

For the NicknameIconsAPI, I will work a bit more in improving PlatformIndicators and then I plan to merge that into dev as well, shouldnt take much

henmalib:
Yop, decorations without trying to override original ones are good
The only thing that seems to be weird for me is that [ADMIN] comes before [APP], but it messages it goes in another direction

image
image

The only thing left is profiles, which currently requires 3 patches
( 1st is to make discord use our tags, 2nd to prop guild/channel to user profile, 3rd is to actually display it )

And it seems that https://github.com/Vendicated/Vencord/pull/3197 has only userId, so... there is a choice
3 patches to have tags in 1 place or just drop user profile support. I don't see other solutions now

TheKodeToad:
> 2nd to prop guild/channel to user profile
Just use SelectedChannelStore - it shouldn't be possible to open a profile for another channel I don't believe

henmalib:
> Just use SelectedChannelStore - it shouldn't be possible to open a profile for another channel I don't believe 

Fire, I didn't knew that this existed or totally forgot about it. Now i should just wait till new API drops and this plugin going to be ~~gluten~~ patch free

>Edit: just remembered the new modal sometimes shows guild profile and sometimes global profile. I suppose that could cause problems.

Eehh, it's still a draft and we have time to see how it behaves, but SelectedStoreChannel should still have the same channel?

henmalib:
Also profiles don't have any tags by default, so [APP] tags are missing there
image

I might just render 2 tags at the same time to stay consistent with every other field, but it is worth it?

Nuckyz:
> Also profiles don't have any tags by default, so [APP] tags are missing there image
> 
> I might just render 2 tags at the same time to stay consistent with every other field, but it is worth it?

It does for me
image

Nuckyz:
Also please merge dev into this branch! Both APIs you used had changes as I mentioned, and I'm pretty sure your tags are going to be off-center because of the classes you are adding, and the fact the APIs do the centering for you now

henmalib:
> It does for me

I forgot that i'm currently replacing the old one

> Also please merge dev into this branch! Both APIs you used had changes as I mentioned, and I'm pretty sure your tags are going to be off-center because of the classes you are adding, and the fact the APIs do the centering for you now
Messages were off-center, I've removed both classes and rem-thingy, they are still off-centered, classes that i've used were from discord itself and that's how they used it

By default now:
image
Manually adding bottom .1rem to the component:
image


Member list seems fine by default anyway
image

henmalib:
So discord usesRemSizies in messages, so should I
image

Manual padding by .1rem from the bottom is not perfect, but without it it looks really off-set

Nuckyz:
They are different sizes, you should look how discord sets the size of the ones in messages and replicate that

henmalib:
> They are different sizes, you should look how discord sets the size of the ones in messages and replicate that

That's exactly how I'm doing it now
useRemSizes sets the height to be .975rem
Without it: 16px

thelordakuma:
did bro die

henmalib:
> did bro die

I will haunt your family till the death of your grandchildre

Crazy-Dog4110:
Any updates?

henmalib:
> Any updates?

I'm still waiting for https://github.com/Vendicated/Vencord/pull/3197 and considering visual refresh it will take longer

henmalib:
Maybe it may be accepted as-is?

Patches, that this plugin uses weren't changed in more than half a year... which seems pretty stable for me

If someone still want's to see less patches, i think people can live without a tag in their profile :/

PR #3172, opened 2025-01-28
Fix: Mangled madness

sadan4:
will push more later, just better folders for now


sadan4 (commented):
fixes better settings

sadan4 (commented):
fixes viewicons

PR #3174, opened 2025-01-28
fix openinapp

sadan4:
just openinapp for now, pushing more as I go


sadan4 (commented):
havent had a chance to test showHiddenChannels or openInApp yet

PR #3177, opened 2025-01-29
Fix everything broken by recent Discord update

Nuckyz:


src/plugins/iLoveSpam/index.ts
```diff
@@ -27,7 +27,7 @@ export default definePlugin({
         {
             find: "hasFlag:{writable",
             replacement: {
-                match: /if\((\i)<=(?:1<<30|1073741824)\)return/,
+                match: /if\((\i)<=(?:0x40000000|1073741824)\)return/,
```
Vendicated, line 30: we should just have both (I added extra group so it's more clear it's the same flag)

```suggestion
                match: /if\((\i)<=(?:(?:0x40000000|1<<30)|1073741824)\)return/,
```

src/utils/modal.tsx
```diff
@@ -99,7 +99,21 @@ export const Modals = findByPropsLazy("ModalRoot", "ModalCloseButton") as {
...
+    ModalContent: filters.byCode(".content,"),
+    ModalFooter: filters.byCode(".footer,"),
+    ModalCloseButton: filters.byCode(".close]:")
+});
+
+export const ModalRoot = LazyComponent(() => Modals.ModalRoot);
+export const ModalHeader = LazyComponent(() => Modals.ModalHeader);
+export const ModalContent = LazyComponent(() => Modals.ModalContent);
+export const ModalFooter = LazyComponent(() => Modals.ModalFooter);
+export const ModalCloseButton = LazyComponent(() => Modals.ModalCloseButton);
```
Vendicated, line 116: we probably don't even need LazyComponent here anymore because lazy allows accessing properties immediately

src/webpack/common/components.ts
```diff
@@ -16,76 +16,77 @@
...
 
 import { waitForComponent } from "./internal";
 import * as t from "./types/components";
 
-export let Forms = {} as {
-    FormTitle: t.FormTitle,
-    FormSection: t.FormSection,
-    FormDivider: t.FormDivider,
-    FormText: t.FormText,
+const FormTitle = waitForComponent<t.FormTitle>("FormTitle", filters.componentByCode(".SELECTABLE),"));
```
Vendicated, line 25: seems a bit too generic

src/webpack/common/components.ts
```diff
@@ -16,76 +16,77 @@
...
+const FormText = waitForComponent<t.FormText>("FormText", filters.componentByCode('["defaultMargin".concat'));
+
+export const Forms = {
+    FormTitle,
+    FormSection,
+    FormDivider,
+    FormText
+};
+
+export const Card = waitForComponent<t.Card>("Card", filters.componentByCode(".editable),"));
```
Vendicated, line 37: seems a bit too generic

src/webpack/common/components.ts
```diff
@@ -16,76 +16,77 @@
...
+    Tooltip: filters.componentByCode("this.renderTooltip()]"),
+    TooltipContainer: filters.componentByCode('="div",')
+}) as {
+    Tooltip: t.Tooltip,
+    TooltipContainer: t.TooltipContainer;
 };
 
-export let Icons = {} as t.Icons;
+export const Tooltip = LazyComponent(() => Tooltips.Tooltip);
+export const TooltipContainer = LazyComponent(() => Tooltips.TooltipContainer);
```
Vendicated, line 50: probably don't need to wrap this in LazyComponent

src/webpack/patchWebpack.ts
```diff
@@ -172,26 +172,36 @@ function patchFactories(factories: Record<string, (module: any, exports: any, re
...
+                nonEnumerableChecking: {
+                    // There are (at the time of writing) 11 modules exporting the window,
+                    // and also modules exporting DOMTokenList, which breaks webpack finding
+                    // Make these non enumerable to improve search performance and avoid erros
+                    if (exports === window || exports[Symbol.toStringTag] === "DOMTokenList") {
+                        shouldMakeNonEnumerable = true;
+                        break nonEnumerableChecking;
+                    }
+
+                    if (typeof exports !== "object") {
```
Vendicated, line 187: just in case (typeof null is object)

```suggestion
                    if (typeof exports !== "object" || exports === null) {
```

src/webpack/patchWebpack.ts
```diff
@@ -172,26 +172,36 @@ function patchFactories(factories: Record<string, (module: any, exports: any, re
...
-                            if (exports[nested] === window) {
-                                foundWindow = true;
-                            }
+                let shouldMakeNonEnumerable = false;
+
+                nonEnumerableChecking: {
+                    // There are (at the time of writing) 11 modules exporting the window,
+                    // and also modules exporting DOMTokenList, which breaks webpack finding
+                    // Make these non enumerable to improve search performance and avoid erros
+                    if (exports === window || exports[Symbol.toStringTag] === "DOMTokenList") {
```
Vendicated, line 182:
```suggestion
                    if (exports === window || exports?.[Symbol.toStringTag] === "DOMTokenList") {
```

src/webpack/patchWebpack.ts
```diff
@@ -172,26 +172,36 @@ function patchFactories(factories: Record<string, (module: any, exports: any, re
...
+                    }
+
+                    if (typeof exports !== "object") {
+                        break nonEnumerableChecking;
+                    }
+
+                    if (exports.default === window || exports.default?.[Symbol.toStringTag] === "DOMTokenList") {
+                        shouldMakeNonEnumerable = true;
+                        break nonEnumerableChecking;
+                    }
```
Vendicated, line 194: should be able to just remove this since it'll be covered by the nested loop (not to mention that default is .Z and .ZP, no?)

src/plugins/_api/contextMenu.ts
```diff
@@ -34,12 +34,21 @@ export default definePlugin({
...
-            replacement: {
-                match: /Menu,{(?<=\.jsxs?\)\(\i\.Menu,{)/g,
-                replace: "$&contextMenuApiArguments:typeof arguments!=='undefined'?arguments:[],"
-            }
+            noWarn: true,
+            replacement: [
+                {
+                    match: /navId:(?=.+?([,}].*?\)))/g,
+                    replace: (m, rest) => {
+                        const destructuringMatch = rest.match(/}=.+/);
```
Vendicated, line 44: you should add a comment explaining what this (and the match) does so it's less confusing

src/webpack/common/components.ts
```diff
@@ -16,76 +16,83 @@
...
+export const Popout = waitForComponent<t.Popout>("Popout", filters.componentByCode("ref:this.ref,preload:"));
+export const Dialog = waitForComponent<t.Dialog>("Dialog", filters.componentByCode('role:"dialog",tabIndex:-1'));
+export const TabBar = waitForComponent("TabBar", filters.componentByCode("ref:this.tabBarRef,className:"));
+export const Paginator = waitForComponent<t.Paginator>("Paginator", filters.componentByCode('rel:"prev",children:'));
+export const Clickable = waitForComponent<t.Clickable>("Clickable", filters.componentByCode("this.context?this.renderNonInteractive():"));
+export const Avatar = waitForComponent<t.Avatar>("Avatar", filters.componentByCode(".size-1.375*"));
+
+export let createScroller: (scrollbarClassName: string, fadeClassName: string) => t.ScrollerThin;
+export let scrollerClasses: Record<string, string>;
+waitFor(filters.byCode('="ltr",orientation:', "onScroll:"), m => createScroller = m);
```
Vendicated, line 68:
```suggestion
waitFor(filters.byCode("forwardRef", '="ltr",orientation:', "onScroll:"), m => createScroller = m);
```

Vendicated (approved):

PR #3179, opened 2025-01-29
CopyEmojiMarkdown: Fix copying animated emojis

Tomsoz:
Discord animated emojis no longer have the extension .gif. They are now .webp with the query parameter animated=true passed to them. I've updated the copyEmojiMarkdown plugin to reflect this as it wasn't copying animated emojis correctly.


Nuckyz (changes requested):
Instead of still checking for the extension this should just check if url.pathname ends with "gif". Take a look at isGifUrl from EmoteCloner

Tomsoz125:
> Instead of still checking for the extension this should just check if url.pathname ends with "gif". Take a look at isGifUrl from EmoteCloner

I checked the contents of url.path and it was **/emojis/996768598938882189.webp**, in the function you mentioned it checks if it ends with .gif but it checks the path as well so there should be no issue with the method
image

Tomsoz125:
Okay, I've made those changes now

Tomsoz:
@Nuckyz 
Any ETA on this getting merged some friends keep complaining that it doesn't copy right.
No rush just wondering?

src/plugins/copyEmojiMarkdown/index.tsx
```diff
@@ -33,11 +33,12 @@ function getEmojiMarkdown(target: Target, copyUnicode: boolean): string {
             : `:${emojiName}:`;
     }
 
-    const extension = target?.firstChild.src.match(
-        /https:\/\/cdn\.discordapp\.com\/emojis\/\d+\.(\w+)/
-    )?.[1];
+    const imgSrc = target?.firstChild.src;
+    const url = new URL(imgSrc);
```
Nuckyz, line 37:
```suggestion
    const url = new URL(target.firstChild.src);
```

Nuckyz (changes requested):

PR #3192, opened 2025-01-31
HideMedia: Add support for automatic hiding of media via defining list of user ids or domains

fumiichan:
This patch includes support to hide attachments with the following configuration:

- Comma separated list of User IDs
- Comma separated list of domains/hostnames
- A toggle to enable/disable the automatic hiding

This patch also includes fixes for forwarded messages with embeds only is not detected.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
 
             - name: Clean up obsolete files
               run: |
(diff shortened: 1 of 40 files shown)


cwireman (commented):
can the user/domain lists have a whitelist/blacklist mode when auto-hide is on? there's definitely [interest in hiding all by default](https://github.com/Vencord/plugin-requests/issues/163).

PR #3195, opened 2025-01-31
New plugin: Mullvad Rich Presence

alessandromrc:
This plugins allows users to show their connection status to Mullvad VPN and lets them show or hide the server name they are connected to so that their privacy is kept.


alessandromrc:
Any reason about why the PR got pretty much instantly closed?

Vendicated:
why in the world would anyone want this 😭

PR #2657, opened 2024-07-02
feat(plugin): EditUsers

Vendicated:
Todo:
- [ ] make nickname replacement work everywhere. need to manually patch messages & member list
- [ ] idk if asking for image urls is a good idea. the fact that cdn links expire made this kinda painful
- [ ] allow editing more stuff:
	- [x] pronouns
	- [ ] avatar decoration
	- [ ] profile effects
- [ ] add user tags (like the APP tag)

PR #3196, opened 2025-02-01
New Plugin:  HideProfilePictures

J3sven:
# HideProfilePictures

This plugin allows the user to hide/unhide the profile pictures of particular users via a new context menu option. It keeps track of users for whom the pictures were hidden so that they remain hidden between sessions, if the users wants to revert the action they simply rightclick on the user again and select 'unhide profile picture'.

By default a hidden user's profile picture is reverted to the default discord avatar image. Optionally, the user can select the "Blur Profile Pictures" setting to make the profile pictures get blurred out instead.

hidepfp


Vendicated:
thank you for your contribution! unfortunately, this contribution doesn't follow the plugin guidelines (namely, using patches instead of raw dom)

and this is better solved by #2657

PR #3198, opened 2025-02-01
Fix lag caused by poorly written CSS rules

Vendicated:


Vendicated:
TODO: fix various plugins that have self-managed css

src/plugins/spotifyControls/spotifyStyles.css
```diff
@@ -30,22 +30,17 @@
...
-.vc-spotify-button svg {
+.vc-spotify-button-icon {
     height: 24px;
     width: 24px;
 }
 
-[class*="vc-spotify-shuffle"] > svg,
-[class*="vc-spotify-repeat"] > svg {
+[class*="vc-spotify-shuffle"] .vc-spotify-button-icon,
+[class*="vc-spotify-repeat"] .vc-spotify-button-icon {
```
Nuckyz, line 39: Why are these wildcards?

Nuckyz (changes requested):

src/plugins/spotifyControls/spotifyStyles.css
Vendicated: cause there are a lot of different states like vc-spotify-shuffle-on/off

Sqaaakoi: i propose using the aria-checked attribute and selecting based upon that
don't nuke the shuffle state className though

PR #3199, opened 2025-02-02
New Plugin: RoleMembersViewer

okisooo:
This plugin lets you quickly view role members through a submenu in two contexts:

### Dev Context Menu
Dev Context Screenshot

### Message Context Menu
Message Context Screenshot

**Key Technical Points:**

- **Chunked Fetching:**  
  Uses FluxDispatcher to dispatch multiple `GUILD_MEMBERS_REQUEST` events, ensuring even offline members are picked up in large servers.

- **Efficient Filtering & Sorting:**  
  Compares current `GuildMemberStore` count to the expected `memberCount` to trigger additional fetches only when needed. Retrieved members are then filtered by role and sorted alphabetically.

- **Context-Specific Integration:**  
  - **Dev Context:** Immediately displays a submenu when right-clicking a role in a user profile—ideal for audits and troubleshooting.  
  - **Message Context:** Deduplicates role mentions to generate a clean, consolidated submenu for quick checks.

- **Live Updates:**  
  Uses `GuildMemberStore`'s dynamic updates to reflect real-time changes as members come online or offline.
  
I would like to add that I've seen a similar plugin as a pull request, however since my plugin is different in functionality and we were pretty much working on it at the same time, I thought it'd be more appropriate to make my own pull request, but I have no problem in collaborating with the other plugin author if necessary!


wheredoweallgo:
this could help out a lot when it comes to massive discord servers esp for moderation and the amount of roles they have

VillainsRule:
If it's a 50,000 member server, does it show every single person? How many dispatcher calls is this? Does this qualify as API spam?

okisooo:
> If it's a 50,000 member server, does it show every single person? How many dispatcher calls is this? Does this qualify as API spam?

It works in a similar way as the "MemberCount" plugin, where it sends a request when needed, in this case it sends a dispatch call in chunks so we can have the same effect for servers of all sizes. It's effectively as if you were loading the member-base of a server by scrolling down indefinitely on the members bar on the right, only it's much more efficient this way. It wouldn't qualify as API spam and it does indeed show each person.

Vendicated:
Stock feature now!!

PR #3200, opened 2025-02-02
New Plugin: NoJoinPermission

Pixeluted:
This plugins automatically detects the 'Join Servers For You' scope in all authorization links you click in your desktop client and strips it out. Effectively not granting that scope to the application.

This plugin can break some apps if they check the scopes you granted them, but that can be easily fixable by just disabling the plugin for little while.

Here is little showcase how it works:
https://github.com/user-attachments/assets/b21965f1-4e9d-4e1a-b1ef-bd413f0c52ef

### DIFF
--- a/src/plugins/noJoinPermission/index.ts
+++ b/src/plugins/noJoinPermission/index.ts
@@ -15,8 +15,8 @@ export default definePlugin({
         {
             find: "No integration type was selected.",
             replacement: {
-                match: /(.includes\(\i\)\);)(return{requestedScopes:\i,accountScopes:\i})/,
-                replace: "$1$self.handleScopes(t, n);$2"
+                match: /(.includes\(\i\)\);)(return{requestedScopes:(\i),accountScopes:(\i)})/,
+                replace: "$1$self.handleScopes($3, $4);$2"
             }
         }
     ],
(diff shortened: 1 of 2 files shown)


src/plugins/noJoinPermission/index.ts
```diff
@@ -0,0 +1,39 @@
...
+export default definePlugin({
+    name: "NoJoinPermission",
+    description: "Removes the annoying 'Join Servers For You' scope from all authorization links (May cause some apps to break)",
+    authors: [Devs.Pixeluted],
+    patches: [
+        {
+            find: "No integration type was selected.",
+            replacement: {
+                match: /(.includes\(\i\)\);)(return{requestedScopes:\i,accountScopes:\i})/,
+                replace: "$1$self.handleScopes(t, n);$2"
```
EepyElvyra, line 19: capture t and n instead of using them directly, they may change

Pixeluted: oh good idea, I completely forgot you can do this

Pixeluted: Alright this was implemented

Pixeluted:
Any opinion on this plugin / chances of it getting approved?
Thanks for any response

Vendicated:
this is way too niche and not really that useful. thanks for your contribution though!

PR #3202, opened 2025-02-03
new plugin AddAttachments: add new files to existing messages

lumap:
Adds a button to add attachments to an existing message of yours

image


src/plugins/addAttachments/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+            async function uploadLoop(file: File) {
+                const attachmentsReq = (await Common.RestAPI.post({
+                    url: `/channels/${channelId}/attachments`,
+                    body: {
+                        files: [
+                            {
+                                filename: file.name,
+                                file_size: file.size,
+                                id: Math.round(Math.random() * 99999999), // apparently this doesn't matter that much ??
```
jamesbt365, line 63: Okay this is just silly... This is the index of the file, you don't need to input a random number just insert the length of the current attachments (that'll be the right number)

lumap: I wish it was, unfortunately it's not. In the client, it's a call to a uniqueId() function that's basically just an incremental counter, and I have no idea how to access/call it with vencord, so random number it is

lumap: nvm figured it out

src/plugins/addAttachments/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+            const uploadPromises = Array.from(input.files).map(async file => {
+                const attachmentsReq = (await Common.RestAPI.post({
+                    url: `/channels/${channelId}/attachments`,
+                    body: {
+                        files: [
+                            {
+                                filename: file.name,
+                                file_size: file.size,
+                                id: findByProps("uniqueId").uniqueId(),
```
lolsuffocate, line 60: You should move this up to the top level and use `const uniqueId = findByPropsLazy(...` instead so you're not running the find every time

src/plugins/addAttachments/index.tsx
dolfies: The index would work fine here

src/plugins/addAttachments/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+            if ((10 - messageAttachments) < input.files.length) {
+                showToast(`You can only add ${10 - messageAttachments} more attachments to this message.`, 2);
+                return input.remove();
+            }
+
+            showToast("Uploading, this can take a while...");
+
+            const uploadPromises = Array.from(input.files).map(async file => {
+                const attachmentsReq = (await Common.RestAPI.post({
```
dolfies, line 55: Why are you making multiple requests to the create attachment route when it accepts an array?

lumap: fr tho i did it cuz that's what the client does when you drag 2+ attachments at once into the chat box, one /attachments request per file

dolfies: u can be better than the discord devs

lumap: maybe..

Vendicated:
Cool but idk about the UX. I think users will struggle finding the button because the plugin doesn't tell you how it works. They'd expect to be able to add attachments when editing

Also only being able to select files from system and not being able to paste clipboard is a bummer

Vendicated:
I made this more user friendly. Now when editing a message
- There will be a button to add attachments
- You can paste files too

image

UX is still a bit bad but I think it's fine

PR #3204, opened 2025-02-05
ConsoleShortcuts: Add functions to switch branch on vesktop

sadan4:


Vendicated:
what's the use case of this? we can't use this for support because people wouldn't have this plugin enabled

src/plugins/consoleShortcuts/index.ts
```diff
@@ -35,6 +35,14 @@ const DESKTOP_ONLY = (f: string) => () => {
...
 };
 
+function makeVesktopSwitcher(branch: string): () => void {
+    return function () {
+        if (Vesktop.Settings.store.discordBranch === branch) throw new Error(`Already on ${branch}`);
+        Vesktop.Settings.store.discordBranch = branch;
+        VesktopNative.app.relaunch();
+    };
+}
+
```
Vendicated:
```suggestion
const makeVesktopSwitcher = (branch: string) => () => {
    if (Vesktop.Settings.store.discordBranch === branch)
        throw new Error(`Already on ${branch}`);

    Vesktop.Settings.store.discordBranch = branch;
    VesktopNative.app.relaunch();
}

```

PR #3205, opened 2025-02-06
feat(translate): preserve emojis and add original text option

flawless-run:
Hey team! 👋  

This PR fixes an issue where the auto-translate plugin was trying to translate emojis and adds a new option to include the original text in the translated message.

#### **What's New?**
1. **Emojis Are Safe Now**
   - The plugin no longer tries to translate emojis (which was breaking things).  
   - Emojis stay exactly where they should be in the message.  

2. **New Setting: Include Original Text**
   - Added a toggle to optionally show the original message below the translation.  
   - Perfect for when you want to double-check the translation!

#### **What Was Fixed?**
Before, if a message had emojis, the plugin would freak out and break. For example:
- **Input**: `Hello <:smile:123456789> world!`  
- **Broken Output**: The translation would fail or look weird because of the emoji.

Now, it works like a charm:
- **Input**: `Hello <:smile:123456789> world!`  
- **Fixed Output**: `Hola <:smile:123456789> mundo!`  

#### **Examples**
Here are some examples to show how it works now:
1. **Simple Message with Emoji**:
   - **Input**: `Hello <:smile:123456789> world!`  
   - **Output**: `Hola <:smile:123456789> mundo!`  

2. **Message with Multiple Emojis**:
   - **Input**: `Hello <:smile:123456789> world! <:wave:987654321> How are you?`  
   - **Output**: `Hola <:smile:123456789> mundo! <:wave:987654321> ¿Cómo estás?`  

3. **With Original Text Enabled**:
   - **Input**: `Hello <:smile:123456789> world!`  
   - **Output**:
     ```
     Hola <:smile:123456789> mundo!
     -# Hello <:smile:123456789> world!
     ```

#### **Why This Matters**
This makes the plugin way more reliable and user-friendly. No more broken messages because of emojis, and the new setting gives users more control over how translations are displayed. Win-win!

---

### **Testing**
I tested it with:
- Messages with emojis at the start, middle, and end.
- Messages with no emojis.
- The new `includeOriginalText` setting.

Everything works as expected!

### DIFF
--- a/src/plugins/translate/index.tsx
+++ b/src/plugins/translate/index.tsx
@@ -87,7 +87,32 @@ export default definePlugin({
         clearTimeout(tooltipTimeout);
         tooltipTimeout = setTimeout(() => setShouldShowTranslateEnabledTooltip?.(false), 2000);
 
-        const trans = await translate("sent", message.content);
-        message.content = trans.text;
+        const emojiRegex = /<:[a-zA-Z0-9_]+:[0-9]+>/g;
+
+        const emojis = message.content.match(emojiRegex) || [];
+        const parts = message.content.split(emojiRegex);
+
+        const translatedParts = await Promise.all(
+            parts.map(async (part) => {
+                const trimmedPart = part.trim();
+                if (!trimmedPart) return part;
+                const trans = await translate("sent", trimmedPart);
+                return trans.text;
+            })
+        );
+
+        const reconstructedMessage = translatedParts.reduce((acc, part, index) => {
+            acc += part;
+            if (emojis[index]) {
+                acc += ` ${emojis[index]} `;
+            }
+            return acc;
+        }, '');
+
+        if (settings.store.includeOriginalText) {
+            message.content = `${reconstructedMessage.trim()}\n-# ${message.content}`;
+        } else {
+            message.content = reconstructedMessage.trim();
+        }
     }
 });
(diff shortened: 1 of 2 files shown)


src/plugins/translate/index.tsx
```diff
@@ -87,7 +87,32 @@ export default definePlugin({
...
 
-        const trans = await translate("sent", message.content);
-        message.content = trans.text;
+        const emojiRegex = /<:[a-zA-Z0-9_]+:[0-9]+>/g;
+
+        const emojis = message.content.match(emojiRegex) || [];
+        const parts = message.content.split(emojiRegex);
+
+        const translatedParts = await Promise.all(
+            parts.map(async (part) => {
```
Vendicated, line 96: splitting the message into parts like this doesn't really make sense because this will mess up the grammar & context. this problem will become even more apparent for RTL languages because you're completely messing up the flow

flawless-run: you make a good point. what if we translate the entire message as a whole while preserving emojis by temporarily replacing them with placeholders instead?

Vendicated: there is a different PR https://github.com/Vendicated/Vencord/pull/3096 which attempts to fix Links sometimes being broken. the solution we ended up with was replacing all links with unique untranslatable (i.e. nothing that even remotely looks like a word) ids and then later replacing them back

Vendicated: we should just unify these PRs. We can likely base everything on 3096 and just extend its regex to also add emojis

flawless-run: yeah that makes sense. would be great to have the option included in that PR instead then.

PR #3207, opened 2025-02-07
New Plugin: Message Jumper

May2Beez:
This plugin jumps into the channels that contains new messages, if your discord is not focused.
Convenient for people, that have Discord on second monitor and still want to read all the messages.
Plugin contains whitelist and blacklist for certain servers and channels.

### DIFF
--- a/src/plugins/messageJumper/index.ts
+++ b/src/plugins/messageJumper/index.ts
@@ -1,0 +1,108 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2025 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { ChannelRouter, ChannelStore, SelectedChannelStore } from "@webpack/common";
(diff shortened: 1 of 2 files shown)


src/plugins/messageJumper/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        options: [
+            {
+                label: "Blacklist",
+                value: "blacklist",
+                default: true
+            },
+            {
+                label: "Whitelist",
+                value: "whitelist"
+            }
```
scattagain, line 35: it'd be better to use `true` and `false` as values here instead of these hardcoded strings

src/plugins/messageJumper/index.ts
```diff
@@ -0,0 +1,108 @@
...
+            const serverIds = servers.split(",").map(id => id.trim());
+            if (whitelistedServers === "whitelist") {
+                if (!serverIds.includes(channel.guild_id)) {
+                    return;
+                }
+            } else {
+                if (serverIds.includes(channel.guild_id)) {
+                    return;
+                }
+            }
```
scattagain, line 91: thought i mentioned this but turns out i didnt, but this could be simplified to
```ts
if (whitelistedServers !== serverIds.include(channel.guild_id)) return;
```
given that you made `whitelistedServers` a bool
this obv works for the channels too

PR #3210, opened 2025-02-09
Added VirusTotal, HybridAnalysis & Tria.ge plugin

nay-cat:
I would like to know possible errors and suggestions, especially to optimize the code, the plugin is already functional.

image
image
image


nin0-dev:
breaks https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md as this requires that the user uses their own API key

also just don't download binaries off discord

nay-cat:
i didn't know that (my bad, i cant fucking read) and you're right about downloading binaries, sorry for the inconvenience

PR #3212, opened 2025-02-10
Add Display Names for Settings

### LINKED ISSUE #3211: CustomRPC: AppID setting has a random space
### Content

minor issue, but in the custom rpc settings, the AppID has a random space in the middle of it

Image

### Request Agreement

lolsuffocate:
Adds display name support for settings and sets display names for some settings that require symbols or capitals not at the beginning of a word.

~~Some of the display names I've set are on string input components which fully capitalise their titles anyway but if we/Discord ever change that behaviour (or for people using themes that unset the uppercase text transform), display names will already be correct.~~
They did change them to not be uppercase.

image
image

Closes #3211


src/plugins/translate/settings.ts
```diff
@@ -63,6 +63,7 @@ export const settings = definePluginSettings({
     },
     deeplApiKey: {
         type: OptionType.STRING,
+        displayName: "DeepL API Key",
         description: "DeepL API key",
```
sadan4, line 67: this description seems a bit redundant

lolsuffocate: It is but a description is required, I'm not sure what else to put there. Maybe a link to DeepL?

nicoluvas: a link to the token page perhaps

lolsuffocate:
As part of just generally tidying up the look of settings I've also changed switch option titles to match every other type.

image

lajczi (commented):
LGTM

PR #3216, opened 2025-02-10
messageTags: Add custom params support (and listing existing tags in commands)

Benjas333:
When the message contains a parameter between $ (`$example$`) the tag command created will include the parameter as an option. The modal is in case the user wants to set default values for optional params.
tag creation
tag params configurator
use only required tag params
use all tag params

Additionally, I added listing existing tags in /tag delete and /tag preview commands:
tag delete listing
tag preview listing


Vendicated:
Thank you for your efforts, but this has been superseded by https://github.com/Vendicated/Vencord/commit/143bb60051b02e50c8536eeb2fc111f573f57dd3

PR #3217, opened 2025-02-11
WebpackPatcher: Use way less closures

Nuckyz:
Decreases the amount of closures WebpackPatcher creates heavily. Before two closures would be created for every module. One for the getter and one for the wrapper. Now none of those creates closures anymore and instead rely on a single proxy handler

### DIFF
--- a/src/webpack/patchWebpack.ts
+++ b/src/webpack/patchWebpack.ts
@@ -61,7 +61,7 @@ export function getFactoryPatchedBy(moduleId: PropertyKey, webpackRequire = wreq
 
 const logger = new Logger("WebpackInterceptor", "#8caaee");
 
-/** Whether we tried to fallback to WebpackRequire of the factory, or disabled patches */
+/** Whether we tried to fallback to the WebpackRequire of the factory, or disabled patches */
 let wreqFallbackApplied = false;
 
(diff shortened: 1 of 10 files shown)


src/webpack/patchWebpack.ts
```diff
@@ -172,57 +170,96 @@ const moduleFactoriesHandler: ProxyHandler<AnyWebpackRequire["m"]> = {
...
+    get(target, p, receiver) {
+        if (target[SYM_ORIGINAL_FACTORY] != null && (p === SYM_PATCHED_SOURCE || p === SYM_PATCHED_BY)) {
+            return Reflect.get(target[SYM_ORIGINAL_FACTORY], p, target[SYM_ORIGINAL_FACTORY]);
+        }
+
+        const v = Reflect.get(target, p, receiver);
+
+        // Make proxied factories `toString` return their original factory `toString`
+        if (p === "toString") {
+            return (...args: unknown[]) => v.apply(target[SYM_ORIGINAL_FACTORY] ?? target, args);
```
Vendicated, line 209:
```suggestion
            return v.bind(target[SYM_ORIGINAL_FACTORY] ?? target);
```

Vendicated (approved):

PR #3219, opened 2025-02-11
BetterFolders: Fix Patch

thororen1234:
they added "data-drop-hovering" in between


thororen1234:
Im blind sorry for the inconvience https://github.com/Vendicated/Vencord/pull/3214

PR #3222, opened 2025-02-13
New plugin: Keybinds

merex27:
Added a plugin that prevents the GIF picker from closing after sending a GIF if the Shift key is held.


sadan4:
I feel like this should at least be merged into gifPaste as an option in settings

Sqaaakoi:
> I feel like this should at least be merged into gifPaste as an option in settings

this is literally an edited version of gifpaste

PR #3225, opened 2025-02-13
Removable Chat Buttons Plugin

mahirox36:
This plugin lets users customize their chat bar by removing default and plugin-added buttons based on their settings. This enhances flexibility and creates a cleaner, clutter-free UI.

Removable Chat Buttons

### DIFF
--- a/src/plugins/removeChatButtons/README.md
+++ b/src/plugins/removeChatButtons/README.md
@@ -1,0 +1,7 @@
+# RemoveChatButtons
+
+**Customize your chat bar by removing default and plugin-added buttons based on your preferences.**
+
+## Installation
+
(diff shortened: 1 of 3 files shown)


jamesbt365:
Your plugin uses DOM manipulation (against our [plugin rules](https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md#plugin-rules)) and only works when they have their Discord set to English.

It will not be accepted as is so please stop submitting it, you are just spamming pull requests at this point.

EepyElvyra:
you dont have to make a new pr for each commit bro 😭

Vendicated:
please you don't need to close and open a new PR every time you make one change... that's not how git works. you just push your changes and they will automatically be added to your existing PR

and this plugin violates our plugin rules so it won't be accepted anyway

PR #3227, opened 2025-02-13
Disable devtools in QuickCSS editor

Fripe070:
It is possible to open devtools in the QuickCSS editor, which serves little purpose since it's impossible to inject anything into it, and can cause confusion in case it is accidentally focused instead of the main app.


nin0-dev:
someone smart enough to use quickcss should be smart enough to know the context of their devtools window

sadan4:
this is just a bad idea, most users, while not using devtools themselves, at least know how to close them. there is no harm in keeping them open.

anyone that could get confused by which window devtools is open in for more than a few seconds, wont be using it them the first place

PR #3230, opened 2025-02-15
FIX: Missing ampersand causing log warning

NeoNyaa:
An ampersand was missing, causing the logs to show a warning, this fixes that.
Yes it's not much of a PR but the warning was annoying me.


Vendicated:
this has nothing to do with missing an &, it's just that our bundler was outdated and didn't support css nesting yet, but that was recently addressed via https://github.com/Vendicated/Vencord/commit/5196c2adc0fe348d66dacde7e04d3247f58dc073

PR #3232, opened 2025-02-16
Fix: Incorrect css syntax with specialCard.css

gyfdb:
Fixes the `vc-special-hyperlink` and its hyperlink text effect.


Vendicated:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting

PR #3238, opened 2025-02-19
Fix FakeNitro loading

T14D3:
Might need some more changes, but seems to work for now


moxie-coder:
does this fix the issue with it not working properly when having nitro?

Nuckyz:
Closing as this fix does not actually change anything. If the length is 0 the iterator will also be empty and none of the loop code will run.

PR #3240, opened 2025-02-19
Added customVoiceFilter plugin

fox3000foxy:
This plugin is in progress, and should add the possibility to let user create their own voices from the Flamingo experiment.

image

No simple slash command plugins like /cat. Instead, make a user installable Discord bot ✅ 
No simple text replace plugins like Let me Google that for you. The TextReplace plugin can do this ✅ 
No raw DOM manipulation. Use proper patches and React ✅ 
No FakeDeafen or FakeMute ✅ 
No StereoMic ✅ 
No plugins that simply hide or redesign ui elements. This can be done with CSS ✅ 
No plugins that interact with specific Discord bots (official Discord apps like Youtube WatchTogether are okay) ✅ 
No selfbots or API spam (animated status, message pruner, auto reply, nitro snipers, etc) ✅ 
No untrusted third party APIs. Popular services like Google or GitHub are fine, but absolutely no self hosted ones ✅ 
No plugins that require the user to enter their own API key ✅ 
Do not introduce new dependencies unless absolutely necessary and warranted ✅


fox3000foxy:
The plugin is ready, apart for the voice modification itself.

fox3000foxy:
Added a comment in the modal which explains that the Voice Filters are in rollout, but that they can still create and download Voicepacks. In this case, the plugin has no reason to be in draft.

fox3000foxy:
Needs to be finished, refactoring voice engine to pass through RVC models

PR #3242, opened 2025-02-20
feat(plugin): StatsfmRPC Plugin

sitescript:
StatsFMRPC is a plugin to display what you are playing on stats.fm. (Based off of Lastfm)

400236067-16ff1e5c-5b0f-4402-b2fc-030d9c777d03
400236065-d664b5f4-2168-4798-8d2d-a283ef825aec


Vendicated:
Why would you use this over spotify status? I don't think this is very useful. If anything it should just be integrated into the last.fm plugin

Thanks for your contribution nevertheless!

PR #3243, opened 2025-02-20
feat(plugin): Added FontLoader

vMohammad24:
# FontLoader

A plugin allowing you to search and use any font from google fonts directly from your discord allowing you to preview it in the app.


Vendicated:
this can be done with css

PR #3249, opened 2025-02-22
LastFM RPC: Option to hide art/album covers

ynot01:
# What

Adds a new option to **LastFMRichPresence** plugin

"Always Hide Art"
\- "Disable downloading album art"

This is a boolean which is checked at `getLargeImage()` and skips returning the track's imageUrl, instead deferring to either the placeholder or null (lastfm logo) depending on other options.

In essence, it acts as if all album art is impossible to find

# Why

Last.fm art is uncensored. I would personally not like to flashbang whoever clicks on my profile with Death Grips' penis from the No Love Deep Web album, but that is not possible if I am using this plugin.

Spotify album art is censored, but Last.fm entries are not required to have Spotify links, so it would be unreliable. For now, a good solution is to allow the end user to disable album art entirely.


src/plugins/lastfm/index.tsx
```diff
@@ -189,6 +189,11 @@ const settings = definePluginSettings({
         description: "show the Last.fm logo by the album cover",
         type: OptionType.BOOLEAN,
         default: true,
+    },
+    alwaysHideArt: {
+        description: "Disable downloading album art",
+        type: OptionType.BOOLEAN,
+        default: false,
     }
```
Vendicated, line 197:
```suggestion
    showAlbumCover: {
        description: "Show album cover. Disabling this will display a placeholder. Useful if your Music has inappropriate art",
        type: OptionType.BOOLEAN,
        default: true,
    }
```

Vendicated (commented):
Honestly not sure what to think about this solution, can't get more nuclear lol

Your current setting is kinda confusing though

PR #3138, opened 2025-01-12
Add TextCorrector plugin

Naseem48:
### Add TextCorrector Plugin

This pull request introduces the **TextCorrector** plugin for Vencord, which enhances the messaging experience by automatically correcting grammar and spelling in user messages. 

#### Features:
- **Toggleable Correction:** Enable or disable message correction directly from the chat bar with a single click.
- **Multiple Correction Services:** Choose between OpenAI and LanguageTool APIs for corrections, offering flexibility based on user preference.
- **Preserves Formatting:** Ensures original quotation marks and text formatting are preserved during corrections.
- **Customizable Settings:**
  - Configure your OpenAI API key directly in the plugin settings.
  - Option to select the correction service (OpenAI or LanguageTool).

#### Usage:
1. **Install the plugin.**
2. **Set your API key in the plugin settings.**
3. Use the chat bar button to enable or disable corrections as needed.

This plugin is designed to improve text clarity while maintaining user intent and formatting.

Let me know if there are any additional changes or improvements you'd like me to make! 😊

PR #3254, opened 2025-02-24
feat(translate): Add OpenAI Compatible API

T1ckbase:
image

### DIFF
--- a/src/plugins/translate/settings.ts
+++ b/src/plugins/translate/settings.ts
@@ -57,7 +57,8 @@ export const settings = definePluginSettings({
         options: [
             { label: "Google Translate", value: "google", default: true },
             { label: "DeepL Free", value: "deepl" },
-            { label: "DeepL Pro", value: "deepl-pro" }
+            { label: "DeepL Pro", value: "deepl-pro" },
+            { label: "OpenAI Compatible", value: "openai-compatible" },
         ] as const,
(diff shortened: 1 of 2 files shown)


jamesbt365:
A similar plugin (#3138) that used AI has already been denied, using AI for stuff like this is a bad idea and once again yet another thing that requires an API key.

Its just not built for something like this and prompt injection is very much possible to do even accidentally, its just not a good idea, use a proper engine built for translation.

T1ckbase:
> A similar plugin (#3138) that used AI has already been denied, using AI for stuff like this is a bad idea and once again yet another thing that requires an API key.
> 
> Its just not built for something like this and prompt injection is very much possible to do even accidentally, its just not a good idea, use a proper engine built for translation.

I totally get your concerns about AI and prompt injection risks, like you mentioned with the previous plugin (#3138).
I opened this PR because I’ve found Google Translate’s quality really lacking—it often mangles code blocks, breaks formatting, and ends up unreadable, which has been frustrating.
In my experience, AI-based translation handles these cases much better and delivers higher-quality results. Even if there’s a risk of injection, the worst case is just some gibberish, which I think is a fair trade-off for the improvement. Adding this as an optional choice alongside Google and DeepL doesn’t hurt—it just gives users more flexibility. I’d really appreciate it if you could reconsider this addition. What do you think?

PR #3257, opened 2025-02-26
Add name and displayName props to known modules

sadan4:
This makes it wildly easier to find what you are looking for in react devtools

## Before
image

## After
image

*not all of those names are provided by this pr, only the ones used in vencord are


sadan4:
I feel like this would be the most useful to people who are new to working with discords code, as they wouldn't feel completely lost with all the minified names

Vendicated:
this api is really weird. especially the option for findComponentLazy is misleading

src/utils/lazyReact.tsx
```diff
@@ -25,5 +25,23 @@ export function LazyComponent<T extends object = any>(factory: () => React.Compo
...
+                    const name: string | undefined = "displayName" in got ? got.displayName : got.name;
+                    return `${name ? `${name}-` : ""}LazyComponent`;
+                } catch (e) {
+                    (IS_DEV ? console.warn : console.debug)("Failed to get name from LazyComponent", e);
+                    return "LazyComponent";
+                }
+            }
+        });
+    } catch (e) {
+        (IS_DEV ? console.warn : console.debug)("Failed to set name on LazyComponent", e);
```
Vendicated, line 43: that should never fail

src/utils/lazyReact.tsx
```diff
@@ -25,5 +25,23 @@ export function LazyComponent<T extends object = any>(factory: () => React.Compo
...
     LazyComponent.$$vencordInternal = get;
 
+    try {
+        Object.defineProperty(LazyComponent, "name", {
+            enumerable: false,
+            get() {
+                try {
+                    const got = get();
+                    const name: string | undefined = "displayName" in got ? got.displayName : got.name;
+                    return `${name ? `${name}-` : ""}LazyComponent`;
```
Vendicated, line 35: I would prefer a display like `LazyComponent(Tooltip)`. Or use some other bracket pair

src/webpack/common/internal.tsx
```diff
@@ -20,18 +20,51 @@ import { LazyComponent } from "@utils/react";
 
 // eslint-disable-next-line path-alias/no-relative
 import { FilterFn, filters, lazyWebpackSearchHistory, waitFor } from "../webpack";
+const SYM_FORWARD_REF = Symbol.for("react.forward_ref");
```
Vendicated, line 23: add a space after imports

src/webpack/common/internal.tsx
```diff
@@ -20,18 +20,51 @@ import { LazyComponent } from "@utils/react";
...
+const SYM_MEMO = Symbol.for("react.memo");
+/**
+ * calls {@link setComponentName} on the given component
+ * @returns maybeComponent
+ */
+export function wrapComponentName<T>(maybeComponent: T, name?: string): T {
+    if (name) setComponentName(maybeComponent, name);
+    return maybeComponent;
+}
+export function setComponentName(maybeComponent: any, name: string): void {
```
Vendicated, line 33: I'm absolutely against polluting Discord objects in any way. If you really wanna do this, it should be part of an optional plugin like ConsoleShortcuts (at this point we could rename it to ConsoleUtils or smth)

src/utils/lazyReact.tsx
sadan4: yea, that does look better

sadan4:
ready for review

src/utils/lazyReact.tsx
rniii: `LazyComponent<Name>` :trolleybus:

rniii:
> this api is really weird. especially the option for findComponentLazy is misleading

all changes to discord's components are `LazyComponent(() => x)` -> `LazyComponent(() => wrapDiscordComponentName(x, "Name"))`, which is also the change inside `findComponentLazy`, it'd make more sense if the argument was in `LazyComponent` no? and no need for an `opts` object in that case, I think

rniii:
in that case you also have a fallback for LazyComponent itself which doesn't need to pollute discord

PR #3260, opened 2025-03-01
vcNarrator: Add ability to use Friend Nickname

peyton2465:
The {{FRIEND_NICKNAME}} tag will narrate the user's friend nickname. In the case that you do not have a friend nickname defined for the user, it will have the same functionality as {{DISPLAY_NAME}}.


Vendicated:
Not a fan of this new variable. I think we should just change the other variables to use friend nickname instead

PR #3265, opened 2025-03-03
Add NoDiscordJunk plugin

Bamboooz:
Adds a NoDiscordJunk plugin that allows the user to turn off annoying Discord features such as: profile effects, avatar decorations, profile themes, nitro buttons, shop buttons, discover button.


Vendicated:
we don't allow css only plugins

NSPC911:
[#css-snippets](https://discord.com/channels/1015060230222131221/1028106818368589824) would be where it might belong

PR #3267, opened 2025-03-04
Fix BadgesAPI not working

sadan4:


src/plugins/_api/badges/index.tsx
```diff
@@ -65,12 +65,12 @@ export default definePlugin({
...
             find: ".FULL_SIZE]:26",
             replacement: {
-                match: /(?<=(\i)=\(0,\i\.\i\)\(\i\);)return 0===\i.length\?/,
+                match: /(?<=(\i)=\(0,\i\.\i\)\(\i\).{0,200};)return 0===\i.length\?/,
                 replace: "$1.unshift(...$self.getBadges(arguments[0].displayProfile));$&"
             }
         },
         {
-            find: ".description,delay:",
+            find: "#{intl::PROFILE_USER_BADGES}",
```
sadan4, line 73:
```suggestion
            find: "let{badges:",
```
if you would rather not use an intl key

PR #3270, opened 2025-03-05
Fix plugins using ApplicationAssetUtils (like CustomRPC)

sadan4:


Arcitec:
Thanks, this fixed the following bug:

#1127

PR #3276, opened 2025-03-07
PluginSettings: stop authors overflow from showing +0

xNasuni:
this is what it looks like before & after

image of 6 authors with +0 showing at the end
image of 6 authors without +0 overflow

### DIFF
--- a/src/components/PluginSettings/styles.css
+++ b/src/components/PluginSettings/styles.css
@@ -81,3 +81,7 @@
 .vc-plugins-info-icon:not(:hover, :focus) {
     color: var(--text-muted);
 }
+
+.user-summary-item svg:last-child>foreignObject {
+    mask: none !important;
+}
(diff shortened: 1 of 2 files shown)


src/components/PluginSettings/styles.css
```diff
@@ -81,3 +81,7 @@
 .vc-plugins-info-icon:not(:hover, :focus) {
     color: var(--text-muted);
 }
+
+.user-summary-item svg:last-child>foreignObject {
+    mask: none !important;
+}
```
Vendicated, line 87: what's this for?

Vendicated:
is this even still relevant on refresh? can't repro

src/components/PluginSettings/styles.css
xNasuni: without this, the other change would still make the last author have the svg clip mask:
image

xNasuni:
i think the behavior only happens when there is exactly 6 authors, that's where i've seen it happen

sadan4:
If vencord is just using discord component, then how does discord handle this issue when it uses the component with exactly six people?

xNasuni:
well i can't reproduce on the visual refresh (without this patch) so i guess it doesn't matter if the old ui will be replaced

Vendicated:
oh i see. thank you anyway!

xNasuni:
okay so turns out i was a little lazy and tried reproducing it incorrectly which gave me a false positive.

this bug still occurs at the limit of the slice thingy whatever but it still occurs even on the visual refresh which is now the main ui.

the changes are still accurate and work for the visual refresh, not outdated, see changes below:

before
after

Vendicated:
i still doubt this is the right solution. it works fine for Discord, we're probably just using the component wrong

PR #3279, opened 2025-03-08
Add vesktop support channel to allowlist for SupportHelper

Overo3:
Relatively trivial change, but has bothered me on occasion


Vendicated:
done via https://github.com/Vendicated/Vencord/commit/d62be1b94ac8a49131bc5adb8c9bcd876ff2706f

PR #3282, opened 2025-03-09
Fix CSS syntax

MoaningTortoise:


Vendicated:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting

PR #3290, opened 2025-03-12
Pass CLI args through pnpm inject script to installer

surgedevs:
Changes the runInstaller.mjs to pass through CLI args.
This would allow for skipping the interaction bit which saves a microseconds when developing a plugin with native functions:

image

very convenient


Vendicated:
Implemented via https://github.com/Vendicated/Vencord/commit/188fd48659251ed72a2fb7b460db19b9b61e27ef

image

PR #3291, opened 2025-03-15
Replace UNPKG with jsDelivr

SukkaW:
UNPKG has not been actively maintained and has been down since `Mar 15, 2025` (https://github.com/unpkg/unpkg/issues/412).

The PR replaces UNPKG usage so that the Monaco Editor and Shiki are loaded from jsDelivr. And since all the UNPKG usages are replaced, the CSP is also changed to `cdn.jsdelivr.net`.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -101,7 +101,7 @@ if (IS_VESKTOP || !IS_VANILLA) {
                 // TODO: Restrict this to only imported packages with fixed version.
                 // Perhaps auto generate with esbuild
                 csp["script-src"] ??= [];
-                csp["script-src"].push("'unsafe-eval'", "https://unpkg.com", "https://cdnjs.cloudflare.com");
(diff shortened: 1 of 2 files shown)


Vendicated:
Seems like unpkg was having scaling issues and increased server count so it should be fixed now, but I guess it's still better to switch. Not just because jsdelivr is way more popular & stable but also to reduce the number of different CDNs used

We should really just stop using CDNs and instead bundle it, but this is good for now, thank you!

SukkaW:
> Seems like unpkg was having scaling issues and increased server count so it should be fixed now, but I guess it's still better to switch. Not just because jsdelivr is way more popular & stable but also to reduce the number of different CDNs used

Also, jsDelivr is more actively maintained. Every time I mail jsDelivr's maintainers I get a response within 4 hours, while the maintainer of the UNPKG has more things to do (Remix, React Router, etc.), that's why it was taking them 18 hours to fix issues.

> We should really just stop using CDNs and instead bundle it, but this is good for now, thank you!

In the long term, yes I agree. But it will require a kind of huge changes with the building process, which can put into the backlog.

PR #3293, opened 2025-03-15
qol: replace spaces in plugin name search

notTamion:
small quality of life change cause i keep typing in plugin names with spaces xd

### DIFF
--- a/src/components/PluginSettings/index.tsx
+++ b/src/components/PluginSettings/index.tsx
@@ -260,7 +260,7 @@ export default function PluginSettings() {
         if (!search.length) return true;
 
         return (
-            plugin.name.toLowerCase().includes(search) ||
+            plugin.name.toLowerCase().includes(search.replace(/\s+/g, "")) ||
             plugin.description.toLowerCase().includes(search) ||
             plugin.tags?.some(t => t.toLowerCase().includes(search))
         );
--- a/src/components/PluginSettings/index.tsx
+++ b/src/components/PluginSettings/index.tsx
@@ -175,7 +175,7 @@ const enum SearchStatus {
 
 function ExcludedPluginsList({ search }: { search: string; }) {
     const matchingExcludedPlugins = Object.entries(ExcludedPlugins)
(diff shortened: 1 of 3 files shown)


sadan4 (commented):
Looks good, this has also been annoying me for a while.

src/components/PluginSettings/index.tsx
```diff
@@ -260,7 +260,7 @@ export default function PluginSettings() {
         if (!search.length) return true;
 
         return (
-            plugin.name.toLowerCase().includes(search) ||
+            plugin.name.toLowerCase().includes(search.replace(/\s+/g, "")) ||
```
Vendicated, line 263: 1. this replacement should be done on line 250 instead of here
2. what if the plugin name contains spaces? now you can't find that anymore

src/components/PluginSettings/index.tsx
```diff
@@ -260,9 +260,9 @@ export default function PluginSettings() {
         if (!search.length) return true;
 
         return (
-            plugin.name.toLowerCase().includes(search) ||
-            plugin.description.toLowerCase().includes(search) ||
-            plugin.tags?.some(t => t.toLowerCase().includes(search))
+            plugin.name.toLowerCase().replace(/\s+/g, "").includes(search) ||
+            plugin.description.toLowerCase().replace(/\s+/g, "").includes(search) ||
+            plugin.tags?.some(t => t.toLowerCase().replace(/\s+/g, "").includes(search))
```
Vendicated, line 265: how does that affect search performance? considering this is gonna be done on every keystroke, let's make sure it doesn't become laaaaaggy

notTamion: image
doesn't really have a huge affect on performance. considering we are only calling this when theres an input i think its fine

sadan4: It still probably wouldn't hurt to memo them or just store them in a top level local variable. Why add extra steps every keystroke if you don't need to

PR #3294, opened 2025-03-15
fix YoutubeAdblock

Vendicated:
replaces adguard with iridium (rewritten) - still needs some testing


Vendicated:
undid for now because it doesn't seem to fully work

PR #3295, opened 2025-03-16
Created README-ru.md

chiter34256:
Created README-ru.md with russian translation


Vendicated:
thank you but this is not really useful and will become outdated quickly

PR #3298, opened 2025-03-18
AlwaysAnimate: Fix patch

sadan4:


sadan4:
For IgnoreActivities, I can not find the module again. Given it was unused, I guess that it was deleted.

src/plugins/alwaysAnimate/index.ts
```diff
@@ -51,7 +51,7 @@ export default definePlugin({
             // Guild Banner
             find: ".animatedBannerHoverLayer,onMouseEnter:",
             replacement: {
-                match: /(?<=guildBanner:\i,animate:)\i(?=}\))/,
+                match: /(?<=guildBanner:\i,animate:)\i(?=,isRefreshEnabled:)/,
```
Nuckyz, line 54:
```suggestion
                match: /(?<=guildBanner:\i,animate:)\i/,
```

Lets just nuke it to not depend on visual refresh stuff

Nuckyz (changes requested):

src/plugins/alwaysAnimate/index.ts
sadan4: if we remove that, and we should, it should be anchored to something that isn't just an identifier so it doesn't match a destructuring assignment and error

sadan4:
```suggestion
                match: /(?<=guildBanner:\i,animate:)\i(?=.{0,100}controller)/,
```
i feel like this is the best i can do to not rely on a visualReferesh prop

sadan4:
Also fixed account panel server profile

PR #3304, opened 2025-03-21
AccountPanelServerProfile: Fix patch issue with onPopoutClose

sadan4:


src/plugins/accountPanelServerProfile/index.tsx
```diff
@@ -81,8 +81,8 @@ export default definePlugin({
                     replace: (_, rest, popoutProps, originalPopout, currentUser) => `${rest}$self.UserProfile({popoutProps:${popoutProps},currentUser:${currentUser},originalRenderPopout:()=>{${originalPopout}}})`
                 },
                 {
-                    match: /\.AVATAR,children:.+?(?=renderPopout:)/,
-                    replace: "$&onRequestClose:$self.onPopoutClose,"
+                    match: /(\.AVATAR,children:.+?onRequestClose:\(\)=>\{)/,
```
sadan4, line 84:
```suggestion
                    match: /(onRequestClose:\(\)=>\{)/,
```

now that we're matching a function, its possible to remove the `\.AVATAR,children:.+?` anchor, but im unsure if it will matter at all.

id lean toward keeping it in as it doesnt hurt anything

Nuckyz (approved):

PR #3307, opened 2025-03-23
Remove code font from callTimer

mochienya:
Why would this be using the code font?? I know I could change it with css but there really should be sensible defaults.


Vendicated:
Because otherwise the numbers have inconsistent width which makes the ui constantly shift. This is very jarring

PR #3308, opened 2025-03-23
ThemeAttributes: Fix crash when avatar src is null

sadan4:
lol discord checks for null right above in their code but we don't :sob:
image


src/plugins/themeAttributes/index.ts
```diff
@@ -54,8 +54,8 @@ export default definePlugin({
         }
     ],
 
-    getAvatarStyles(src: string) {
-        if (src.startsWith("data:")) return {};
+    getAvatarStyles(src: string | null) {
+        if (src == null || src.startsWith("data:")) return {};
```
Vendicated, line 58:
```suggestion
        if (!src || src.startsWith("data:")) return {};
```

PR #3313, opened 2025-03-26
ClientTheme: Fix for visual refresh

CodeF53:
image
image
image

I realize my screenshots dont show the redesign at all but Im in a rush


sadan4:
does this work with pre visual refresh

CodeF53:
> does this work with pre visual refresh

No

mochienya:
> does this work with pre visual refresh

the visual refresh is a global update, not an experiment. you would have to refresh your client to apply this update, so the refresh would get applied anyway

CodeF53:
I'll just fix it so it generates offsets for old and new color vars
No big deal

CodeF53:
> I'll just fix it so it generates offsets for old and new color vars No big deal

I dont have access to the old design anymore, cant do this

NorthLake:
> I dont have access to the old design anymore, cant do this

Enable the Experiments plugin, then set the "Desktop Visual Refresh" experiment to Treatment 0

NEVARLeVrai:
it will be fixed ?

CodeF53:
> it will be fixed ?

Im lazy, I'll just wait for the experiment to die :3, either they revert the redesign or force everyone onto it.
- In the first case this PR gets closed
- In the second case this PR gets merged

CodeF53:
this PR has been "Merged" into equicord https://github.com/Equicord/Equicord/commit/ca4b63c3862d14058b89b74dc5b65f6f0bce0aa2#diff-4e674bfc2a79162de5b1b648c2f0e7890fd32cbd85f00b940995f6dd703519c6R123 for anyone who wants to use it.

src/plugins/clientTheme/index.tsx
```diff
@@ -160,48 +156,41 @@ function generateColorOffsets(styles) {
...
 }
 
-function captureOne(str, regex) {
-    const result = str.match(regex);
-    return (result === null) ? null : result[1];
-}
-
-function mapReject(arr, mapFunc) {
-    return arr.map(mapFunc).filter(Boolean);
+// F53: It seems like this isn't needed anymore with the redesign, but I don't like testing light mode...
```
Vendicated, line 165: please do test this so we can remove this code

Vendicated:
To clarify, this only works on the new UI, not the old UI?

That's okay, I don't think it's realistic to maintain support for both. It would be different if the rollout wasn't 100%, but it is, so 🤷🏻

jacelli:
Hello

Before discord update UI 'refreshed', the ClientTheme plugin was perfect, and it looked like this:

Screenshot_1c2871

but now it's like this: (everything changes to 1 color)

Screenshot (13191)

I'm using Default discord 'Dark' Theme i tried everything
I couldn't fix the texts font color back to default (QuickCSS worked but wasn't perfect)
Is it possible to fix it like it was before?
Ty


(Version:
stable 388298 (fe304ad) Host 1.0.9188 x64 (61444) Build Override: N/A Windows 10 64-bit (10.0.19044)
Vencord 77bed92 (Standalone)
Electron 33.4.0
Chromium 130.0.6723.191)

CodeF53:
> Hello
> 
> Before discord update UI 'refreshed', the ClientTheme plugin was perfect, and it looked like this:
> 
> Screenshot_1c2871
> 
> but now it's like this: (everything changes to 1 color)
> 
> Screenshot (13191)
> 
> I'm using Default discord 'Dark' Theme i tried everything I couldn't fix the texts font color back to default (QuickCSS worked but wasn't perfect) Is it possible to fix it like it was before? Ty
> 
> (Version: stable 388298 (fe304ad) Host 1.0.9188 x64 (61444) Build Override: N/A Windows 10 64-bit (10.0.19044) Vencord [77bed92](https://github.com/Vendicated/Vencord/commit/77bed92c575ced3228bf3e7ba8a1a6cd24362acf) (Standalone) Electron 33.4.0 Chromium 130.0.6723.191)

The new color schemes in discord is are longer well-suited to the way the internal client theme logic works. I gave a lot of effort to new approaches, but this is the best I could get with the generative approach client theme takes.

CodeF53:
if you want, you can try using the tinting system used by nitro themes for the non-gradient parts of the app
```css
:root {
  --theme-base-color: #e700ff;
  --theme-base-color-amount: 25%;
}
```
but in general this is underwhelming and not really the same aesthetic that clientTheme was either

jacelli:
Thanks it's working, it's not as perfect as the plugin it looks too washed out in contrast but maybe i'll get used to it

In discord default themes i liked 'Onyx' better but the left side panel of channels it's always darker than the rest 
So i'll just stick with default theme 'Dark'

and i'll use this:

```
:root {
  --theme-base-color: #21013570;
  --theme-base-color-amount: 100%;
}
```

The result:

Screenshot z

PR #3320, opened 2025-03-26
feat(plugin): implement message options on spotifyShareCommands plugin

IchiiDev:
For consistency, I implemented optional arguments for the commands added by the plugin to include a message to append the data to.

e.g:
image
image

Other built-in commands like `/tableflip` or `/shrug` also have this behavior, so it made sense to implement it for the Spotify share plugin.
image


src/plugins/spotifyShareCommands/index.ts
```diff
@@ -81,54 +81,82 @@ export default definePlugin({
...
             inputType: ApplicationCommandInputType.BUILT_IN,
-            options: [],
-            execute: (_, ctx) => {
+            options: [
+                {
+                    name: "message",
+                    description: "Appends the Spotify link to your message",
+                    type: ApplicationCommandOptionType.STRING,
+                    required: false
+                }
```
Vendicated, line 90: use `OptionalMessageOption` from the commands api instead
image

src/plugins/spotifyShareCommands/index.ts
```diff
@@ -81,54 +81,82 @@ export default definePlugin({
...
+            ],
+            execute: (options, ctx) => {
                 const track: Track | null = Spotify.getTrack();
+                const message = options.find(option => option.name === "message")?.value;
```
Vendicated, line 94: use `findOption` from the commands api instead

Vendicated:
Decided to refactor/rewrite the plugin

Vendicated:

PR #3324, opened 2025-03-27
Plugin: VisualDefresh

Inbestigator:
This plugin makes Discord look mostly like it did before the visual overhaul.

*Classes special to the refresh may still be applied*

I noticed that a ton of users are annoyed by the new look, and can't seem to find the experiments tab. So this plugin automagically de-applies the experiment for them.


4wincode:
also, when the old interface is removed eventually this plugin straight up won't work and would also have to be removed so there's no point

Inbestigator:
Yeah makes sense, **I** probably wouldn't even use it lol.

PR #3325, opened 2025-03-27
feat(plugin): 7TV

prodbyeagle:
This plugin adds support for 7tv emotes. the default emote set is mine but can be changed with the settings ( will need their twitch id but i could change it to use the discord id (needs linking).


sadan4:
This could be done with an emote server and fake nitro

PR #3327, opened 2025-03-27
Add the FixUpX plugin

Rainnny7:
This plugin makes it so when you send or edit a message that contains a Twitter/X URL, it will replace it with the replacement URL in the config (that being https://fixupx.com) for better embeds, because Twitter doesn't know how to embed properly 😭

### DIFF
--- a/patch.bat
+++ b/patch.bat
@@ -1,0 +1,5 @@
+@echo off
+git pull
+pnpm install --frozen-lockfile
+pnpm build
+pnpm inject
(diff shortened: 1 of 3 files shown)


lolsuffocate:
This can be done with the TextReplace plugin already

Vendicated:
TextReplace plugin:
```
Find: https://x.com
Replace: https://fixupx.com
Only if includes: /status/
```

done

PR #3332, opened 2025-03-30
feat: Option to blacklist/whitelist servers in noReplyMention

scattagain:
As the title says, this lets you change the default for pinging within specific servers


src/plugins/noReplyMention/index.tsx
```diff
@@ -28,7 +29,7 @@ const settings = definePluginSettings({
         type: OptionType.STRING,
         default: "1234567890123445,1234567890123445",
     },
-    shouldPingListed: {
+    shouldPingListedUsers: {
```
jamesbt365, line 32: You should use `migratePluginSetting` so that users do not lose the setting.

scattagain: Wasn't aware there was a way to do that, should be fixed now!

PR #3333, opened 2025-03-30
[Plugin] KeybindMute

Voxxin:
This plugin adds keybinds for muting, unmuting, deafening, and undeafening yourself.  
While Discord provides toggle options by default, this plugin lets you assign keybinds to set a specific state directly—for example, instantly deafening yourself with a single press.  

These keybinds are integrated into the Keybinds page action list, making setup and customization simple and intuitive.

### DIFF
--- a/src/plugins/keybindMute/index.tsx
+++ b/src/plugins/keybindMute/index.tsx
@@ -1,0 +1,132 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2025 Vendicated and contributors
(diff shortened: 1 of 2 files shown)


Vendicated:
Patches should never hardcode minified variable names. Those will change very frequently and likely lead to crashes. You can use Vencord's special `\i` escape sequence to match identifiers 

This plugins also seems very niche, I don't think many people would use this, so I'll close this, sorry

Thank you for your contribution nevertheless!

PR #3189, opened 2025-01-31
New Plugin: CustomTimestamps

lolsuffocate:
# CustomTimestamps

This plugin allows you to customize the timestamps in chat messages and tooltips.

This is a rework of [Rini and nvhhr's plugin](https://gist.github.com/nvhhr/7385814810cad48be5ffa4483fe4d2f8) and as such both are included as devs too.

## Formats

- Cozy: The timestamp by your username when chat is in cozy mode.
cozy

- Compact: The timestamp to the left of messages when chat is in compact mode and when you send multiple messages in cozy mode.
compact

- Tooltip: The timestamp in the tooltip when hovering over a message.
tootip

## Placeholders

- **[calendar]**: Replaced with moment.js's calendar time, which dynamically adjusts based on the date.
- **[relative]**: Replaced with moment.js's relative time, such as "4 hours ago".

## Timeframes for Calendar Time

- **Same day**: Format for today's dates (e.g., "Today, 2:30 PM").
- **Last day**: Format for yesterday's dates (e.g., "Yesterday, 2:30 PM").
- **Last week**: Format for dates within the last week (e.g., "Monday, 2:30 PM").
- **Same else**: Format for older dates (e.g., "01/01/2024, 2:30 PM").

For more information on formatting options, refer to the [Moment.js formatting documentation](https://momentjs.com/docs/#/displaying/format/).

PR #3334, opened 2025-03-30
[Plugin] ShowSeconds

leafyzito:
This plugin simply adds the option to have seconds shown in message timestamps, as some people (like me) might like it

image
image

### DIFF
--- a/src/plugins/showSeconds/index.tsx
+++ b/src/plugins/showSeconds/index.tsx
@@ -13,8 +13,45 @@ import { useEffect } from "@webpack/common";
 import { useForceUpdater } from "@utils/react";
 
 const timeFormats = {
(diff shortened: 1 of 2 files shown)


lolsuffocate:
https://github.com/Vendicated/Vencord/pull/3189 does this too, albeit more complicated

leafyzito:
> #3189 does this too, albeit more complicated
Yes, I saw that and found it very confusing
This one is more straight to the point and just adds the seconds to the timestamps, also with a more direct plugin name

Vendicated:
thanks for your contribution but this is too overly specific!

 a more useful plugin would be one that allows you to customise the timestamp format to your liking, aka that linked PR

Issue #878, opened 2023-04-12
Patcher ignoring custom MIN_WIDTH/HEIGHT


MeguminSama:
### Discord Account

me

### What happens when the bug or crash occurs?

when disableMinSize is false, it doesn't check if the user has put in a custom MIN_WIDTH or MIN_HEIGHT

https://github.com/Vendicated/Vencord/blob/main/src/main/patcher.ts#L115-L124

### What is the expected behaviour?

The patcher should respect the user's MIN_WIDTH and MIN_HEIGHT, only setting defaults if it doesn't already exist. 

### How do you recreate this bug or crash?

Set custom MIN_WIDTH and MIN_HEIGHT and turn off the "disable min height" toggle.
Vencord will ignore your settings and set it to a default of 940x500

### Errors

meow

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well

PR #3335, opened 2025-03-31
Fix "Disable minimum window size" option

### LINKED ISSUE #3330: [Bug] Disable minimum window size no longer works

Enabling "Disable minimum window size" in Settings > Vencord no longer disables the minimum window size. This stopped working seemingly alongside the updated Discord UI refresh.

I expect Vencord to still be able to disable the minimum window size via Electron tinkering or some other patch rather than [relying on the MIN_WIDTH and MIN_HEIGHT settings from appSettings](https://github.com/Vendicated/Vencord/blob/b3bff83dd5040950c55e09bed9e47a60490f81d8/src/main/patcher.ts#L121-L130).

1. Enable Settings > Vencord > Disable minimum window size
2. Restart Vencord

### Errors

No related errors

### Request Agreement

- [x] I am a Vencord Developer

iminlikewithyou:
Fixes #3330.

it seems like something's going on with Discord not respecting the appSettings MIN_WIDTH and MIN_HEIGHT anymore, so what I've implemented is simply disabling the electron call to setMinimumSize entirely. not sure if there's a technically better way - but it works and i'm very happy about it.


thc282:
May i ask a question?
will this effect the window is disappear in the `Alt+tab / Win+tab`

My discord always disapper, but it still running in the background

iminlikewithyou:
no, all this does is make the Discord windows able to be resized smaller. i'm not sure what you're describing

vantiss:
thank you so much for this btw - I've been using it since the day you submitted this and it's been a lifesaver

thc282:
> no, all this does is make the Discord windows able to be resized smaller. i'm not sure what you're describing

i mean 
when u press the Alt + Tab, there are some windows 


but the discord.exe always disappeared with no where
Discord is running at the background for sure

each time i need to quit it and relaunch to fix it
Dunno why it will disappear

iminlikewithyou:
pretty sure that's Discord working as intended - if you press the "X" on Discord instead of "-", the window will close, but Discord keeps running in the background. that's due to this setting here:
image

but if not, you should open a new issue. this is unrelated

snoggles:
@Vendicated what are next steps for getting the test workflow to run and for getting this merged?

Vendicated:
should also fix https://github.com/Vendicated/Vencord/issues/878

PR #3338, opened 2025-04-01
fix(ImplicitRelationships): update logic

dolfies:
Discord updated their logic so you no longer need a mutual guild for a user to be considered an implicit relationship. This makes the plugin logic quite a bit simpler (it should now be a lot faster as it loads asynchronously). Unfortunately, it also means some users may not show up in the list if the user object isn't cached, but I don't think it's worth adding HTTP fetching logic for these rare edge cases.

### DIFF
--- a/src/plugins/implicitRelationships/index.ts
+++ b/src/plugins/implicitRelationships/index.ts
@@ -20,7 +20,7 @@ import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 import { findStoreLazy } from "@webpack";
-import { Constants, FluxDispatcher, GuildStore, RelationshipStore, SnowflakeUtils, UserStore } from "@webpack/common";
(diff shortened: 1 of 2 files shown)


src/plugins/implicitRelationships/index.ts
```diff
@@ -121,55 +121,61 @@ export default definePlugin({
...
         };
 
-        // @ts-ignore
         FluxDispatcher.subscribe("GUILD_MEMBERS_CHUNK_BATCH", callback);
         for (let i = 0; i < toRequest.length; i += 100) {
             FluxDispatcher.dispatch({
                 type: "GUILD_MEMBERS_REQUEST",
                 guildIds: allGuildIds,
                 userIds: toRequest.slice(i, i + 100),
+                presences: true,
```
Vendicated, line 178: what's that for?

dolfies: It also requests `presences` so the user list isn't mostly offline as it is rn

Vendicated:
thanks pookie

PR #3340, opened 2025-04-01
ReplaceAnything plugin

smoukey:
This plugin allows users to replace any specified text with another chosen text before sending a message.

### DIFF
--- a/src/plugins/ReplaceAnything/index.tsx
+++ b/src/plugins/ReplaceAnything/index.tsx
@@ -1,0 +1,47 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { addMessagePreSendListener, removeMessagePreSendListener } from "@api/MessageEvents";
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+
+const settings = definePluginSettings({
+    target: {
+        type: OptionType.STRING,
+        description: "Text to replace",
+        default: "?"
+    },
+    replace: {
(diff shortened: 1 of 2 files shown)


jamesbt365:
1. you don't need to reopen your PR to push new commits
2. you have just reinvented TextReplace, but made it a *lot* worse

smoukey:
I didnt know about TextReplace but i just went thru it and it’s kinda overcomplicated. It has regex support, multiple rule types, and a bunch of UI elements that most people don’t need. Mine is way simpler, just pick what you want to replace and what to replace it with, and you’re done. No unnecessary settings, no clutter, just works.

lolsuffocate:
You still have to go to the settings modal and set the text for yours too, which is exactly the same as if you just put a single rule into TextReplace, it just allows you to add more than one

jamesbt365:
> and it’s kinda overcomplicated

It really isn't that complicated, you get your find, your replace and an option to only replace in certain situations

> and a bunch of UI elements that most people don’t need

I don't get this at all? What?

> Mine is way simpler

Yes, it is simpler but in a way that makes this unusable for any person who wants to replace *multiple* things or more advanced things, your plugin has *1* replacement option, thats it, you can only have *1*, TextReplace lets you have as many as you want, which is great because most people will want to replace more than one thing.

Absolutely nothing is unnecessary about TextReplace and your plugin doesn't improve on anything at all, it is just way more expansive and provides way more usability than this plugin of yours ever does.

Vencord is not a fan of implementing duplicate functionality, and this just lacks code quality and usability in comparison to every other plugin in this client mod.

(also, you build your regex you use the replace the text every time somebody sends a message including the target, which is just inefficient and completely unnecessary as you can do string replacements without regex)

This just lacks usability and quality, you can't argue TextReplace is overcomplicated here.

I'm all for improvements to Vencord, but this just isn't it.

PR #3341, opened 2025-04-01
Mock Plugin

smoukey:
Fun little plugin that adds a "Mock Message" button to repeat messages in mocking SpOnGeBoB case.


jamesbt365:
https://vencord.dev/plugins/MoreCommands

smoukey:
> https://vencord.dev/plugins/MoreCommands

this only adds a command, instead of copy pasting and using that command u can just press a button

PR #3349, opened 2025-04-03
Fix patches for discord update

sadan4:
only on canary as of now

will add more fixes as I get to them

### DIFF
--- a/src/plugins/consoleJanitor/index.tsx
+++ b/src/plugins/consoleJanitor/index.tsx
@@ -117,6 +117,7 @@ export default definePlugin({
         this.settings.store.whitelistedLoggers?.split(";").map(x => x.trim()).forEach(logAllow.add.bind(logAllow));
     },
 
+    Noop,
     NoopLogger: () => NoopLogger,
 
     shouldLog(logger: string, level: keyof AllowLevels) {
@@ -149,14 +150,14 @@ export default definePlugin({
             find: "is not a valid locale.",
             replacement: {
                 match: /\i\.error(?=\(""\.concat\(\i," is not a valid locale."\)\))/,
-                replace: "$self.NoopLogger"
+                replace: "$self.Noop"
             }
         },
         {
             find: '"AppCrashedFatalReport: getLastCrash not supported."',
             replacement: {
                 match: /console\.log(?=\("AppCrashedFatalReport: getLastCrash not supported\."\))/,
-                replace: "$self.NoopLogger"
+                replace: "$self.Noop"
             }
         },
         {
--- a/src/plugins/consoleJanitor/index.tsx
+++ b/src/plugins/consoleJanitor/index.tsx
@@ -148,15 +148,15 @@ export default definePlugin({
         {
             find: "is not a valid locale.",
             replacement: {
-                match: /\i\.error\(""\.concat\(\i," is not a valid locale."\)\)/,
-                replace: '""'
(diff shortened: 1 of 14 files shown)


sadan4:
thanks to @thororen1234 this covers everything now

src/webpack/common/components.ts
```diff
@@ -74,7 +74,7 @@ export const ScrollerNone = LazyComponent(() => createScroller(scrollerClasses.n
 export const ScrollerThin = LazyComponent(() => createScroller(scrollerClasses.thin, scrollerClasses.fade, scrollerClasses.customTheme));
 export const ScrollerAuto = LazyComponent(() => createScroller(scrollerClasses.auto, scrollerClasses.fade, scrollerClasses.customTheme));
 
-const { FocusLock_ } = mapMangledModuleLazy("attachTo:null!==", {
+const { FocusLock_ } = mapMangledModuleLazy('document.getElementById("app-mount"))', {
```
Nuckyz, line 77: This find is very weak,  needs to be more specific

Nuckyz (changes requested):

Nuckyz:
Ideally these fixes should also cover the old versions, seems like another weird bundler change. Is it possible?

src/plugins/customidle/index.ts
```diff
@@ -49,7 +49,7 @@ export default definePlugin({
                     replace: "$self.handleOnline()"
                 },
                 {
-                    match: /(setInterval\(\i,\.25\*)\i\.\i/,
+                    match: /(setInterval\(\i,30\*)\i\.\i\.Millis\.SECOND/,
```
sadan4, line 52: I would be weary about merging this without through testing as we are multiplying by a different constant (.25 -> 30)

Vendicated: it's just the interval time. lower number = polls more often

sadan4: yeah, thats why i looked into it.

src/plugins/consoleJanitor/index.tsx
```diff
@@ -148,15 +148,15 @@ export default definePlugin({
...
+                replace: "$self.NoopLogger"
             }
         },
         {
             find: '"AppCrashedFatalReport: getLastCrash not supported."',
             replacement: {
-                match: /console\.log\("AppCrashedFatalReport: getLastCrash not supported\."\);/,
-                replace: ""
+                match: /console\.log(?=\("AppCrashedFatalReport: getLastCrash not supported\."\))/,
+                replace: "$self.NoopLogger"
```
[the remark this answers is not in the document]
sadan4: thanks, misread what NoopLogger was yesterday when i did that

src/plugins/ircColors/index.ts
```diff
@@ -66,7 +66,7 @@ export default definePlugin({
         {
             find: '="SYSTEM_TAG"',
             replacement: {
-                match: /(?<=\i.gradientClassName]\),style:.{0,80}:void 0,)/,
+                match: /(?<=\i.gradientClassName]\),style:.+?,)(?=onClick:)/,
                 replace: "style:{color:$self.calculateNameColorForMessageContext(arguments[0])},"
```
Vendicated, line 70:
```suggestion
                match: /(?<=\i.gradientClassName]\),style:/,
                replace: "style:{color:$self.calculateNameColorForMessageContext(arguments[0])},_style:"
```

PR #2957, opened 2024-10-18
New plugin: WeatherRPC, a RPC to show the temperature in a city.

iakzs:
Hello!

I just did this plugin which lets you show the temperature and chance of raining on your RPC. All using open sourced and free APIs.

image

image

image
(Sorry for the different sizes of the images 🙏 )

If there's any needed changes please tell me 

Thanks, iakzs

PR #3355, opened 2025-04-05
New Plugin: WeatherPeek

tiso99:
Simple and easy way to quickly take a peek at weather information via hotkey.

- BX

### DIFF
--- a/src/plugins/weatherPeek/index.ts
+++ b/src/plugins/weatherPeek/index.ts
@@ -1,0 +1,151 @@
+/*
+ * Vencord Plugin: WeatherPeek
+ * Description: Pops up a weather overlay in Discord UI when a hotkey is pressed.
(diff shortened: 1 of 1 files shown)


jamesbt365:
This doesn't have any use really? Use actual weather widgets instead of packing your chat application with every non chat related feature?

#2957, a plugin that added RPC for weather was denied for simply just having no use either, Weather information has no place in this client mod.

This also goes against the [plugin submission rules](https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md#plugin-rules) as you are using DOM manipulation.

Also, your git skills are lacking a lot, you made a second PR just for your changes to the dev constant? It will not pass CI if its not within the same PR.

jamesbt365:
Also note that this is the second time you have ignored the plugin rules when submitting a plugin

PR #3358, opened 2025-04-06
ShowHiddenChannels: show 'Last message sent' instead of 'Last message created'

Rayanzay:
This clears up confusion especially when viewing a locked text channel

Previous Text Channel:
Same thing as the below screenshot, but with "Last message created."

New Text Channel:
image

Forum Channel is kept the same:
image


Inbestigator:
How is that confusing?

PR #3361, opened 2025-04-06
New Vencord Plugin (Server Access)

RlxChap2:
## New Server Access Plugin

folder path: `Vencord\src\plugins\serverAccess\serverAccess.js`

# Image before and after Usage:

## Before:

image

## After:

image
image
image
image

### DIFF
--- a/src/plugins/serverAccess/serverAccess.js
+++ b/src/plugins/serverAccess/serverAccess.js
@@ -66,7 +66,7 @@ module.exports = class ServerAccessPlugin {
 module.exports.meta = {
     name: "ServerAccess",
     description: "Bypass server permission checks and gain admin access",
-    authors: [Devs.YourNameHere],
+    authors: [Devs.RlxChap2],
     dependencies: [],
     settings
 };
--- a/src/plugins/serverAccess/serverAccess.js
+++ b/src/plugins/serverAccess/serverAccess.js
@@ -1,0 +1,72 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2025 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+const { definePluginSettings } = require("@api/Settings");
+const { Devs } = require("@utils/constants");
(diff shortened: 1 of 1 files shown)


src/plugins/serverAccess/serverAccess.js
```diff
@@ -0,0 +1,72 @@
...
+const { definePluginSettings } = require("@api/Settings");
+const { Devs } = require("@utils/constants");
+const { find, findByProps } = require("@webpack");
+const { Patcher, Webpack: { getModule } } = require("@vencord/types");
+
+const settings = definePluginSettings({
+    enabled: {
+        type: "boolean",
+        description: "Enable server access",
+        default: true,
```
RlxChap2, line 16: also you can change the default button from `true` to `false`

nyakowint:
You didnt even try did you

RlxChap2:
> You didnt even try did you

Nah i tried it

I tried it with `userPlugins` folder then moved to `plugins` folder

src/plugins/serverAccess/serverAccess.js
jamesbt365, line 16: Why would this be enabled by default?

jamesbt365 (commented):
Not only that, but this is unsafe and will send a fuck ton of invalid requests, this has no reason to be in Vencord and will just be a potential danger to peoples discord account, this will never be an official plugin.

Also, you aren't even in the dev constants.

src/plugins/serverAccess/serverAccess.js
RlxChap2: > Why would this be enabled by default?

to enable plugin when you start

RlxChap2:
> Not only that, but this is unsafe and will send a fuck ton of invalid requests, this has no reason to be in Vencord and will just be a potential danger to peoples discord account, this will never be an official plugin.
> 
> Also, you aren't even in the dev constants.

actually yeah i tried to do something new in vencord

Vendicated:
AI cannot write vencord plugins.

PR #3363, opened 2025-04-06
Make NoDevtoolsWarning required

JuniorWolfgamingDE:
Since NoDevtoolsWarning also fixes session issues, making NoDevtoolsWarning a required plugin would get rid of reoccurring issues being opened up on logouts.

The plugin's impact is minimal aside from the session fix and removing the console banner, so nothing else is changed that could potentially get in the way of the user.


Vendicated:
No. I will add a patch to Vesktop instead

PR #3365, opened 2025-04-06
VcNarrator: Fix Voice selection setting

Vendicated:
Need people to test this on Windows & Mac and let me know if it works fine


Marocco2:
image
I don't see any options

PR #3374, opened 2025-04-10
feat(plugin): Show Role Icons in compact mode

jamesbt365:
Relatively simple plugin, only submitting it because Discord as per usual broke (either intentionally or unintentionally) being able to do this before.

Before discord released the new UI, you could just enable avatars in compact mode, then just see out the avatar element if you didn't want it. Now even while showing the avatar, you don't see the role icon.

Could be seen as niche, but I'd like this feature back and it doesn't seem like a burden for Vencord, so here we go.


mja00:
Would be great to have this plugin

PR #3376, opened 2025-04-12
youtubeAdblock: fix plugin so it can find the youtube embed

Impqxr:
After new changes, the plugin is broken for Watch Together. This PR fixes it

### DIFF
--- a/src/plugins/youtubeAdblock.desktop/native.ts
+++ b/src/plugins/youtubeAdblock.desktop/native.ts
@@ -5,16 +5,28 @@
  */
 
 import { RendererSettings } from "@main/settings";
-import { app } from "electron";
+import { app, WebFrameMain } from "electron";
 import adguard from "file://adguard.js?minify";
 
 app.on("browser-window-created", (_, win) => {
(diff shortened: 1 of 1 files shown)


Impqxr:
i think this solution will be better

Impqxr:
@Vendicated any thoughts? maybe it's better to just close it and leave this plugin in a broken state?

Oddward:
Damn, your fix works well, been waiting ages for it to get merged 🗿 
I think you should create an open issue describing the problem and then link the pull request that solves it, better workflow for it to be seen.

Impqxr:
> Damn, your fix works well, been waiting ages for it to get merged 🗿 I think you should create an open issue describing the problem and then link the pull request that solves it, better workflow for it to be seen.

vendicated doesn't want to fix it, use custom vencord

Kodarru:
You could always bring up the PR in the core-development channel. Maybe it got lost

Impqxr:
> You could always bring up the PR in the core-development channel. Maybe it got lost

i don't have access to it and i don't see "I want my PR to be reviewed and merged" in modmail

PR #3379, opened 2025-04-13
Plugin: InlineForward

surgedevs:
Swaps out the foward model to display the to be forwarded message in the message bar

image


surgedevs:
needs review, also taking better plugin name suggestions

PR #3381, opened 2025-04-14
SpotifyControls: fix SeekBar not updating

Vendicated:


src/webpack/common/internal.tsx
```diff
@@ -16,19 +16,19 @@
...
-export function waitForComponent<T extends React.ComponentType<any> = React.ComponentType<any> & Record<string, any>>(name: string, filter: FilterFn | string | string[]): T {
+export function waitForComponent<T extends React.ComponentType<any> = React.ComponentType<any> & Record<string, any>>(name: string, filter: FilterFn | string | string[]) {
     if (IS_REPORTER) lazyWebpackSearchHistory.push(["waitForComponent", Array.isArray(filter) ? filter : [filter]]);
 
     let myValue: T = function () {
         throw new Error(`Vencord could not find the ${name} Component`);
     } as any;
 
-    const lazyComponent = LazyComponent(() => myValue) as T;
+    const lazyComponent = LazyComponent(() => myValue) as LazyComponentWrapper<T>;
```
Nuckyz, line 31:
```suggestion
    const lazyComponent = LazyComponent(() => myValue);
```

I think this isn't needed anymore

Nuckyz (changes requested):

src/webpack/common/internal.tsx
Vendicated: typescript is goofball

PR #3121, opened 2025-01-03
feat(Settings)!: implement array types

EepyElvyra:
This PR implements: #615 #2210

New option types:
- `OptionType.ARRAY`
- `OptionType.USERS`
- `OptionType.CHANNELS`
- `OptionType.GUILDS`

these replace the previously used strings with separators.
Additionally, context menu entries are generated for all user, channel and guild options.

Since IDs are not user-friendly, this PR additionally implements a Modal replicating Discord's forwarding Modal to search for users/channels/guilds by text.

**Warning: make a backup of your settings before trying this for testing as the changes made are incompatible with the current codebase and things will break if you rollback.**

Issue #2210, opened 2024-02-26
[API PROPOSAL] Allow/Deny list


pwall2222:
### Content

So, I was on the Vencord discord server, and brought an idea for a plugin.
An allow/deny list for the `NoReply` plugin, which brought \@sqaaakoi to comment that there should be an API.

I decided to bring it upon myself to create such request, which since I've kinda dealt with this issue in the past I wanna try my best to help.

## What would be included?

Allow/Deny listing:

- By ID
- By Name Substring
- By Guild Role

## Priority

I think a sensible priority would be:

ID > Name > Guild Role

Since we want to always respect user decision first, then the most notable part of the personalization (people forget changing their roles a lot), roles and then default.

## Techincal details

I have little to no idea about the other Vencord API, but we could store this info in a per plugin store or just let plugins handle it themselves.

## API Examples

```js
const default = true;
const allowed = isAllowListed(userId, default);
```

or 

```js
const allowed = isAllowListed(configStore, userId);
```

### Request Agreement

- [X] I have read the requirements for opening an issue above

Sqaaakoi:
When I mentioned API, I was also thinking
- Settings component for managing these lists (maybe we could steal FriendRow or the report menu user component as it has a button)
- Auto-generated context (sub)menu on users (and roles) to manage them being in lists

I am not sure how it would be stored, but I think storing roles should be separate from the list of users

Kyuuhachi:
I think the `default` parameter sounds pretty useless — just do `isUserListed(userId) ?? true`.

Also if there's a context menu, that would require users of such lists to be registered somehow so that you can choose which filter to add them to. Not necessarily an issue, but needs to be addressed.

Vendicated:
yep we already discussed about this in the past

my basic idea was to allow plugins to define white-/blacklists with a function similar to defineSettings:
```js
const lists = defineLists({
	userWhitelist: {
		type: "user",
		name: "Users to log messages from"
	}
})

const isUsedWhitelisted = lists.userWhitelist.includes(someId);
```

this would then automatically add a context menu and settings ui for you and
![image](https://github.com/Vendicated/Vencord/assets/45497981/b8659677-9790-47e0-9a91-708c09808ade)

Sqaaakoi:
> yep we already discussed about this in the past
> 
> my basic idea was to allow plugins to define white-/blacklists with a function similar to defineSettings:
> 
> ```js
> const lists = defineLists({
> 	userWhitelist: {
[...]

PR #3382, opened 2025-04-14
feat(settings)!: Implement an Array OptionType

sillyElv:
This aims to replace the previously used string settings with separators, as well as generate checkboxes into context menus for easier access if the array is one of users, guilds, channels or roles.

This leads to new option types being added:
- `OptionType.ARRAY`
- `OptionType.USERS`
- `OptionType.CHANNELS`
- `OptionType.GUILDS`
- `OptionType.ROLES`

Furthermore, this PR adds a user-friendly way to select the objects without forcing the user to manually find the IDs they need by using the SearchableSelect component. If the user has developer mode enabled, they will be shown an additional text-input field for the ID.

The checkboxes come from the proposal in 2210, but I've felt making a different API for lists would be introducing too much overhead, as essentially it would just be an exact copy of the existing API for settings. Therefore we can just keep the current API and add the functionality of Arrays into it, which was requested anyway. 

implements: #615  #2210 (kinda)

<sub>This is a remake of #3121, but better.</sub> 

**This PR modifies your settings file in a way that is incompatible with the current code. If you wish to try it out, back up your settings!** 
<sup> There are no implications for end-users affected by this in a merge, as long as no rollback happens.</sup>


EepyElvyra:
this should be done now :3 this time it uses selects to display the values instead of whatever ugly thing i was cooking last time

EepyElvyra:
@CodiumAI-Agent /review

sillyElv:
added a role array type as well now

sillyElv:
i split the array types into different files now for so its more maintainable and more readable

PR #3383, opened 2025-04-15
Fix clipboard api, IrcColors, and MemberListDecoratorsAPI

sadan4:
The fix for IrcColors could be made in a way where it doesn't match the exact value; however, I don't think it will change.

### DIFF
--- a/src/plugins/ircColors/index.ts
+++ b/src/plugins/ircColors/index.ts
@@ -73,8 +73,8 @@ export default definePlugin({
         {
             find: "#{intl::GUILD_OWNER}),children:",
             replacement: {
-                match: /(typingIndicatorRef:.+?},)(\i=.+?)color:null!=.{0,50}?(?=,)/,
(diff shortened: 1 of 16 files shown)


sadan4:
They pushed stable

src/plugins/ircColors/index.ts
```diff
@@ -73,8 +73,8 @@ export default definePlugin({
         {
             find: "#{intl::GUILD_OWNER}),children:",
             replacement: {
-                match: /(typingIndicatorRef:.+?},)(\i=.+?)color:null!=.{0,50}?(?=,)/,
-                replace: (_, rest1, rest2) => `${rest1}ircColor=$self.calculateNameColorForListContext(arguments[0]),${rest2}color:ircColor`
+                match: /(?<=color:)null!=\i\?\i:void 0/,
+                replace: (_, rest1, rest2) => "$self.calculateNameColorForListContext(arguments[0])"
```
Vendicated, line 77:
```suggestion
                match: /(?<=roleName:\i,)color:/,
                replace: "color:$self.calculateNameColorForListContext(arguments[0]),originalColor:"
```

PR #3384, opened 2025-04-15
fix: exclude whole support category from venbot in messagelogger

khcrysalis:
image

# 

So this doesn't happen, doesn't happen when a new channel is created here with vencord-support

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -23,7 +23,7 @@ import { updateMessage } from "@api/MessageUpdater";
 import { Settings } from "@api/Settings";
 import { disableStyle, enableStyle } from "@api/Styles";
 import ErrorBoundary from "@components/ErrorBoundary";
-import { Devs } from "@utils/constants";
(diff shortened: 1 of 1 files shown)


src/plugins/messageLogger/index.tsx
```diff
@@ -295,8 +295,8 @@ export default definePlugin({
             ignoreChannels.includes(ChannelStore.getChannel(message.channel_id)?.parent_id) ||
             (isEdit ? !logEdits : !logDeletes) ||
             ignoreGuilds.includes(ChannelStore.getChannel(message.channel_id)?.guild_id) ||
-            // Ignore Venbot in the support channel
-            (message.channel_id === "1026515880080842772" && message.author?.id === "1017176847865352332");
+            // Ignore Venbot in the support channels
+            (ChannelStore.getChannel(message.channel_id)?.parent_id === SUPPORT_CATEGORY_ID && message.author?.id === VENBOT_USER_ID);
```
Vendicated, line 299:
```suggestion
            (message.author?.id === VENBOT_USER_ID && ChannelStore.getChannel(message.channel_id)?.parent_id === SUPPORT_CATEGORY_ID);
```

PR #3386, opened 2025-04-18
Fix missing background on notifications

VanillaSixtySix:
I've noticed from a couple different devices/platforms that notifications no longer have backgrounds. I don't think this intentional! I added a `background-color` for what I think looks right from memory.

Screenshot 2025-04-18 at 8 47 10 AM

Screenshot 2025-04-18 at 8 37 11 AM


Vendicated:
there already is a background but it doesn't work on nitro themes cause the variable is transparent. Just gotta use some other variable

VanillaSixtySix:
Is there a reason you're using `--bg-overlay-floating` or would it be fine to just use `--background-base-low`? Stylistic choice?

PR #3387, opened 2025-04-18

sadan4:


src/plugins/accountPanelServerProfile/index.tsx
```diff
@@ -19,8 +19,7 @@ interface UserProfileProps {
     originalRenderPopout: () => React.ReactNode;
 }
 
-const UserProfile = findComponentByCodeLazy("UserProfilePopoutWrapper: user cannot be undefined");
-const styles = findByPropsLazy("accountProfilePopoutWrapper");
```
sadan4, line 23: They no longer wrap the profile with a class

src/plugins/accountPanelServerProfile/index.tsx
```diff
@@ -121,14 +120,18 @@ export default definePlugin({
         }
 
         const currentChannel = getCurrentChannel();
-        if (currentChannel?.getGuildId() == null) {
+        if (currentChannel?.getGuildId() == null || !UserProfile) {
```
sadan4, line 123: If the find ever fails again, the original profile will still be usable

Vendicated: this check does nothing lmao it will never be true even if the find fails

Vendicated:
```js
!UserProfile.$$vencordGetWrappedComponent()
```

src/plugins/_api/serverList.ts
```diff
@@ -34,7 +34,7 @@ export default definePlugin({
         {
             find: "#{intl::SERVERS}),children",
             replacement: {
-                match: /(?<=#{intl::SERVERS}\),children:)\i\.map\(\i\)/,
+                match: /(?<=#{intl::SERVERS}\),children:)\i\.map\(.{0,50}\.length\)/,
```
sadan4, line 37: No plugins use this, we could just remove the patch

sadan4:
the fix for SHC rn does not work, just pushing my work before i go incase anyone else has time before i get back

sadan4:
for SHC, the patches work fine, but im not sure the result is what's desired.

sadan4:
Discord has a new experiment for their sidebar (2025-04_sidebar_pinned_section), and this also fixes better folders for both treatments

src/plugins/_api/memberListDecorators/index.tsx
```diff
@@ -36,9 +36,13 @@ export default definePlugin({
                     match: /let\{[^}]*lostPermissionTooltipText:\i[^}]*\}=(\i),/,
                     replace: "$&vencordProps=$1,"
                 },
+                {
+                    match: /decorators:.{0,100}?(?=user:)/,
+                    replace: "$&vencordProps:typeof vencordProps!=='undefined'?vencordProps:void 0,"
+                },
```
Vendicated, line 42: why do we need this extra patch

sadan4: The decorators component is now it's own function so we need to pass the props needed

sadan4:
This covers everything i think i know about and should be ready for review

note: I was not able to fix this issue with mutual group dms (don't have the energy to fix it), but I did fix the issue with the tab going off the screen in the modal
image

src/plugins/_api/serverList.ts
```diff
@@ -32,9 +32,9 @@ export default definePlugin({
             }
         },
         {
-            find: "#{intl::SERVERS}),children",
+            find: '#{intl::SERVERS}),gap:"xs",children',
             replacement: {
-                match: /(?<=#{intl::SERVERS}\),children:)\i\.map\(\i\)/,
+                match: /(?<=#{intl::SERVERS}\),gap:"xs",children:)\i\.map\(.{0,50}\.length\)/,
                 replace: "Vencord.Api.ServerList.renderAll(Vencord.Api.ServerList.ServerListRenderPosition.In).concat($&)"
```
sadan4, line 38: this find/match could prob be better, but I'm too sick rn.

src/plugins/ircColors/index.ts
```diff
@@ -66,8 +66,8 @@ export default definePlugin({
         {
             find: '="SYSTEM_TAG"',
             replacement: {
-                match: /\i.gradientClassName]\),style:/,
-                replace: "$&{color:$self.calculateNameColorForMessageContext(arguments[0])},_style:"
+                match: /(style:)(.{0,500}textDecorationColor)/,
```
Vendicated, line 69: this is very bad

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -325,7 +325,7 @@ export default definePlugin({
             ]
         },
         {
-            find: '})},"overflow"))',
+            find: "overflowCountVariant",
```
Vendicated, line 328: this is a very bad find. they don't pass this prop rn so it works, but the second they add it anywhere, it will break

the only unique find i could come up for this module is `disablePopout:"function`, but that sucks too

src/plugins/mutualGroupDMs/style.css
```diff
@@ -0,0 +1,3 @@
+.vc-mutual-gdms-tab-bar {
+    gap: 0;
```
Vendicated, line 2:
```suggestion
    gap: 0;
    justify-content: space-between;
```

src/plugins/_api/serverList.ts
```diff
@@ -32,9 +32,9 @@ export default definePlugin({
             }
         },
         {
-            find: "#{intl::SERVERS}),children",
+            find: '#{intl::SERVERS}),gap:"xs",children',
```
Vendicated, line 35: we can use something like this instead
```suggestion
            find: '.setGuildsTree(',
```

src/plugins/reviewDB/index.tsx
```diff
@@ -76,25 +76,27 @@ export default definePlugin({
...
             }
         },
+        // places like the user profile on the right in dms
         {
-            find: 'location:"UserProfilePanel"',
+            find: 'location:"UserProfileSiebar"',
             replacement: {
-                match: /{profileType:\i\.\i\.PANEL,children:\[/,
-                replace: "$&$self.BiteSizeReviewsButton({user:arguments[0].user}),"
+                match: /(children:\[)(.{0,50}shouldShowTooltip:)/,
```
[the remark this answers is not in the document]
sadan4:
```suggestion
                match: /(children:\[)([^[]+shouldShowTooltip:)/,
```
this is the best i can come up with instead

src/plugins/showHiddenChannels/index.tsx
sadan4: the only better thing i could come up with is `.SIZE_24,overflowCountVariant:`

src/plugins/forceOwnerCrown/index.ts
```diff
@@ -29,7 +29,8 @@ export default definePlugin({
         {
             find: "#{intl::GUILD_OWNER}),children:",
             replacement: {
-                match: /,isOwner:(\i),/,
+                // The isOwner prop is used in more tha one component in this module. Make sure we patch the right one
+                match: /,isOwner:(\i),(?=[^}]+guildId)/,
                 replace: ",_isOwner:$1=$self.isGuildOwner(e),"
```
Vendicated, line 34: HORROR why does this hardcode e

sadan4: <:trolley:1024751352028602449>

src/plugins/forceOwnerCrown/index.ts
```diff
@@ -29,8 +29,9 @@ export default definePlugin({
         {
             find: "#{intl::GUILD_OWNER}),children:",
             replacement: {
-                match: /,isOwner:(\i),/,
-                replace: ",_isOwner:$1=$self.isGuildOwner(e),"
+                // The isOwner prop is used in more tha one component in this module. Make sure we patch the right one
+                match: /,isOwner:(\i),(?=[^}]+guildId)/,
+                replace: ",_isOwner:$1=$self.isGuildOwner(arguments[0]),"
```
Vendicated, line 34: maybe we can do something like this instead

```suggestion
                match: /(?<=decorators:.{0,200}?isOwner:)\i/,
                replace: "$self.isGuildOwner(arguments[0])"
```

sadan4: yeah, works fine.

src/plugins/ircColors/index.ts
sadan4:
```suggestion
                match: /(style:)(\(\(\)=>\{)/,
```

this is what i did, pick your poison

PR #3389, opened 2025-04-19
Added a slider to vcDoubleClick

Muminwilmer:
Added a setting to customize double-click speed for joining VCs

This update adds a small quality-of-life slider that allows you to change how fast you need to double-click to join a voice channel.

The default is still 0.5 seconds (same as before), so nothing changes unless you want it to. But now you can tweak it between 0.1 and 2 seconds if you're a slow clicker or just want it to feel a bit different.

tl;dr - more control for people who care, no difference for people who don't.

### DIFF
--- a/src/plugins/vcDoubleClick/index.ts
+++ b/src/plugins/vcDoubleClick/index.ts
@@ -17,7 +17,8 @@
 */
 
 import { Devs } from "@utils/constants";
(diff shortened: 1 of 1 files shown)


Vendicated:
i dislike this change. we should instead just change the plugin to use the doubleclick browser event which respects your OS preference as mentioned above

PR #3390, opened 2025-04-20
Vencord New Plugin

ItsAlexIK:
**Basically a new ChatBar Button for short time Reminders.**

image
image
image

### DIFF
--- a/src/plugins/messageReminder/index.tsx
+++ b/src/plugins/messageReminder/index.tsx
@@ -15,6 +15,9 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
+
+import "./styles.css";
+
 import { addChatBarButton, ChatBarButton, ChatBarButtonFactory, removeChatBarButton } from "@api/ChatButtons";
 import { sendBotMessage } from "@api/Commands";
 import { Devs } from "@utils/constants";
@@ -22,7 +25,6 @@ import { localStorage } from "@utils/localStorage";
 import { ModalContent, ModalFooter, ModalHeader, ModalRoot, openModal } from "@utils/modal";
 import definePlugin from "@utils/types";
 import { Button, Forms, SelectedChannelStore, TextInput, useState } from "@webpack/common";
-import "./styles.css";
 
 interface Reminder {
     message: string;
--- a/src/plugins/messageReminder/index.tsx
+++ b/src/plugins/messageReminder/index.tsx
@@ -15,9 +15,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
-
-import "./styles.css";
-
 import { addChatBarButton, ChatBarButton, ChatBarButtonFactory, removeChatBarButton } from "@api/ChatButtons";
 import { sendBotMessage } from "@api/Commands";
 import { Devs } from "@utils/constants";
@@ -25,6 +22,7 @@ import { localStorage } from "@utils/localStorage";
 import { ModalContent, ModalFooter, ModalHeader, ModalRoot, openModal } from "@utils/modal";
 import definePlugin from "@utils/types";
 import { Button, Forms, SelectedChannelStore, TextInput, useState } from "@webpack/common";
+import "./styles.css";
 
 interface Reminder {
     message: string;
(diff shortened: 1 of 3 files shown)


src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+        triggered: reminder.triggered
+    }));
+    localStorage.setItem("equicord-reminders", JSON.stringify(remindersToSave));
+}
+
+/**
+ * Loads saved reminders from localStorage
+ */
+function loadRemindersFromStorage() {
+    const savedReminders = localStorage.getItem("equicord-reminders");
```
jamesbt365, line 257: Remnants of your attempts to put this in Equicord. Equicord/Equicord#237

src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+            if (document.visibilityState === "visible") {
+                checkReminders();
+            }
+        });
+    },
+    stop: () => {
+        removeChatBarButton("Reminder");
+        if (globalCheckInterval) clearInterval(globalCheckInterval);
+        activeReminders = [];
+        localStorage.removeItem("equicord-reminders");
```
jamesbt365, line 361: Ditto

src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+    if (type === "trigger") {
+        try {
+            const audio = new Audio();
+            audio.src = "https://discord.com/assets/dd920c06a01e5bb8b09678581e29d56f.mp3";
+            audio.volume = 0.5;
+            audio.play();
+        } catch (e) {
+            console.error("Could not play notification sound:", e);
+        }
+    }
```
jamesbt365, line 303: This will not respect notification settings

src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+        },
+        timestamp: new Date(exprireAt).toISOString(),
+    };
+
+    sendBotMessage(channelId, {
+        content: "",
+        username: `${emoji} Reminder ${emoji}`,
+        avatar_url: "",
+        embeds: [embed]
+    } as any);
```
jamesbt365, line 292: Theres a decent chance the user will just change channel before noticing this, or it'll just run off the screen when the user isn't looking.

I get that you only check when the window is visible (if i read that right), what use are reminders if they don't proc at the time they are supposed to be delivered?

src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+
+        globalCheckInterval = setInterval(() => {
+            checkReminders();
+        }, 1000);
+
+        document.addEventListener("visibilitychange", () => {
+            if (document.visibilityState === "visible") {
+                checkReminders();
+            }
+        });
```
jamesbt365, line 355: insane

src/plugins/messageReminder/index.tsx
```diff
@@ -0,0 +1,363 @@
...
+        <ChatBarButton tooltip="Set Reminder" onClick={handleClick}>
+            <svg width="25" height="25" viewBox="0 0 24 24">
+                <path fill="currentColor" d="M12 22c-.825 0-1.413-.587-1.413-1.412h4c0 .825-.588 1.412-1.413 1.412H12Zm-9-3v-2h2V9c0-2.925 1.95-5.075 4.95-5.075 3.1 0 5.05 2.15 5.05 5.075v7h2v2H3Z" />
+            </svg>
+        </ChatBarButton>
+    );
+};
+
+export default definePlugin({
+    name: "Message Reminder",
```
jamesbt365, line 340: This doesn't fit the current naming scheme

jamesbt365 (commented):
A plugin to this has already been denied, #2269, so this won't be accepted for the same reasons.

Also, IMO this is just worse, you're attempting to emulate an app, so why not just use an app which can do this better, or just use a native program on your computer?

This doesn't really have a spot in Vencord.

src/plugins/messageReminder/index.tsx
ItsAlexIK: You get notification when u click the channel back

ItsAlexIK: It can be any channel you want but yes you will get the notification

src/plugins/messageReminder/index.tsx
ItsAlexIK: Yeah cuz it's supposed to remind you even with DND

ItsAlexIK:
> This doesn't really have a spot in Vencord.

Welp that's okay can close it if there's no spot for it.

PR #3397, opened 2025-04-23
Fix spelling mistake with plugin description

zcraftelite9495:
Noticed the error in client so wrote up a quick pull request so a fix can be included in the next Vencord update.


jamesbt365:
#3380, #3369.

Vendicated:
this is not a spelling mistake

PR #3400, opened 2025-04-24
Added Plugin hideBlockedUsersNameplate

LunaLu-dev:
Added a plugin to hide all users blocked by the user from the server member list


lolsuffocate:
Would it be better to add this as a setting for NoBlockedMessages instead?

Vendicated:
image

PR #3403, opened 2025-04-25
New plugin: sortMessagesByReactions

bvoq:
As the description says:
Adds a button to the message bar. If you click it, it will load 100 more messages (Discord API limit, see: https://discord-api-types.dev/api/next/discord-api-types-v10/interface/RESTGetAPIChannelMessagesQuery) and sort them by reaction count. Pressing the button multiple times, will load more messages and sort all of the already loaded messages. It further displays how far back messages are loaded.

This is particularly useful for browsing memes on a meme channel :-)
Here's an example of the plugin in action (note the date is backwards).
image

### DIFF
--- a/src/plugins/sortMessagesByReactions/README.md
+++ b/src/plugins/sortMessagesByReactions/README.md
@@ -1,0 +1,1 @@
+Adds a button to the message bar. If you click it, it will load 100 more messages (Discord API limit) and sort them by reaction count. Pressing the button multiple times, will load more messages. It further displays how far back messages are loaded.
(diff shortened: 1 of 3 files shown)


Vendicated:
thank you for your contribution, but this seems way too niche and also doesn't really do things in a good way. you're
- polluting their stored data instead of just patching the message list renderer
- dispatching MESSAGE_UPDATE events with the wrong datatype. Discord has both api messages (raw json as received by the api with snake_case) and message "records" (the same messages converted to js classes with snakeCase). MESSAGE_UPDATE expects api messages, but you're dispatching it with records. This will break attachments, embeds and maybe more

PR #3405, opened 2025-04-27
Add setting to Spotify Controlls

FroggyPanda:
Added the "Show Album Info On Album Expand" setting to show track name, artist name, and album name when the album image is expanded.


Vendicated:
what's the point? if you want big cover always, just use css

FroggyPanda:
> what's the point? if you want big cover always, just use css

when you have the big cover it removes the title, artist, and the album of the song id personally like to see those things still

FroggyPanda:
> what's the point? if you want big cover always, just use css

true, just thought a setting would be good too

PR #3396, opened 2025-04-23
Fix betterFolders, reviewDB, showHiddenChannels

sadan4:

PR #3406, opened 2025-04-28
Fix BetterFolders plugin

alfuwu:
This PR fixes the Better Folders plugin's sidebar, and additionally turns the Sidebar Anim boolean config value into a slider determining the time it takes for the animation to finish for more customizability.

For the past few days, whenever I opened a folder with the Better Folders plugin while having the custom sidebar enabled, it would simply provide me with a blank bar and no guilds inside it.
image

This PR restores the functionality to how it worked before:
image

After writing the fix I found #3396, so if this is a duplicate, please close it.

### DIFF
--- a/src/plugins/betterFolders/index.tsx
+++ b/src/plugins/betterFolders/index.tsx
@@ -317,11 +317,21 @@ export default definePlugin({
         return child => {
             if (!isBetterFolders) return true;
 
-            if (child?.type === "ul" && child.props?.children != null) {
+            if (child?.type === "ul" && child.props?.children?.props?.children?.props?.children !== null) {
                 // Filter out everything but the guild list (removes the DM button & favorites button)
-                child.props.children.props.children.props.children = child.props.children.props.children.props.children.filter(child => child?.props?.renderTreeNode != null)[0];
-                // Assign properties of the sidebar somewhere where the other functions can read them
-                child.props.children.props.children.props.children.props = { ...child.props.children.props.children.props.children.props, isBetterFolders, betterFoldersExpandedIds };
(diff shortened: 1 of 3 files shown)


sadan4:
Does this cover both experiment treatments

lem0n0fficial:
Still active like that too;
image

alfuwu:
I believe it should support the experiments now, but if it doesn't please lmk w/ what experiment is breaking it still

src/plugins/betterFolders/index.tsx
```diff
@@ -311,13 +313,25 @@ export default definePlugin({
...
+                    // The bottomSection has some junk inside it, so we need to clean it out
+                    child.props.children.props.children.props.children.props.children.props.children = child.props.children.props.children.props.children.props.children.props.children.filter(child => child?.props?.renderTreeNode != null)[0];
+                    // Now we can assign the properties we want to the guilds list so that the other functions can read them
+                    child.props.children.props.children.props.children.props.children.props.children.props = { ...child.props.children.props.children.props.children.props.children.props.children.props, isBetterFolders, betterFoldersExpandedIds };
+                } else {
+                    // No experiment is enabled, so we just go straight to assigning properties to the guilds list
+                    child.props.children.props.children.props.children.props = { ...child.props.children.props.children.props.children.props, isBetterFolders, betterFoldersExpandedIds };
+                }
```
Vendicated, line 334: just no

alfuwu: Okay, fair, it looks ugly. However, the only alternative I see is to apply many more patches (like Sadan did in their fix).

PR #3407, opened 2025-04-28
fix ImageZoom for animated webp images

hexa0:
Animated webps won't animate without `?animated=true` being appended to the URL,
this doesn't currently check for the type of image before appending but from my testing this hasn't broken anything so it should be good enough


Nuckyz:
Thanks for the contribution!

PR #3409, opened 2025-04-30
PlainFolderIcon: Fix plugin not working

sadan4:
Discord now uses css to switch bewteen the plain and mini guild folder icons.

However, the folder icons are transparent and look terrible, so set a sane background color (same as home icon)

### DIFF
--- a/src/plugins/betterFolders/index.tsx
+++ b/src/plugins/betterFolders/index.tsx
@@ -61,7 +61,7 @@ function filterTreeWithTargetNode(children: any, predicate: (node: any) => boole
             return true;
         }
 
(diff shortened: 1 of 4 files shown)


Etsi0:
I've added this to my QuickCSS until it's fixed:
```css
[class*="folderButtonContent"] {
    transform: none;
    background-color: hsl(from var(--custom-folder-color) h s sqrt(l));
}
```

ItsPasi:
I still don't see this as working, has this really been fixed?

sadan4:
> I still don't see this as working, has this really been fixed?

use our support channel so we can help you

PR #3415, opened 2025-05-05
Allow LastFM plugin to load credentials from path

roobscoob:
Storing secrets in my config.json is less than ideal!

This PR moves to allow the last.fm rich presence plugin to load secrets from a path. Used in combination with a credential manager it can provide a more secure way to store credentials.


Vendicated:
You just added an arbitrary file read which is much much more servere than storing an api key in the settings

There is nothing stopping a potential attacker (who is trying to steal your lastfm key?) from just reading this alternative file instead

Also lastfm api keys are essentially public information because of how their api works. As soon as you make an oauth app, any user authorising your app gets your key. Try authorising openscrobbler or fmbot and pay close attention to the lastfm URL it provides you. (You could even use their key)

PR #3420, opened 2025-05-10
feat(plugin): FixAudioFiles

prodbyeagle:
this plugin ensures that only one audio source plays at a time. when a new sound starts, any currently playing sound is automatically paused. this can especially be useful for quickly previewing sound effects or samples.


Nuckyz:
Does this really need to monkey patch Audio.play or is this issue fixable by a patch?

Vendicated:
Thanks for your contribution, but we will have to reject it because the way it is written is bad. There are other audios inside Discord (for example notification sounds) that will conflict with this, and polluting global prototypes like this is very bad practice

PR #3191, opened 2025-01-31
feat(plugin): CopyStickerLinks

byeoon:
This plugin adds support to copy and open sticker urls from the context menu. Normal emojis have this feature, but stickers do not. This makes it much more complicated to do *simple* things such as download stickers or open them in a browser. Discord makes it very hard to find the sticker URLs, so this plugin's goal is to get around that.

There have been many times where I myself and others have wanted to access a sticker url, but have not been able to. This is a QoL feature that discord should have added since a long time ago.

**Why is this useful?**
- Being able to directly copy sticker links into the chat (if they are animated, you can even favorite them into gifs)
- Opening them into your browser in 1 click to download or do whatever 
- Emojis **already have this feature** and the fact stickers can't do this is very annoying.
- Discord overcomplicates finding Sticker URLs, so this is the most direct way of getting to them.

PR #3423, opened 2025-05-13
New Plugin: stickerLink

Commandtechno:
this what it look like
image

always got annoyed that they only added it for emojis
dylan


Commandtechno:
dupe of #3191 mb
cat boom

PR #3431, opened 2025-05-16
Vencord language | Vencord Translation 

Starebel:
Sorry for reopening the PR! I closed the previous one myself.

This is a working version of Vencord with added support for the Russian language. A new option has been added in the settings to choose the interface language.

A directory structure for language files has been implemented, along with a new settings tab for selecting the language. Currently, the main interface sections have been translated.

Translations are loaded using the `getLanguage()` function from `Language.ts`. This function takes the category type as an argument (`api`, `components`, or `plugins`), and for plugins — also the plugin name. Each category has a dedicated JSON file for its translations.

To use a custom local JSON file, the `defineLanguage()` function has been added. It accepts the plugin name and an object in the format `{ language: file }`. Currently, plugin translations are stored in `/src/language/{language}/plugins`, but they can also reside in the plugin’s local directory.

There is also a plan to integrate automatic translations via Google Translate. However, since machine translations can be inaccurate and affect clarity, this is currently being considered as a temporary solution.


khcrysalis:
This is an extremely unacceptable pull request. Mainly due by the fact that theres *many* unneccessary file deletions, including permission changes, and gitignore modifications additions that have important source code inside of them.

Vendicated:
Sorry but we really can't merge this :( Way too many files changed and if we were to add localisation it would be done by a core maintainer

We were actually thinking about & working on localisation but we decided against it

Sorry and thank you for your attempt & efforts!!

PR #3436, opened 2025-05-19
Updater: fix network errors triggering popups

Vendicated:


src/Vencord.ts
```diff
@@ -134,7 +134,11 @@ async function init() {
 
     if (!IS_WEB && !IS_UPDATER_DISABLED) {
         runUpdateCheck();
-        setInterval(runUpdateCheck, 1000 * 60 * 30); // 30 minutes
+
+        // this tends to get really annoying, so only do this if the user has auto-update without notification enabled
+        if (Settings.autoUpdate && !Settings.autoUpdateNotification) {
+            setInterval(runUpdateCheck, 1000 * 60 * 30); // 30 minutes
```
Nuckyz, line 140: What if the user changes one of these settings after startup, should it matter?

src/main/updater/common.ts
```diff
@@ -35,7 +35,10 @@ export function serializeErrors(func: (...args: any[]) => any) {
                 ok: false,
                 error: e instanceof Error ? {
                     // prototypes get lost, so turn error into plain object
-                    ...e
+                    ...e,
```
Nuckyz, line 38: What's the reason for adding each one manually?

src/main/updater/http.ts
```diff
@@ -32,7 +32,7 @@ const API_BASE = `https://api.github.com/repos/${gitRemote}`;
 let PendingUpdates = [] as [string, string][];
 
 async function githubGet(endpoint: string) {
-    return get(API_BASE + endpoint, {
+    return checkedFetch(API_BASE + endpoint, {
```
Nuckyz, line 35: Should just use fetchJson and avoid having to later call .json on each

Nuckyz (changes requested):

src/main/updater/common.ts
Vendicated: non enumerable properties. no clue how this code ever worked lol
image

src/Vencord.ts
Vendicated: eh it's whatever

PR #3437, opened 2025-05-23
feat: Add more kaomojis and variants

Harshcombo50548:
Updated plugin description to note that all kaomojis are slash commands and added some more kaomojis.


Vendicated:
honestly we should just remove this plugin cause it can easily be done with textreplace

PR #3446, opened 2025-05-29
ShowMeYourName: Fix emojis getting filled with the gradient role colour

lolsuffocate:
Prevents emojis getting filled with colour when a user has gradient/holographic role colours.
Also fixes the glow behind the text not matching the actual text and the muted colour of the suffix.

image

image


src/plugins/showMeYourName/index.tsx
```diff
@@ -42,6 +42,18 @@ const settings = definePluginSettings({
     },
 });
 
+function wrapEmojis(name: string | undefined) {
```
Nuckyz, line 45: does Discord not export a function to do this?

src/plugins/showMeYourName/index.tsx
```diff
@@ -68,18 +70,46 @@ export default definePlugin({
...
+                return <>{prefix}{parsedNick} <span className="vc-smyn-suffix">{parsedUsername}</span></>;
 
-            return <>{prefix}{username}</>;
+            return <>{prefix}{parsedUsername}</>;
         } catch {
             return <>{author?.nick}</>;
         }
     }, { noop: true }),
+
+    getUsername: ({ author, message, isRepliedMessage, withMentionPrefix, userOverride }: UsernameProps) => {
```
Vendicated, line 108: why do we need this? and why so much code repetition?

lolsuffocate: It provides the text for the glow effect behind the username, without this if you use any setting that isn't displayname first it won't match up. Code repetition is just laziness because one needs the suffix span and the other just needs text

lolsuffocate:
Stopped being lazy and did it properly

PR #3447, opened 2025-05-29
Add BlockedMessageSound Plugin

Zekerocks011:
made a plugin that will play a sound effect in ur current channel whenever someone you blocked sends a message.

i personally use the mr washee washee bing bong sound effect


Vendicated:
amazing

PR #3448, opened 2025-05-31
LastFMRichPresence: Attempt to get images for singles

BlockBuilder57:
On last.fm, scrobbles from services like Spotify and Apple Music scrobble singles with the single name as the album name. This creates albums with the same name as the single. Users can add cover art to singles this way. However, this method does not show up with the `user.getrecenttracks` method, which means "missing" art for said singles.

This PR fixes this by checking for an album of the same name as the single.

### DIFF
--- a/src/plugins/lastfmRichPresence/index.tsx
+++ b/src/plugins/lastfmRichPresence/index.tsx
@@ -251,13 +251,38 @@ export default definePlugin({
             if (!trackData?.["@attr"]?.nowplaying)
                 return null;
 
+            let imageUrl = trackData.image?.find((x: any) => x.size === "large")?.["#text"] || "";
+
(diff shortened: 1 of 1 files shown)


Vendicated:
what? if it doesn't get the cover, it means you're scrobbling wrong (i.e. scrobbling with no album info)

Vendicated:
yes album name is optional. but that means there is no album attached to your track and thus no album art. lastfm itself will also not show any album art on your feed. i'm really not a fan of the plugin attempting to fix broken scrobbles like this, you really should just be scrobbling properly instead!

you might be interested in https://picard.musicbrainz.org to fix your tags instead, pretty sure it can do bulk tagging of your library

Vendicated:
why not just make a playlist? or what I sometimes do is make an unofficial compilation and upload a cover I like to lastfm

Vendicated:
maybe time to look for a new player :P

and all good, thank you for your effort anyway!

PR #3458, opened 2025-06-03
fix: \\i crashes patchHelper

cootshk:
Putting a `\\i` into the `match` box with when it is already showing a diff will crash Patch Helper

image

### DIFF
--- a/src/utils/patches.ts
+++ b/src/utils/patches.ts
@@ -40,7 +40,12 @@ export function canonicalizeMatch<T extends RegExp | string>(match: T): T {
         return partialCanon as T;
     }
 
-    const canonSource = partialCanon.replaceAll("\\i", String.raw`(?:[A-Za-z_$][\w$]*)`);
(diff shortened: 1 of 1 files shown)


cootshk:
~~am dumb and broke something, 1 sec~~

being more dumb, it works just fine

cootshk:
video: https://discord.com/channels/1015060230222131221/1015063227299811479/1379577041749278861

Vendicated:
thanks but there's no use case for this. vencord's patcher api has the exact same flaw so this wouldn't even be a valid patch. there's no reason to have \\i in a patch

cootshk:
> thanks but there's no use case for this. vencord's patcher api has the exact same flaw so this wouldn't even be a valid patch. there's no reason to have \i in a patch

I write my patches in the patchhelper window so it's something I run into quite a bit because i'm editing text directly before a `\i`

cootshk:
also I'm pretty sure this fixes the underlying patcher as well

PR #3460, opened 2025-06-04
ShowHiddenChannels: Fix showing lock icon for special channels

cootshk:
`Server Guide`, `Channels & Roles`, and `Browse Channels` now correctly show their icons in the chat channel selection bar.

Before:

image

After:

image

### DIFF
--- a/src/plugins/showHiddenChannels/index.tsx
+++ b/src/plugins/showHiddenChannels/index.tsx
@@ -498,6 +498,7 @@ export default definePlugin({
 
             if (channel.channelId) channel = ChannelStore.getChannel(channel.channelId);
             if (!channel || channel.isDM() || channel.isGroupDM() || channel.isMultiUserDM()) return false;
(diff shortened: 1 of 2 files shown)


src/plugins/showHiddenChannels/index.tsx
```diff
@@ -498,6 +498,7 @@ export default definePlugin({
 
             if (channel.channelId) channel = ChannelStore.getChannel(channel.channelId);
             if (!channel || channel.isDM() || channel.isGroupDM() || channel.isMultiUserDM()) return false;
+            if (["browse", "customize", "guide"].includes(channel.id)) return false;
 
             return !PermissionStore.can(PermissionsBits.VIEW_CHANNEL, channel) || checkConnect && !PermissionStore.can(PermissionsBits.CONNECT, channel);
```
Nuckyz, line 503: do they not have those permissions, why are they returning true

cootshk: The special channels don't have those perms because discord

PR #3461, opened 2025-06-05
.userscript and .extension targets + setting a CSP in the userscript

cootshk:
Edit the Content Security Policy on load to allow `vencord.dev`, `*.vencord.dev`, and `localhost` (required for devCompanion) domains to load in Firefox

Make a request to `https://discord.com/app` and get the `Content-Security-Policy` header, falling back to a hardcoded one if it's missing for some reason

Then just add the urls and add it to a `<meta>` tag in the header


Vendicated:
I don't think this even works?

From mdn:
> Adding additional [CSP] policies can only further restrict the capabilities of the protected resource

cootshk:
works on my machine™

cootshk:
well so we now have `.userscript` and `.extension` plugin targets

cootshk:
Will set this as able to review because 1. the .userscript and .extension targets and also 2. it does work if you have disabled CSP in tampermonkey

to disable csp in tampermonkey (don't do this unless you only use tampermonkey for discord):
- go to tampermonkey's settings
- go to general and set config mode to advanced
- scroll down to security
- set Modify existing content security policy (CSP) headers to Remove entirely
Also, you'll most likely want to:
- set page filter mode to both
- add `*discord.*/*` to your whitelist

cootshk:
\*note that this applies to both Chrome and Firefox

scripts/build/common.mjs
```diff
@@ -164,6 +164,8 @@ export const globPlugins = kind => ({
 
                     if (target && !IS_REPORTER) {
                         const excluded =
+                            (target === "userscript" && IS_EXTENSION && kind === "web") ||
+                            (target === "extension" && !IS_EXTENSION && kind === "web") ||
```
Vendicated, line 168: this isn't right

scripts/build/common.mjs
```diff
@@ -131,7 +131,7 @@ export const makeAllPackagesExternalPlugin = {
 /**
  * @type {(kind: "web" | "discordDesktop" | "vesktop") => import("esbuild").Plugin}
  */
-export const globPlugins = kind => ({
+export const globPlugins = (kind, IS_EXTENSION = false) => ({
```
Vendicated, line 134: make sure to follow conventional naming also the typing needs to be updated

```suggestion
export const globPlugins = (kind, isExtension = false) => ({
```

scripts/build/buildWeb.mjs
```diff
@@ -26,6 +26,14 @@ import Zip from "zip-local";
 
 import { BUILD_TIMESTAMP, commonOpts, globPlugins, IS_DEV, IS_REPORTER, VERSION, commonRendererPlugins, buildOrWatchAll, stringifyValues } from "./common.mjs";
 
+function makePlugins(is_extension) {
```
Vendicated, line 29: make sure to follow conventional naming, also this function is missing typing
```suggestion
function makePlugins(isExtension) {
```

PR #3464, opened 2025-06-07
fix(csp): allow connecting to configured cloud URL

ryanccn:
Adds the host of the configured Vencord cloud URL to `connect-src` in the new Content Security Policy implementation.

### DIFF
--- a/src/main/csp.ts
+++ b/src/main/csp.ts
@@ -6,6 +6,9 @@
 
 import { session } from "electron";
 
+import { RendererSettings } from "./settings";
+
+
 type PolicyMap = Record<string, string[]>;
 
 export const ConnectSrc = ["connect-src"];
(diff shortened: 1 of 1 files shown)


Vendicated:
this is not safe technically. but yes I am aware and will fix this properly soon

Vendicated:
any value coming from the renderer is inherently unsafe and shouldn't be trusted, otherwise context isolation is pointless

a safe way to implement this would be to have ipc methods for whitelisting domains, showing an explicit native confirmation modal and then storing this in a separate settings file the renderer has no access to (NativeSettings)

that's exactly what I was planning on adding

ryanccn:
Makes sense, thanks!

Vendicated:
#3476 

image

PR #3468, opened 2025-06-07
Allow cdnb.artstation.com as Media source

Taner314:
Adds ArtStation CDN to the CSP policies map as a MediaSrc.


nin0-dev:
we aren't willing to play the game of allowing a source for only a few users, reuploading to imgur/catbox takes like two seconds

PR #3470, opened 2025-06-07
MessageClickActions: make delete key detection consistent on lost focus

xNasuni:
when holding backspace and alt-tabbing out, the `isDeletePressed` variable stays true, which causes messages to be deleted unintentionally on click.

this bug has happened to me like **30** times on accident and it's annoying. it’s not super frequent, but when it does happen, it’s really frustrating.

this probably impacts a lot of other users. in fact i know it impacts a few of my friends using vencord.

cat

### DIFF
--- a/src/plugins/messageClickActions/index.ts
+++ b/src/plugins/messageClickActions/index.ts
@@ -20,7 +20,7 @@ import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
 import { findByPropsLazy } from "@webpack";
-import { FluxDispatcher, PermissionsBits, PermissionStore, UserStore } from "@webpack/common";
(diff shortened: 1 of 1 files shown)


src/plugins/messageClickActions/index.ts
```diff
@@ -62,11 +63,13 @@ export default definePlugin({
     start() {
         document.addEventListener("keydown", keydown);
         document.addEventListener("keyup", keyup);
+        window.addEventListener("blur", blur);
```
Vendicated: please use WindowStore instead

```js
WindowStore.addChangeListener(() => console.log(WindowStore.isFocused()))
```

Vendicated (commented):
good catch!

src/plugins/messageClickActions/index.ts
xNasuni: i see. i didn't know this existed. i'll remember for future cases

Vendicated:
thank you mia

PR #3473, opened 2025-06-08
QuickReply: correctly handle new messages

Vendicated:
Basically the plugin stores your current reply position for up/down functionality to work

Previously it would store a simple int (index), but this meant that if someone sent a new message it would skip a message (the more new messages, the more severe)

Now it instead stores the current message id and then dynamically finds the next message

The logic for this is kinda complicated so I'll have to test this a bit more to ensure there aren't any bugs


Vendicated:
tested more and seems to work well

PR #3481, opened 2025-06-12
Add the DuckDuckGo image mirror to the CSP policy

colleirose:
e.g.: https://external-content.duckduckgo.com/iu/?u=https://avatars.githubusercontent.com/u/167055039?v=4

Any image can be fetched through the DDG mirror, but they're sanitized and proxied through DDG servers to not reveal the user's IP address.


claration:
There was a recent commit which allows you to whitelist custom domains if they're blocked https://github.com/Vendicated/Vencord/commit/ed5ed4b80a7b3a50858eefef083b385449147486 should probably wait for when this gets merged to main.

PR #3486, opened 2025-06-14
Update dependencies

heytkowskii:
## changes:
- updated pnpm from `10.4.1` to `10.12.1`
- bumped minimum node.js version to v20, because v18 has reached EOL
- updated node.js version to v22 in github actions
- updated dependencies


heytkowskii:
why? code has been tested - `pnpm test` works

PR #3491, opened 2025-06-15
feat(Plugin): BetterBanModal

crepnf:
Discord removed the media slot in the ban modal. Consequently, the BANger plugin was [removed from Vencord](https://github.com/Vendicated/Vencord/commit/7f2c4a35660002cbf31df72ad43bff67b7026fc0). 

This plugin reintroduces the media slot and provides users with the ability to customise the URL to a GIF or video. Additionally, it moves the modal title from the modal content to the modal header, where it should logically be positioned.

Without this plugin, the modal header title slot would be empty, with the title appearing in the modal content instead[^1]. The media slot would also be unavailable, since Discord's product managers decided that we can't have fun anymore.

The comparison below shows Discord's ban modal circa 2022 (left) versus the current ban modal with BetterBanModal activated (right):

[^1]: This might be a bug that Discord fixes eventually, which would simplify this plugin and leave less room for error. Any improvements to the current patching logic would be most appreciated.

banger

### DIFF
--- a/src/plugins/betterBanModal/index.tsx
+++ b/src/plugins/betterBanModal/index.tsx
@@ -1,0 +1,133 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


Vendicated:
you can achieve this with css

ghost:
You cannot achieve this via CSS; I tried. The following does work, somewhat, for images or GIFs:
```css
div:has(> [role="radiogroup"] [role="radio"]:nth-child(4):last-child)::before {
  content: "";
  display: block;
  width: 100%;
  height: auto;
  aspect-ratio: 420/236;
  background: url('https://i.imgur.com/O3DHIA5.gif') center/contain no-repeat;
  margin: 0 auto 24px;
}
```
However, it does not work for videos. BANger did work for videos because Discord used to provide a video element in this modal, rather than a simple image tag that only works for GIFs.

I understand that you might not be a big fan of this plugin, and the handful of people who would appreciate this feature will have to give up on it if they don't want to maintain their own installation.

I won't put any further energy into this here, but I wanted to mention that some people had customised this with their own video URLs, which is impossible to achieve with CSS alone.

Cheers!

PR #3496, opened 2025-06-17
Prevent onChange setting option from being fired twice.

EtorixDev:
Currently any declared `onChange` options on a plugin setting are being fired twice.


Vendicated:
thanks for the fix

PR #3476, opened 2025-06-10
Allow users to manually whitelist Domains for use in themes

Vendicated:
https://github.com/user-attachments/assets/c499e82e-d483-4203-8376-3801bba14ce8

PR #3499, opened 2025-06-19
Fix CspViolation process to catch online theme with blocked domain

Maix0:
This small patch allow the already existing CspViolation (added by #3476) to also catch csp violation created by online themes.

This souldn't have any other side effect since no plugins should inject remote stylesheet in discord, but it would also catch thoses (and display them as all others).

tested on my machine with my own domain, and it works.
The workaround is to either edit the settings file manually, or to make a dummy (local) themes that tries to load something from the desiered url. as this will trigger the event (and not be ignored)m allowing the user to choose weither they want to allow/deny the domain.

Cheers :)

### DIFF
--- a/src/utils/cspViolations.ts
+++ b/src/utils/cspViolations.ts
@@ -8,7 +8,7 @@ import { useLayoutEffect } from "@webpack/common";
 
 import { useForceUpdater } from "./react";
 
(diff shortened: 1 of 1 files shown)


Vendicated:
thanks for the fix! I ended up fixing this a good while ago (i didn't see this PR sorry) so this PR is obsolete

PR #3504, opened 2025-06-20
feat(Plugin): addServerTitlesSidebar

NewsGuyTor:
I made a plugin that adds the titles of servers/folders to the sidebar. 

It can easily be expanded/collapsed with a toggle button. It has a slick animation by default, but that can be disabled in settings.

https://github.com/user-attachments/assets/914a6b5d-d929-4047-b546-81c62010ec18

All input and improvements are welcome!


jamesbt365:
https://github.com/Vendicated/Vencord/blob/main/CONTRIBUTING.md
image
Additionally, just hover over the server name why would you need to see all of them at once?

NewsGuyTor:
Because that's extremely cumbersome, there's a reason you don't have to hover over every file in a file explorer to see their filenames.

I'll look into patches.

Vendicated:
this breaks the plugin guidelines and seems ai generated

it's also not very useful, just hover over servers use the quickswitcher. Type `*` into the quick switcher and it will list your servers by name and even let you search

PR #3517, opened 2025-06-26
Fix patches for latest discord update

sadan4:


sadan4:
tested and these work on both stable and canary

PR #3520, opened 2025-06-28
add @vencord/discord-types package

Vendicated:
See the [README file](https://github.com/Vendicated/Vencord/blob/@vencord/discord-types/packages/discord-types/README.md) for info

TODO:
- Migrate all of discord-types to this new package and update them
- Find all places in Vencord code where due to outdated types it either casts or extends types and remove them

### DIFF
--- a/src/plugins/favEmojiFirst/index.ts
+++ b/src/plugins/favEmojiFirst/index.ts
@@ -18,8 +18,8 @@
 
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
-import { EmojiStore } from "@webpack/common";
 import { Emoji } from "@vencord/discord-types";
+import { EmojiStore } from "@webpack/common";
 
 interface EmojiAutocompleteState {
     query?: {
--- a/src/plugins/memberCount/index.tsx
+++ b/src/plugins/memberCount/index.tsx
@@ -23,8 +23,8 @@ import { classNameFactory } from "@api/Styles";
 import ErrorBoundary from "@components/ErrorBoundary";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
-import { findStoreLazy } from "@webpack";
 import { FluxStore } from "@vencord/discord-types";
+import { findStoreLazy } from "@webpack";
 
 import { MemberCount } from "./MemberCount";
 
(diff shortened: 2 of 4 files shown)


Vendicated:
this error is driving me crazy. doesn't happen on a fresh setup with @types/node and DOM so idk what's causing it...
```
node_modules/.pnpm/typescript@5.8.3/node_modules/typescript/lib/lib.dom.d.ts:28689:13 - error TS2451: Cannot redeclare block-scoped variable 'self'.

28689 declare var self: Window & typeof globalThis;
                  ~~~~
```

Vendicated:
@sadan4 should we just merge this as is for now to avoid inevitably running into more conflicts?

and then just make a second pr for updating the types more

Vendicated:
I added an enums folder with real .ts files where we can export const enums for use in actual code. for now it only contains command enums (because otherwise we would have had duplicated types), but we could move more enums over

https://github.com/Vendicated/Vencord/commit/36acda0670a3299f372e87f702d011e6d79f26e9

sadan4:
> @sadan4 should we just merge this as is for now to avoid inevitably running into more conflicts?
> 
> and then just make a second pr for updating the types more

yeah, that sounds good to me

sadan4:
> this error is driving me crazy. doesn't happen on a fresh setup with @types/node and DOM so idk what's causing it...
> 
> ```
> node_modules/.pnpm/typescript@5.8.3/node_modules/typescript/lib/lib.dom.d.ts:28689:13 - error TS2451: Cannot redeclare block-scoped variable 'self'.
> 
> 28689 declare var self: Window & typeof globalThis;
>                   ~~~~
> ```

i'd hazard a guess that's its caused by this: https://github.com/Vendicated/Vencord/pull/3520/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0L6-R6

we might be able to fix this by making a separate lint & tsconfig for the discord-types package

Vendicated:
well yes it happens because I enabled libcheck again.

but we should have that enabled, without it enabled, ts won't check .d.ts files at all. that's why your changes to this branch had so many errors. i only disabled it nack then to work around this very issue

Vendicated:
it's some incompatibility between whatever types packages we have, gotta figure out what exactly

Vendicated:
so turns out the `cannot redeclare const` was cause some random file had a `const self` lmfao

https://github.com/Vendicated/Vencord/pull/3520/commits/73ab9928a7a947879cd7635f355000010055989f

PR #3523, opened 2025-06-28
messageLatency: fix bot check

unsigned-32-bit-integer:
Accurately checks if a message is sent from a bot, preventing the bot form extracting bogus timestamps from bots that use their own format for nonces.


rniii (commented):
lgtm

PR #3527, opened 2025-06-29
Change spotifyControls album cursors

AnubisNekhet:
Low effort but improves legibility


Vendicated:
seems good on paper but these cursors look like shit imo

PR #3528, opened 2025-06-29
AlwaysAnimate: animate gradient roles

lumap:
just needed one match in alwaysAnimate


src/plugins/alwaysAnimate/index.ts
```diff
@@ -54,6 +54,14 @@ export default definePlugin({
                 match: /(\.headerContent.+?guildBanner:\i,animate:)\i/,
                 replace: "$1!0"
             }
+        },
+        {
+            // Gradient roles in chat
+            find: "value:{animate:",
```
Vendicated: I don't really like this, kinda generic

maybe `.contentOnly]:`

PR #3529, opened 2025-06-30
Add YouTube music desktop rich presence

MarcoSin42:
Request for comments.


Vendicated:
that app literally has native discord presence according to their website

MarcoSin42:
At severe risk of "it works on my machine" can you try it on linux?  

E: Anyways--whether or not you find Rich presence working on your machine or not--I think the advantage of websockets is clear in this case as they're going to be agnostic to not only the Linux distro/setup but also the OS.

https://github.com/ytmdesktop/ytmdesktop/issues/1549

MarcoSin42:
If you're still looking into this or whatever, I've decided to forgo web sockets and decided to target MPRIS instead as it will be supported by including but not limited to: the following [Linux media players.](https://wiki.archlinux.org/title/MPRIS).

Vendicated:
yes mpris makes much more sense as it works for every player

PR #3530, opened 2025-06-30
Add UTC

yofukashino:
Adds UTC (Universal Theming Classes) that appends stable class names [For example `utc_xyz`] to elements, so theme devs don’t have to worry about class rerolls, only structural changes.
Helps with theming by removing the need for wildcard selectors, which also improves performance.

The idea is to get it adopted across other client mods too, to make theming easier and consistent. Already PR'ed at [replugged](https://github.com/replugged-org/replugged/pull/701)


sadan4:
this should probably be a part of themeattributes instead

yofukashino:
sure, the only thing that matters is `utc` prefix on classes, so that CSS would work out of the box on other clients as well.
 
 so should I change location or wait for someone else to review first?

ZapGaming:
This would save me so much effort vencord pls add this

PR #3532, opened 2025-06-30
add hideTopBar plugin

Lutitious:
Signed-off-by: Rahul <notlutitious@gmail.com>


Lutitious:
This plugin hides the top bar of discord cause its dumb and useless

Vendicated:
this is just css and thus has no reason to be a plugin

Lutitious:
> this is just css and thus has no reason to be a plugin

I couldn't figure out how to set data-fullscreen = true without using Javascript

PR #3536, opened 2025-07-03
Fix plugins for removal of methods from the Guild class

sadan4:
Discord removed all methods from the guild class. I fixed all uses that tsserver could find for me.


src/utils/discord.tsx
```diff
@@ -218,3 +219,15 @@ export function getEmojiURL(id: string, animated: boolean, size: number) {
...
+
+export function getGuildAcronym(guildName?: string): string {
+    return guildName ?
+        guildName.replace(/'s /g, " ")
+            .replace(/\w+/g, m => m[0])
+            .replace(/\s/g, "")
+        : "";
+}
+
+export function guildHasFeature(guild: Guild, feature: keyof GuildFeatures["GuildFeatures"]): boolean {
```
Vendicated: kinda useless imo

why not just do this inline?

sadan4: could be inlined

Kept it as its own function in case Discord ever changes the way features are done.

your call

Vendicated: seems fine

I don't like this name though, `hasGuildFeature` would probably be better

src/utils/discord.tsx
```diff
@@ -218,3 +219,17 @@ export function getEmojiURL(id: string, animated: boolean, size: number) {
     const url = IconUtils.getEmojiURL({ id, animated, size });
     return animated ? url.replace(".webp", ".gif") : url;
 }
+
+// Discord has a similar function in their code
+export function getGuildAcronym(guildName?: string): string {
```
Nuckyz: Make the argument the guild and use the name from it

src/plugins/pauseInvitesForever/index.tsx
```diff
@@ -27,8 +27,7 @@ function showDisableInvites(guildId: string) {
     if (!guild) return false;
 
     return (
-        // @ts-ignore
-        !guild.hasFeature("INVITES_DISABLED") &&
+        !hasGuildFeature(guild, "INVITES_DISABLED" as any) &&
```
Nuckyz: Use // @ts-ignore and remove `as any`

sadan4: ts-ignore suppress all errors, not just the one with the string type. eg: guild is null, or other things that could be missed in a refactor

src/utils/discord.tsx
sadan4: should be good, haven't tested

Vendicated:
looks good

nin0-dev:
LBTM

Nuckyz (approved):

Vendicated:
LFTM (looks fire to me)

PR #3538, opened 2025-07-04
feat: Vencord.Webpack.Common.GuildActionCreators

darwinx64:
adds GuildActionCreators to Vencord.Webpack.Common, mugman helped me find this


Vendicated:
the point of commons is to provide common modules used by various plugins. this doesn't actually seem to be used by any current plugins so it makes little sense to add it

just have this find inside your plugin and if in the future more plugins make use of it, it should be moved to commons

PR #3542, opened 2025-07-07
RoleColorEverywhere: Implement the new Gradient role colors

Ezzud:
Implemented Gradient colors:
- Member list
- Threads list
- Voice chat users
- Poll voters

Edited **getColorStyle, getColorInt, getColorString and RoleGroupColor** to handle variable colorStrings instead of colorString. 

Added **getColorClass & getPollResultColorClass** to apply the required classes for the gradient to appear. 

Edited patches to handle class modification


lelanceflamer:
I tested it out, it's fantastic. However, in the case of the member list headers, if it contains an emoji, the emoji would also suffer the same fate as the text
image
I looked at how Discord handles emojis in gradient roles (I can say they forgot this case too) but in usernames, they thought about it and separated emojis and text by putting each emoji into their own span :)
So here's the function I made in order to replicate this:
```tsx
// function to avoid the emojis from suffering the same fate as the text by putting each of them into a span
function writeContent(text: string) {
    const parts = [...text].map(char => ({ char, isEmoji: /\p{Extended_Pictographic}/u.test(char) }));

    return parts.map(({ char, isEmoji }, i) =>
        isEmoji
            ? <span key={i} className="emoji_e5de78">{char}</span>
            : char
    );
}
```
And all is left to do is add in the span children:
```js
{writeContent(title ?? label)} &mdash; {count}
```
I hope this will turn out useful 🙏  it has been a long while since I've written typescript

PR #3545, opened 2025-07-10
Refactor Settings UI

Vendicated:


Vendicated:
I HATE CIRCULAR DEPENDENCIES 
image

src/api/Badges.ts
```diff
@@ -16,11 +16,13 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+// DO NOT REMOVE UNLESS YOU WISH TO FACE THE WRATH OF THE CIRCULAR DEPENDENCY DEMON!!!!!!!
+import "~plugins";
```
Vendicated: circular imports are too funny

We should really look into fixing this properly so we don't need to commit this atrocity

Vendicated: like why in the world does this have to be precisely here lmao

i tried adding it at the top of the api index.ts file but no dice

Vendicated:
I'm considering changing how settings work. I want to remove the Save button and instead make all changes immediately apply, to reduce complexity and because personally I don't think it's very useful to have

claration:
Wouldn't that require a restart for every single toggle though?

Vendicated:
no

sadan4:
> I'm considering changing how settings work. I want to remove the Save button and instead make all changes immediately apply, to reduce complexity and because personally I don't think it's very useful to have

I like this idea. More often than I'd like to admit, I click out instead of clicking the save button, then wonder why nothing is working.

Vendicated:
I refactored the Settings and removed the saving process

Would be useful if people could test it :+1:

Vendicated:
Okay, should be ready I think!

Vendicated:
I need people to test this change and just play around with settings & plugin settings a bunch and make sure everything still works fine

Vendicated:
no one tested it's over, time to test in prod (real)

PR #3548, opened 2025-07-11
Add Plugin: ReEmbed by TheCat

FedericoSlongo:
Converts unsupported embeds to supported embeds

For example twitter.com (or x.com) get converted to fxtwitter.com

Current supports instagramm, twitter (x) and tiktok

Also gives the user the choice of which twitter mirror they want to use


nin0-dev:
textreplace exists

FedericoSlongo:
Yes but this is more convenient and easier for the user but I understand

PR #3549, opened 2025-07-13
New Plugin: LaliluleNO

hideki2k02:
Name: LaliluleNO
Author: Me (AKA Duckouji)
Description: This plugin replaces any blacklisted words (separated by comma) with "lalilulelo" on the client-side
(Blacklisted words are configurable, but by default they are: gregtech, gt, gtnh, greg)

sample text

### DIFF
--- a/src/plugins/laliluleNO/index.tsx
+++ b/src/plugins/laliluleNO/index.tsx
@@ -55,7 +55,14 @@ export default definePlugin({
                 "2": "z"
             };
 
-            return [...text].map(char => leet_to_letter[char] || char).join('');
+            // This makes sure leet_convert only replace "word" tokens
+            return text.replace(/\b[\w]+\b/g, word => {
+                // Only process if the word contains at least one letter 
+                // (not just digits or punctuation)
+                if (!/[a-zA-Z]/.test(word)) return word;
+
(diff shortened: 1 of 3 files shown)


Vendicated:
Closed because this is not useful / seems like an inside joke, and mutating real data in place like this is very bad

If you actually want this, you should add affecting all messages as a feature to the TextReplace plugin and implement it without mutating discord's data

hideki2k02:
Thanks for the explanation

> this is not useful / seems like an inside joke

this plugin was heavily inspired by inside jokes (as seen from the default parameters) but its more a "Streaming Mode+" than just a silly plugin, its actual intended use is to replace "bad words" with something silly (currently lalilulelo as referencing Metal Gear, but it would've been configurable eventually).

> you should add affecting all messages as a feature to the TextReplace plugin

I don't think this should be in TextReplace as AFAIK it was made to replace your message before they were actually sent to Discord, This plugin replaces blacklisted words from any message client-side, so it only affects you.

With that in mind though, will try to make a "safer" version that doesn't mutate Discord Data in a foreseeable future.

Vendicated:
The only form of this plugin that would be accepted is as an addition to TextReplace

This way users can change whatever text in the app they want with their own replacement

PR #3551, opened 2025-07-15
"Quick Userscript" setting

20kdc:
I develop and maintain a number of personal userscripts for browser-based Discord. These are, effectively, implementing something slightly more powerful than QuickCSS, capable of containing some dynamic logic and element injection.

I can't run these on current Vencord without reimplementing them as plugins, but this increases maintenance effort (I use the browser to save RAM when possible) for what are ultimately rather simple scripts, not particularly more complicated than `oneko.js`.

Something I did consider was the idea of implementing a loader similar to the oneko.js plugin - I decided against this. I have assumed that Vencord's current model of plugin distribution, while user-unfriendly, is for security purposes to prevent rogue updates.

However, regardless of if it's in the devtools or in a userscript box, a user who wishes to directly inject external code can. Adding the userscript box does not add much persistence that cannot be achieved already by, for instance, rewriting Vesktop files for their own purposes. To a user who knows what they are doing, but does not wish to manually track Vencord updates, it adds a great degree of flexibility.

### DIFF
--- a/src/Vencord.ts
+++ b/src/Vencord.ts
@@ -155,6 +155,12 @@ async function init() {
     }
 }
 
+/* The quickUserscript runs here, so it may add custom plugins and patches. */
+try {
+    new Function(Settings.quickUserscript)();
+} catch (exception) {
+    PMLogger.error("Error in QuickUserscript", exception);
+}
 startAllPlugins(StartAt.Init);
 init();
 
--- a/src/api/Settings.ts
+++ b/src/api/Settings.ts
@@ -32,6 +32,7 @@ export interface Settings {
     autoUpdate: boolean;
     autoUpdateNotification: boolean,
     useQuickCss: boolean;
+    quickUserscript: string;
     eagerPatches: boolean;
     enabledThemes: string[];
     enableReactDevtools: boolean;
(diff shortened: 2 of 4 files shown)


Vendicated:
We considered something like this a while ago but decided against it for security reasons. users who want this should just build from source and add their script as a userplugin. Or in your case you can use a userscript manager. You can even even use vencord apis from it, vencord exposes all its apis as `window.Vencord`

Many users share their settings backups and sometimes people even ask strangers to send them theirs for whatever reason. This pr would turn that into a security hazard as someone could send you a settings backup with a malicious script in it

20kdc:
There are some situations where browser Discord is insufficient, so a userscript manager is not always available.

Secondly, after all the issues I've repeatedly had with various vague Node cryptography API breakage and so forth, I am _extremely_ skeptical of "just build from source" as a statement in regards to any Node/NPM-based project.

This is not to mention that, as Discord is a moving target, a source build is also a moving target.
Do you have some reasonably deployable rebase-and-build system for forks? If not, I'm not sure what, exactly, you're expecting people who want to deploy custom code to actually do other than figure out how to automate it themselves (which in practice means essentially starting a fork of the entire Vencord project).

Vendicated:
yes.

20kdc:
Uhh... oooookay then! Thank you for the clarity, I guess...???

nin0-dev:
assuming dependencies don't change (happens very rarely in which case just run `pnpm i`), the remote is `Vendicated/Vencord`, and `git pull` would normally work (which is the case if there are no merge conflicts (there never are if you only put things in the `userplugins` folder)) the vencord updater can handle devbuild updates

20kdc:
Ah, thank you. That's the sort of solution that should work for my use-case.

PR #3560, opened 2025-07-22
ViewRaw: Add ViewRaw option to message context menu

TomW1605:
add a setting to hide the view raw button in the message popover as well as add the option to the message context menu


TomW1605:
think i could get someone to take a look at this?

Vendicated:
Plugins shouldn't have to implement this themselves. I'm working on a solution that will work for all plugins (not final UI):
 
image
image

Thank you for your contribution regardless!

TomW1605:
Hi, i think having a centralised setting for including things in the hover menu is a great idea but this PR also includes adding the option to the message context menu.
would it be possable to merge it for now then remove the plugin spesific setting later when the new UI is released?

Vendicated:
you can PR the message context menu addition. i would rather not add a temporary toggle for the hover button though. you can just hide it with css for now:
```css
[class*=hoverBarButton_][aria-label*="View Raw"] {
	display: none;
}
```

TomW1605:
will do

TomW1605:
removed the settings stuff. PR is now just for adding it to the context menu

PR #3563, opened 2025-07-24
feat: add splitting long messages patches + logic

celgost:
Basically a lightweight recreation of Split Large Messages from Devil Bro but in Vencord. I might refactor the splitting logic, even though it's pretty solid and light already.

Maybe need some feedback on the send message timeout I am using. I put 20 milliseconds a bit arbitrarily.


celgost:
[selfbotting](https://github.com/Vencord/plugin-requests/issues/77)

PR #3564, opened 2025-07-25
VcNarrator: make nickname fallback to displayName

TomW1605:
change it so that if there is no nickname, it will fallback to the display name instead of directly back to the username. this more acuratly reflects what is displayed in the VC


AntiApple4life:
I'd love to see this as well!

Vendicated:
yep yep

PR #3569, opened 2025-07-28

sadan4:


sadan4:
one patch still needs to be fixed in SHC but everything else seems good

packages/discord-types/src/stores/GuildRoleStore.d.ts
```diff
@@ -1,7 +1,13 @@
 import { FluxStore, Role } from "..";
 
+// TODO: add the rest of the methods for GuildRoleStore
 export class GuildRoleStore extends FluxStore {
     getRole(guildId: string, roleId: string): Role;
-    getRoles(guildId: string): Record<string, Role>;
-    getAllGuildRoles(): Record<string, Record<string, Role>>;
+    /**
+     * Prefer {@link getSortedRoles} to this
```
Vendicated: why?

sadan4: It was changed to that in most places.

Vendicated: why not add `getRolesSnapshot`? that method seems stable, whereas the unsafe one might disappear

src/plugins/_api/messageEvents.ts
```diff
@@ -27,7 +27,7 @@ export default definePlugin({
         {
             find: "#{intl::EDIT_TEXTAREA_HELP}",
             replacement: {
-                match: /(?<=,channel:\i\}\)\.then\().+?(?=return \i\.content!==this\.props\.message\.content&&\i\((.+?)\))/,
+                match: /(?<=,channel:\i\}\)\.then\().+?\i\.content!==this\.props\.message\.content&&\i\((.+?)\)\}(?=return)/,
```
sadan4: i have not tested it on canary but it looks good.

only works on canary rn.

pre edit listeners are also only used by fakenitro rn

PR #3571, opened 2025-07-30
Fix petpet transparency issues

unsigned-32-bit-integer:
Currently the petpet plugin uses the darkest color in the palette to represent transparency, this means that if used with an image that has pixels close to `#000000`, these pixels will be rendered as transparent. This pr resolves this issue by using the last color in the pallet to represent transparency, and manually assigning this to any pixel with an alpha value of 0.
(`applyPaletteTransparent` is loosely based off the `applyPalette` function from gifenc)

### DIFF
--- a/src/plugins/petpet/index.ts
+++ b/src/plugins/petpet/index.ts
@@ -21,7 +21,6 @@ import { Devs } from "@utils/constants";
 import { makeLazy } from "@utils/lazy";
 import definePlugin from "@utils/types";
 import { CommandArgument, CommandContext } from "@vencord/discord-types";
-import { findByPropsLazy } from "@webpack";
 import { DraftType, UploadAttachmentStore, UploadHandler, UploadManager, UserUtils } from "@webpack/common";
 import { GIFEncoder, nearestColorIndex, quantize } from "gifenc";
 


src/plugins/petpet/index.ts
```diff
@@ -150,7 +174,8 @@ export default definePlugin({
 
                 const canvas = document.createElement("canvas");
                 canvas.width = canvas.height = resolution;
-                const ctx = canvas.getContext("2d")!;
+                canvas;
```
Vendicated:
```suggestion
```

Vendicated:
Thanks for the fix

can this be solved by using a different format? https://github.com/mattdesl/gifenc#index--applypalettergba-palette-format--rgb565

If not, it would be best if you can also report it to gifenc via an issue

src/plugins/petpet/index.ts
unsigned-32-bit-integer: oops lol

unsigned-32-bit-integer:
> can this be solved by using a different format? https://github.com/mattdesl/gifenc#index--applypalettergba-palette-format--rgb565

the gif format doesnt actually support alpha channels, so idk why rgba4444 is an option there

unsigned-32-bit-integer:
ok, looks like i misunderstood this a bit earlier (and made a small mistake when attempting to test), the purpose of rgba4444 is to allow the alpha channel to be considered during quantization and application so that `rgba(0, 0, 0, 0)` is considered to be far from `rgba(0, 0, 0, 255)`, and the reduction in rgb color space is just for optimization. although i still dont understand the reasoning for this approach when the gif format only allows for a single fully transparent color.

ive made [a branch](https://github.com/unsigned-32-bit-integer/Vencord/tree/fix/petpet-rgba4444) to fix this using this much simpler method at the expense of color depth, and seemingly causing slight flickering with some images (this could potentially be addressed by reusing the same palette across frames).

Vendicated:
oh yes I honestly have no clue why it doesn't reuse the palette

you could also try playing with the alpha options here, not sure if they would solve the issue https://github.com/mattdesl/gifenc#palette--quantizergba-maxcolors-options--

unsigned-32-bit-integer:
the flickering appears to be due to mattdesl/gifenc#13, ive found that it can be avoided by using the same palette and cache for every frame, but that requires making a custom `applyPalette` function, and if we need to do that anyway it probably makes the most sense to continue using rgb565.

src/plugins/petpet/index.ts
```diff
@@ -143,17 +166,32 @@ export default definePlugin({
...
 
                 const gif = GIFEncoder();
 
+                const paletteImageSize = Math.min(120, resolution);
+
                 const canvas = document.createElement("canvas");
-                canvas.width = canvas.height = resolution;
-                const ctx = canvas.getContext("2d")!;
+                canvas.width = resolution;
+                canvas.height = Math.max(resolution, 2 * paletteImageSize);
```
Vendicated, line 179: why is the height different from the width? won't this lead to possibly stretched / cut off images?

src/plugins/petpet/index.ts
```diff
@@ -143,17 +166,32 @@ export default definePlugin({
 
                 const avatar = await loadImage(url);
 
-                const delay = findOption(opts, "delay", DEFAULT_DELAY);
+                let delay = findOption(opts, "delay", DEFAULT_DELAY);
+                // Frame delays < 20ms don't function correctly on chromium and firefox
+                if (delay < 20) delay = 20;
```
Vendicated: it would be better user experience to be more transparent on this. the option description should note this limitation and this should probably be an error instead

src/plugins/petpet/index.ts
```diff
@@ -167,14 +205,22 @@ export default definePlugin({
...
+                            palette,
+                            delay,
+                        });
+                    } else {
+                        gif.writeFrame(index, resolution, resolution, {
+                            transparent: true,
+                            transparentIndex: 255,
+                            delay,
+                        });
+                    }
```
Vendicated:
```suggestion
                    gif.writeFrame(index, resolution, resolution, {
                        transparent: true,
                        transparentIndex: 255,
                        delay,
                        palette: i === 0 ? palette : null,
                    });
```

src/plugins/petpet/index.ts
```diff
@@ -143,17 +166,32 @@ export default definePlugin({
...
 
                 UploadManager.clearAll(cmdCtx.channel.id, DraftType.SlashCommand);
 
+                // Generate palette from an image where hand and avatar are fully visible
+                ctx.drawImage(avatar, 0, paletteImageSize, 0.8 * paletteImageSize, 0.8 * paletteImageSize);
+                ctx.drawImage(frames[0], 0, 0, paletteImageSize, paletteImageSize);
+                const { data } = ctx.getImageData(0, 0, paletteImageSize, 2 * paletteImageSize);
+                const palette = quantize(data, 255);
+
+                const cache = new Array(65536);
```
Vendicated: u8array seems better suited, as that is also what you're using above

```suggestion
                const cache = new Uint8Array(65536);
```

Vendicated: also to be less magic numbery, it'd probably be better to change this from `65536` to `2 ** 16` or `0xFFFF + 1`

Vendicated: actually isn't this array straight up way too big? this is your palette index cache and our palette only contains 256 colours at most, so this should just need to be 256 bytes (and also probbaly be renamed)

Vendicated: oh nevermind, palette is a 2d array. so the size used is actually `256 * 256`. then that is the most intuitive way to write this size

```suggestion
                const cache = new Uint8Array(256 * 256);
```

src/plugins/petpet/index.ts
```diff
@@ -85,10 +85,33 @@ async function resolveImage(options: CommandArgument[], ctx: CommandContext, noS
...
+}
+
+function applyPaletteTransparent(data: Uint8Array | Uint8ClampedArray, palette: number[][], cache: number[], threshold: number): Uint8Array {
+    const index = new Uint8Array(Math.floor(data.length / 4));
+
+    for (let i = 0; i < data.length; i += 1) {
+        const r = data[4 * i];
+        const g = data[4 * i + 1];
+        const b = data[4 * i + 2];
+        const a = data[4 * i + 3];
```
Vendicated:
```suggestion
    for (let i = 0; i < data.length; i += 4) {
        const r = data[i];
        const g = data[i + 1];
        const b = data[i + 2];
        const a = data[i + 3];
```

Vendicated: i'm honestly confused by the old code, unless I'm missing something, it will spend 1/4th of the loop doing all elements, then 3/4 of the loop is just accessing out of bounds indexes

src/plugins/petpet/index.ts
unsigned-32-bit-integer: this array caches the palette values that correspond to a given rgb565 value in the same way that gifenc does internally, i have changed `65536` to `2 ** 16` in 46944316f74f243e742817370777ae28c54aa529 to make it less magic numbery

src/plugins/petpet/index.ts
unsigned-32-bit-integer: 0738f36ccd1daad3a0ab6ec18fe9ad34680dac5c

src/plugins/petpet/index.ts
unsigned-32-bit-integer: Using a `Uint8Array` breaks the behavior of `in` when applying the palette, since `new Uint8Array(length)` initializes all values to 0.

src/plugins/petpet/index.ts
unsigned-32-bit-integer: the extra height of the canvas is only used when generating the palette, the extra space is needed so that the hand doesnt obscure part of the pfp. the pallete generation image looks like this: 
image
but will not be encoded into the gif

Vendicated: I see, that makes sense

probably worth adding a short comment

unsigned-32-bit-integer: theres already a comment [here](https://github.com/Vendicated/Vencord/pull/3571/files#diff-320ffca22acbdaf48c52911db6cf7b7b938a8a09eef9bc69acbf03d93581c9adR187) but i'll add another on this line too

src/plugins/petpet/index.ts
unsigned-32-bit-integer: 2d654bbc12ac030d09ea2783bc5bd851780c9e74

unsigned-32-bit-integer:
i discovered that this plugin errors on userscript vencord due to a CSP violation, is it worth b64 encoding the hand images so it can work w userscript or should i avoid jank and just make it show a proper error for userscript users?

unsigned-32-bit-integer:
for now i've just made it show a proper error when the command is run with Userscript Vencord

unsigned-32-bit-integer:
since the scope of this pr has increased considerably since it was opened, is there anything that should be split into a separate pr or is it good as it is?

Vendicated:
Thanks a lot again for looking into this issue and the fix!

I removed your UserScript error again, because some userscript managers (namely TamperMonkey) support bypassing csp, so this plugin can potentially work there

PR #3574, opened 2025-07-31

sadan4:


Vendicated:
canary only for now

src/plugins/showHiddenChannels/index.tsx
sadan4, line 1: from what i can tell, this is the previous behavior, however this *feels wrong* to me

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -492,6 +492,16 @@ export default definePlugin({
         }
     ],
 
+
+    fixPermCheck(originalPerms: any, channel: Channel) {
```
Vendicated: you're doing math on this so it's either a number or this code is wrong

```suggestion
    fixPermCheck(originalPerms: number, channel: Channel) {
```

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -492,6 +492,16 @@ export default definePlugin({
         }
     ],
 
+
+    fixPermCheck(originalPerms: number, channel: Channel) {
```
Vendicated: okay it's actually bigint
```suggestion
    fixPermCheck(originalPerms: bigint, channel: Channel) {
```

PR #3575, opened 2025-08-01
feat(lastfmRichPresence): placeholder support in statusName

controlado:
Allows the user to use placeholders to make a customized status, as shown below

## Preview

| Settings | Result |
| --- | --- |
| Settings | Result |
| Settings 2 | Result 2 |

## Code overview

```ts
const trackPlaceholderRegex = /\$\{(name|artist|album)\}/gi;

...

function applyPlaceholders(tmpl: string, t: TrackData) {
    return tmpl.replace(trackPlaceholderRegex, (_, key: keyof TrackData) => {
        return t[key] ?? "";
    }).trim();
}

```


Vendicated:
This has been added

PR #3577, opened 2025-08-03
SpotifyControls: properly show nitro theme background

nin0-dev:


Vendicated:
this was fixed in the commit removing old compat stuff

PR #3580, opened 2025-08-04
New Plugin: autoTimestamp

catsonluna:
Description: Automatically convert times in messages to Discord timestamps.

features:
- automatically converts time and date in text to timestamps
- supports both am/pm and military time (18:00)
- append ":" to the start of the time to make it relative
-- all date, time, and datetime support relative
- converts all times and dates in a single text
- if you say "10pm" at 11pm for example, it will automatically make it 10pm tomorrow 

examples:
- "i will be there at 10pm" the "10pm" will be turned into a timestamp
- "i will be there 20/08/2025" the "20/08/2025" get converted into a date
- "i will be there at 10pm 20/08/2025" the whole "10pm 20/08/2025" will be turned into a single timestamp, if time is placed right before or after the date
- "i will be there at :10pm 20/08/2025" the whole "10pm 20/08/2025" will be turned into a relative timestamp

settings:
- toggle time parsing
- toggle date parsing
- toggle datetime parsing
- should a time be wrapped in something, for example if \` is placed, it will only convert times if it is \`10pm\`
- relative time prefix, if its empty, relative time is disabled
- date format, change from the default DD/MM/YYYY

Before:
Screenshot 2025-08-04 at 20 55 57

Sent:
Screenshot 2025-08-04 at 20 56 16

### DIFF
--- a/src/plugins/autoTimestamp/index.tsx
+++ b/src/plugins/autoTimestamp/index.tsx
@@ -74,10 +74,12 @@ function parseDate(dateStr: string, format: string, relativePrefix: string): str
     let day = 1, month = 0, year = 1970;
     let match: RegExpMatchArray | null = null;
     let isRelative = false;
+
     if (relativePrefix && dateStr.startsWith(relativePrefix)) {
         isRelative = true;
         dateStr = dateStr.slice(relativePrefix.length);
     }
+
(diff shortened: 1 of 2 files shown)


Vendicated:
already exists as SendTimestamps

PR #3582, opened 2025-08-04
:)

Starry0Wolf:


celestial-skies:
edits the translate plugin to support auto translating specific channels

celestial-skies:
Excluding channels and supporting full servers is a WIP

PR #3583, opened 2025-08-05
Fix patches for latest discord update

sadan4:


src/plugins/vencordToolbox/index.tsx
```diff
@@ -140,9 +140,9 @@ export default definePlugin({
 
     patches: [
         {
-            find: "toolbar:function",
+            find: '["isAuthenticated"]',
```
sadan4: this find seems like it will break easily, should be replace with a better one

src/plugins/vencordToolbox/index.tsx
```diff
@@ -140,9 +140,9 @@ export default definePlugin({
 
     patches: [
         {
-            find: "toolbar:function",
+            find: "=!0,toolbar:",
```
Vendicated: I don't like this, but I can't think of many better alternatives

The best I could come up with was `.controlButtonWrapper,`

src/plugins/showAllMessageButtons/index.ts
```diff
@@ -28,8 +28,9 @@ export default definePlugin({
         {
             find: "#{intl::MESSAGE_UTILITIES_A11Y_LABEL}",
             replacement: {
-                match: /isExpanded:\i&&(.+?),/,
-                replace: "isExpanded:$1,"
+                // isExpanded = isShiftPressed && other conditions...
+                match: /(?<=(\i)=)\i(?=&&.+isExpanded:\1)/,
```
Vendicated: this looks like it might be slow, we should verify it isn't before merging

Vendicated: okay we're good

image

Vendicated:
no stable compat is rough, I'll probably extract the compatible fixes and we can merge the others later

image

Vendicated:
actually besides the SHC stuff there's nothing stable compatible so let's just wait in general

src/plugins/showAllMessageButtons/index.ts
Nuckyz: The wildcard in this fine should be changed to be lazy, I forgot to do it yesterday

sadan4:
> actually besides the SHC stuff there's nothing stable compatible so let's just wait in general

it looks like stable got pushed see https://canary.discord.com/channels/1015060230222131221/1337479880849362994/1402694238184931388

PR #3584, opened 2025-08-05
Use lazy scroller for Notification log

Vendicated:
When you have hundreds of Notifications, the Notification log becomes very laggy due to rendering so many Elements

This migrates it to Discord's lazy scroller


Vendicated:
For some reason when you start scrolling it goes crazy and scrolls you all the way to the bottom

Vendicated:
Okay turns out missing react key caused the scrolling issues, crazy

PR #3587, opened 2025-08-06
fix hideAttachments not working for forwarded embeds

zordythezordan:
Fix for hideMedia (in plugins folder named hideAttachments) plugin to check embeds in forwarded messages

The solution is not perfect ("Media hidden" shows both on messages, forwarded and the message that shows the forwarded message), but working


wdgasterdeltarune:
oh, and also i fixed silly misspell in "shapshots" instead of "snapshot"

Vendicated:
I ended up reworking how this works but thank you for your contribution!

wdgasterdeltarune:
Thanks! :D

PR #3591, opened 2025-08-07
New Plugin: EmbedReplace

lolsuffocate:
# EmbedReplace

This plugin allows you to right-click a link to replace its embed in a message via TextReplace style rules.

[Preview](https://github.com/user-attachments/assets/029bb473-1b03-4668-89ba-631368ff301b)

### DIFF
--- a/src/plugins/embedReplace/README.md
+++ b/src/plugins/embedReplace/README.md
@@ -2,4 +2,4 @@
 
 This plugin allows you to right-click a link to replace its embed in a message via TextReplace style rules.
 
-![Preview](https://github.com/user-attachments/assets/f1533528-7269-4423-8511-2aecdcdad528)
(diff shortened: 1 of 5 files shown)


Minemetero:
Seems interesting, but the function would be rarely used

Minemetero:
You'd prefer to click on the tweets in 1 second instead of replacing with fxtwitter spending 3 seconds to see the same thing

xl-spooky:
> You'd prefer to click on the tweets in 1 second instead of replacing with fxtwitter spending 3 seconds to see the same thing

Honestly, i prefer to have that directly inside of discord, would take more time to open twitter, and me personally i don't even have an account there, So this definitely is a cool feature.

Vendicated:
don't think this is particularly useful if it's manual (which is the only option we have, I know). at that point, you can just use a user app (right click > apps > replace embeds) and have it reply with an ephemeral message

PR #3593, opened 2025-08-08
fix(ChatButtons): Use ButtonProps for ChatBarButton instead of HTMLButton props

tgayle:
Small typing fix I noticed when trying to add `buttonRef` on a `ChatBarButton`.


Vendicated:
Obsolete I think? The underlying types have all been changed since. Thanks though!

PR #3598, opened 2025-08-09
Plugin: Custom Screenshare Preview

FawazTakahji:
Introduce a new plugin that allows users to set any image up to 200kb in size as their screenshare preview.


Vendicated:
this is way too niche

PR #3600, opened 2025-08-10
AccountPanelServerProfile: fix patch

sadan4:
i think the added patch isn't _great_, open to feedback

### DIFF
--- a/src/plugins/accountPanelServerProfile/index.tsx
+++ b/src/plugins/accountPanelServerProfile/index.tsx
@@ -11,7 +11,7 @@ import { getCurrentChannel } from "@utils/discord";
 import definePlugin, { OptionType } from "@utils/types";
 import { User } from "@vencord/discord-types";
 import { findComponentByCodeLazy } from "@webpack";
-import { ContextMenuApi, Menu, useEffect, useRef } from "@webpack/common";
+import { ContextMenuApi, Menu } from "@webpack/common";
 
 interface UserProfileProps {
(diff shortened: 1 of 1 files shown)


Vendicated:
you're grabbing the ref to the dom element just to click it

why not match the onClick listener instead?

sadan4:
> you're grabbing the ref to the dom element just to click it
> 
> why not match the onClick listener instead?

We don't have the DOM event to pass to it that it expects. afaik this is the easiest way

src/plugins/accountPanelServerProfile/index.tsx
```diff
@@ -85,7 +80,11 @@ export default definePlugin({
                 },
                 {
                     match: /(?<=#{intl::SET_STATUS}\),)/,
-                    replace: "ref:$self.accountPanelRef,onContextMenu:$self.openAccountPanelContextMenu,"
+                    replace: "onContextMenu:$self.openAccountPanelContextMenu,"
+                },
+                {
+                    match: /(?<=innerRef:)\i(?=,style:\i(?:,|\}))/,
+                    replace: "$self.grabRef($&)"
```
Vendicated:
```suggestion
                    match: /innerRef:(\i)(?=,style:)/,
                    replace: "innerRef:$self.grabRef($1)"
```

Vendicated: actually we don't need a separate patch for this, we can do this above

PR #3604, opened 2025-08-11
optimize: reduce memory usage in webpack, plugins, and settings

Hungryy2K:
• Plugin Flux Cleanup - Added explicit event handler cleanup to prevent memory leaks during plugin lifecycle
• Proxy Caching - Implemented WeakMap-based proxy caching in SettingsStore to eliminate duplicate proxy creation
• Webpack Search Limiting - Capped search history at 100 entries with automatic cleanup to prevent unbounded growth
• Settings Debounce - Increased cloud sync debounce from 60s to 120s to reduce memory pressure
• Plugin Defaults Cache - Added caching for SELECT option defaults to avoid redundant calculations


Vendicated:
aislop garbage. literally every single change is useless or straight up breaks the feature

PR #3606, opened 2025-08-12
FavGifSearch: Encode search query

FawazTakahji:
Find gifs with non ascii characters in the link

before:
image
after:
image


rniii:
is this better than doing the opposite thing? `decodeURI(getTargetString(gif.url ?? gif.src))`

FawazTakahji:
> is this better than doing the opposite thing? `decodeURI(getTargetString(gif.url ?? gif.src))`

You'd have to decode every item instead of encoding the query once, but I don't know if it's that much of a meaningful difference.

PR #3607, opened 2025-08-14
fix empty messages dont get logged (modal interaction responses)

Egemenbdk:
If a slash command directly returns a modal, for example: `interaction.response.send_modal()`, the slash command will get logged as a deleted message even though no message was sent and since its contentless, it overlaps with other elements shown below and doesn't appear right, this 4 line update will stop logging contentless messages which are only possible if a slash command returns a modal.

image

<sub>edit: fixed my grammar its 5 am</sub>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.12.10",
+    "version": "1.12.11",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
     "bugs": {
--- a/packages/discord-types/src/components.d.ts
+++ b/packages/discord-types/src/components.d.ts
@@ -43,12 +43,7 @@ export type FormDivider = ComponentType<{
 }>;
 
 
-export type FormText = ComponentType<PropsWithChildren<{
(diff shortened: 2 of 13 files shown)


src/plugins/messageLogger/index.tsx
```diff
@@ -259,6 +259,11 @@ export default definePlugin({
                 const msg = cache.get(id);
                 if (!msg) return;
 
+                if (!msg.content || msg.content.trim() === "") {
+                    cache = cache.remove(id);
+                    return;
+                }
```
jamesbt365: I don't like this? Messages without content but have embeds, attachments, components (esp CV2), polls, message snapshots, nested interaction data or any other future field will be ignored because content will be empty.

There must be a better way you can handle this.

Egemenbdk: I'll add a check to make sure it was a slash command that directly returned a modal

jamesbt365: See if message.state is `SENDING` on these modal spawns and if it is check if its a bot or from an interaction, if it does then you should be able to reliably go off that.

Hardcoding this is not future proof at all and should be avoided, especially over something so minor.

PR #3621, opened 2025-08-21
fix(utils): improve `wordsFromCamel` correctness

ryanccn:
Currently settings keys such as `"customEngineURL"` get parsed into `["custom", "engine", "u", "r", "l"]`, and when turned into title case become "Custom Engine U R L". The new implementation, despite being slightly verbose, makes sure that acronyms stay intact.

### DIFF
--- a/src/utils/text.ts
+++ b/src/utils/text.ts
@@ -21,7 +21,8 @@ import { moment } from "@webpack/common";
 // Utils for readable text transformations eg: `toTitle(fromKebab())`
 
 // Case style to words
-export const wordsFromCamel = (text: string) => text.split(/(?=[A-Z])/).map(w => w.toLowerCase());
+export const wordsFromCamel = (text: string) =>
+    text.split(/(?=[A-Z][a-z])|(?<=[a-z])(?=[A-Z])/).map(w => w.match(/^[A-Z]+$/) ? w : w.toLowerCase());
 export const wordsFromSnake = (text: string) => text.toLowerCase().split("_");
 export const wordsFromKebab = (text: string) => text.toLowerCase().split("-");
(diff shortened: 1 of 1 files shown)


sadan4:
related: https://github.com/Vendicated/Vencord/pull/3212

src/utils/text.ts
```diff
@@ -21,7 +21,8 @@ import { moment } from "@webpack/common";
 // Utils for readable text transformations eg: `toTitle(fromKebab())`
 
 // Case style to words
-export const wordsFromCamel = (text: string) => text.split(/(?=[A-Z])/).map(w => w.toLowerCase());
+export const wordsFromCamel = (text: string) =>
+    text.split(/(?=[A-Z][a-z])|(?<=[a-z])(?=[A-Z])/).map(w => w.match(/^[A-Z]+$/) ? w : w.toLowerCase());
```
Vendicated:
```suggestion
    text.split(/(?=[A-Z][a-z])|(?<=[a-z])(?=[A-Z])/).map(w => /^[A-Z]+$/.test(w) ? w : w.toLowerCase());
```

Vendicated:
this still isn't fully correct cause one letter words exist like `ryanIsACutie`

we should see if there are any cases like this (i think A and I are the only single letter words in English) in Vencord

Vendicated:
```js
Object.values(Vencord.Plugins.plugins).flatMap(p => {
    if (!p.options) return [];
    return Object.keys(p.options)
        .filter(k => /[AI][A-Z]/.test(k))
        .map(k => `${p.name} ${k}`)
})
```

there's one such key: `CtrlEnterSend sendMessageInTheMiddleOfACodeBlock`

PR #3625, opened 2025-08-24
TextReplace: Add support for changing other users' messages client side.

EtorixDev:
- Modifies the message content of rendered messages to apply TextReplace rules.
- Preserves the original content in a new attribute to prevent recursive replacements.
- Also added a new setting to choose whether to apply rules to your own messages, to others', or to all. Default is your own and existing rules are updated as such.


Nuckyz:
Scope could be a bitfield/enum with the setting being two checkboxes, allowing for a rule to apply to both at once.

Modifying the content property of a message is not acceptable and the patch needs to be adjusted to only modify the local reference where the content is used.

PR #3633, opened 2025-08-28
Add FirstMessage plugin

liasaysorry:
This plugin adds a slash command (/firstmessage) that jumps the user to the first message of whatever channel it was used in. I know that the contributing guidelines say "No simple slash command plugins" but I think this is something that could be useful to many people.

### DIFF
--- a/src/plugins/firstMessage/index.ts
+++ b/src/plugins/firstMessage/index.ts
@@ -4,11 +4,23 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+import { ApplicationCommandInputType } from "@api/Commands";
 import { Devs } from "@utils/constants";
 import definePlugin from "@utils/types";
+import { NavigationRouter } from "@webpack/common";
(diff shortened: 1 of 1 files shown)


lolsuffocate:
This is pretty easy to do stock, just ctrl+f in the channel and switch order to oldest first

image

liasaysorry:
> This is pretty easy to do stock, just ctrl+f in the channel and switch order to oldest first
> 
> image

true in most cases except to my knowledge that is not possible in places like vc chats

lolsuffocate:
You're right, looks like you can't ctrl+f in those channels but you can just type `in:"voice chat"` and do the same that way unfortunately

image

liasaysorry:
> You're right, looks like you can't ctrl+f in those channels but you can just type `in:"voice chat"` and do the same that way unfortunately
> 
> image

I didn't know this thank you for letting me know I will now close this since i no longer think there is a use case

PR #3638, opened 2025-09-01
New Plugin CompleteDiscordQuests: Init

Aqualunem:
Hi! I developed a plugin for complete discord quests easy nwn, inspired on amia's original script

### DIFF
--- a/src/plugins/completeDiscordQuests/README.md
+++ b/src/plugins/completeDiscordQuests/README.md
@@ -1,28 +1,110 @@
-# CompleteDiscordQuests
+# CompleteDiscordQuests ✨
 
-Completes your active Discord Quest automatically.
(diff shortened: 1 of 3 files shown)


Vendicated:
sorry but no

1. Including this in Vencord will just make Discord more likely to patch it
2. This code is owned by aamia. This PR is copyright infringement
3. Your code does not follow our plugin rules / best practices

Zaikonurami:
I credited who the code came from bruh, it was fair use, also the code is refactored, that's why it says I was inspired by it, in any case thanks for the answer, I understand the other 2 points

Vendicated:
that is not how fair use works

Vendicated:
actually I'm mistaken, seems like aamia released the code under GPL3.0. so yes, you're right, the first point is not true

PR #3648, opened 2025-09-05
Fix FakeNitro, OnePingPerDM, NoPendingCount Patches

thororen1234:
FakeNitro does as its just removed
OnePingPerDM doesn't have stable comp
NoPendingCount doesn't have stable comp


thororen1234:
added a nopendingcount fix not sure if thats how you all would want it tho 👍

PR #3650, opened 2025-09-05
New Plugin: PrefixSuffix

psikoo:
> ## PrefixSuffix
> A plugin to add any prefix and suffix to your messages.

This can be use for simple things like automatically adding "-# " to your messages so they show as subtext, or for complex and annoying things to type, like ansi codes.

### Features:

- Chat bar button: 
  - Left click can enable/disable the plugin. 
  - Right click opens the configuration modal
- Configuration Modal:
  - Allows for convenient editing of the prefix and suffix

### Example and Modal:

yellow

```
prefix > "```ansi\n[33m"
suffix > "[0m\n```"
```

settings


Vendicated:
useless spam + text replace can do this

PR #3601, opened 2025-08-10
feat(plugin): AlwaysExpandProfiles

thororen1234:
This plugin replaces user or bot's profile popouts with the full modal

PR #3651, opened 2025-09-06
feat(plugin): AlwaysExpandProfiles

thororen1234:
This plugin replaces user, bot, or system's profile popout with the full modal

Moved from https://github.com/Vendicated/Vencord/pull/3601 so it has proper rules


thororen1234:
apparently accountpanelserverprofile would just spam open the modal one after another once it was closed so I added a little check for it but I need to add it opening the server profile not just user profile

PR #3652, opened 2025-09-06
feat(plugin): StatusWhilePlaying

thororen1234:
This plugin automatically updates your status to the status you have selected in the settings of the plugin when playing a game. When the game closes the status will revert back to its previous state.

Moved from https://github.com/Vendicated/Vencord/pull/2988 so it has proper rules

### DIFF
--- a/packages/discord-types/CONTRIBUTING.md
+++ b/packages/discord-types/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Hint: https://docs.discord.food is an incredible resource and allows you to copy paste complete enums and interfaces
--- a/packages/discord-types/enums/messages.ts
+++ b/packages/discord-types/enums/messages.ts
@@ -11,3 +11,586 @@ export const enum StickerFormatType {
     LOTTIE = 3,
     GIF = 4
 }
+
+export const enum MessageType {
+    /**
+     * A default message (see below)
+     *
+     * Value: 0
+     * Name: DEFAULT
+     * Rendered Content: "{content}"
+     * Deletable: true
+     */
+    DEFAULT = 0,
+    /**
+     * A message sent when a user is added to a group DM or thread
+     *
+     * Value: 1
+     * Name: RECIPIENT_ADD
+     * Rendered Content: "{author} added {mentions [0] } to the {group/thread}."
+     * Deletable: false
+     */
+    RECIPIENT_ADD = 1,
+    /**
+     * A message sent when a user is removed from a group DM or thread
+     *
+     * Value: 2
+     * Name: RECIPIENT_REMOVE
+     * Rendered Content: "{author} removed {mentions [0] } from the {group/thread}."
+     * Deletable: false
+     */
+    RECIPIENT_REMOVE = 2,
+    /**
+     * A message sent when a user creates a call in a private channel
+     *
+     * Value: 3
+     * Name: CALL
+     * Rendered Content: participated ? "{author} started a call{ended ? " that lasted {duration}" : " — Join the call"}." : "You missed a call from {author} that lasted {duration}."
+     * Deletable: false
+     */
+    CALL = 3,
+    /**
+     * A message sent when a group DM or thread's name is changed
+     *
+     * Value: 4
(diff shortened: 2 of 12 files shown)


src/plugins/statusWhilePlaying.desktop/index.ts
```diff
@@ -0,0 +1,61 @@
...
+            {
+                label: "Invisible",
+                value: "invisible",
+            }
+        ]
+    }
+});
+
+export default definePlugin({
+    name: "StatusWhilePlaying",
```
Vendicated: i think it should be named AutoDNDWhilePlaying even if it has the other options, as that's the main use case

```suggestion
    name: "AutoDNDWhilePlaying",
```

src/plugins/statusWhilePlaying.desktop/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    description: "Automatically updates your online status (online, idle, dnd) when launching games",
+    authors: [Devs.thororen],
+    settings,
+    flux: {
+        RUNNING_GAMES_CHANGE({ games }) {
+            const userId = UserStore.getCurrentUser().id;
+            const status = PresenceStore.getStatus(userId);
```
Vendicated:
```suggestion
            const status = PresenceStore.getStatus(AuthenticationStore.getId());
```

src/plugins/statusWhilePlaying.desktop/index.ts
```diff
@@ -0,0 +1,61 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { definePluginSettings } from "@api/Settings";
+import { getUserSettingLazy } from "@api/UserSettings";
+import { Devs } from "@utils/constants";
+import definePlugin, { OptionType } from "@utils/types";
+import { PresenceStore, UserStore } from "@webpack/common";
+
+let savedStatus = "";
```
Vendicated:
```suggestion
let savedStatus: string | null = null;
```

src/plugins/statusWhilePlaying.desktop/index.ts
```diff
@@ -0,0 +1,61 @@
...
+const StatusSettings = getUserSettingLazy("status", "status");
```
Vendicated:
```suggestion
const StatusSettings = getUserSettingLazy("status", "status")!;

```
and don't use optional chaining below

src/plugins/autoDndWhilePlaying.discordDesktop/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+});
+
+export default definePlugin({
+    name: "AutoDNDWhilePlaying",
+    description: "Automatically updates your online status (online, idle, dnd) when launching games",
+    authors: [Devs.thororen],
+    settings,
+    flux: {
+        RUNNING_GAMES_CHANGE({ games }) {
```
Vendicated, line 48: this won't work on Vesktop. Not sure how to properly solve that

Vendicated: maybe mark as discord desktop only for now and a future goal should be for Vesktop to handle RPC better

src/plugins/autoDndWhilePlaying/index.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export default definePlugin({
+    name: "AutoDNDWhilePlaying",
+    description: "Automatically updates your online status (online, idle, dnd) when launching games",
+    authors: [Devs.thororen],
+    settings,
+    flux: {
+        RUNNING_GAMES_CHANGE({ games }) {
+            const status = PresenceStore.getStatus(AuthenticationStore.getId());
```
Nuckyz: You can use StatusSettings.getSetting() here

naturalsnowbuildings:
This is very useful, thanks! As a suggestion I would love to see an option so that it does the same thing with Listening status instead of Playing

Issue #1527, opened 2023-07-27
[Bug] BetterFolders sidebar doesn't close when removing a folder


Akitakek:
### Discord Account

Akitake_

### What happens when the bug or crash occurs?

The "opened folder" sidebar doesn't close when emptied .

### What is the expected behaviour?

I expect the sidebar to dissapear when a folder also does, much like it does when closing a folder.

### How do you recreate this bug or crash?

1. Group up servers in a folder
2. Take servers out one by one (must have the sidebar open there obviously)
3. When there is none left, the sidebar will remain opened yet empty

Pressing CTRL+R doesn't fix it.
Opening another folder then closing it doesn't fix it.

The only thing that seems to update it is re-creating or moving a folder in that same position.

### Errors

```
VM230 C:\Users\Akitake\AppData\Local\Discord\app-1.0.9015\modules\discord_desktop_core-1\discord_desktop_core\core.asar\app\node_modules\@sentry\utils\cjs\instrument.js:144 Refused to connect to 'sentry-ipc://sentry-electron.scope/sentry_key' because it violates the document's Content Security Policy.
```

This is an error I see while testing it and it may be unrelated.

### Request Agreement

- [X] I am using Discord Stable or tried on Stable and this bug happens there as well

AutumnVN:
the empty sidebar auto closes for me after few seconds without doing anything

Shinigami-Senpai99:
The empty sidebar doesn't close for me till I do this extremely specific thing, I have drag a server till that green bar when rearranging server pops up and just put it in that position even if the position is the exact same, then the Sidebar goes away.

ColbyHF:
> The empty sidebar doesn't close for me till I do this extremely specific thing, I have drag a server till that green bar when rearranging server pops up and just put it in that position even if the position is the exact same, then the Sidebar goes away.

This just saved me a life long struggle thank you

jtoya85:
> The empty sidebar doesn't close for me till I do this extremely specific thing, I have drag a server till that green bar when rearranging server pops up and just put it in that position even if the position is the exact same, then the Sidebar goes away.

Had this problem just now as well. This fixed it. Thank you.

sadan4:
fixed via https://github.com/Vendicated/Vencord/commit/7c839be64fbef65b4825ec1d17acd0264aadcd64

PR #3658, opened 2025-09-08
fix(BetterFolders): Fix Closing With 0 Servers

thororen1234:
Fixes https://github.com/Vendicated/Vencord/issues/1527

Not sure if line 35-36 is good and feedback is welcome


src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -21,13 +21,28 @@ import { findComponentByCodeLazy, findStoreLazy } from "@webpack";
 import { Animations, useStateFromStores } from "@webpack/common";
 import type { CSSProperties } from "react";
 
-import { ExpandedGuildFolderStore, settings } from ".";
+import { ExpandedGuildFolderStore, settings, SortedGuildStore } from ".";
 
 const ChannelRTCStore = findStoreLazy("ChannelRTCStore");
 const GuildsBar = findComponentByCodeLazy('("guildsnav")');
 
+function getExpandedGuildIds() {
```
Nuckyz: This naming is weird, a guild can't be expanded, the folder it belongs to can. Maybe getGuildIdsInExpandedFolders?

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -21,13 +21,28 @@ import { findComponentByCodeLazy, findStoreLazy } from "@webpack";
...
+    const expanded = ExpandedGuildFolderStore.getExpandedFolders();
+    const folders = SortedGuildStore.getGuildFolders();
+    const expandedGuilds = new Set<string>();
```
Nuckyz: There's no reason for it to be a Set, make it just an array instead

src/plugins/betterFolders/FolderSideBar.tsx
Nuckyz: This function could also return the ids of the folders which are expanded and include a guild inside them, which then can be used in the Sidebar component instead of the other useStateFromStore

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -21,24 +21,41 @@ import { findComponentByCodeLazy, findStoreLazy } from "@webpack";
...
         <GuildsBar
             {...guildsBarProps}
             isBetterFolders={true}
-            betterFoldersExpandedIds={expandedFolders}
+            betterFoldersExpandedIds={folderIds}
         />
     );
 
-    const visible = !!expandedFolders.size;
+    const visible = !!guildIds.size;
```
Nuckyz: guildIds can be discarded, as folderIds is now guaranteed to not be empty only if one folder is expanded and it contains a guild. Just use folderIds.size here

src/plugins/betterFolders/FolderSideBar.tsx
```diff
@@ -21,24 +21,39 @@ import { findComponentByCodeLazy, findStoreLazy } from "@webpack";
 import { Animations, useStateFromStores } from "@webpack/common";
 import type { CSSProperties } from "react";
 
-import { ExpandedGuildFolderStore, settings } from ".";
+import { ExpandedGuildFolderStore, settings, SortedGuildStore } from ".";
 
 const ChannelRTCStore = findStoreLazy("ChannelRTCStore");
 const GuildsBar = findComponentByCodeLazy('("guildsnav")');
 
+function getExpandedFoldersData() {
```
Nuckyz: Rename the function

Issue #3289, opened 2025-03-12
QuickReply: Selects blocked messages


sadan4:
### Content

When using quick reply, it will include blocked messages, without expanding the message being replied to.

### Request Agreement

- [x] I have read the requirements for opening an issue above

Vendicated:
this is probably very hacky to fix, have to yoink their setState

PR #3659, opened 2025-09-08
feat(QuickReply): Fix Replying To Blocked Users

thororen1234:
Fixes https://github.com/Vendicated/Vencord/issues/3289

image

### DIFF
--- a/src/plugins/quickReply/index.ts
+++ b/src/plugins/quickReply/index.ts
@@ -40,13 +40,18 @@ const settings = definePluginSettings({
         description: "Ping reply by default",
         options: [
             {
-                label: "Follow NoReplyMention",
+                label: "Follow NoReplyMention plugin (if enabled)",
                 value: MentionOptions.NO_REPLY_MENTION_PLUGIN,
                 default: true
             },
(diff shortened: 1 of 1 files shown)


src/plugins/quickReply/index.ts
```diff
@@ -134,6 +134,7 @@ function getNextMessage(isUp: boolean, isReply: boolean) {
         if (m.deleted) return false;
         if (!isReply && m.author.id !== meId) return false; // editing only own messages
         if (hasNoBlockedMessages && NoBlockedMessagesPlugin.shouldIgnoreMessage(m)) return false;
+        if (RelationshipStore.isBlocked(m.author.id)) return false;
```
Nuckyz: How about isBlockedOrIgnored? I dont remember how messages from ignored users show up

Vendicated:
should we really skip? it might be better to instead auto open it

thororen1234:
I figured just skipping it would be fine as if you have them blocked the chances you are actually wanting to reply to them are low imo

Vendicated:
maybe just make it a setting then, enabled by default

PR #3665, opened 2025-09-09
DiscordProximity: Proximity Chat Support for Discord

Siriusmart:
Taken from the plugin README

---

Support for a client side proximity chat protocol I designed yesterday.

## Why?

There are previously no Discord-based proximity chat clients, so I might as well as write it.

I have read the code of conduct:
- Nothing similar to this existed for Discord
- This plugin has not been previously rejected
- The plugin does not depend on any external APIs or bots.
- I'm not opening a request because I've already wrote the plugin, I'm here just to ask if Vencord would like to include it.
- I have checked that changing local volume 20 times a second will not spam Discord with API requests, setting update requests are only triggered by changes made by the user.

### Showcase
(unmute the video)

https://github.com/user-attachments/assets/fa02ff46-4a46-470b-a71d-e58ddadc79d5

### How This Works

The main idea behind it is similar to ARRPC.
1. The game client (e.g. Minecraft) starts a websocket server at `ws://127.0.0.1:25560/api/subscriptions`
2. This plugin connects to the websocket server
3. This plugin sends the list of users in the VC to the game client.
4. The game client calculates the **volume multiplier**, and sends it back to the plugin.
5. The plugin updates the user volumes accordingly.

The plugin resets all use volumes to their original volumes when the game closes, or when the user leaves the VC.

### The Protocol

Messages are JSON formatted
```json
{
	"t": "string",
	"c": "any"
}
```

Here are the message types that can be sent:

#### connected

WS server -> plugin when the connection is made

```json
{
  "t": "connected",
  "c": null
}
```

#### sub

Plugin -> WS server when you join a VC, or someone new joined a VC.

```json
{
  "t": "sub",
  "c": [ "12345678910", "list of discord ids..."]
}
```

#### unsub

Plugin -> WS server when someone leaves a VC.

```json
{
  "t": "unsub",
  "c": [ "12345678910", "list of discord ids..."]
}
```

#### clear

Plugin -> WS server when a connection is opened, to clear all previous subscription info.

```json
{
  "t": "clear",
  "c": null
}
```

#### set

WS server -> plugin to set a user volume by giving a volume multiplier, the multiplier is typically between 0 and 1.

```json
{
  "t": "set",
  "c": 0.456
}
```

All other stuff not specified above are up to client implementation.

### Example Implementation

[DiscordProximity for Minecraft Forge 1.8.9](https://github.com/Siriusmart/DiscordProximity).

Unless your game has a Discord linking feature, you might want a **nameserver** to map Discord IDs to in-game names in your implementaiton.

For the Forge mod, I used [this single-source-file nameserver](https://github.com/Siriusmart/proximity-nameserver) and a [Discord bot](https://github.com/siriusmart/merlin) to allow users to set their IGN.

### DIFF
--- a/src/plugins/discordProximity/index.tsx
+++ b/src/plugins/discordProximity/index.tsx
@@ -5,6 +5,7 @@
  */
 
 import { Devs } from "@utils/constants";
+import { Logger } from "@utils/Logger";
 import definePlugin from "@utils/types";
 import { findByPropsLazy } from "@webpack";
 import { ChannelStore, SelectedChannelStore, UserStore } from "@webpack/common";
@@ -16,7 +17,7 @@ let originalVolumes = {};
 // connection to game
 // undefined if not connected or closed
 // Websocket otherwise
-let ws;
+let ws : WebSocket | undefined;
 // bindWS() is called with a generation
 // generation is incremented when a connection closes
 // if the global generation != the generation bindWS() is ran with
@@ -44,7 +45,7 @@ function restore(): void {
         generation++;
     }
 
-    console.log("Restored user volumes");
+    new Logger("DiscordProximity").log("Restored user volumes");
 }
 
 // runs after connection opened
@@ -60,10 +61,10 @@ function bindWS(localGen: number): void {
 
     try {
         ws = new WebSocket("ws://127.0.0.1:25560/api/subscription");
-        ws.onopen = () => {
-            console.log("Connected to proximity websocket.");
+        ws.onopen = function () {
+            new Logger("DiscordProximity").log("Connected to proximity websocket.");
             connect();
-            ws.send(JSON.stringify({
+            this.send(JSON.stringify({
                 t: "clear",
                 c: 0
             }));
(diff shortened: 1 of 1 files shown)


src/plugins/discordProximity/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+function bindWS(localGen: number): void {
+    const myId = UserStore.getCurrentUser().id;
+    if (generation !== localGen) return;
+    // no longer in VC
+    if (whosThereReverse[myId] === undefined) return;
+
+    try {
+        ws = new WebSocket("ws://127.0.0.1:25560/api/subscription");
+        ws.onopen = () => {
+            console.log("Connected to proximity websocket.");
```
sillyElv: You should probably use the logger instead of console.log

Siriusmart: updated to use logger instead

Siriusmart:
@lolsuffocate no way proximity chat was so good they added it to real life

Siriusmart:
the metioned minecraft mods will also need to be installed on the server, which is not possible for most public servers

currently there are mods to for client side proximity chat using Mumble, but nobody uses Mumble

Vendicated:
this is very cool, but at the same time it's way too niche to be included

you should just distribute it to your users yourself, they have to install your mod anyway

PR #3667, opened 2025-09-10
BadgesAPI: Add badges to user settings account profile card

thororen1234:
image

### DIFF
--- a/packages/discord-types/enums/activity.ts
+++ b/packages/discord-types/enums/activity.ts
@@ -1,0 +1,30 @@
+export const enum ActivityType {
+    PLAYING = 0,
+    STREAMING = 1,
+    LISTENING = 2,
+    WATCHING = 3,
+    CUSTOM_STATUS = 4,
+    COMPETING = 5,
+    HANG_STATUS = 6
+}
+
+export const enum ActivityFlags {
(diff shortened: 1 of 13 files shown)


src/plugins/_api/badges/index.tsx
```diff
@@ -111,6 +111,13 @@ export default definePlugin({
                     replace: "...$self.getBadgeMouseEventHandlers($1),$&"
                 }
             ]
+        },
+        {
+            find: "profileCardUsernameRow,children",
```
Nuckyz:
```suggestion
            find: "profileCardUsernameRow,children:",
```

src/plugins/_api/badges/index.tsx
```diff
@@ -111,6 +111,13 @@ export default definePlugin({
...
+            replacement: {
+                match: /(?<=displayProfile:(\i).+?badges:)(\i)/,
```
Nuckyz: Dangerous quantifier inside lookbehind, if you need the displayProfile place the lookbehind after the actual match and adjust the match to not depend on it

Nuckyz: For example: `/badges:(\i)(?<=displayProfile:(\i).+?)/`

PR #3669, opened 2025-09-10
AppleMusicRichPresence: add status display type

nin0-dev:


nin0-dev:
y'all who do i have to pay to get this plugin updated and merged into the main thing i'm so serious. do i have to become a sponsor of vencord? like genuinely the only thing that's stopping me from making the switch from BD is the lack of this plugin

PR #3670, opened 2025-09-17
ShowConnections: Cap max number of connections displayed

sadan4:
continuation of https://github.com/Vendicated/Vencord/pull/3156 because git is dumb and couldn't merge dev into that branch


src/plugins/showConnections/index.tsx
```diff
@@ -59,6 +59,14 @@ const settings = definePluginSettings({
             { label: "Cozy", value: Spacing.COZY }, // US Spelling :/
             { label: "Roomy", value: Spacing.ROOMY }
         ]
+    },
+    maxNumberOfConnections: {
+        type: OptionType.SLIDER,
+        description: "Max number of connections to show",
+        markers: makeRange(6, 48, 7),
```
Vendicated, line 66: These markers and the default value are probably unreasonable considering iconSize and iconSpacing settings will change how many icons can be displayed per row. By default it will be 7 items but it could be any number of items

It'd probably make more sense to either make this not configurable or change it to max number of rows of connections to show

I think just always making it max 2 rows without setting would be best. If you want to keep the configurability, add a toggle "Limit max number of connections" so people can just have the old behaviour if they'd like

Vendicated: Sorry I had a stroke

sadan4: i'd prefer the setting being the number of rows myself, but that is more work to measure the width and calculate the number of icons per row. I can do that if thats what you want but i'd rather not.

Vendicated: Instead of calculating you can probably just do this:
```tsx
const containerRef = useRef(null);
const [visibleCount, setVisibleCount] = useState(connections.length);
const [showAll, setShowAll] = useState(false);

useLayoutEffect(() => {
  const container = containerRef.current;
  if (!container) return;

  let rows = [];
  let i = 0;
  for (const child of container.children) {
    if (!rows.includes(child.offsetTop)) {
      rows.push(child.offsetTop);
    }

    if (rows.length > 2) {
      setVisibleCount(i - 1);
      return;
    }
	i++;
  }
}, [connections]);

const hasTooManyConnections = connections.length > visibleCount;
const connectionsToRender = showAll || !hasTooManyConnections
								? connections
								: connections.slice(0, visibleCount);

return (
	<div ref={containerRef}>
		{connectionsToRender.map(...)}
		{hasTooManyConnections && <ToggleShowAllButton />}
	</div>
)
```

sadan4: ill figure something out and come back to this later

PR #3677, opened 2025-09-21
MessageLatency: add option to ignore own messages

qwertyquarty:
Would be a great feature if your own time is wrong but you still want to see other people's latency without being bothered from your own

### DIFF
--- a/src/plugins/messageLatency/index.tsx
+++ b/src/plugins/messageLatency/index.tsx
@@ -11,7 +11,7 @@ import { isNonNullish } from "@utils/guards";
 import definePlugin, { OptionType } from "@utils/types";
 import { Message } from "@vencord/discord-types";
 import { findComponentByCodeLazy } from "@webpack";
-import { SnowflakeUtils, Tooltip } from "@webpack/common";
+import { SnowflakeUtils, Tooltip, UserStore } from "@webpack/common";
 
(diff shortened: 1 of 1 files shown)


src/plugins/messageLatency/index.tsx
```diff
@@ -100,6 +105,8 @@ export default definePlugin({
         // Bots basically never send a nonce, and if someone does do it then it's usually not a snowflake
         if (message.author.bot) return null;
 
+        if (ignoreSelf && message.author.id === UserStore.getCurrentUser().id) return null;
```
Vendicated:
```suggestion
        if (ignoreSelf && message.author.id === AuthenticationStore.getId()) return null;
```

src/plugins/messageLatency/index.tsx
```diff
@@ -11,7 +11,7 @@ import { isNonNullish } from "@utils/guards";
 import definePlugin, { OptionType } from "@utils/types";
 import { Message } from "@vencord/discord-types";
 import { findComponentByCodeLazy } from "@webpack";
-import { SnowflakeUtils, Tooltip } from "@webpack/common";
+import { SnowflakeUtils, Tooltip, UserStore } from "@webpack/common";
```
Vendicated:
```suggestion
import { AuthenticationStore, SnowflakeUtils, Tooltip } from "@webpack/common";
```

PR #3680, opened 2025-09-24
arRPC: increase connection timeout to 5s

sadan4:
1s frequently times out


Vendicated:
this simply shouldnt use a timeout, just use `ws.on("open")`, `ws.on("error")`, and `ws.on("close")`

PR #3684, opened 2025-09-26
fix(quickReply): prevent replying to system pinned messages

YashRajCodes:
fixes the following issue which occurs when trying to reply to system messages caused by people pinning messages

image


src/plugins/quickReply/index.ts
```diff
@@ -134,6 +134,7 @@ function getNextMessage(isUp: boolean, isReply: boolean) {
         if (m.deleted) return false;
         if (!isReply && m.author.id !== meId) return false; // editing only own messages
         if (hasNoBlockedMessages && NoBlockedMessagesPlugin.shouldIgnoreMessage(m)) return false;
+        if (m.type === 6) return false; // prevents replying to CHANNEL_PINNED_MESSAGE
```
Vendicated: don't use magic numbers. there should be a MessageType enum in the discord-types package. if not, please add it yourself

Vendicated:
you can find the missing enum keys here: https://docs.discord.food/resources/message#message-type

Vendicated:
Thank you, merged!

I made it better by using Discord's own "is this message replyable?" logic, which properly covers all message types and also made it ignore ephemeral messages

I don't have permission to directly push to your branch so I did it manually

Vendicated:
https://github.com/Vendicated/Vencord/commit/f040133412f1edc7ccfa3e4c3cc8309b46ff9aed

PR #3685, opened 2025-09-26
feat: add AutoSystemTheme plugin for syncing Discord color theme with system theme

mahirox36:
This plugin automatically switches discord color themes based on the system's theme preference (light/dark).

the problem was when I change the system theme, discord doesn’t update my color theme, so I always have to open settings and change it manually

With this plugin, you can pick your favorite light and dark themes once, and it will switch them automatically with your system. If you just want it to only flip between light/dark without changing the colors, there’s an option for that in the plugin settings.


nin0-dev:
you know that’s a stock feature right 
IMG_1535

mahirox36:
Yes i know but it only works with the default themes: light, dark, darker, midnight, but never the color themes like sunshine, sunset, etc

PR #3687, opened 2025-09-27
New plugin: Share Active Window

ipasechnikov:
# Rationale

I've been really enjoying Vencord recently. I use it daily at work for pair programming sessions. Increased FPS rate provided by FakeNitro plugin helps a lot for my teammates to keep track of what's going on the screen.

However, when you work in a group, it might be not the best option to share the entire screen. I have an ultrawide monitor and some of my colleagues feel more comfortable working on laptops with smaller screens which makes it quite hard for them to see a clear picture because of resolution difference.

The solution is to track an active window and auto-switch sharing to the active window. Being a seasoned developer myself, I decided to give it a try and contribute to this amazing community hoping that someone might find the plugin useful.

# Approach

There is no easy way to track an active window across multiple OSes in a unified way. I ended up partially including a [vendor library](https://github.com/paymoapp/node-active-window) with a few custom changes. It does not introduce any NPM dependencies. Everything is contained within the plugin directory. However, it does add a few megabytes of base64-encoded platform-specific binaries to the final build.

The plugin supports Windows, Linux and MacOS.

# Screenshots

You can easily enable/disable the feature on the fly in the stream management context menu.
image

A couple of settings to control the plugin behavior
image

---

Please feel free to comment and suggest any improvement on the plugin and the code itself.


Vendicated:
Thank you for your contribution! Unfortunately this can in no way be merged because it has native binaries which
1. I cannot trust to be safe
2. The way they are included as base64 then written to a file is just way too cursed

Plus this plugin seems niche!

Sorry :(

ipasechnikov:
@Vendicated Thanks for the response. Appreciate it.

I agree with all of your points. Honestly, I didn't have high expectations on this to be merged. What's more, if I were you, I would not approve this as well.

Hopefully someone might find the plugin useful someday and despite all precautions not to grab random binary from a random dude on the web would still compile the thing.

Anyway, it was a nice experience writing a plugin for your this project. I really enjoyed it. It is amazing what you have done here both from the user and developer perspective. Keep it up.

PR #3688, opened 2025-09-27
feat: translation v3

lewisakura:
supersedes #2528


lewisakura:
@Vendicated need some input, how should things like this be translated?

https://github.com/Vendicated/Vencord/blob/a3850c130e6b2bf776d515a2fabc3e80d2bd3d55/src/components/settings/tabs/vencord/NotificationSettings.tsx#L59-L65

i think the ideal way would be all in one go, but atm it's not possible to do so because `<Translate>` only translates direct children (which lines up with how other translation libraries work right now), so we'd need to rewrite that component in a way that supports this which is going to be very difficult

alternatively, we just make all of these separate translations and provide context to the translators

lewisakura:
> need some input, how should things like this be translated?

official stance: give up and cry

we'll cross the bridge when we get there, we might just change how this part of the UI looks anyway

Vendicated:
Tbh I don't think translations are ever happening

PR #3694, opened 2025-09-30
VGP encryption and decryption of messages with pgp/gpg

FedericoSlongo:
This plugin adds the ability to encrypt and decrypt (with signature verification) discord messages using pgp/gpg encryption

Made by [Nikicoraz](https://github.com/Nikicoraz) and [me](https://github.com/FedericoSlongo)

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -606,6 +606,14 @@ export const Devs = /* #__PURE__*/ Object.freeze({
         name: "thororen",
         id: 848339671629299742n
     },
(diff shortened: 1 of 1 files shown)


Vendicated:
We won't add plugins like this, as they offer questionable privacy (Discord has access to your private key) and don't really fit into Discord; you will always have a bad experience if you're looking at the messages without the plugin. Users who want encrypted messaging should just use Signal, Matrix or XMPP

Thanks for the contribution nevertheless

PR #3697, opened 2025-10-01
RevealAllSpoilers: use cmd key for macOS

wjdgustn:


src/plugins/revealAllSpoilers/index.ts
sadan4, line 28: Should the description be adjusted to reflect that Mac uses `cmd`?

wjdgustn: Seems other plugins are not including that in description?
[CtrlEnterSend](https://github.com/Vendicated/Vencord/blob/8943c90cb0b8be7c6c268ec4fb6931d30ad1dc59/src/plugins/ctrlEnterSend/index.ts#L14) (included in setting)
[QuickReply](https://github.com/Vendicated/Vencord/blob/8943c90cb0b8be7c6c268ec4fb6931d30ad1dc59/src/plugins/quickReply/index.ts#L56)

I think mac users will realize cmd+click, because ctrl+click shows context menu in mac.

Vendicated:
thanks for the fix!

PR #3698, opened 2025-10-01
Add Transcript plugin for richly rendered channel exports

Naseem48:
- add new Transcript plugin to export HTML, Markdown, or JSON transcripts from any text channel, thread, DM, or group DM
- render exports with Discord-like UI (avatars, nicknames, embeds, playable media) and an inline context menu for copying IDs or message links
- provide configurable time ranges (from first message, up to now, ISO/relative inputs) plus filters for bots, pinned messages, media-only, authors, and keywords


Vendicated:
this is api abuse, also way too much code

just use DiscordChatExporter or make your own program for this

PR #3699, opened 2025-10-01
fix reporter on canary

Vendicated:
The old code stringifies wreq.u to get the export map

But it now is a wrapper function so it doesn't include the object inline anymore

image

Simple solution is to just use a getter on the `Object.prototype` for a unique symbol, then call the method with that symbol to get the object it's using. Compatible with both the new and old code

### DIFF
--- a/src/debug/loadLazyChunks.ts
+++ b/src/debug/loadLazyChunks.ts
@@ -174,14 +174,13 @@ export async function loadLazyChunks() {
         const chunkMap = getWebpackChunkMap();
         if (!chunkMap) throw new Error("Failed to get chunk map");
 
-        const allChunks = Object.keys(chunkMap);
+        const allChunks = Object.keys(chunkMap).map(id => Number.isNaN(Number(id)) ? id : Number(id));
         if (allChunks.length === 0) throw new Error("Failed to get all chunks");
 
         // Chunks which our regex could not catch to load
         // It will always contain WebWorker assets, and also currently contains some language packs which are loaded differently
         const chunksLeft = allChunks.filter(id => {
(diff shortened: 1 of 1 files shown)


src/debug/loadLazyChunks.ts
```diff
@@ -153,23 +171,17 @@ export async function loadLazyChunks() {
...
-            const id = currentMatch[1] ?? currentMatch[2];
-            if (id == null) continue;
-
-            const numId = Number(id);
-            allChunks.push(Number.isNaN(numId) ? id : numId);
-        }
+        const chunkMap = getWebpackChunkMap();
+        if (!chunkMap) throw new Error("Failed to get chunk map");
 
+        const allChunks = Object.keys(chunkMap);
```
Nuckyz:
```suggestion
        const allChunks = Object.keys(chunkMap).map(id => Number.isNaN(Number(id)) ? id : Number(id));
```

src/debug/loadLazyChunks.ts
```diff
@@ -153,23 +171,17 @@ export async function loadLazyChunks() {
...
         if (allChunks.length === 0) throw new Error("Failed to get all chunks");
 
         // Chunks which our regex could not catch to load
         // It will always contain WebWorker assets, and also currently contains some language packs which are loaded differently
         const chunksLeft = allChunks.filter(id => {
-            return !(validChunks.has(id) || invalidChunks.has(id));
+            const key = Number.isNaN(Number(id)) ? id : Number(id);
+            return !(validChunks.has(key) || invalidChunks.has(key));
```
Nuckyz:
```suggestion
            return !(validChunks.has(id) || invalidChunks.has(id));
```

PR #3704, opened 2025-10-04
Added MessageRecall plugin

pIxne:
This PR adds a new plugin: MessageRecall.

- Lets you recall your previously sent messages into the text box
- Default keybinds: Shift+ArrowUp / Shift+ArrowDown
- Keybinds can be customized via plugin settings
- Works across DMs, groups, and servers
- Resets recall when you send a new message (so you always start from the latest)

This is useful for quickly reusing or past messages without needing to copy them manually.

Made for this plugin request: https://github.com/Vencord/plugin-requests/issues/1103

Closes vencord/plugin-requests#1103


Vendicated:
this conflicts with quickReply

also just
1. press up to edit the last message
2. press ctrl a to mark all text
3. press ctrl c to copy
4. press escape to stop editing
5. press ctrl v to paste

pIxne:
This uses Shift + Up/Down I tested it along with quickReply and don't see an issue, also like the request said it's more of a QoL thing It makes grabbing older messages easier since you can go back further than just one message

PR #3707, opened 2025-10-05
Add CleanCord plugin

MinecraftBedrockArabic:
Adds CleanCord, a Vencord plugin that cleans up Discord’s UI by hiding all the flashy stuff you might not care about.

With this plugin, you can:

- Hide Nitro nameplates
- Hide avatar decorations
- Hide profile effects
- Hide booster icons
- Hide server tags
- Hide role icons
- Hide Super Reactions and their burst effects

Everything is toggleable in the settings, so you can hide as much or as little as you like.


Vendicated:
No plugins that simply hide or redesign ui elements. This can be done with CSS

PR #3710, opened 2025-10-07
showMeYourName: Add "Nickname Only" mode

TipsyTheCat:
Allows showing friend nicks only :3

image

### DIFF
--- a/src/plugins/showMeYourName/index.tsx
+++ b/src/plugins/showMeYourName/index.tsx
@@ -30,7 +30,7 @@ const settings = definePluginSettings({
             { label: "Username then nickname", value: "user-nick", default: true },
             { label: "Nickname then username", value: "nick-user" },
             { label: "Username only", value: "user" },
-            { label: "Nickname only", value: "nick" },
+            { label: "Nickname only (fallback to username)", value: "nick" },
         ],
     },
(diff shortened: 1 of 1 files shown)


ben-mcginty:
+1
Ironically I made my own patch for this exact issue and was about open a PR before I saw this here.

src/plugins/showMeYourName/index.tsx
```diff
@@ -30,6 +30,7 @@ const settings = definePluginSettings({
             { label: "Username then nickname", value: "user-nick", default: true },
             { label: "Nickname then username", value: "nick-user" },
             { label: "Username only", value: "user" },
+            { label: "Nickname only", value: "nick" },
```
ben-mcginty:
```suggestion
            { label: "Nickname only (falls back to username if no nickname specified)", value: "nick" },
            I would like to add this note here however that it falls back to the username if there is no nickname assigned.
```

TipsyTheCat: i shortened the fallback notice so it would fit better

Tiebe:
@Vendicated would you be able to take a look at this? Very small change, but I'd love to have this in Vencord!

PR #3711, opened 2025-10-07
added auto idle on LISTENING

Enqvy:
### listioning.


Vendicated:
i closed for a reason, you know..

low effort copy paste and not particularly useful

4clsd:
:( well if i rewrite the whole code but functionality is same, is that any different at the end?

PR #3712, opened 2025-10-07
AutoIdleWhenLISTENING plugin, Fully recoded edition

Enqvy:
fully recoded not 1 line of code is reused from the original paste
edit: This could go right


4clsd:
hmm still open? thats promising you are thinking about it

Vendicated:
https://youtu.be/l60MnDJklnM

PR #3713, opened 2025-10-08
AutoDNDWhilePlaying: add option to prevent status changes when invisible

KK2-5:
Added as toggle to keep it simple and preserve the current behavior. Thought about letting you ignore specific statuses instead of just invisible, but that felt overkill for such a niche feature.


src/plugins/autoDndWhilePlaying.discordDesktop/index.ts
```diff
@@ -48,6 +53,10 @@ export default definePlugin({
         RUNNING_GAMES_CHANGE({ games }) {
             const status = StatusSettings.getSetting();
 
+            if (settings.store.excludeInvisible && status === "invisible") {
+                return;
+            }
```
thororen1234: It should be checking if saved status is invisible as well because the user might set the status to go invis during a game

```suggestion
            if (settings.store.excludeInvisible && (savedStatus ?? status) === "invisible") return;
```

giiiiiiiiiibb:
Hi — I opened #3755. CI is green; is anything blocking merge?

Vendicated:
I don't think there is any scenario where someone who is invisible would like the plugin to change their status, so I removed the option

Vendicated:
tbh this plugin should also handle manual status changes

1. Open game while online
2. Plugin changes you to DND
3. Change to invisible manually
4. Close game
5. Plugin will now change status to online, which is probably not what you want

I think if the user manually changes their status we should just set savedStatus to null again

Vendicated:
https://github.com/Vendicated/Vencord/commit/5d4d2447ff69d286bb57ed8db54bf7b3e3081c10

PR #3719, opened 2025-10-12
Fix(stickerPaste): Resolve conflict with fakeNitro

5c0:


src/plugins/stickerPaste/index.ts
```diff
@@ -17,8 +17,8 @@ export default definePlugin({
             find: ".stickers,previewSticker:",
             replacement: {
                 match: /if\(\i\.\i\.getUploadCount/,
-                replace: "return true;$&",
+                replace: "try { return true; } finally { if (arguments[2]?.stickers) arguments[2].stickers.length = 0; }$&",
```
Vendicated: don't do this cursed try finally, just put it before the return

Vendicated:
I wanted to give you the benefit of the doubt but this is clearly AI-slop, and doesn't even work

sadan4:
You might have thought the issue was resolved because i did a poor job of describing it in my issue. I'll update it in a bit to be more clear on how to reproduce the issue.

PR #3720, opened 2025-10-12
ViewIcons: fix conflicts with profile/group modal

sadan4:


src/plugins/viewIcons/index.tsx
```diff
@@ -220,7 +220,8 @@ export default definePlugin({
             replacement: {
                 match: /null==\i\.icon\?.+?src:(\(0,\i\.\i\).+?\))(?=[,}])/,
                 // We have to check that icon is not an unread GDM in the server bar
-                replace: (m, iconUrl) => `${m},onClick:()=>arguments[0]?.size!=="SIZE_48"&&$self.openAvatar(${iconUrl})`
+                // stop propagation to avoid opening the edit gdm name/icon modal
+                replace: (m, iconUrl) => `${m},onClick:(e)=>arguments[0]?.size!=="SIZE_48"&&($self.openAvatar(${iconUrl}), e?.stopPropagation?.())`
```
Vendicated: we should consider changing it to `openAvatar(event, url)` and calling stopPropagation inside that function, and pass event from all places, to ensure it doesnt double click anywhere else either

sadan4: did this, haven't tested it yet

src/plugins/viewIcons/index.tsx
```diff
@@ -24,6 +24,7 @@ import { openImageModal } from "@utils/discord";
 import definePlugin, { OptionType } from "@utils/types";
 import type { Channel, Guild, User } from "@vencord/discord-types";
 import { GuildMemberStore, IconUtils, Menu } from "@webpack/common";
+import type * as React from "react";
```
Nuckyz: Import only what's needed

sadan4: I am using a type only namespace import because mouse event is also a global and I specifically want the react version and not the DOM version

Nuckyz: You can rename the import from the react

Vendicated: just import directly and shadow the global

src/plugins/viewIcons/index.tsx
```diff
@@ -67,10 +68,11 @@ const settings = definePluginSettings({
     }
 });
 
-const openAvatar = (url: string) => openImage(url, 512, 512);
-const openBanner = (url: string) => openImage(url, 1024);
+const openAvatar = (url: string, ev?: MouseEvent) => openImage(url, 512, 512, ev);
+const openBanner = (url: string, ev?: MouseEvent) => openImage(url, 1024, undefined, ev);
 
-function openImage(url: string, width: number, height?: number) {
+function openImage(url: string, width: number, height?: number, ev?: MouseEvent) {
```
Nuckyz: Change this one to accept an object with the params named

sadan4: done

src/plugins/viewIcons/index.tsx
```diff
@@ -198,18 +206,21 @@ export default definePlugin({
...
-            find: "return{avatarProps:{",
-            replacement: {
-                match: /(?<=avatarProps:(\i),eventHandlers:(\i).{0,50}?)return null==/,
-                replace: 'Object.assign($2,{style:{cursor:"pointer"},onClick:()=>$self.openAvatar($1.src)});$&',
-            }
+            find: ".overlay:void 0,status:",
+            replacement: [
+                {
+                    match: /avatarSrc:(\i),eventHandlers:(\i).+?"div",.{0,100}className:\i,/,
+                    replace: '$&style:{cursor:"pointer"},onClick:e=>{$self.openAvatar($1,e)},',
```
Vendicated: Introducing `e` here while using `$1` will explode if `$1` happens to be called `e` too. You should use a safer name like vcEvent
```suggestion
                    replace: '$&style:{cursor:"pointer"},onClick:vcEvent=>{$self.openAvatar($1,vcEvent)},',
```

Vendicated: same for all other patches

sadan4:
This also fixes it not working in the user profile sidebar in DMs and replaces discords worse implementation in the user profile modal

PR #3727, opened 2025-10-17
fix(BetterSettings): prevent "Other options" from appearing

sillyElvy:
The "Other options" section only contains 
- "Developer Education", which is a blank page for users
- The Log Out button, which isn't really a setting
- The merch button, who needs that
- The changelog button, which isn't a setting either

Currently, the other settings options would get rendered thrice and cause display and behaviour issues.
As none of the sections within it are relevant as settings themselves, removing all of them seems like the best (and easiest) option


Vendicated:
WHEN YOU SEE IT

PR #3736, opened 2025-10-24
fix(FixYoutubeEmbed): Fix Error 15 in Youtube Embeds

emabrey:
**TLDR:**
This should fix the "Error 15" in youtube embeds related to youtube identity verification. This also adds detection and workaround for the Youtube "you look like a bot" blocks and attempts to reload the embed in that case. Also modifes the adblock plugin to be aware of the non-standard (but google official) `youtube-nocookie` domain used to bypass the identity error.

**More in depth:**
When attempting to play [a blocked video](https://www.youtube.com/watch?v=qXRCe0njp3I) with the FixYoutubeEmbed plugin, I kept getting error despite the plugin technically "working" (as in, the UMG block was bypassed but the video still would not play).
Image showing error 15

Upon investigation, this was related to Youtube's identity verification policies as shown in the debug message for the javascript player:

```
"debug_error": "{\"errorCode\":\"embedder.identity.denied\",\"errorDetail\":\"0\",\"errorMessage\":\"This video is unavailable\",\"hF\":\"Error code: 15\",\"ZV\":\"0;a6s.0\",\"BE\":2,\"cpn\":\"\"}"
```

Internet research revealed that Youtube is looking for a cookie to be set in order to validate age, and it defaults to failing you if the hueristics that generate the cookie have not run and provided you an identity cookie. After switching to the `youtube-nocookie.com` domain, which is Youtube's official GDPR compliant replacement for `youtube.com`, I was able to solve this problem because in that environment no cookie can be expected, which side steps the issue entirely. However, I also needed to add it to the CSP for Vencord to prevent `frame-src` mismatch related CSP errors. After fixing that problem, I ran into a new issue - Youtube was repeatedly detecting me as a bot and refusing to play the video without login (which is impossible, or at minimum extremely difficult in the embedded context). I added additional code to handle the bot blocking in addition to the original block workaround for the UMG style blocks. Finally, I noticed that the ad blocking wasn't always working because of the new domain, so I modified the youtubeAdblock plugin to be aware of the possibility of the `youtube-nocookie.com` domain.

### DIFF
--- a/src/plugins/fixYoutubeEmbeds.desktop/index.ts
+++ b/src/plugins/fixYoutubeEmbeds.desktop/index.ts
@@ -10,5 +10,5 @@ import definePlugin from "@utils/types";
 export default definePlugin({
     name: "FixYoutubeEmbeds",
     description: "Bypasses youtube videos being blocked from display on Discord (for example by UMG)",
-    authors: [Devs.coolelectronics]
+    authors: [Devs.coolelectronics, Devs.cutestbunny],
 });
--- a/src/plugins/fixYoutubeEmbeds.desktop/index.ts
+++ b/src/plugins/fixYoutubeEmbeds.desktop/index.ts
@@ -11,5 +11,5 @@ export default definePlugin({
     name: "FixYoutubeEmbeds",
     description: "Bypasses youtube videos being blocked from display on Discord (for example by UMG)",
     tags: ["Media", "Utility"],
-    authors: [Devs.coolelectronics]
+    authors: [Devs.coolelectronics, Devs.cutestbunny],
 });
--- a/src/plugins/fixYoutubeEmbeds.desktop/native.ts
+++ b/src/plugins/fixYoutubeEmbeds.desktop/native.ts
@@ -4,24 +4,66 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
-import { RendererSettings } from "@main/settings";
-import { app } from "electron";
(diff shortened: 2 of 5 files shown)


emabrey:
Made a small modification to the PR so it won't generate a plugin error message in the event the plugin is disabled and we run into a `null` value we cant handle.

emabrey:
Sorry for force-push spam. I didn't realize that the github ui merge would be non fast forward and I had to undo what github did to the repo lol. sigh >.>

emabrey:
I made some changes to remove the logging statements that weren't previously part of the plugins, and to move the CSP editing from the main CSP file into the plugin itself so that users who aren't opted-in to using the plugin don't have any modifications to their CSP. I also added additional checks to ensure the plugin doesnt have null dereferences or array out of bounds accesses done within the frame javascript context.

emabrey:
Just updated the branch to match main. If there is anything that needs to be done to make this mergeable please let me know.

JimCarlTay:
> Just updated the branch to match main. If there is anything that needs to be done to make this mergeable please let me know.

It currently says that this pull request's workflow needs to be approved by a maintainer before it can start. There is also an issue notice that this has to be up-to-date with base.

emabrey:
> > Just updated the branch to match main. If there is anything that needs to be done to make this mergeable please let me know.
> 
> It currently says that this pull request's workflow needs to be approved by a maintainer before it can start. There is also an issue notice that this has to be up-to-date with base.

I've updated the branch it to bring it in line with base/upstream.

PR #3737, opened 2025-10-25
feat(replaceGoogleSearch): option to set custom engine as default

norowachi:
added plugin option to set your custom search engine as the default one instead of having google or a drop down menu for searching, with related menu logic.

### DIFF
--- a/packages/discord-types/src/common/Guild.d.ts
+++ b/packages/discord-types/src/common/Guild.d.ts
@@ -47,7 +47,7 @@ export class Guild extends DiscordRecord {
     };
     premiumProgressBarEnabled: boolean;
     premiumSubscriberCount: number;
-    premiumTier: number;
+    premiumTier: 0 | 1 | 2 | 3;
     profile: {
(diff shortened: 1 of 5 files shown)


src/plugins/replaceGoogleSearch/index.tsx
```diff
@@ -32,6 +32,11 @@ const settings = definePluginSettings({
         description: "The URL of your Engine",
         type: OptionType.STRING,
         placeholder: "https://google.com/search?q="
+    },
+    setDefaultEngine: {
+        description: "Set your custom engine as the default search engine",
```
Vendicated: Good idea! Just some feedback to make it more user friendly:
 
1. this should be changed to better reflect that this will make the engine the only option and replace the menu
2. it would be nicer if this were a dropdown where the user can choose between	- Off
	- One of the default engines
	- Their custom engine

norowachi: done! thanks for the suggestion!

PR #3741, opened 2025-10-27
VencordToolbox: fix patch

sadan4:


Vendicated:
not a fan of this patch, I'll fix it myself

PR #3744, opened 2025-10-28
Silence Call Plugin

YONN2222:
Added a SilenceCalls plugin that lets you mute an incoming call using a button in the chat bar.

image


Vendicated:
this doesn't actually do what it says

also doesn't pass tests

YONN2222:
Where does it not do what it says, and which tests does it fail?

PR #3745, opened 2025-10-28
Updater: Move to full githash

thororen1234:


Vendicated:
Need to also update workflows https://github.com/Vendicated/Vencord/blob/main/.github/workflows/build.yml#L50-L56

and maybe more places where it uses the git hash

src/main/updater/http.ts
```diff
@@ -49,8 +49,7 @@ async function calculateGitChanges() {
     const data = await githubGet(`/compare/${gitHash}...HEAD`);
 
     return data.commits.map((c: any) => ({
-        // github api only sends the long sha
-        hash: c.sha.slice(0, 7),
+        hash: c.sha,
         author: c.author.login,
```
Vendicated:
```suggestion
        author: c.author?.login ?? c.commit?.author?.name ?? "Ghost",
```

src/plugins/_core/supportHelper.tsx
```diff
@@ -22,6 +22,7 @@ import ErrorBoundary from "@components/ErrorBoundary";
 import { Flex } from "@components/Flex";
 import { Link } from "@components/Link";
 import { openUpdaterModal } from "@components/settings/tabs/updater";
+import { gitHashShort } from "@shared/vencordUserAgent";
```
Vendicated:
```suggestion
import { gitHash, gitHashShort } from "@shared/vencordUserAgent";
```
and remove from below

thororen1234: just added them

PR #3747, opened 2025-10-29
fix(openInApp): Tidal regex

Aztup:
Tidal has changed their link format recently and it broke the open in app plugin this PR simply update the regex to the new format.

### DIFF
--- a/src/plugins/openInApp/index.ts
+++ b/src/plugins/openInApp/index.ts
@@ -52,7 +52,7 @@ const UrlReplacementRules: Record<string, URLReplacementRule> = {
         description: "Open Epic Games links in the Epic Games Launcher",
     },
     tidal: {
-        match: /^https:\/\/tidal\.com\/(?:browse\/)?(track|album|artist|playlist|user|video|mix)\/([^/?]+)(?:[/?].*)?$/,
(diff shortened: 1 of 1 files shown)


src/plugins/openInApp/index.ts
```diff
@@ -52,7 +52,7 @@ const UrlReplacementRules: Record<string, URLReplacementRule> = {
         description: "Open Epic Games links in the Epic Games Launcher",
     },
     tidal: {
-        match: /^https:\/\/tidal\.com\/browse\/(track|album|artist|playlist|user|video|mix)\/(.+)(?:\?.+?)?$/,
+        match: /^https:\/\/tidal\.com\/(track|album|artist|playlist|user|video|mix)\/(\d+).+/,
```
lewisakura: i'm not sure what you're trying to do at the end here? you need to capture the full uuid, you're only capturing the first couple of digits and discarding everything else, i'm not sure why this was changed

lewisakura: also this should probably just have the `browse/` bit be an optional so old links still work

Aztup: Oh woops completely forgot that Tidal use different id depending on what type of content. Sorry about that!

PR #3751, opened 2025-10-30
ci: auto rebase new PRs to dev if based on main

lewisakura:
if it's set back to main it'll leave the PR alone, this is specifically for new PRs so we (and by we i mean me) don't accidentally merge stuff to main by accident


Vendicated:
meh, I don't really think this would be good to have

there are valid reasons for targeting main

PR #3752, opened 2025-10-30
Decor: Fix Avatar Decor Hook

thororen1234:


thororen1234:
they removed isStaff from `isDecoupledGameClippingEnabled(){` which experiments used so I removed it as well

PR #3753, opened 2025-10-30
fix(openInApp): Tidal regex (again)

Aztup:
The previous PR for the Tidal regex (https://github.com/Vendicated/Vencord/pull/3747) does not support the `listen` subdomain and would also break on the universal link this PR fixes that too so hopefully it supports all the new format and links now.


src/plugins/openInApp/index.ts
```diff
@@ -52,7 +52,7 @@ const UrlReplacementRules: Record<string, URLReplacementRule> = {
         description: "Open Epic Games links in the Epic Games Launcher",
     },
     tidal: {
-        match: /^https:\/\/tidal\.com\/(?:browse\/)?(track|album|artist|playlist|user|video|mix)\/(.+)(?:\?.+?)?$/,
+        match: /^https:\/\/(?:listen\.)?tidal\.com\/(?:browse\/)?(track|album|artist|playlist|user|video|mix)\/([a-f0-9-]+).*$/,
```
Vendicated: `.*$` doesn't really make much sense, they essentially cancel each other out. At that point you could just drop all of it including the $.

Star-Made: As this is substitution we need to drop the "/u" in the link, this was accomplished through the .\*, granted the $ is not necessary .* should suffice. https://tidal.com/track/68689136/u -> tidal://track/68689136 and not tidal://track/68689136/u

PR #3754, opened 2025-10-31
Performance improvements to settings store and plugin system

omnizs:
## Performance Improvements

- Optimized SettingsStore.notifyListeners to reduce iterations and cache listener references
- Optimized useSettings hook to prevent unnecessary re-subscriptions using useMemo for stable path references
- Added React.memo to PluginCard and ExcludedPluginsList components to prevent unnecessary re-renders
- Cached plugin lookups in startAllPlugins and subscribeAllPluginsFluxEvents functions to avoid multiple property accesses
- Improved path resolution in SettingsStore by using direct property access instead of reduce


nin0-dev:
aislop, also

omnizs:
> aislop, also
> 
> 

I noticed this after I published my PR and immediately started editing it. And now I've published it and you left me a comment at the same time. Thank you, I'll keep that in mind.

PR #3757, opened 2025-11-01
[feat] plugin: silentMentions

chri-k:
A very simple plug-in which modifies the `allowed_mentions` field of sent messages based on which mentions in the message had a set prefix.

Since `@everyone` and `@here` can already be escaped with a `\`, their behaviour is not modified.


31a05b9c:
it is worth noting that since only bots can normally do this, there is a possibility of this causing detection issues

PR #3742, opened 2025-10-28
feat: unlimited stickers plugin

chev0004:
### Summary

This PR adds the `UnlimitedStickers` plugin which lets Nitro users send local images as native Discord stickers.

### Features

-   **Local Sticker Library**: Send any supported local image as a Discord sticker.
-   **Sticker Categories**: Uses the sub-folders in your designated sticker directory as categories.
-   **Favorites & Recents**:
    -   Mark stickers as Favorites.
    -   Keeps track of your Recently Used stickers.
-   **Search**: Find a sticker by searching for its filename or category name.
-   **Internationalization**: Currently only translated in English and Japanese.

### How It Works

1.  **Stickers Server**: On its first run, the plugin creates a new server named "Stickers". This server acts as a temporary storage space for your stickers.

2.  **The "One-Slot" Method**: Discord servers have a limited number of sticker slots. To work around this, the plugin uses a single, reusable sticker slot within the "Stickers" server.

3.  **The Sending Process**: When you select a sticker to send:
    a. The plugin reads the image file from your local disk and converts it into a Base64 data string.
    b. An API request is sent to delete the previously uploaded sticker from the reusable slot in the "Stickers" server (if there's any).
    c. A new API request is immediately sent to upload your selected image into that now-empty slot.
    d. The plugin receives the ID of this newly created sticker and uses it to send the sticker in your current channel.

### Configuration

1.  **Set Your Sticker Path**: In the plugin settings, you must provide the **absolute path** to the folder where you store your stickers.
    -   *Example (Windows):* `C:\Users\chev\Documents\Stickers`
    -   *Example (Linux/macOS):* `/home/chev/Documents/Stickers`

2.  **Organize Your Folders**: Inside your main sticker path, create sub-folders to act as categories. Place your image files inside these sub-folders. The plugin will use the folder names as the category names in the sticker picker.

    *Example folder structure:*
    ```
    Stickers/
    ├── Kurumi-chan/
    │   ├── good_morning.png
    │   └── sobbing.png
    └── Menhera-kun/
        ├── sharpening_knife.png
        └── 了解.png
    ```

### Disclaimers

-   **Nitro Is Required**: You are essentially sending a sticker from another server, which is a feature exclusive to Discord Nitro subscribers.
-   **Server Creation**: The plugin will create a server named "Stickers". This is essential for its operation. Even if you delete it, it'll just come back. Make sure that you have enough server slots.

### Optimization Efforts

[...]

PR #3761, opened 2025-11-03
feat(plugin): UnlimitedStickers

chev0004:
Nitro exclusive plugin that lets you upload images and send them as stickers by temporarily uploading them in a server when sending

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.13.4",
+    "version": "1.13.5",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
     "bugs": {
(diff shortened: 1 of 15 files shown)


nin0-dev:
this plugin is a security hazard, you should never be able to read an arbitrary directory from renderer

also the implementation is just insane 😭

chev0004:
> this plugin is a security hazard, you should never be able to read an arbitrary directory from renderer
> 
> also the implementation is just insane 😭

There is no file access from the renderer and it's not reading anything from any arbitrary directories. Folders are uploaded, not read from the disk

Vendicated:
why did you open a duplicate pull request? https://github.com/Vendicated/Vencord/pull/3742

chev0004:
> why did you open a duplicate pull request? https://github.com/Vendicated/Vencord/pull/3742

I fixed the issues you mentioned on the last PR, and the other was kinda getting filled with so much commits so I thought it'd be better to have a cleaner one.

PR #3765, opened 2025-11-04
feat(Plugin): Add MessageExpiry

extremq:
I have created a plugin that lets you set an expiration time for your messages. It supports exclusions for DMs (all), channel ids, guild ids. It has different options such as editing the message before deleting, getting notified for messages that have been deleted, forceful cleanup of message queue.

It also has handy slash commands that let you add exclusion directly from the channel/guild you are in.

When this plugin is toggled or the user forcefully flushes the message queue, it may mass delete messages depending on how long the expiration time is. I don't know the rate limits discord uses, so I did not implement any yet because I want input from more experienced people.

The functions that delete messages are all inside the `MessageStore` class, more specifically `pruneMessages` and `flushMessages`. 

I would also want to know if there is such thing as the plugin doing something before the app exists in order to make sure messages are deleted.

Either way, If this gets merged, I have plans to properly integrate this into the UI so it is way easier to toggle stuff and visually see how many messages are left to be deleted! There is also a `README.md` with more details :D

### DIFF
--- a/src/plugins/messageExpiry/index.tsx
+++ b/src/plugins/messageExpiry/index.tsx
@@ -220,8 +220,6 @@ export default definePlugin({
                     break;
             }
 
-            console.dir(message);
-
             msgStore.addMessage({
                 id: message.id,
                 channelId: message.channel_id,


extremq:
image
:thinking: 

Do let me know if you think such plugin can be implemented (especially with proper rate limiting). My ideal scenario is that in the end, you would have some counter in the UI that tells you something like "Messages waiting to expire: x" and when you disable the script/forcefully delete everything, it would batch it in groups of 10-20. :+1:

Vendicated:
deleting all your messages is extremely annoying, just don't do it

PR #3770, opened 2025-11-07
protexBlocks Plugin

noxitechdk:
Adds protexBlocks plugin for integration with the Protex moderation bot API.
The plugin automatically checks users against the API to detect blocked users.

-# No API Key is required


lewisakura:
> No untrusted third party APIs. Popular services like Google or GitHub are fine, but absolutely no self hosted ones

PR #3773, opened 2025-11-07
Decor: Fix not displaying in multiple places

thororen1234:
for whatever reason the patch with `getAvatarDecorationURL:` arguments[0] is undefined in chat so I made it use the actual arg

using the value of avatarDecorationOverride in `$1??vcDecorAvatarDecoration??($&)` broke decor removing the `$1??` fixes it


thororen1234:
added a fix for decor not applying in select/create modal

thororen1234:
This apperantly always shows the decor in the shop so I'll need to fix that

Vendicated:
yop fix also stable compat yay or nay?

thororen1234:
> yop fix also stable compat yay or nay?
yes 👍 

I 100% forgot to respond

PR #3776, opened 2025-11-12
Fix ChatBarButtons

thororen1234:
They added keys so if the key is null its ignored


Vendicated:
I don't like this implementation, we should just push into the children

Vendicated:
thank you anyway!!

https://github.com/Vendicated/Vencord/commit/e90b31e717277b8142bed5c3afcc1654570b9433

PR #3777, opened 2025-11-13
RoleColorEverywhere: Fix patch

sadan4:


sadan4:
should:tm: be stable compatible

PR #3781, opened 2025-11-15
fix contributor/donor badges, fix member list in group dms for IrcColors plugin

alfuwu:
the Vencord Contributor/Donor badges currently seem to be broken:
image
this pr fixes that, restoring the badges:
image

additionally, the IrcColors plugin wasn't coloring in the member list of group DMs when set to `Only Apply Color In Dms`, which this pr also fixes

### DIFF
--- a/src/plugins/ircColors/index.ts
+++ b/src/plugins/ircColors/index.ts
@@ -59,7 +59,7 @@ const settings = definePluginSettings({
 export default definePlugin({
     name: "IrcColors",
     description: "Makes username colors in chat unique, like in IRC clients",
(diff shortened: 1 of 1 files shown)


nin0-dev (commented):
avoid adding yourself to devs for very minor changes like this

Vendicated:
Thank you for the fixes! ❤️ 

For future reference, avoid sending pull requests from the main branch. Instead, always use [feature branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow). This is better for both you and us

PR #3784, opened 2025-11-16
Add port of Pocket Bird as a plugin

IdreesInc:
Hello, This is a port of my open source desktop pet "Pocket Bird". The pet is usually run as a bookmarklet or browser extension, so this port allows it to run automatically within the discord client.
The full source code of the project is available here: https://github.com/IdreesInc/Pocket-Bird 
The plugin adds a small pixel-art bird that hops around your discord and lands on images, as well as the ability to insert sticky notes that the bird can land on. New birds can be unlocked by catching the feathers that fall every so often. It's a relatively simple pet and does not store any data elsewhere besides the Vencord datastore.

Note: I added the code locally in this PR rather than importing externally like with the other desktop pet "Oneko" as It seemed safer for end users that way. Let me know if it is preferred to host the code externally for programs that minimally affect the application. The changes specific for vencord can be seen here: https://github.com/IdreesInc/Pocket-Bird/blob/main/src/context.js#L279

### DIFF
--- a/src/plugins/pocketBird/README.md
+++ b/src/plugins/pocketBird/README.md
@@ -1,6 +1,5 @@
 # Pocket Bird
 
-![License](https://img.shields.io/github/license/IdreesInc/Pocket-Bird)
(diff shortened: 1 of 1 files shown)


Vendicated:
sorry but no

PR #3786, opened 2025-11-16
Re-add moreUserTags plugin completely reworked

StellarosYT:
Re-add moreUserTags plugin completely reworked without any crash


Vendicated:
Thanks for your contribution!

However, this is a duplicate and also uses multiple bad practices (polluting the global object, abusing an interval, not using proper plugin lifecycle methods, etc)

PR #3787, opened 2025-11-17
Improve Toolbox

Vendicated:
1. Allows plugins to add any menu item, not just simple actions
2. Adds Theme settings submenu
3. Adds Plugin settings submenu

https://github.com/user-attachments/assets/0b84a78f-d14a-4dd5-bbc1-41c31fe56fa7

### DIFF
--- a/src/components/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary.tsx
@@ -20,6 +20,7 @@ import { LazyComponent, LazyComponentWrapper } from "@utils/lazyReact";
 import { Logger } from "@utils/Logger";
 import { Margins } from "@utils/margins";
 import type { React } from "@webpack/common";
+import { ComponentType } from "react";
 
 import { ErrorCard } from "./ErrorCard";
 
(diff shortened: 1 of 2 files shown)


src/components/settings/tabs/themes/index.tsx
```diff
@@ -80,6 +81,23 @@ function UserscriptThemesTab() {
...
 }
 
-export default IS_USERSCRIPT
+const Tab = IS_USERSCRIPT
     ? wrapTab(UserscriptThemesTab, "Themes")
     : wrapTab(ThemesTab, "Themes");
+
+export default Tab;
+
+export function openThemeModal() {
```
Vendicated: fixme: this should probably be a more generic function that supports opening all settings pages as modal. and it should use more reasonable classnames

src/plugins/vencordToolbox/index.tsx
```diff
@@ -20,40 +20,227 @@ import "./styles.css";
...
+                                label={"Open Settings"}
+                                action={() => openPluginModal(p)}
+                            />
+
+                            <Menu.MenuCheckboxItem
+                                id={`${p.name}-enabled`}
+                                label={"Enable Plugin"}
+                                checked={isPluginEnabled(p.name)}
+                                action={() => {
+                                    // TODO
```
Vendicated: TODO

We already have a function for this in the plugin settings code. that function should probably just be made more generic and moved to the PluginManager file

Vendicated:
I really like the plugin menu and would use it myself for a few plugins but I feel like having all plugins in there is too insane. Maybe we could allow users to pick a few plugins to display there, or Vencord could track (locally) which plugin's settings you open the most and automatically display those

Vendicated:
It would be helpful if people could suggest new Toolbox entries that would be useful, either for a specific plugin or for a feature in Vencord

Kyuuhachi:
BetterSettings' settings cog submenu lists only enabled plugins. Whatever filter is ultimately chosen, the two should preferably be synced.

mantikafasi:
ready to merge

mantikafasi:
@copilot this is ready to merge

Vendicated:
yop it is

PR #3788, opened 2025-11-17
[oneko] Add support for different followers and hot-reload

autodistries:
- Add settings for this plugin with a dropdown list of usable followers (currently Oneko (the original), Tora (a cat), and Dog (...))
- This selects different sprites generated from the original oneko program (I can release the tool that assembles the images too) (the images have then been edited manually so that i.e. the Zzzs are more visible) (the images were uploaded to imgur but that can be changed)
- The JS script is loaded on plugin startup, then is cached and reused when changing selected follower


Vendicated:
I'm not interested in any modifications to the plugin

PR #3797, opened 2025-11-24
Unify styles into single root node & support preload sandboxing 

Vendicated:
Unifies and organises all Vencord styles into one central root node. The main motivation behind this is that previously, styles were all over the place and it was easy to accidentally change the order in which they were inserted, which breaks specificity. That issue happened in an earlier version of this PR. Thus I decided to also reorganise.

image

---

Also refactors the preload script to support [Electron process sandboxing](https://www.electronjs.org/docs/latest/tutorial/sandbox), which heavily restricts what the preload can access.

This will only be used on Vesktop because the Discord desktop app does not support sandboxing. But once they eventually enable it, we will already be prepared :)

This is needed for https://github.com/Vencord/Vesktop/pull/1206

___
Same PR because they are tangentially related (both refactor Vencord css logic to use IPC instead of doing it fully in the preload)


Vendicated:
Essentially sandboxing makes it so that the preload script is also sandboxed and has similar access to the renderer (plus a few electron apis). So we have to move all file operations and such to the main process via IPC call. That's what this PR does

browser/background.js
```diff
@@ -12,7 +12,7 @@
     ({ responseHeaders, type, url }) => {
         if (!responseHeaders) return;
 
-        if (type === "main_frame") {
+        if (type === "main_frame" && url.includes("discord.com")) {
```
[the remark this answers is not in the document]
Vendicated: DIDNT ASK

browser/modifyResponseHeaders.json
```diff
@@ -15,7 +15,8 @@
             ]
         },
         "condition": {
-            "resourceTypes": ["main_frame", "sub_frame"]
+            "resourceTypes": ["main_frame", "sub_frame"],
+            "urlFilter": "https://*.discord.com/*"
```
Vendicated: this needs testing to ensure no regression

browser/VencordNativeStub.ts
```diff
@@ -55,7 +55,14 @@ window.VencordNative = {
 
     native: {
         getVersions: () => ({}),
-        openExternal: async (url) => void open(url, "_blank")
+        openExternal: async (url) => void open(url, "_blank"),
+        getRendererCss: async () => {
+            await metaReady;
+
+            return fetch(RENDERER_CSS_URL)
```
Vendicated: TODO: userscript implementation

src/api/Styles.ts
```diff
@@ -135,31 +186,4 @@ export const compileStyle = (style: Style) => {
...
-/**
- * @param prefix The prefix to add to each class, defaults to `""`
- * @returns A classname generator function
- * @example
- * const cl = classNameFactory("plugin-");
- *
- * cl("base", ["item", "editable"], { selected: null, disabled: true })
- * // => "plugin-base plugin-item plugin-editable plugin-disabled"
- */
-export const classNameFactory = (prefix: string = "") => (...args: ClassNameFactoryArg[]) => {
```
Vendicated, line 157: moved to avoid circular dependency issues with BaseText and these do make more sense to be in utils anyway

Vendicated (commented):
Some Notes:

PR #3798, opened 2025-11-24
add tabs to notification log modal

rushy-me:
adds tabs to the notification log so plugins can have their own filtered view. plugins can register a tab via the notificationLogTab property

already added it to betterSessions and relationshipNotifier as examples

Screenshot 2025-11-24 141611


Vendicated:
This should just be a prop passed to the showNotification function

```js
showNotification({
	category: "Example Category",
	title: "Cool Title",
	body: "wowie"
});
```

PR #3803, opened 2025-11-27
Complete new settings UI support, add sidebar icons, fix version info, and add option to use the old UI

vvareo:
Finished the new settings UI support. The existing branch only added a placeholder,
so I wired up the full Vencord section and tabs, added proper icons, fixed the
version/info rows, and added an option to force the old UI if people prefer it.

First time opening a PR here, so if anything should be done differently, just let me know.

image

### DIFF
--- a/src/plugins/_core/settings.tsx
+++ b/src/plugins/_core/settings.tsx
@@ -103,14 +103,17 @@ export default definePlugin({
             find: ".versionHash",
             replacement: [
                 {
-                    match: /\.compactInfo.+?\[\(0,\i\.jsxs?\)\((.{1,10}),(\{[^{}}]+\{.{0,20}.versionHash,.+?\})\),/,
-                    replace: "$&$self.makeInfoElements($1,$2),"
+                    match: /\.compactInfo.+?(?=null!=(\i)&&(.{0,20}\i\.Text.{0,200}?,children:).{0,15}?("span"),({className:\i\.versionHash,children:\["Build Override: ",\1\.id\]\})\)\}\))/,
+                    replace: (m, _buildOverride, makeRow, component, props) => {
+                        props = props.replace(/children:\[.+\]/, "");
+                        return `${m},$self.makeInfoElements(${component},${props}).map(e=>${makeRow}e})),`;
+                    }
                 },
                 {
                     match: /\.info.+?\[\(0,\i\.jsxs?\)\((.{1,10}),(\{[^{}}]+\{.{0,20}.versionHash,.+?\})\)," "/,
                     replace: (m, component, props) => {
                         props = props.replace(/children:\[.+\]/, "");
-                        return `${m},$self.makeInfoElements(${component}, ${props})`;
+                        return `${m},$self.makeInfoElements(${component},${props})`;
                     }
                 },
                 {


vvareo:
Moved the settings icons into the Icons component for consistency and adjusted their size/alignment to match Discord’s native settings icons better

src/plugins/_core/settings.tsx
```diff
@@ -76,11 +107,122 @@ export default definePlugin({
             find: "2025-09-user-settings-redesign-1",
             replacement: {
                 match: /enabled:![01],showLegacyOpen:/g,
-                replace: "enabled:false,showLegacyOpen:"
+                replace: (m: string) =>
```
Vendicated: we shouldn't include this. let's just remove it

src/plugins/_core/settings.tsx
```diff
@@ -28,6 +29,30 @@ import gitHash from "~git-hash";
...
+    icon?: () => React.ReactNode;
+    render?: () => React.ReactNode;
+}
+
+interface SettingsLayoutBuilder {
+    key?: string;
+    buildLayout(): SettingsLayoutNode[];
+}
+
+function getSettingsCfg(): any { try { return Settings.plugins.Settings; } catch { return null; } }
```
Vendicated: this shouldn't be necessary. it can never error unless you're calling it too early. this shouldn't be a problem, but if it is, you can just use PlainSettings.plugins.Settings instead

src/plugins/_core/settings.tsx
```diff
@@ -76,11 +107,122 @@ export default definePlugin({
...
+        const vencordEntries: SettingsLayoutNode[] = [
+            makeEntry("vencord_main", "Vencord", VencordTab, MainSettingsIcon),
+            makeEntry("vencord_plugins", "Plugins", PluginsTab, PluginsIcon),
+            makeEntry("vencord_themes", "Themes", ThemesTab, PaintbrushIcon),
+            makeEntry("vencord_cloud", "Cloud", CloudTab, CloudIcon),
+            makeEntry("vencord_backup_restore", "Backup & Restore", BackupAndRestoreTab, BackupRestoreIcon)
+        ];
+
+        if (!IS_UPDATER_DISABLED && UpdaterTab) {
+            vencordEntries.push(makeEntry("vencord_updater", "Updater", UpdaterTab, UpdaterIcon));
```
Vendicated: please keep the order unchanged. Updater should be above cloud. consider using
```js
const vencordEntries = [
	makeEntr(),
	!IS_UPDATER_DISABLED && UpdaterTab && makeEntry()
].filter(isTruthy)
```

src/plugins/_core/settings.tsx
```diff
@@ -76,11 +107,122 @@ export default definePlugin({
...
+                break;
+            }
+        }
+
+        layout.splice(insertIndex, 0, vencordSection);
+
+        return layout;
+    },
+
     customSections: [] as ((SectionTypes: SectionTypes) => any)[],
```
Vendicated, line 286: customSections should remain respected. this is used by Vesktop and some userplugins

src/plugins/_core/settings.tsx
```diff
@@ -76,11 +107,122 @@ export default definePlugin({
...
+            }
+            case "belowActivity": {
+                const idx = findIndexByKey(layout, "activity_section");
+                insertIndex = idx === -1 ? layout.length : idx + 1;
+                break;
+            }
+            case "bottom":
+            default: {
+                const idx = findIndexByKey(layout, "logout_section");
+                insertIndex = idx === -1 ? layout.length : idx;
```
Vendicated: it would be better to use something like 2 as the fallback value. we want to keep it more towards the top because users will not know that they should scroll all the way down

src/plugins/_core/settings.tsx
```diff
@@ -28,6 +29,30 @@ import gitHash from "~git-hash";
 type SectionType = "HEADER" | "DIVIDER" | "CUSTOM";
 type SectionTypes = Record<SectionType, SectionType>;
 
+const LayoutType = { SECTION: 1, ENTRY: 2, PANEL: 3, PANE: 4 } as const;
```
Vendicated:
```suggestion
const enum LayoutType {
	SECTION = 1,
	ENTRY = 2,
	PANEL = 3,
	PANE = 4
};
```

src/plugins/_core/settings.tsx
```diff
@@ -28,6 +29,30 @@ import gitHash from "~git-hash";
...
+interface SettingsLayoutNode {
+    key?: string;
+    type: number;
+    legacySearchKey?: string;
+    useLabel?: () => string;
+    useTitle?: () => string;
+    buildLayout?: () => SettingsLayoutNode[];
+    icon?: () => React.ReactNode;
+    render?: () => React.ReactNode;
+}
```
Vendicated, line 64:
```suggestion
interface SettingsLayoutNode {
	type: LayoutType;
    key?: string;
    legacySearchKey?: string;
    useLabel?(): string;
    useTitle?(): string;
    buildLayout?(): SettingsLayoutNode[];
    icon?(): React.ReactNode;
    render?(): React.ReactNode;
}
```

Vendicated:
Right now all titles are duplicated. The titles inside our components should be removed and the ones in this PR should be adjusted to match the old titles

image

src/components/Icons.tsx
```diff
@@ -444,3 +444,99 @@ export function PlaceholderIcon(props: IconProps) {
...
+    );
+}
+
+export function CloudIcon(props: IconProps) {
+    return (
+        <Icon
+            {...props}
+            viewBox="0 0 24 24"
+        >
+            <g transform="translate(1 5) scale(1.1)">
```
Vendicated: all these transforms are really weird. where did you take these svgs from?

vvareo: The icons are based on the Discord assets, which I found in a Figma file (https://www.figma.com/community/file/1445040225357227161), except the cloud (I made that one) and a slightly adjusted plugins icon. The odd transforms came from how I exported them originally, for some reason I decided to fix it that way rather than re exporting them which I’ve now done properly

src/plugins/_core/settings.tsx
vvareo: By this I assume you meant the entire toggle for the old UI, which would have been nice to have but I imagine it would get removed eventually

Vendicated:
Thanks again! Great first PR

PR #3807, opened 2025-11-30
Plugin "TF2 Old English Translator"

nfsmaniac:
As a little fun project for my friend, I am opening this PR as to publish new plugin "TF2 Old English Translator", which aims to reproduce a Medieval (Chat) Mode from the game "Team Fortress 2".
When this mode is activated, all messages sent into chat gets automatically translated into old English -- with randomly added prepended or appended phrases; and vocabulary replaced by ancient words or phrases, based on the `autorp.txt` dictionary file from TF2.

This is my first Vencord plugin, and while I have tried to be as careful as possible and integrate my plugin into Vencord as best as I can, I was forced to do a few patches to the Vencord's base components for plugin settings:

First, I noticed, that `NumberSetting.tsx` cannot handle decimal numbers stored inside the NumericUpDown-like component, as it was explicitly casting every `Number` datatype to `BigInt`, causing console error and component's malfunction, whenever non-integer `Number` type was present.

Then, I really wanted to implement a "Reset Settings" feature inside of the Plugin's settings Modal, but I struggled to enforce UI's refresh upon changing component's current value programatically, outside of user's clicking interaction. After so many attempts, some React's `.useEffect()` magic added to each of the individual Plugin Setting component (like `BooleanSetting`, `SelectSetting`, etc.) finally worked for me. I will leave that for review, as I don't really have much of experience with React nor TypeScript, but so far it seems to work flawlessly for me.

These Vencord patches mentioned above were mandatory to make the plugin possible, but I believe other plugin developers can benefit from them too.

You may find pre-compiled Vencord browser extensions in the attachments below for testing:

[TF2_Old_English_Translator_extension-firefox_RC1.zip](https://github.com/user-attachments/files/23837190/TF2_Old_English_Translator_extension-firefox_RC1.zip)
[TF2_Old_English_Translator_extension-chrome_RC1.zip](https://github.com/user-attachments/files/23837191/TF2_Old_English_Translator_extension-chrome_RC1.zip)


Vendicated:
cute idea but this is way too much code for something that is very gimmicky and not particularly useful

PR #3809, opened 2025-11-30
Fix plugin chat buttons not appearing in sidebar chats

### LINKED ISSUE #3715: [Bug] `PreviewMessage` plugin doesn't work on forums

The preview message feature does not work in a forum channel

The preview message button should appear inside forums

1. Enable the PreviewMessage plugin
2. Go inside a forum post
3. Observe that the PreviewMessage button doesn't appear if you want to preview your message

### Errors

```
Replace this text with your crash-log.
```

### Request Agreement

- [x] I am a Vencord Developer

vvareo:
Previously plugins that use the `isMainChat` (PreviewMessage, SendTimestamps, SilentMessageToggle, SilentTyping, and Translate) didn’t appear in the sidebar composer for threads, forum posts, or VC text
Discord labels those as "sidebar" instead of "normal", so they were being skipped
This update treats "sidebar" the same as "normal" so all plugins show up as expected

Fixes #3715.

### DIFF
--- a/src/components/Flex.tsx
+++ b/src/components/Flex.tsx
@@ -23,9 +23,10 @@ export interface FlexProps extends HTMLAttributes<HTMLDivElement> {
     gap?: CSSProperties["gap"];
     justifyContent?: CSSProperties["justifyContent"];
     alignItems?: CSSProperties["alignItems"];
+    flexWrap?: CSSProperties["flexWrap"];
 }
 
-export function Flex({ flexDirection, gap = "1em", justifyContent, alignItems, children, style, ...restProps }: FlexProps) {
+export function Flex({ flexDirection, gap = "1em", justifyContent, alignItems, flexWrap, children, style, ...restProps }: FlexProps) {
     style ??= {};
     Object.assign(style, {
         display: "flex",
@@ -33,6 +34,7 @@ export function Flex({ flexDirection, gap = "1em", justifyContent, alignItems, c
         gap,
         justifyContent,
         alignItems,
+        flexWrap
     });
 
     return (
--- a/src/components/settings/tabs/themes/ThemeCard.tsx
+++ b/src/components/settings/tabs/themes/ThemeCard.tsx
@@ -35,7 +35,7 @@ export function ThemeCard({ theme, enabled, onChange, onDelete }: ThemeCardProps
                 )
             }
             footer={
-                <Flex flexDirection="row" style={{ gap: "0.2em" }}>
+                <Flex flexDirection="row" gap="0.2em">
                     {!!theme.website && <Link href={theme.website}>Website</Link>}
                     {!!(theme.website && theme.invite) && " • "}
                     {!!theme.invite && (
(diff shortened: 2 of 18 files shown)


src/plugins/previewMessage/index.tsx
```diff
@@ -89,8 +89,8 @@ const PreviewIcon: IconComponent = ({ height = 20, width = 20, className }) => {
     );
 };
 
-const PreviewButton: ChatBarButtonFactory = ({ isMainChat, isEmpty, type: { attachments } }) => {
-    const channelId = SelectedChannelStore.getChannelId();
+const PreviewButton: ChatBarButtonFactory = ({ isMainChat, isEmpty, type: { attachments }, channel }) => {
+    const channelId = channel.id ?? SelectedChannelStore.getChannelId();
```
Nuckyz: Is SelectedChannelStore still needed or can channel.id be always used?

src/api/ChatButtons.tsx
```diff
@@ -104,7 +104,7 @@ function VencordChatBarButtons(props: ChatBarProps) {
                 .filter(([key]) => chatBarButtons[key]?.enabled !== false)
                 .map(([key, { render: Button }]) => (
                     <ErrorBoundary noop key={key} onError={e => logger.error(`Failed to render ${key}`, e.error)}>
-                        <Button {...props} isMainChat={props.type.analyticsName === "normal"} />
+                        <Button {...props} isMainChat={props.type.analyticsName === "normal" || props.type.analyticsName === "sidebar"} />
```
Nuckyz:
```suggestion
                        <Button {...props} isMainChat={["normal", "sidebar"].includes(props.type.analyticsName)} />
```

src/plugins/previewMessage/index.tsx
vvareo: SelectedChannelStore wasn’t reliable in for the sidebar since it sometimes pointed at the parent channel. Switching to channel.id fixed the inconsistent loading and the wrong target behaviour, so the fallback wasn’t needed, not sure why I left it there tbh

Vendicated:
idk about this, discord also doesn't show their buttons here 

if anything, only plugins that actually insert something in the chatbox should show here. Translate, SilentMessageToggle and SilentTyping shouldn't show here. InsertTimestamps could

Julienraptor01:
> idk about this, discord also doesn't show their buttons here
> 
> if anything, only plugins that actually insert something in the chatbox should show here. Translate, SilentMessageToggle and SilentTyping shouldn't show here. InsertTimestamps could

isn't silentmessagetoggle the same as inserting `@silent` in front of the message ?
also preview message would be incredibly useful
being able to see how your message will look formatted is useful everywhere

vvareo:
At the moment the only thing that appears there is emojis, even though you'd expect any chat related plugin to show up as well. Just thought it would be nice to also be able to use them in these places
image
image

Vendicated:
I still think only buttons that apply to this specific chat should be shown.

Shown:
- PreviewMessage
- InsertTimestamps

Not shown:
- SilentMessageToggle
- Translate
- SilentTyping

Vendicated:
did just that via https://github.com/Vendicated/Vencord/commit/9aa3f380333e9bf9f170126c8bd7de94400c4e61

PR #3812, opened 2025-12-02
Add a "MuteSync" plugin

roli2py:
Made a plugin that mutes a microphone when it's physically muted. A main idea is to show for other members that you can't talk without a need of always remember to mute yourself in the app.

Not fully tested, because firstly it was the plugin to satisfy my need, but then I decide to open a PR.


roli2py:
Want to notify, that I won't continue to improve the plugin, because of an absence of a wish to do this, so if someone want to improve, then go ahead.

PR #3817, opened 2025-12-04
Fix: Use cancelText intead of secondaryConfirmText, because it does not show.

gaxolotl:
Fix: Use cancelText intead of secondaryConfirmText, because it does not show.
New:
image
Old:
image


Vendicated:
Thanks but I don't really like this fix.

Also, it seems that Discord just fully removed secondaryConfirm stuff. SupportHelper is not the only plugin that uses that. I will probably just make our own custom alert modal

PR #3830, opened 2025-12-10
FakeNitro: fix animated webp emojis not animating

sophiethefox:
All emojis (regardless of extension) may be sent with a `.webp` extension, as outlined [here](https://discord.com/developers/docs/resources/emoji#emoji-object-emoji-formats). When sent as such, with the `animated=true` parameter, the fake nitro plugin will not animate the (originally `.gif`) emoji.
This PR fixes this behavior and sets the emoji to `animated` if the `animated` parameter is set to `true`.

Written due to a change in a mobile plugin (unrelated to Vencord) sending fake emojis with `.webp`. This fixes compatibility and broadens support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.13.7",
+    "version": "1.13.8",
     "description": "The cutest Discord client mod",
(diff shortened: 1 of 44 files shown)


Vendicated:
thanks!

PR #3831, opened 2025-12-10
SHC: Fix Find

thororen1234:
class names went brrrrrr


Vendicated:
oh i fixed myself already i didnt see

but thanks

PR #3834, opened 2025-12-10
feat: configurable sync direction for cloud settings

lewisakura:
This will let you change your local client's cloud sync directionality.

For instance, if you are using a laptop, you may only want to pull your Vencord settings from the cloud, but you don't want to push back. Your changes on the laptop would remain intact until another device pushes to the cloud, at which point it would sync from the cloud again with the new changes.

The four options are:
- Sync bidirectionally (the current method, and the default)
- Push settings to cloud
- Pull settings from cloud
- Disable automatic sync (so you will have to manually sync your settings)

The setting for this is stored in LocalStorage since this is a per-client option, and not something that should be synchronized even in settings backups.

### DIFF
--- a/src/components/settings/tabs/sync/CloudTab.tsx
+++ b/src/components/settings/tabs/sync/CloudTab.tsx
@@ -120,8 +120,6 @@ function CloudSetupSection() {
                     Reauthorise
                 </ButtonWithIcon>
             </Grid>
-
-            <Divider className={Margins.top16} />
         </section>
     );
 }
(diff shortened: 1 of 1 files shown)


src/components/settings/tabs/sync/CloudTab.tsx
```diff
@@ -144,8 +177,8 @@ function CloudTab() {
                         Reauthorise
                     </Button>
                     <Button
-                        size={Button.Sizes.MEDIUM}
-                        color={Button.Colors.RED}
+                        size={"medium"}
+                        color={"red"}
```
Vendicated: this should be variant="dangerPrimary"

PR #3853, opened 2025-12-17
feat (plugin): Add mute notify plugin

BothimTV:
Notifies you via a sound when speaking whilst being muted.


Vendicated:
Thank you! But this is too niche (and way too much code for something so simple)

PR #3854, opened 2025-12-20
ShowAllMessageButtons: add setting to show frequently used reactions

LoganDark:
image

This new setting copies the JSX element from the non-expanded path used to show the frequently used reactions and pastes it into the expanded path as well.

image

### DIFF
--- a/src/plugins/showAllMessageButtons/index.ts
+++ b/src/plugins/showAllMessageButtons/index.ts
@@ -16,22 +16,44 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import { canonicalizeMatch } from "@utils/patches";
+import definePlugin, { OptionType } from "@utils/types";
+
+export const settings = definePluginSettings({
+    showFrequentlyUsedReactions: {
+        description: "Show frequently used reactions",
+        type: OptionType.BOOLEAN,
+        default: true
+    }
+})
 
 export default definePlugin({
     name: "ShowAllMessageButtons",
     description: "Always show all message buttons no matter if you are holding the shift key or not.",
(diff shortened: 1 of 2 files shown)


src/plugins/showAllMessageButtons/index.ts
```diff
@@ -16,22 +16,44 @@
...
-            }
+            replacement: [
+                {
+                    // isExpanded: isShiftPressed && other conditions...
+                    match: /isExpanded:\i&&(.+?),/,
+                    replace: "isExpanded:$1,"
+                },
+                {
+                    predicate: () => settings.store.showFrequentlyUsedReactions,
+                    match: /function \i\(\i\){let{[^}]+}=function\(\i\).*?}(?=function \i\(\i\))/,
```
Vendicated: this match is very generic. please try making it more specific. Instead of matching generic functions signatures, try matching keywords, just like how the patch above matches `isExpanded`

src/plugins/showAllMessageButtons/index.ts
```diff
@@ -16,22 +16,44 @@
...
 import { Devs } from "@utils/constants";
-import definePlugin from "@utils/types";
+import { canonicalizeMatch } from "@utils/patches";
+import definePlugin, { OptionType } from "@utils/types";
+
+export const settings = definePluginSettings({
+    showFrequentlyUsedReactions: {
+        description: "Show frequently used reactions",
+        type: OptionType.BOOLEAN,
+        default: true
```
Vendicated, line 28: I don't think this setting is needed honestly. Could just remove it

If you really want to make this a setting, you should add `requiresRestart: true` as this will only apply after restarting

LoganDark: My bad, I thought `requiresRestart: true` was the default

I made it not only apply after restarting

src/plugins/showAllMessageButtons/index.ts
LoganDark: I tried to make it more specific

PR #3859, opened 2025-12-20
feat(plugins/oneko): added the hability to change the oneko script and/or gif

KeyZox71:
Added two settings to set the oneko script and gif


Vendicated:
This is not safe

PR #3864, opened 2025-12-24
plugin: copyReactionsAsJSON

Rames507:
I built this mainly for myself due to an admittedly pretty obscure use case. 

I figured this could be useful to others that might need to import reaction data into external scripts so I made a PR for it.

I apologize if I made any mistakes regarding the process of submitting PRs, I am not that familiar with the Vencord repo.


Vendicated:
This is way too niche to be added, sorry!

Rames507:
Yup, I figured it would probably be too niche. 
Still wanted to try ;)

Thank you for maintaining vencord!

PR #3865, opened 2025-12-24
add new plugin MoreQuickReactions

programminglaboratorys:
This plugin allows you to modify the max number of reactions for the frencency reaction menu

3 was never enough, 5 is the plugin default value, configurable through plugin settings 

preview


Vendicated:
thanks, cool plugin!

PR #3867, opened 2025-12-27
WebScreenShareFixes: improve video codec selection

LeviSnoot:
So I realized the native Discord client tends to screenshare in VP9 or VP8, leading to better image quality than Vencord. This commit extends the existing webScreenShareFixes plugin by re-ordering the priority of the codec selection in discord web clients. Without the patch, the codec priority is 1. H.264, 2. VP8, 3. VP9. This patch reverses that order so it picks whichever codec is more bitrate efficient first. I tested this with Vesktop and it seems to work great!

image


src/plugins/webScreenShareFixes.web/index.ts
```diff
@@ -27,6 +27,10 @@ export default definePlugin({
                 {
                     match: /;usedtx=".concat\((\i)\?"0":"1"\)/,
                     replace: '$&.concat($1?";stereo=1;sprop-stereo=1":"")'
+                },
+                {
+                    match: /:\[(\i\.ad\.H264),(\i\.ad\.VP8),(\i\.ad\.VP9)\]/,
```
Vendicated: `ad` is a minified identifier and `\i` should be used instead

Also the code you're patching here is
```js
(experimentalBrowserHevcEnabled ? [p.ad.H265, p.ad.H264, p.ad.VP8, p.ad.VP9] : [p.ad.H264, p.ad.VP8, p.ad.VP9])
```

This patch should either patch both arrays or we could just always choose the HEVC branch

Vendicated:
I pushed a change that improves the patch and uses the best possible codec:
AV1 > HEVC > VP9 > H264 > VP8

Personally I couldn't get AV1 or HEVC to work despite them working fine in OBS, it always uses VP9, but still good and maybe those codecs will work for some people

LeviSnoot:
> This patch should either patch both arrays or we could just always choose the HEVC branch

Yeah at first I patched both but I wasn't sure what conditions triggered one or the other so I chose to patch the condition that applied to my environment to avoid potentially extraneous code.  

Your changes look good to me, thanks for looking it over and fixing my mistakes!

Vendicated:
Thank you for your PR, great find!

For future reference, you can use breakpoints to your advantage

We can see that the value `t` controlling which array is used is passed as second argument
image

Then you can just
1. Place a breakpoint at the start of the function
2. Cause it to be called
3. Use the callstack to find the caller (the place calling this function)
4. Wait 5 weeks for Chrome Devtools to load
5. See what it's passing

Example video:

https://github.com/user-attachments/assets/44f45bbd-0ba6-4ba7-9af8-6366ef304d6d

LeviSnoot:
Ooh thanks for taking the time to explain! Super kind!

Also, just tested your version and I am getting AV1 when screensharing! No idea why it works for me and not for you though...
image

Vendicated:
okay it's magically giving me AV1 too now lmao
image

LeviSnoot:
@Vendicated A friend of mine noticed that he can't watch my streams when he's on his mobile device, but can do so when on his PC. I tried forcing VP9 and he reported it working on both devices then. Should we add a config for this plugin which will let the user toggle, and maybe re-order codecs to their own preference with sane defaults? This should definitely have crossed my mind before making the PR but for some reason it totally slipped.

Vendicated:
I can't reproduce. Can you try messing with the list until it works? Maybe we can't use AV1 afterall

imjustprism:
maybe related to browser if av1 not supported then sdp negotiation fails
```js
RTCRtpSender.getCapabilities?.("video")?.codecs?.some(c => c.mimeType.includes("AV1")) ?? false
```

edit: tested without av1 support and it just fell back to h265, no errors at all, so i have no idea

LeviSnoot:
> I can't reproduce. Can you try messing with the list until it works? Maybe we can't use AV1 afterall

It appears that Discord might switch to a supported video codec depending on who joins the stream _first_? When I've encountered the issue I was looking at the debug output and I was definitely streaming AV1 at the time. Now that I'm trying to reproduce by joining a screenshare session from my iPad, the stream automatically switches to H264. I can't fully test this because I only have so many devices available to me, and I'd rather not ask my friends for help debugging. 

You can see in the screen recording attached that the codec switches when a viewer joins/leaves the stream (discord's stream join/leave sound effects are a bit delayed for some reason, so for clarity it switches to H264 when the stream is joined, and back to AV1 when there are no viewers)

https://github.com/user-attachments/assets/3a9e51ae-5b10-4f0a-b823-8dd261c262c6

Vendicated:
so you can't reproduce the issue anymore?

LeviSnoot:
> so you can't reproduce the issue anymore?

Not by myself, I don't know what device was used to join the stream when the issue occurred, only that it was a phone. No idea whether it was iOS or Android. I only have iOS/iPadOS devices available as mobile devices, and it seems that they will force a compatible codec, at the very least if they are the _first client to join a stream_. Every computer I've tested has no trouble decoding AV1, even an old 2011 iMac was able to do it. 

To reproduce, I'd expect you would have to start a stream, then have one AV1-decode capable user join the stream first so AV1 persists, and then have someone else join the stream from a non-AV1-decode capable device.

imjustprism:
i dont think forcing av1 is the right approach here cuz discord already enables av1 automatically when conditions are met.

on desktop, gpu detection passes av1 encode flag is only set for nvidia/amd gpus on windows

if (tF?.startsWith("NVIDIA") || tF?.startsWith("AMD")) {
    e.setExperimentFlag(eL.V8.SIGNAL_AV1, true);
} else {
    e.setExperimentFlag(eL.V8.SIGNAL_AV1_DECODE, true);
}

on browser its a different, av1 support comes from the browsers actual webrtc capabilities via getCodecCapabilities which queries what the browser can actually encode/decode

either way codec capabilities are checked and av1 only added to available codecs if supported

if (e.has(a.V8.SIGNAL_AV1)) {
    c.push({name: "AV1", encode: true, decode: true});
}

server decides codec based on all participants, sends codec choice via _handleCodecs considering what everyone can handle

the issue with forcing av1 in sdp priority is the codec gets locked when first viewer joins. if that viewer supports av1 the stream uses av1. when second viewer joins without av1 decode support like phone, intel gpu, older device, they cant decode the stream

handleVideoEncoderFallback only triggers on encoder failures not decoder failures on viewer side
av1 never parsed from browser sdp for a reason ig

Vendicated:
I don't really think that's relevant. It's just a list of formats that Discord will try in order, they already handle support etc.

By default it doesn't include AV1 ever, but we added it. That is what I suspect to be the problem. However, since I can't reproduce, I can't test this

LeviSnoot:
Sounds to me like omitting AV1 from the plugin is the best way forward then. I did already test without it and that definitely stopped the issue from occurring back when I originally encountered it.  H265 and VP9 seems more than sufficient enough for a discord screencast imo, if people really want AV1 it could be toggled as an experimental flag in config?

vermeeren:
@Vendicated Could AV1 support be turned into an option within the plugin? It works well for me on CPU encoding (GPU accel crashes because of chromium bug reported in https://github.com/Vencord/Vesktop/issues/1207#issuecomment-3704775521 which is https://issues.chromium.org/issues/471780477, once that is fixed it would be great)

When sharing CPU-heavy games though AV1 software encoder is just too heavy for my relatively old CPU resulting in a lot of performance loss. I would like AV1 support to be possible as an option instead of the removal suggested above.

Thanks!

LeviSnoot:
> I would like AV1 support to be possible as an option instead of the removal suggested above.

That's exactly what I suggested. see:

> if people really want AV1 it could be toggled as an experimental flag in config?

LeviSnoot:
Here's a proposed solution, giving the user full control with sane defaults.

```ts
/*
 * Vencord, a Discord client mod
 * Copyright (c) 2024 Vendicated and contributors
 * SPDX-License-Identifier: GPL-3.0-or-later
 */

import { definePluginSettings } from "@api/Settings";

    av1: {
        type: OptionType.BOOLEAN,
        default: false,
        description: "Enable AV1 (experimental)"
    },
    h265: {
        type: OptionType.BOOLEAN,
        default: true,
        description: "Enable H.265"
    },
    vp9: {
        type: OptionType.BOOLEAN,
        default: true,
        description: "Enable VP9"
    },

    h264: {
        type: OptionType.BOOLEAN,
        default: true,
        description: "Enable H.264"
    },
    vp8: {
        type: OptionType.BOOLEAN,
        default: true,
        description: "Enable VP8"
    },
    av1Priority: {
        type: OptionType.SLIDER,
        description: "(lower = higher priority)",
        markers: [1, 2, 3, 4, 5],
        default: 1
    },
    h265Priority: {
        type: OptionType.SLIDER,
        description: "(lower = higher priority)",
        markers: [1, 2, 3, 4, 5],
        default: 2
    },
    vp9Priority: {
        type: OptionType.SLIDER,
        description: "(lower = higher priority)",
        markers: [1, 2, 3, 4, 5],
        default: 3
    },
    h264Priority: {
        type: OptionType.SLIDER,
        description: "(lower = higher priority)",
        markers: [1, 2, 3, 4, 5],
        default: 4
    },
    vp8Priority: {
        type: OptionType.SLIDER,
        description: "(lower = higher priority)",
        markers: [1, 2, 3, 4, 5],
        default: 5
    }
});

function buildCodecArray() {
    const codecMap = [
        { name: "AV1", enabled: settings.store.av1, priority: settings.store.av1Priority, key: "AV1" },
        { name: "H265", enabled: settings.store.h265, priority: settings.store.h265Priority, key: "H265" },
        { name: "VP9", enabled: settings.store.vp9, priority: settings.store.vp9Priority, key: "VP9" },
        { name: "H264", enabled: settings.store.h264, priority: settings.store.h264Priority, key: "H264" },
        { name: "VP8", enabled: settings.store.vp8, priority: settings.store.vp8Priority, key: "VP8" }
    ];

    return codecMap
        .filter(codec => codec.enabled)
        .sort((a, b) => a.priority - b.priority)
        .map(codec => codec.key);
}

export default definePlugin({
    name: "WebScreenShareFixesDev",
    authors: [Devs.Kaitlyn],
    description: "Removes 2500kbps bitrate cap on chromium and vesktop clients.",
    enabledByDefault: true,
    settings,
    getCodecs(codecEnum) {
        const codecMap = [
            { name: "AV1", enabled: settings.store.av1, priority: settings.store.av1Priority, key: "AV1" },
            { name: "H265", enabled: settings.store.h265, priority: settings.store.h265Priority, key: "H265" },
            { name: "VP9", enabled: settings.store.vp9, priority: settings.store.vp9Priority, key: "VP9" },
            { name: "H264", enabled: settings.store.h264, priority: settings.store.h264Priority, key: "H264" },
            { name: "VP8", enabled: settings.store.vp8, priority: settings.store.vp8Priority, key: "VP8" }
        ];

        const filtered = codecMap.filter(codec => codec.enabled);
        const sorted = filtered.sort((a, b) => a.priority - b.priority);
        const result = sorted.map(codec => {
            const value = codec.key === "AV1" ? "AV1" : codecEnum[codec.key];
            return value;
        });
        return result;
    },
    patches: [
        {
            find: "x-google-max-bitrate",
            replacement: [
                {
                    match: /"x-google-max-bitrate=".concat\(\i\)/,
                    replace: '"x-google-max-bitrate=".concat("80_000")'
                },
                {
                    match: ";level-asymmetry-allowed=1",
                    replace: ";b=AS:800000;level-asymmetry-allowed=1"
                },
                {
                    match: /;usedtx=".concat\((\i)\?"0":"1"\)/,
                    replace: '$&.concat($1?";stereo=1;sprop-stereo=1":"")'
                },
                {
                    match: /\i\?\[(\i\.\i)\.H265,\i\.\i\.H264,\i\.\i\.VP8,\i\.\i\.VP9\]/,
                    replace: "true?$self.getCodecs($1)"
                }
            ]
        }
    ]
});
```

This is working as expected in my own limited testing, but I'd like to get some feedback before creating a new PR.

pheiduck:
This Plugin update causes Screensharing loading issues.

LeviSnoot:
> This Plugin update causes Screensharing loading issues.

Thanks for the report, but as you can see above, we are very well aware of this and resolving it.

powerofthe69:
For this plugin, I know that it's not available in the official client (since it doesn't have the bitrate cap issue like web version) but I was wondering if it could be refactored in a way that the reversed codec preference could be applied in the official client sans the web-specific fixes?

AMD defaults to using H.264 on VAAPI right now, but it'd be nice to use the plugin to get better quality and bitrate efficiency with a better codec if possible

LeviSnoot:
> I was wondering if it could be refactored in a way that the reversed codec preference could be applied in the official client sans the web-specific fixes?

I'm not aware of any (safe) way to modify the official discord desktop client, or inject JS to the webpack of it. But maybe someone with more knowledge can provide a more concrete answer.

teodorjuravlea:
> For this plugin, I know that it's not available in the official client (since it doesn't have the bitrate cap issue like web version) but I was wondering if it could be refactored in a way that the reversed codec preference could be applied in the official client sans the web-specific fixes?
> 
> AMD defaults to using H.264 on VAAPI right now, but it'd be nice to use the plugin to get better quality and bitrate efficiency with a better codec if possible

I can't say I'm very experienced with how everything works, but from my testing it seems to me that the official client (at least on Linux) uses custom code for encoding/decoding, not directly the Electron/Chromium implementations, so just specifying the codec probably won't work.

I am using [Loukious' fork of Vencord](https://github.com/Loukious/Vencord) which actually has a plugin that lets you change more screenshare parameters, including codec, but I've tried both AV1 and VP9 on it and the encoder wasn't initializing on the official client, as opposed to the web version where these codecs work.

Another thing that makes me believe they're doing something different on the official client is the fact that the VAAPI H264 screenshare can be viewed by anyone without problems.
On the web version, streaming with VAAPI H264 leads to issues on the viewer's side if they are on the official client (infinite loading, color glitches etc.) - regression since Mesa 24.3.0.

I've also noticed AV1 streams from other people don't load if I'm on the web version, regardless of HW and SW accelerated decoding (Chromium on Linux supports both), while on the official client they do. Don't know if it's just on my setup or if it happens to everyone.

PR #3868, opened 2025-12-27
feat(plugin) IaCorrector — message correction via Mistral AI

HyperBeats:
Adds the IaCorrector plugin (src/plugins/IaCorrector) to correct spelling/grammar right before a message is sent using the Mistral API.

Features

Chat bar button to enable/disable auto-correct (click or right-click) with a clear on/off visual state.
Settings: Mistral API key, target language (auto or ISO code), optional success toast.
Error handling (missing/invalid key, network/API failures).
Privacy
When enabled, message content is sent to Mistral’s API for processing.

How to test

Enable IaCorrector and set a valid Mistral API key in plugin settings.
Toggle auto-correct from the chat bar button.
Send a message with mistakes → it should be corrected before sending.
Try an invalid API key → you should get an error toast.

### DIFF
--- a/src/plugins/IaCorrector/ApiKeyHelpButton.tsx
+++ b/src/plugins/IaCorrector/ApiKeyHelpButton.tsx
@@ -1,0 +1,30 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2025 Vendicated and contributors
+ *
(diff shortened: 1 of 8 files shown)


nin0-dev:
such plugins have been rejected many times

also you cannot have a plugin where the user has to enter their own API Key, see CONTRIBUTING.md

PR #3870, opened 2025-12-28
feat(plugin): ConfirmInviteToServer

aqxorus:
Shows a confirmation modal before inviting a user to a server, preventing accidental invites.

image


Vendicated:
too niche, just hide the context menu entry with css

PR #3871, opened 2025-12-29
feat(plugin): VolumeProfiles

lk3n-0:
A plugin that lets users save and switch between custom volume configurations for people in a voice chat. They can create multiple volume presets that have specific users assigned to them at specific volume levels which can then be applied using keybinds or through context menus.
image
image


Vendicated:
Thanks for your contribution, but I think this is too niche and also complicated

PR #3872, opened 2025-12-29
feat(notrack): make launch signature clean

wavedevgit:
[Launch signature](https://docs.discord.food/reference#launch-signature) is a way that discord can tell if you are using client mods.

It basically goes through each client mod, checking a list of properties tied to that mod on window["..."]. I changed it so all those checks just return undefined, keeping Vencord hidden.

since I couldn't find a working patch, I used a monkey patch

### DIFF
--- a/src/plugins/_core/noTrack.ts
+++ b/src/plugins/_core/noTrack.ts
@@ -140,5 +140,16 @@ export default definePlugin({
                 Reflect.deleteProperty(window, "DiscordSentry");
             }
         });
+        const originalReflectGet = Reflect.get;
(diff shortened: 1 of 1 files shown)


arHSM:
This PR is nugatory, it will only turn into a cat-and-mouse.

Most of the time Discord doesn't care if you're using a client mod, they will care if said client mod is doing something malicious. I can only think of this change coming off as negative and hostile towards Discord.

P.S. If you're so concerned and/or paranoid about it then you shouldn't use mods.

wavedevgit:
Okay

PR #3874, opened 2025-12-30
Fix vencord badges not rendering when a user doesn't have any discord badges

unsigned-32-bit-integer:


unsigned-32-bit-integer:
since the colors are currently slightly different between stable and canary, i've changed platform indicators to evaluate the styles rather than relying on hex codes

Vendicated:
Why remake patches instead of just reusing the previous ones?

unsigned-32-bit-integer:
> Why remake patches instead of just reusing the previous ones?

The old patches for adding badges do still work, but i dont see the reason for using 2 patches to accomplish something that can be done with one

src/plugins/_api/badges/index.tsx
```diff
@@ -108,6 +104,13 @@ export default definePlugin({
                     replace: "...$self.getBadgeMouseEventHandlers($1),$&"
                 }
             ]
+        },
+        {
+            find: "getLegacyUsername(){",
+            replacement: {
+                match: /getBadges\(\){[^}]{0,400}?return\[/,
```
Vendicated:
```suggestion
                match: /getBadges\(\)\{.{0,100}?return\[/,
```

PR #3880, opened 2026-01-02
feat: add inline to codeblock, add icon-only to button, fix game overlay being darkened by themes

imjustprism:


src/components/Button.tsx
```diff
@@ -24,16 +24,17 @@ export type ButtonSize = "min" | "xs" | "small" | "medium";
...
+    iconOnly?: boolean;
 };
 
 export type LinkButtonProps = ComponentPropsWithRef<"a"> & {
     size?: ButtonSize;
     variant?: ButtonVariant;
 };
 
-export function Button({ variant = "primary", size = "medium", children, className, ...restProps }: ButtonProps) {
+export function Button({ variant = "primary", size = "medium", iconOnly, children, className, ...restProps }: ButtonProps) {
```
Vendicated: this should just be a button variant, no? or size

Vendicated: probably `size="icon"`

imjustprism:
@Vendicated any better way to handle game overlay or is skipping it entirely fine?

Vendicated:
uhhh idk

Vendicated:
i dont use that shit

imjustprism:
same, i dont see a reason to apply themes to the overlay anyway, so i just disabled it entirely

nexpid:
> same, i dont see a reason to apply themes to the overlay anyway, so i just disabled it entirely

square
image

src/components/Button.css
```diff
@@ -62,6 +62,16 @@
     line-height: 1.25;
 }
 
+.vc-btn-icon {
+    min-width: auto;
+    min-height: 38px;
```
Vendicated: idk about this height, why such a specific number? is that from discord?

imjustprism: yeah, 38px is discord min-height for medium buttons

imjustprism: ill try something

imjustprism:
i think using it as a variant is better

PR #3883, opened 2026-01-03
Fix: Allow users to specify a custom path name within the vencloud instance

bWlrYQ:
While attempting to self-host a Vencloud instance, I realised that I could not specify a backend API in the following format: `https://foo.bar/vencloud/`

Despite the backend API URL being stored locally properly by Vencord, the call to Vencloud would be: `https://foo.bar/v1/oauth/settings`. 

/vencloud/ path getting stripped

Due to the current structure of the code, the path is always removed because the following calls begin with a leading slash: 

https://github.com/Vendicated/Vencord/blob/c123efd65923bb183850242affdb1d7b0d228e50/src/api/SettingsSync/cloudSetup.tsx#L17
```js
new URL("/v1/foo/bar", getCloudUrl())
// returns https://foo.bar/v1/foo/bar because of leading slash
```

https://github.com/Vendicated/Vencord/blob/c123efd65923bb183850242affdb1d7b0d228e50/src/api/SettingsSync/cloudSetup.tsx#L96
```js
export const getCloudUrl = () => new URL(Settings.cloud.url);
// returns https://foo.bar/vencloud/
```

In order to allow users to specify a custom path in the API backend URL, I've enforced the addition of a slash at the end of the URL submitted by the user if it was omitted and removed the slashes when calling the Vencloud API.

### DIFF
--- a/src/api/SettingsSync/cloudSetup.tsx
+++ b/src/api/SettingsSync/cloudSetup.tsx
@@ -93,7 +93,7 @@ export async function authorizeCloud() {
     if (!await checkCloudUrlCsp()) return;
 
     try {
-        const oauthConfiguration = await fetch(new URL("/v1/oauth/settings", getCloudUrl()));
+        const oauthConfiguration = await fetch(new URL("v1/oauth/settings", getCloudUrl()));
         var { clientId, redirectUri } = await oauthConfiguration.json();
     } catch {
         showNotification({
--- a/src/api/SettingsSync/cloudSync.ts
+++ b/src/api/SettingsSync/cloudSync.ts
@@ -28,7 +28,7 @@ export async function putCloudSettings(manual?: boolean) {
     if (!await checkCloudUrlCsp()) return;
 
     try {
-        const res = await fetch(new URL("/v1/settings", getCloudUrl()), {
+        const res = await fetch(new URL("v1/settings", getCloudUrl()), {
             method: "PUT",
             headers: {
                 Authorization: await getCloudAuth(),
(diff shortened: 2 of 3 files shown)


Vendicated:
I don't really understand the use case. Why are you trying to host this on a subpath instead of using a subdomain? Doing so is generally discouraged as it will mix cookies etc with other things on the same domain. Subdomains offer better isolation and more reasonable request logs on both ends

bWlrYQ:
> I don't really understand the use case. Why are you trying to host this on a subpath instead of using a subdomain? Doing so is generally discouraged as it will mix cookies etc with other things on the same domain. Subdomains offer better isolation and more reasonable request logs on both ends

I totally understand your point and whilst I can't really "disagree" with it, hosting tooling under a subpath isn't inherently wrong or unusual (it's actually quite common, or maybe I'm outdated...). Many already existing solutions ([HedgeDocs](https://docs.hedgedoc.org/configuration/#hedgedoc-location), [Gitea](https://docs.gitea.com/administration/reverse-proxies#use-a-sub-path)...) handle this use case. 

As you said, subdomains usually offer "better isolation", but that only depends on how the users configure stuff and handle their security (I assume people who decide to self-host Vencloud do have a slight bit of experience and are aware of what they're doing). It is also possible to scope cookies to the path, and add eventual protections, but is there a point for the 4 endpoints proposed by Vencloud? 

To me, subpaths are a legitimate way of hosting apps, supporting them improves flexibility without bothering users who don't need them. If you really feel like this is a waste of energy and time, just close the PR and I'll end up using a whole subdomain for Vencloud (which seems a bit too excessive to me for 4 API endpoints...)

PR #2601, opened 2024-06-19
MessageTags: Add templating & edit

Luna-devv:
- Adds a way to put one tag into another:
```bash
# set via /tags create:
haha main

sub: [insert-tag:sub]
sub2: [insert-tag:sub2]
```
```bash
# when being ran:
haha main

sub: hello from sub tag
sub2: even
deeper
sub cmds
```

- Adds a way to edit existing tags with `/tags edit`
- Improves command descriptions

PR #3885, opened 2026-01-04
Improve usage of MessageTags

abb3v:
## MessageTags Modal Patch
I was always turned away from the existing MessageTags plugin on Vencord purely because it could not reasonably provide a way to have longer, multi-lined, or formatted messages.

The MessageTags plugin expects tags to be created through a command:
image

the problem is you cannot have multiple lines without manually typing `\n`, which isnt feasible for hand typed messages. 

#### proposed fix
Have a modal for MessageTags instead which allows these tags to be created in an editor which supports native formatting, editing after their creation, importing and exporting, and the ability to take an existing message and create a tag from it.

#### Import / Export

While Vencord handles global "Backup & Restore," adding independent Import/Export logic to a single plugin might seem unusual. But, although anecdotal, i found this useful for scenarios where:
* You may run a support server and want to share a list of canned responses (tags) with other staff members.
* This allows you to share your tags without exposing your entire Vencord configuration or forcing others to overwrite their own settings.

### Screenshots for the deaf

|muha | muja  | ebds  | image  |
|---|---|---|---|

### DIFF
--- a/src/plugins/messageTags/index.tsx
+++ b/src/plugins/messageTags/index.tsx
@@ -19,6 +19,7 @@ import { findGroupChildrenByChildId, NavContextMenuPatchCallback } from "@api/Co
 import { definePluginSettings } from "@api/Settings";
 import { Button } from "@components/Button";
 import { Heading, HeadingSecondary } from "@components/Heading";
+import { DeleteIcon } from "@components/Icons";
 import { Paragraph } from "@components/Paragraph";
 import { Devs } from "@utils/constants";
 import { classNameFactory } from "@utils/css";


abb3v:
fixed deprecation issues

abb3v:
Forgot to mention this ties in with #2601 and specifically the ability to import and export JSON. Although I do believe I may have changed how existing tags are handled with this... May need review.

https://github.com/Vendicated/Vencord/pull/2601#issuecomment-2182102966

Vendicated:
Thank you for your efforts, but this has been superseded by https://github.com/Vendicated/Vencord/commit/143bb60051b02e50c8536eeb2fc111f573f57dd3

You actually implemented it quite similarly to me but I wasn't really aware of this PR until now, very sorry about this :C 

image

image

image

If you can think of any further improvement to my change, feel free to adjust your PR to that and let me know. However, I don't think import/export should be a plugin specific feature, if anything it should just be a Vencord feature for every plugin

PR #3886, opened 2026-01-04
feat (plugin): RemoveShortcuts

MartinPrograms:
This plugin can disable arbitrary shortcuts from discord!

Discord's shortcuts are handled through a few independent systems, I've tried to implement this plugin using "combokeys", but a lot of shortcuts (like `ctrl+,` or `ctrl+/`) are handled elsewhere and can not be reliably disabled through that mechanism.

Because of that the plugin captures the keydown event, to intercept the shortcuts before they ever reach discord's internal handlers, this avoids incomplete coverage and is less brittle. 

This does involve adding a scoped DOM event listener. The implementation does not mutate the DOM, does not patch React components, and removes the listener when the plugin is disabled.


src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,175 @@
...
+            default: true,
+            onChange(value) {
+                reloadRemoveCallShortcut(value);
+            },
+        },
+        userRemovedShortcuts: {
+            type: OptionType.STRING,
+            name: "Custom Removed Shortcuts",
+            description:
+                "A list of custom keyboard shortcuts to remove from Discord. Seperated by |, in the format `CTRL + Slash` or `ALT + KeyB`. Modifiers are optional. Example: `CTRL + SHIFT + KeyA | ALT + KeyC`",
```
imjustprism: read that "Seperated" again

the description feels waaaaay too long to me, also why you added ":3" 😭

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,175 @@
...
+        .split("|")
+        .map(l => l.trim())
+        .filter(l => l.length > 0);
+
+    combos = combos.filter(c => !c.userDefined);
+
+    for (const line of lines) {
+        const combo = parseCombo(line);
+        if (combo) {
+            (combo as KeyCombo).userDefined = true;
```
imjustprism: already checked above, cast does nothing

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,175 @@
...
+                    e.preventDefault();
+                    return;
+                }
+            }
+        };
+        document.addEventListener("keydown", keydownHandler, true);
+    },
+    stop() {
+        document.removeEventListener("keydown", keydownHandler!, true);
+        keydownHandler = null;
```
imjustprism: also clear the combos array here

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,175 @@
...
+function reloadRemoveCallShortcut(value: boolean) {
+    if (value) {
+        for (const combo of callShortcuts) {
+            if (!combos.find(c => matchesCombo(c, combo))) {
+                combos.push(combo);
+            }
+        }
+    } else {
+        for (const combo of callShortcuts) {
+            combos = combos.filter(c => !matchesCombo(c, combo));
```
imjustprism: filter runs every iteration

src/plugins/removeShortcuts/index.ts
MartinPrograms: Damn let me be silly man 😔

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,170 @@
...
+                if (combo.code !== "") return null;
+                combo.code = part;
+        }
+    }
+
+    return combo.code ? combo : null;
+}
+
+let keydownHandler: ((e: KeyboardEvent) => void) | null = null;
+let combos: KeyCombo[] = [];
```
imjustprism: move these inside plugin object

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,170 @@
...
+        a.ctrl === b.ctrl &&
+        a.shift === b.shift &&
+        a.alt === b.alt &&
+        a.meta === b.meta
+    );
+}
+
+function reloadRemoveCallShortcut(value: boolean) {
+    if (value) {
+        combos.push(...callShortcuts);
```
imjustprism: i think this adds duplicates every onchange

src/plugins/removeShortcuts/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    description:
+        "Remove keyboard shortcuts from discord. By default `CTRL + '` and `CTRL + SHIFT + '` to call is disabled. This plugin blocks input from reaching Discord's keyboard shortcut handler.",
+    authors: [Devs.Kim],
+    start() {
+        if (keydownHandler) return;
+        reloadCombos(this.settings.store.userRemovedShortcuts);
+        reloadRemoveCallShortcut(this.settings.store.removeCallShortcut);
+        keydownHandler = (e: KeyboardEvent) => {
+            for (const combo of combos) {
+                if (matchesCombo(combo, {
```
imjustprism: probably its cleaner to use IS_MAC ? e.metaKey : e.ctrlKey
you can import it from @utils/constants

maddyuwu123:
this  would  be  a  nice  feature  to  have!!!!

PR #3888, opened 2026-01-05
Fix error when updatePopoutWindows is called before PopoutWindowStore is initialized

unsigned-32-bit-integer:


src/api/Themes.ts
```diff
@@ -92,6 +92,7 @@ function applyToPopout(popoutWindow: Window | undefined) {
 }
 
 function updatePopoutWindows() {
+    if (PopoutWindowStore === undefined) return;
```
thororen1234, line 95: `if (!PopoutWindowStore) return;`

imjustprism: https://github.com/Vendicated/Vencord/pull/3880 i have a fix here already

Vendicated:
thanks but fixed via https://github.com/Vendicated/Vencord/pull/3880!

PR #3889, opened 2026-01-06
AutoAffix Plugin

Falcon249:
Adds customizable prefix and/or suffix to messages.

### Features:
- Prefix/Suffix addition with space control
- Excludes slash commands
- Custom pattern filtering (skip messages containing specific words)
- Optional empty message handling
- Lightweight with minimal performance impact


Vendicated:
TextReplace can do this

PR #3890, opened 2026-01-06
feat(plugin): PopOut Plus

funteaqueue:
Pop out streams and cameras independently from each other.


funteaqueue:
Closing PR since discord released native version

PR #3895, opened 2026-01-09
fix(RCE): Don't recolor messages that failed to send

sillyElvy:
Recoloring messages that failed to send causes them to become indistinguishable from sent messages without hovering over them. They will appear sent, as if you're partaking in the current conversation, not showing you there was an issue, which makes this unwanted behaviour.
This PR resolves this and lets them appear red, as they do by default.
Earlier return is not possible due to react rules.

### DIFF
--- a/src/plugins/roleColorEverywhere/index.tsx
+++ b/src/plugins/roleColorEverywhere/index.tsx
@@ -105,8 +105,8 @@ export default definePlugin({
             find: 'tutorialId:"whos-online',
             replacement: [
                 {
-                    match: /null,\i," — ",\i\]/,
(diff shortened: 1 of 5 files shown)


src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -192,6 +192,9 @@ export default definePlugin({
             const { messageSaturation } = settings.use(["messageSaturation"]);
             const author = useMessageAuthor(message);
 
+            // Do not apply role color if the send fails, otherwise it becomes indistinguishable if the message is sent
+            if (message.state === "SEND_FAILED") return;
```
Vendicated:
```suggestion
            if (message.state !== "SENT") return;
```

PR #3898, opened 2026-01-10
vcNarrator: stop narrating other sessions

unsigned-32-bit-integer:
For example prevents "joined" message when vencord is started while you are in call in another session


Vendicated:
Thanks! But using SessionsStore for this is suboptimal, which also makes all the other changes redundant, so I did this myself via https://github.com/Vendicated/Vencord/commit/d86423cae3a52b4daef70038d8a507c1c4dd9e37

PR #3899, opened 2026-01-10
feat(plugin): add CopyMentions plugin

naystie:
-


nin0-dev:
aislop and useless (you can just copy id and <@id>) or even better, @username

PR #3906, opened 2026-01-12
fix(PluginModal): update author fetching to include all authors

EhDaYaGhaly:


Vendicated:
why? plugins should really not have more than 6 authors, and we don't want to fetch too many users

PR #3908, opened 2026-01-13
Dearrow plugin change

evaz1995:
Hello,

I noticed that the Dearrow plugin does not properly work when the API for Dearrow throws a 404 error, even when the embedded link is indeed a Youtube video. Hence I am proposing this change: if no title nor thumbnail is provided by the main API, the thumbnail API is used instead, which should provide a response for all YouTube videos unless the API is down.

### DIFF
--- a/src/plugins/dearrow/index.tsx
+++ b/src/plugins/dearrow/index.tsx
@@ -54,14 +54,11 @@ async function embedDidMount(this: Component<Props>) {
         if (!videoId) return;
 
         const res = await fetch(`https://sponsor.ajay.app/api/branding?videoID=${videoId}`);
-        if (!res.ok) return;
 
         const { titles, thumbnails } = await res.json();
 
         const hasTitle = titles[0]?.votes >= 0;
         const hasThumb = thumbnails[0]?.votes >= 0 && !thumbnails[0].original;
-
-        if (!hasTitle && !hasThumb) return;
 
 
         embed.dearrow = {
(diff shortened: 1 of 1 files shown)


Vendicated:
Hi!
The way the plugin currently works is that it will only replace the thumbnail if the replacement has at least one upvote. I think this was done to only use manually created replacements.
I am not opposed to also using automatic replacements but I think many people probably don't want these. We should make it configurable with a new setting, defaulting to false

Besides that, it'd be much better if you refactored your changes so that it won't use a failed request's response as if it succeeded, that's asking for trouble! Something like the following would be much more robust
```js
const res = await fetch(normalApi);
if (res.ok) {
	// existing code ...
} else if (settings.store.useAutomaticThumbnails) {
	const res = await fetch(thumbnailApi);
	// ...
}
```

Besides that I do like this idea, so thank you!

src/plugins/dearrow/index.tsx
```diff
@@ -171,9 +174,11 @@ const settings = definePluginSettings({
 export default definePlugin({
     name: "Dearrow",
     description: "Makes YouTube embed titles and thumbnails less sensationalist, powered by Dearrow",
-    authors: [Devs.Ven],
+    authors: [Devs.Ven, {
+        name: "ichi0995",
```
thororen1234, line 178: You should add yourself to the devs export like how devs.ven is

PR #3909, opened 2026-01-13
feat: add plugin to munge opus codec WebRTC SDP

nerdwave-nick:
This plugin enables stereo audio when watching streams

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "vencord",
     "private": "true",
-    "version": "1.13.11",
+    "version": "1.13.12",
     "description": "The cutest Discord client mod",
     "homepage": "https://github.com/Vendicated/Vencord#readme",
     "bugs": {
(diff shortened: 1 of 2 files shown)


nerdwave-nick:
Been looking into the patch that's already being applied in the webScreenShareFixes.web plugin. The only difference I can tell between mine and that one is that the webScreenShareFixes.web plugin doesn't apply the transformation to offers. I've spent an hour or so now looking if I can find a reason. The offers aren't generated with the `usedtx` part in the SDP. 

But I can't find any code for their generation. This is I believe handled in the native code based on the registered devices? Maybe, maybe not. I've been trying to look into it. The stuff is created with a `createOffer` call, but breakpoints show that the patched code does not run before that for an offer.

So current theory - this stuff is being inferred from streams which are registered? Or devices? Or whatever. I do not know the details enough. What I do know is that my plugin works so I'll use it for now and just be happy.

PR #3916, opened 2026-01-15
Fix AlwaysExpandRoles, PermissionsViewer, ShowConnections

thororen1234:
🧇 

(not on stable yet)

### DIFF
--- a/src/plugins/showConnections/index.tsx
+++ b/src/plugins/showConnections/index.tsx
@@ -160,10 +160,10 @@ export default definePlugin({
 
     patches: [
         {
-            find: ".hasAvatarForGuild(null==",
+            find: "UserProfilePopoutBody",
             replacement: {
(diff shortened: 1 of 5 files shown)


thororen1234:
we've hit stable  🔥

src/plugins/showConnections/index.tsx
```diff
@@ -160,10 +160,10 @@ export default definePlugin({
 
     patches: [
         {
-            find: ".hasAvatarForGuild(null==",
+            find: "UserProfilePopoutBody",
             replacement: {
-                match: /currentUser:\i,guild:\i[^}]*?\}\)(?=])(?<=user:(\i),bio:null==(\i)\?.+?)/,
-                replace: "$&,$self.profilePopoutComponent({ user: $1, displayProfile: $2 })"
+                match: /userId:\i\.id,guild:\i\}\)(?=])/,
```
thororen1234, line 165: Was wondering if this was okay to do arguments or should I just capture what we need?

PR #3935, opened 2026-01-24
Switch: migrate to new style

imjustprism:
image

with icon:

image


Vendicated:
these look so bad

imjustprism:
i said the same thing, but i got a bit used to it over time

PR #3936, opened 2026-01-25
Plugin disable start call keybind

LovesScotch:
Simple plugin to disable the global keybind for starting a voice call when in a DM or Group DM.

I often find myself accidentally hitting CTRL+' instead of SHIFT+ENTER (I can't explain it either) when typing up longer messages that need line breaks.


Vendicated:
hmm thanks but this seems too niche

you can post it in our third party plugins channel if you want

PR #3941, opened 2026-01-26
feat(plugins): add InvisibleMessage plugin

Hikkywannafly:
A fun plugin to share hidden notes with friends. It automatically hides any text wrapped in > < syntax using invisible Unicode characters.

How to use: Hello >  I love you  :3 <

Others see: "Hello"
Plugin users see: "Hello Encrypted message: I love you  :3 "
Messages remain fully invisible on mobile and for anyone not using the plugin.

d1

d2

d3


nin0-dev:
this used to exist as InvisibleChat but it got removed because it was too niche and gave a false sense of security

PR #3944, opened 2026-01-27
HideAttachments: Support Discord Components V2

MrCRACKDev:
First time contributing here, not sure if adding myself to Devs for something this small is ok or if that should be removed.


Nuckyz:
Hiding accessories seems outside of the plugin purposal

Vendicated:
yeah this contribution is too minor to add yourself as author

besides that, looks good, thanks!

PR #3946, opened 2026-01-28
SelectToQuote: Add plugin

Mikofoxie:
Adds `SelectToQuote`, a plugin to quote specific parts of a message.

**Features:**
- Adds "Reply with Quote" to the context menu when text is selected.
- **Smart Formatting:** Detects if selection is inside a codeblock (`hljs`/`language-`) and preserves the language + auto-fencing (N+1 backticks).
- **Diff Mode:** specific Shift+Click behavior to create a `diff` block for corrections.

**Technical:**
- Uses `FluxDispatcher` (`CREATE_PENDING_REPLY`) for native reply state (no DOM text injection hacks).
- Zero external dependencies.
- Strict DOM checks to prevent menu pollution (only shows when text is selected inside a message).

### DIFF
--- a/src/plugins/selectToQuote/README.md
+++ b/src/plugins/selectToQuote/README.md
@@ -1,0 +1,8 @@
+# SelectToQuote
+
+Adds a "Reply with Quote" button to the context menu when selecting text in a message.
+
+## Features
+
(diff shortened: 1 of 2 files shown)


Vendicated:
aislop

PR #3948, opened 2026-01-29
Fix persistent dummy author.

EtorixDev:
Fixes the dummy author from remaining as the first author of every plugin forever.

### DIFF
--- a/src/components/settings/tabs/plugins/PluginModal.tsx
+++ b/src/components/settings/tabs/plugins/PluginModal.tsx
@@ -78,22 +78,18 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
 
     useEffect(() => {
         (async () => {
-            const loadedAuthors: Partial<User>[] = [];
-
-            for (const user of plugin.authors.slice(0, 6)) {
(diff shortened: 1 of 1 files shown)


src/components/settings/tabs/plugins/PluginModal.tsx
```diff
@@ -73,22 +73,27 @@ export default function PluginModal({ plugin, onRestartNeeded, onClose, transiti
...
                         : makeDummyUser({ username: user.name });
 
-                    setAuthors(a => [...a, author]);
+                    loadedAuthors.push(author);
                 } catch (e) {
                     continue;
                 }
             }
+
+            setAuthors(loadedAuthors.length ? loadedAuthors : [dummyAuthor]);
```
Nuckyz: This will only set the authors once all of the promises resolved. It should set each author as soon as their individual promise resolves

EtorixDev: Changed it to update during the loop.

Vendicated:
https://github.com/Vendicated/Vencord/commit/dc27c4da76e37b55b242d65bbb2b822654e54c4d

PR #3949, opened 2026-01-29
feat: Add Discord Hidden Ink plugin

dex593:
- Invisible zero-width encoding for message content with auto-decode for users who have this tool.
- Mention preservation for user, role, and channel pings.
- Optional "shared key" encryption for message payloads and supported attachments. Messages you send will be encoded and should decode for users with the same key
- Attachment encryption/decryption for images and `.txt` files.
- "Show more..." expands long decrypted text for easier reading.


Vendicated:
We had a plugin like this before but removed it because it's fairly niche, gives a wrong impression of privacy and annoying for anyone without the plugin

PR #3950, opened 2026-01-29
Role color everywhere compatibility patch

EnergoStalin:
Old way not compatible with this

```js
{
  find: "#{intl::GUEST_NAME_SUFFIX})]",
  replacement: {
    match: /#{intl::GUEST_NAME_SUFFIX}[^"]+""(?<=user:(\i).+?)/,
    replace: "$&,$self.showInjection($1.id)"
  }
}
```

### DIFF
--- a/src/plugins/roleColorEverywhere/index.tsx
+++ b/src/plugins/roleColorEverywhere/index.tsx
@@ -128,7 +128,7 @@ export default definePlugin({
             find: "#{intl::GUEST_NAME_SUFFIX})]",
             replacement: [
                 {
-                    match: /#{intl::GUEST_NAME_SUFFIX}.{0,50}?""\](?<=guildId:(\i),.+?user:(\i).+?)/,
+                    match: /#{intl::GUEST_NAME_SUFFIX}[^"]+""[^}]+(?<=guildId:(\i),.+?user:(\i).+?)/,
                     replace: "$&,style:$self.getColorStyle($2.id,$1),"
                 }
             ],


src/plugins/roleColorEverywhere/index.tsx
```diff
@@ -127,7 +127,7 @@ export default definePlugin({
             find: "#{intl::GUEST_NAME_SUFFIX})]",
             replacement: [
                 {
-                    match: /#{intl::GUEST_NAME_SUFFIX}.{0,50}?""\](?<=guildId:(\i),.+?user:(\i).+?)/,
+                    match: /#{intl::GUEST_NAME_SUFFIX}[^"]+""[^}]+(?<=guildId:(\i),.+?user:(\i).+?)/,
```
Nuckyz, line 131:
```suggestion
                    match: /#{intl::GUEST_NAME_SUFFIX}[^"]+?""\](?<=guildId:(\i),.+?user:(\i).+?)/,
```

Would this work?

EnergoStalin: Just tested. No.

Should be like this after patching.
```js
{
  className: a()(c, T.Xh, q, {
    [T.Pi]: !U && !G && R
  }),
  children: [
    null != z ? z : j.Ay.getName(Z),
    F ? (0, r.jsxs)("span", {
      className: T.IW,
      children: ["\xa0", E.intl.string(E.t["pFO/Ph"])]
    }) : "",
    Vencord.Plugins.plugins["AllCallTimers"].showInjection(Z.id)
  ],
  style: Vencord.Plugins.plugins["RoleColorEverywhere"].getColorStyle(Z.id, W),
}
```

It's not RoleColorEverywhere preventing the patch it can't patch after AllCallTimers patch. Cause current patch wont skip over `Vencord.Plugins.plugins["AllCallTimers"].showInjection(Z.id)` line. Actually could be ordering issue but i dont see order option anywhere so just fall back to something familiar like fixing the patch.

Vendicated:
What's the purpose of this change? I don't understand

EnergoStalin:
Current patch cant match when something patches end of the children array since it expect `]` right after `""` which is not true in snippet above.

PR #3951, opened 2026-01-29
feat(plugins): Add plugin for automatically declining friend requests of new discord accounts

MaxBosse:
This plugin allows to set a minimum account age for friend requests and automatically declines them, if a friend request does not fulfill the requirement.

I tend to get random friend requests from very new accounts regularly. Currently Discord only allows to disable friend requests based on friendships and common servers or allow everyone to send you a friend requests. Because I sometimes need to get a friend request from someone I do not share a friend/server with, I have to keep my setting on "Allow everyone".

This plugin helps reduce noise by filtering out new (spam) accounts. There is a log in the plugin settings in case you are waiting for a friend request of someone who is actually new to Discord, so you can add them yourselves.

Screenshot_20260129_203847

### DIFF
--- a/src/plugins/autoDeclineFriendRequests/index.ts
+++ b/src/plugins/autoDeclineFriendRequests/index.ts
@@ -1,0 +1,188 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2024 Vendicated and contributors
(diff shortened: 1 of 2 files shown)


Vendicated:
We don't want to add any sort of automation / selfbot behaviour

PR #3961, opened 2026-02-01
feat(plugins): Add NoLinkEmbed Plugin

vqpe:
Automatically wraps links in <> to prevent them from embedding.


Vendicated:
textreplace can do this

PR #3975, opened 2026-02-06
Improved messageLatency and added viewAs Plugin

fox3000foxy:
WIP: Message Latency Graph (may be removed later)
Added ViewAs plugin, a plugin that uses dispatcher to use Discord "View As" feature.
- Uses Role context menu to add "View As" button
- Add a User context menu to dispatch as the user


Vendicated:
View as role is already a feature of the PermissionViewer plugin. It's just gated behind a hidden setting because it may have a risk for your account standing, since it has the potential to trigger many invalid requests

I don't really understand the Graph change for MessageLatency, I don't think it's needed

PR #3976, opened 2026-02-07
Update messageUpdater to handle reference changes

x86disasm:
Force re-render when message reference changes


Vendicated:
you just changed the comment...?

PR #3983, opened 2026-02-09
trigger message re-renders

x86disasm:
when used by other plugins to trigger message re-renders,

I am  new to this to lmk if i am doing something wrong or absolutely nothing


Nuckyz:
An API for this already exists

x86disasm:
> An API for this already exists

:OO

PR #3990, opened 2026-02-13
plugin ViewRaw: add message context menu option

BoatFloater:
I'd like to be able to view raw messages, but I don't need it enough to justify the popover button.

This PR adds a setting (default: `false`) to provide "View Raw" as a context menu option on the message. 
I did not make this setting also disable the popover as the popover can be disabled in the global "**Manage UI elements**" menu.


src/plugins/viewRaw/index.tsx
```diff
@@ -185,7 +193,12 @@ export default definePlugin({
         "thread-context": MakeContextCallback("Channel"),
         "gdm-context": MakeContextCallback("Channel"),
         "user-context": MakeContextCallback("User"),
-        "dev-context": devContextCallback
+        "dev-context": devContextCallback,
+        "message": (children, { message }) => {
+            if (settings.store.messageContextMenu) {
+                MakeContextCallback("Message")(children, { message });
```
Vendicated: don't recreate it every time

just store in a variable or move the boolean check into the callback itself

BoatFloater: Sure, I moved the check into the callback

BoatFloater:
@Vendicated is this able to be merged?

PR #3991, opened 2026-02-15
[iLoveSpam] feat: add optional notification when a user is a likely spammer

C1200:
This adds an optional notification when a user is a likely spammer but will still keep the message unhidden as is default. Also, I used the same patch as messageLatency so I made a slight change to that to avoid conflict.


Vendicated:
I don't really get the point of this change. If you want a notice you'd just not enable this plugin

C1200:
The reason I wanted this change is to see that someone is marked as a spammer but not have to click unhide every time I want to view a block of messages from a potential spammer.

PR #3992, opened 2026-02-15
appleMusic: move back to `music.apple.com`

VillainsRule:
the fix previously introduced by nin0 was understandable, but came with multiple issues, primarily incorrect album covers. i observed these nearly globally, even if songs WERE on itunes, which made very little sense to me. for example, jackboys 2 by travis scott would show up as a sza album. i have utterly no idea why.

the fix that seems to be a good end-all is to just match the entire JWT auth token and avoid using too strong a regex; this exact solution has worked for 9 months: http://github.com/villainsrule/applemusic

### DIFF
--- a/src/plugins/appleMusic.desktop/native.ts
+++ b/src/plugins/appleMusic.desktop/native.ts
@@ -12,6 +12,40 @@ import type { TrackData } from ".";
 
 const exec = promisify(execFile);
 
+let authorization: string = '';
+
+(async () => {
+    const mainReq = await fetch('https://music.apple.com/us/new', { headers: { 'user-agent': VENCORD_USER_AGENT } });
+    const mainRes = await mainReq.text();
+    const scriptURL = mainRes.match(/<script type="module" crossorigin src="(.*?)"/)?.[1];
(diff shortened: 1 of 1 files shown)


nin0-dev:
itunes seems to find the sza song with travis scott on it, that is weird

you should definitely set a macos user agent if you don't wanna get blocked at some point, also setup a proper dev environment using vscode and all the recommended extensions (https://docs.vencord.dev) so that your code is formatted

nin0-dev:
tbh we should just get rid of this and use last fm for album covers, it will work better than using apple's api

VillainsRule:
> tbh we should just get rid of this and use last fm for album covers, it will work better than using apple's api

didn't know last.fm had an API, i can go look at that later

VillainsRule:
> tbh we should just get rid of this and use last fm for album covers, it will work better than using apple's api

unless i'm wrong, last fm's api requires a key and they don't have a normal search endpoint(?)

nin0-dev:
are you in the vencord server? if not can you join and open a modmail for dev channel access, this will be easier to discuss than over github

VillainsRule:
closing this at least for now given the possibility of a last.fm API

PR #3997, opened 2026-02-18
feat(plugins): add FixupX plugin

haumlab:
Adds a new `FixupX` plugin (`src/plugins/fixupX`) that
  rewrites `x.com` links to `fixupx.com` on message
  send/edit for more reliable embeds.


lolsuffocate:
This can be done with the TextReplace plugin already

evanwhitt:
ah i see, so it is realistically unneeded haha oof lol, oh well

PR #3999, opened 2026-02-19
fix(plugins): fix multiple broken plugins

sillyElvy:
currently only on canary

### DIFF
--- a/src/plugins/showHiddenChannels/index.tsx
+++ b/src/plugins/showHiddenChannels/index.tsx
@@ -438,11 +438,11 @@ export default definePlugin({
             },
         },
         {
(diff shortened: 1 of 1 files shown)


src/plugins/typingTweaks/index.tsx
```diff
@@ -102,7 +102,7 @@ export default definePlugin({
 
     patches: [
         {
-            find: "#{intl::THREE_USERS_TYPING}",
+            find: "handleDismissInviteEducation=",
```
imjustprism:
```suggestion
            find: "#{intl::SEVERAL_USERS_TYPING_STRONG}",
```

sillyElv:
it hit stable

PR #4002, opened 2026-02-21
fix(translate): add Korean to DeepL language list

seomin0610:


Vendicated:
Thanks, seems like this was a regression in https://github.com/Vendicated/Vencord/commit/0126090c8f06849cde9df18030f52e31931ece49

PR #4010, opened 2026-02-25
GifCaptioner

AlphaX50:
Adds a built in caption editor for GIFs, images and videos directly inside Discord.
You can quickly add meme style captions or speech bubbles with a live preview, then the media is rendered to a ready-to-send GIF and uploaded automatically to the current channel. Supports multiple entry points (GIF picker button, media context menu and attachment action) so you can caption almost anything without leaving the chat.


Vendicated:
This can be done with a bot like esmbot

PR #4011, opened 2026-02-25
CtrlEnterSend: add option to disable call keybind

nullsoepic:
Problem:

Misclicking Ctrl + ' and starting a call accidentally when trying to send a message
It might be bit niche but also might just be the most annoying "feature" I've ever seen

Solution:


Vendicated:
doesn't really fit in this plugin and it's too niche / specific

PR #4015, opened 2026-02-27
Translate: add LibreTranslate support + native requests (CSP-safe)

MrPatates:
## Summary

This PR adds LibreTranslate support to the Translate plugin and makes LibreTranslate requests CSP-safe by sending them through native plugin helpers (IPC) instead of renderer `fetch`.

## Changes

- Added `LibreTranslate` as a translation service option.
- Added LibreTranslate settings:
  - `libreTranslateUrl` (instance base URL)
  - `libreTranslateApiKey` (optional)
- Added LibreTranslate language loading (`/languages`) with caching per base URL.
- Added Google <-> Libre language code normalization/mapping.
- Added native helper requests in `src/plugins/translate/native.ts`:
  - `makeLibreTranslateLanguagesRequest`
  - `makeLibreTranslateRequest`
  - URL protocol validation (`http`/`https`)
- Updated Translate modal language selectors to refresh when Libre settings change.

## Why

Renderer-side requests to local/self-hosted LibreTranslate instances can be blocked by CSP (example: `Refused to connect because it violates the document's Content Security Policy.`).  
Routing LibreTranslate requests through native helpers fixes this while keeping the existing UX.

## Validation

- `pnpm build` succeeds.
- Verified in app:
  - `Translate -> service -> LibreTranslate` is available.
  - Local instance URL works for both language fetch and translation.
  - Existing Google/DeepL behavior remains unchanged.

### DIFF
--- a/src/plugins/translate/TranslateModal.tsx
+++ b/src/plugins/translate/TranslateModal.tsx
@@ -20,25 +20,34 @@ import { Divider } from "@components/Divider";
 import { FormSwitch } from "@components/FormSwitch";
 import { Margins } from "@utils/margins";
 import { ModalCloseButton, ModalContent, ModalHeader, ModalProps, ModalRoot } from "@utils/modal";
-import { Forms, SearchableSelect, useMemo } from "@webpack/common";
+import { Forms, SearchableSelect, useEffect, useMemo, useState } from "@webpack/common";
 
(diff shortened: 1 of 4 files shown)


Vendicated:
aislop & unsafe

PR #4016, opened 2026-02-27
Translate: add LibreTranslate support (HTTPS-only, CSP-aware)

MrPatates:
- Adds LibreTranslate to the Translate plugin service list.
- Adds libreTranslateUrl and libreTranslateApiKey settings (API key is optional).
- Enforces https:// instance URLs (invalid URLs and http:// are rejected with a clear error).
- Implements LibreTranslate language loading (/languages) and translation requests (/translate).
- Uses the existing CSP permission flow (requestAddOverride / isDomainAllowed) for connect-src.


Vendicated:
?

PR #4019, opened 2026-03-01
FakeNitro: Enable toggling Favorites server

PrinceBunBun981:
With Discord making changes to the Favorites experiment and planning to release it to Nitro Standard, this patch makes the function `isPremiumExactly` always return true, enabling both the Show Favorites Server setting in User Settings > Chat and the server appearing in the server list.

Currently, you'd need the experiment enabled, but it should hopefully work if they roll things out as they are, since it's just modifying what that function returns, not the actual Favorites logic. However, I wouldn't be surprised if they change how the client handles making sure people without Standard won't be able to access it, lol

Do tell if I shouldn't PR patches for experiments that haven't rolled out yet though


electricsteve:
I agree this probably will get fixed by changing favorites logic, but I think for now this should work and is acceptable

PR #4021, opened 2026-03-01
fix(ShowMeYourName): add optional chaining to nick

Clarence212:
just added ?. on nick so it wont crash if it’s missing and wrapped the condition in parentheses to make it easier to read.


Vendicated:
Is this actually a real issue? I have never seen any complaints or experienced an issue with this

If it really is an issue, then the other code would also sometimes render `undefined` as name, which again has never happened to me nor been complained about

Clarence212:
mb, it was an isolated issue

PR #4032, opened 2026-03-06
fix(betterFolders): fix broken guildsnav regex

haruki-py:
fixed BetterFolders failing to inject vars into guildsnav
it was causes because discord updated `guildsnav` which added className and themeOverride 
in this commit, i updated the regex to use `[^}]*` and it fixes the issue.
(I reopened this because my last PR was a disaster and had messed up)


haruki-py:
already fixed

PR #4034, opened 2026-03-07
fix(PlatformIndicators): support VR platform icon and tooltip label

### LINKED ISSUE #4022: [Bug] PlatformIndicators plugin does not support VR clients
## The Issue
As of recently Discord now reports the platform of Android clients on Meta and potentially other VR headsets as `vr` instead of `mobile`. PlatformIndicators should use a VR headset icon and proper tooltip capitalization for this platform. 

- The plugin falls back to using the same icon as for desktop, which is incorrect. 
- The plugin makes the assumption that only the first letter of the platform identifier is capitalized, so the tooltip appears as `Vr` instead of `VR`. 

Image

## Additional Information

Discord already has a dedicated status icon for this platform, similar to mobile, which could probably be reused for the PlatformIndicators icon. Unfortunately, I could not figure out where the `svg-mask-status-online-vr` svg is sourced from to patch Vencord and make a pull request myself. 

Image

## Why?

This recent change in reported platform is due to Discord working on [a dedicated application for Meta  VR headsets](https://www.meta.com/blog/discord-account-connect-meta-horizon-profile-quest-vr-games-share-status/) to be released later this year. 

### Request Agreement

ghost:
Fixes #4022.

### DIFF
--- a/packages/discord-types/src/stores/PresenceStore.d.ts
+++ b/packages/discord-types/src/stores/PresenceStore.d.ts
@@ -6,7 +6,7 @@ export interface UserAndActivity {
     activity: Activity;
 }
 
-export type DiscordPlatform = "desktop" | "mobile" | "web" | "embedded";
+export type DiscordPlatform = "desktop" | "mobile" | "web" | "embedded" | "vr";
 
 export interface PresenceStoreState {
     presencesForGuilds: Record<string, Record<string, { status: OnlineStatus; activities: Activity[]; clientStatus: Partial<Record<DiscordPlatform, OnlineStatus>>; }>>;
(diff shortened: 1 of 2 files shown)


ghost:
sorry about the messy pr

nin0-dev:
why would node modules be missing in your testing? can’t you just run `pnpm i` and be able to build your pr?

also how does the icon look like in place?

Vendicated:
Every morning begins the same way: coffee, inbox, and—without fail—another **AISLOP GitHub pull request** waiting for review. At first it was mildly amusing, then mildly annoying, and now it has evolved into a daily ritual of frustration. I open the notification with a tiny, irrational spark of hope—maybe this one will be different. Maybe this one will actually make sense. Maybe this one will solve a real problem. But it never does. It never even comes close.

Instead, what greets me is the same pattern: poorly thought-out changes, unnecessary refactors, questionable naming, and code that somehow manages to be both overcomplicated and underconsidered at the same time. It’s the kind of code that makes you stare at the diff for a few seconds wondering whether you’re missing some brilliant hidden intention—before realizing, no, it really is just bad.

The worst part is not the existence of bad pull requests. Everyone writes bad code sometimes. That’s part of development. The real exhaustion comes from the **relentless frequency**. Daily AISLOP pull requests arrive like clockwork, as if somewhere there is a machine whose sole purpose is to generate mediocre contributions and deliver them straight to my notification panel. There is no pause, no reflection, no sign that anyone stopped for a moment to ask the fundamental question: “Should this actually be a pull request?”

Instead, the review queue grows, and with it grows the quiet dread of opening GitHub. Reviewing code should be an opportunity for collaboration—an exchange of ideas, improvements, and shared ownership of a project. But when the majority of contributions feel rushed, careless, or unnecessary, the process turns into something else entirely: a tedious cycle of pointing out obvious issues, requesting changes that should never have been needed, and wondering why any of this exists in the first place.

Sometimes I imagine an alternate reality in which people pause before submitting these pull requests. They run tests. They read the surrounding code. They consider whether their change improves the project in any meaningful way. In that reality, a pull request would be something exciting—something worth opening. Instead of bracing myself for another round of avoidable problems, I could look forward to seeing thoughtful work from collaborators who actually care about the quality of what they’re submitting.

But that is not the reality I live in. The reality is daily AISLOP pull requests, each one arriving with the same quiet promise of disappointment. And at this point, I find myself wishing for something very simple: fewer pull requests, better pull requests, or ideally both.

Until then, the routine continues. Notification. Open diff. Sigh. Review. Repeat.

PR #4037, opened 2026-03-07
Created the fix webcam shortcut plugin

M4rga:
When using the webcam shortcut in Discord, module 87203 calls the camera toggle function with a 
hardcoded third argument true that always forces the preview modal to open — completely ignoring 
the "Don't show this again" setting. The plugin overrides that function by removing the forced flag, 
so the shortcut respects the user's preference exactly like the manual button click does.
This is most likely a bug in Discord, this plugin serves as a workaround until Discord hopefully fixes it, i've already opened an issue!.


Vendicated:
too niche and doesn't meet our plugin guidelines

PR #4038, opened 2026-03-08
EmbedXivModArchive v1.0.0 Initial Release

mashirochan:
A simple plugin to embed XIV Mod Archive links from the popular MMORPG, FFXIV.

image

It adds a MessageAccessory if the specific regex for a site's link is detected, upon which it will fetch the mod's information JSON from xivmodarchive.com, using a 1-second delayed queue and a session cache to prevent excessive calls if dozens of XMA links are present in the renderer at the same time.

Features markings for content that XMA marks NSFW and automatically spoilers the images, requiring the user to click to reveal them.

image

### DIFF
--- a/src/plugins/embedXivModArchive/index.tsx
+++ b/src/plugins/embedXivModArchive/index.tsx
@@ -7,20 +7,20 @@
 import { addMessageAccessory, removeMessageAccessory } from "@api/MessageAccessories";
 import { LinkButton } from "@components/Button";
 import { Devs } from "@utils/constants.js";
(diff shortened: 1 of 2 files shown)


Vendicated:
Hi! This should really just be a bot instead. Either one that people add to their server and it looks for messages and does the exact same with normal embeds/components, or manually via a slash command

This plugin is too niche to be added to Vencord and doesn't do anything a regular bot can't do.

Sorry! Thank you anyway ❤️

mashirochan:
That's fair, thank you for the feedback!

PR #4039, opened 2026-03-08
New Plugin: Deterministic User Panning

regulad:
Hi everyone. This is my first contribution to this project.

This is a simple plugin which pans the audio of every user to a different place between left and right, creating a simple soundstage  where you can pick out voices by position in addition to timbre.

Unfortunately, while the logic for setting the `setLocalPanning` is sound, it would appear that Discord removed the support for panning behind the scenes as nothing happens even when running `setLocalPanning` and then confirming it with `getLocalPanning`. The panning functionality is documented as an RPC method (https://docs.discord.com/developers/topics/rpc#setuservoicesettings).

This PR is left as a draft as I am attempting to patch panning support back into `MediaEngineNative` (It lists it as supported, but it doesn't actually work).

### DIFF
--- a/src/plugins/deterministicUserPanning/index.ts
+++ b/src/plugins/deterministicUserPanning/index.ts
@@ -1,0 +1,104 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2026 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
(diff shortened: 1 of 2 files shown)


regulad:
If anybody has experience tinkering with the `MediaEngineNative` engine, please get in touch! I'm at a standstill.

Vendicated:
I think Discord might be working on this themselves?

https://x.com/DiscordPreviews/status/2069184753547235732?s=20

image

PR #4043, opened 2026-03-09
Added 'ThemeSwitcher' Plugin

SupremeMuhit:
This plugin lets which between themes instantly with a keyboard shortcut or a button


Vendicated:
Too niche, just use Toolbox or Settings context menu

image

image

PR #4044, opened 2026-03-09
NowPlaying Plugin

benzokones:
Hi, this plugin allows the user to do /now to display their currently playing track similar to nicotine+. There is planned support for other services such as last.fm.

image
image

I am aware that the contributing guidelines specify against simple slash commands but creating this as a discord bot would not support foobar's nowplaying feature and can be disabled by servers. By having it as a plugin more services can be supported 🙂.

Let me know if there are any problems or issues.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -629,6 +629,10 @@ export const Devs = /* #__PURE__*/ Object.freeze({
         name: "prism",
         id: 390884143749136386n,
     },
+    benzokones: {
+        name: "benzokones",
+        id: 1319511318109093949n,
+    },
 } satisfies Record<string, Dev>);
 
 // iife so #__PURE__ works correctly


src/plugins/nowPlaying/native.ts
```diff
@@ -0,0 +1,25 @@
...
+    return path.join(...paths);
+}
+
+export async function stat(event: any, filePath: string): Promise<{ size: number; }> {
+    const stats = await fs.stat(filePath);
+    return { size: stats.size };
+}
+
+export async function readFile(event: any, filePath: string): Promise<string> {
+    return await fs.readFile(filePath, "utf8");
```
Vendicated, line 24: this is really unsafe never do this

Vendicated:
Thanks! But this plugin is too niche and your implementation is unsafe. This plugin already exists for Spotify. You could add a /share command to the lastfm-rpc plugin though, it already has everything you need

PR #4049, opened 2026-03-13
feat(plugin): add listing existing commands in CustomCommands

Benjas333:
Add listing tags in the `/tags delete` command:
imagen


Vendicated:
users should use the settings ui for deleting tags

PR #4050, opened 2026-03-14
feat: Add Pronouns Autofetch for userMessagesPronouns

patttterson:
Adds option for pronouns auto-fetching, as before the plugin would only work with user's profiles that you yourself clicked on and opened. 

For the plugin [userMessagesPronouns](https://github.com/Vendicated/Vencord/tree/main/src/plugins/userMessagesPronouns)


Vendicated:
api abuse

PR #4051, opened 2026-03-14
Add ChannelTracker plugin

MrPaXe:
This plugin adds icons next to the channel name based on whether you've sent messages or viewed the channel.
It saves the state using the DataStore.


Vendicated:
way too niche and obvious ai readme which makes it likely that the code is also ai

PR #4052, opened 2026-03-15
feat(types): update & add types

thororen1234:


packages/discord-types/enums/application.ts
```diff
@@ -0,0 +1,43 @@
...
+    EMBEDDED = 131072,
+    GATEWAY_MESSAGE_CONTENT = 262144,
+    GATEWAY_MESSAGE_CONTENT_LIMITED = 524288,
+    EMBEDDED_FIRST_PARTY = 1048576,
+    APPLICATION_COMMAND_BADGE = 8388608,
+    SOCIAL_LAYER_INTEGRATION = 134217728,
+    PROMOTED = 536870912,
+    PARTNER = 1073741824,
+    PARENT = 8589934592,
+    DISABLE_RELATIONSHIPS_ACCESS = 17179869184,
```
Vendicated: are u sure these dont use bigint?

these are meant for bitwise operations which operate on 32 bit integers, which is way too small for the bigger numbers here

Vendicated: yeah its bigint
Image

packages/discord-types/enums/user.ts
```diff
@@ -20,3 +26,36 @@ export const enum ReadStateType {
...
+    VERIFIED_BOT = 65536,
+    VERIFIED_DEVELOPER = 131072,
+    CERTIFIED_MODERATOR = 262144,
+    BOT_HTTP_INTERACTIONS = 524288,
+    SPAMMER = 1048576,
+    DISABLE_PREMIUM = 2097152,
+    PROVISIONAL_ACCOUNT = 8388608,
+    QUARANTINED = 17592186044416,
+    COLLABORATOR = 1125899906842624,
+    RESTRICTED_COLLABORATOR = 2251799813685248,
```
Vendicated: this one is weird. these numbers are too large so they have to be bigint yet discord uses normal int for user flags

I don't think the client will ever use these higher ones

packages/discord-types/src/modules/CloudUpload.d.ts
```diff
@@ -41,7 +40,6 @@ export class CloudUpload extends EventEmitter {
     filename: string;
     id: string;
     isImage: boolean;
-    isRemix: boolean | undefined;
```
Vendicated, line 44: RIPPP

Vendicated:
Looks good for the most part but I'm not sure about adding the quests type

Vencord isn't going to use these and they are likely subject to regular change

PR #4056, opened 2026-03-16
feat(plugin): add searchRole plugin

MunchyG:
###   **SearchRole Plugin**

  Adds a Search Messages from Role context menu option when right-clicking a role badge in Discord (requires Developer
  Mode to be enabled in Discord settings.)

###   **What it does**

  Lets you search through all messages sent by members of a specific role in the current server, without needing to
  manually look up each member individually.

###   **Features**

  - Role context menu — right-click any role badge to trigger a search
image

  - Filtered results modal — view matching messages with author avatar, username, channel, and timestamp
image

  - Channel filter — narrow results to specific channels
image

  - Content filter — search by message text (press Enter or click Search)
  - Has filter — filter by attachment type (Image, Video, File, Embed, Link)
  - Pagination — load more results on demand, with a running count shown
image

  - Jump to message — click Jump on any result to navigate directly to it in the channel
  - Graceful member fetching — if role members aren't cached yet, triggers a member fetch and prompts you to try again
 
###   **Notes**

  - Results are capped at the first 50 members with the role to stay within Discord's search API limits. A notice is
  shown when results are truncated.
image

  - Uses Discord's native guild message search API — no external requests.
  - Claude was used to write this and is likely not ideal. 
  - This is meant primarily as a proof of concept 
 
###   **My Use Case**
  - Keeping track of messages from Arrowhead Studios developers, as they often message in the general chat that gets buried in seconds. With this plugin, I can easily see all messages that any AH dev has made without having to manually search each dev's user ID

### DIFF
--- a/src/plugins/searchRole/index.tsx
+++ b/src/plugins/searchRole/index.tsx
@@ -1,0 +1,307 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2024 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { getCurrentGuild } from "@utils/discord";
+import { ModalContent, ModalHeader, ModalProps, ModalRoot, ModalSize, openModal } from "@utils/modal";
+import definePlugin from "@utils/types";
+import { Avatar, Button, ChannelStore, FluxDispatcher, GuildChannelStore, GuildMemberStore, GuildRoleStore, IconUtils, Menu, NavigationRouter, Parser, RestAPI, ScrollerThin, Select, showToast, Text, TextInput, Timestamp, Toasts, UserStore, useEffect, useMemo, useState } from "@webpack/common";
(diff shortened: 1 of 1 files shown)


0xmortuex:
A few issues:

1. CI is failing — missing license header and unsorted imports. Run `pnpm lint --fix` to resolve both.
2. Double check that no personal user/role IDs are left in the default templates.
3. This was submitted from a renamed repo, not a proper Vencord fork — that could cause problems.
4. The Co-Authored-By: Claude line in the commit might raise flags with maintainers.

Nice idea though, templates from the chat bar is handy.

MunchyG:
How was this not a proper Vencord fork? What did I rename incorrectly?
And as stated in the notes, I was not expecting my code to be the final product. I was primarily just making a proof of concept, and since there is no plugin request option, I figured this was the best course of action to get such a feature added to vencord.

PR #4059, opened 2026-03-18
fix(plugins): multiple broken plugins

sillyElvy:
this is probably suboptimal but i couldnt find a place to patch it anywhere. Apparently there's no method thats called when the guild request is approved (or i couldn't find it) but this is the nicest solution i could come up with.

### DIFF
--- a/src/plugins/newGuildSettings/index.tsx
+++ b/src/plugins/newGuildSettings/index.tsx
@@ -144,7 +144,7 @@ export default definePlugin({
     applyDefaultSettings,
     flux: {
         GUILD_JOIN_REQUEST_UPDATE({ guildId, request, status }) {
-            if (status === "APPROVED" && UserStore.getCurrentUser().id === request.user_id)
+            if (status === "APPROVED" && request.user_id === UserStore.getCurrentUser().id)
                 applyDefaultSettings(guildId);
         }
     }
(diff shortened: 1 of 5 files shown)


src/plugins/newGuildSettings/index.tsx
```diff
@@ -141,5 +141,10 @@ export default definePlugin({
         }
     ],
     settings,
-    applyDefaultSettings
+    applyDefaultSettings,
+    flux: {
+        GUILD_JOIN_REQUEST_UPDATE({ guildId, status }) {
```
dolfies: check user id as this will fire for other peoples' acceptances for mods

sillyElv: ty i missed that, fixed

sillyElvy:
all of these are stable compat

src/plugins/newGuildSettings/index.tsx
```diff
@@ -141,5 +141,11 @@ export default definePlugin({
         }
     ],
     settings,
-    applyDefaultSettings
+    applyDefaultSettings,
+    flux: {
+        GUILD_JOIN_REQUEST_UPDATE({ guildId, request, status }) {
+            if (status === "APPROVED" && UserStore.getCurrentUser().id === request.user_id)
```
Vendicated: Yoda

```suggestion
            if (status === "APPROVED" && request.user_id === UserStore.getCurrentUser().id)
```

Vendicated:
thanks!

PR #4062, opened 2026-03-18
ReviewDB: Profile Component Improvements

thororen1234:
Restrict limit of 4 users only in profile component
before:
image
after:
image

Newest to oldest review sort & filter out system in profile component
image


thororen1234:
added support for disabled reviews and users with no reviews
image
image

PR #4065, opened 2026-03-20
feat(last-active): display when users were last active

Antrubtor:
Adds a plugin that displays how long ago a user went offline based on presence updates, similar to an old Discord experiment.

image


nin0-dev:
was rejected: https://github.com/Vencord/plugin-requests/issues/300

PR #4067, opened 2026-03-21
fix(PauseInvitesForever): add missing null check in disableInvites

watchthelight:
\`showDisableInvites\` already guards against a null guild from \`GuildStore.getGuild\`, but \`disableInvites\` does not, which can crash when spreading \`guild.features\`.

Added the same \`if (!guild) return;\` guard that the sibling function already uses.


Vendicated:
This shouldn't be necessary, as this function shouldn't be reached if guild is null. Even if it somehow ends up null, it'll just be a silent error

PR #4068, opened 2026-03-21
MessagePopover: use proper React components for hook support

watchthelight:
## Summary
- Render functions in `VencordPopoverButtons` were called as plain
  functions inside `.map()`, which violates React's Rules of Hooks
- Extracted a `PluginPopoverButton` wrapper component so the render
  call happens inside a proper React component body
- Follows the same pattern already used by `ChatButtons.tsx`
- Error handling moved from try/catch to ErrorBoundary (consistent
  with every other similar API)

All 4 existing consumers (viewRaw, translate, quickMention,
hideAttachments) work without changes.

## Tests run
- [x] viewRaw renders and responds to left/right click
- [x] Translate renders and opens modal
- [x] QuickMention renders and works
- [x] No console errors

### DIFF
--- a/src/api/MessagePopover.tsx
+++ b/src/api/MessagePopover.tsx
@@ -64,6 +64,17 @@ export function removeMessagePopoverButton(identifier: string) {
     MessagePopoverButtonMap.delete(identifier);
 }
 
+function PluginPopoverButton({ render, message, Component }: {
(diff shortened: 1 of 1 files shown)


Vendicated:
Please don't send ai generated pull requests

PR #4070, opened 2026-03-22
fix(showMeYourName): Exclude emojis from gradients

sillyElv:
issue is, as shown below, that emojis become part of the gradient making them a colored blob without keeping the emoji recognizable
with this fix the plugin uses the same behaviour discord does for emojis in nicknames

before
after


sillyElvy:
regex is super pretty now

src/plugins/showMeYourName/index.tsx
```diff
@@ -104,8 +107,9 @@ export default definePlugin({
             if (mode === "user-nick")
                 return <>{prefix}{username} <span className="vc-smyn-suffix">{nick}</span></>;
```
Vendicated, line 108: I think all of them need wrapping. `username` can also correspond to user.globalName

PR #4071, opened 2026-03-22
Create plugin "microslop"

Revilo0509:
**TL:DR This plugin NEEDS direct DOM manipulation for its primary function.** In the case that this is **COMPLETELY** unacceptable *just close this PR* without a second thought.

I'm aware of the rule "No raw DOM manipulation. Use proper patches and React". But the primary function of this plugin is to replace *ALL* occurrences of "microsoft" to "microslop". This means the plugin has to be able to walk the DOM tree and replace the occurrences.

I originally got the idea from this [chome web extension](https://chromewebstore.google.com/detail/microsoft-to-microslop/hlkljlkdinjnbfmclionhbefbnefcgll) and wanted a discord/vencord equivalent.

I've tried this as thoroughly as I could. It works on server names, usernames, nicknames and messages. But doesn't break links and is only visual.

I'm mainly making this PR so everybody has the chance to experience the beauty of seeing *microslop* instead of the horrifying "microsoft" (Eww)

### DIFF
--- a/src/plugins/microslop/index.ts
+++ b/src/plugins/microslop/index.ts
@@ -1,0 +1,61 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2026 Revilo0509
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+export default definePlugin({
+    name: "microslop",
+    description: "Replaces any and all occurrences of \"microsoft\" to \"microslop\" (case-insensitive)",
+    authors: [Devs.revilo0509],
+
+    start() {
+        this.scan();
+
+        const observer = new MutationObserver(() => {
+            this.scan();
+        });
+
+        observer.observe(document.body, {
+            childList: true,
+            subtree: true
+        });
+
+        this.observer = observer;
+    },
+
+    stop() {
+        this.observer?.disconnect();
+    },
+
+    scan() {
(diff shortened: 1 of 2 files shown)


SupItsZaire:
replacing all visible microsoft strings does not require mutating text nodes in the DOM. 

in a react app, the correct implementation is the element/string creation path afaik: 
transform string children and selected string props before react renders them

i get the intent and it's really fucking funny ngl but this should be implemented as a render-time transformation, not a document-wide text rewrite

thororen1234:
Text replace
First box: Microsoft
Second box: Microslop
And that's it

Revilo0509:
> Text replace First box: Microslop Second box: Microslop And that's it

I found and tried that plugin, but it did not work for other ppls messages, nicknames or server names.

Revilo0509:
> replacing all visible microslop strings does not require mutating text nodes in the DOM.
> 
> in a react app, the correct implementation is the element/string creation path afaik: transform string children and selected string props before react renders them
> 
> i get the intent and it's really fucking funny ngl but this should be implemented as a render-time transformation, not a document-wide text rewrite

I did try to hook/patch the React.createElement function but I did not get it to replace all of the occurrences of "microsoft". I tried a bit more after this PR got closed and I'd read your reply. But I still can't get it to replace all of them.

I found that just walking the DOM tree is the working, easiest and most reliable implementation.

Vendicated:
both ways to implement this are bad and this plugin is very pointless so it won't be accepted either way

PR #4074, opened 2026-03-23
Add Chrome extension ID and fix mapping function in customSections

THEBOSS9345:
feat: add Chrome extension ID for React Developer Tools installation
fix: update mapping function in customSections to use underscore for unused parameter


Vendicated:
don't send spam prs

PR #4076, opened 2026-03-24
feat: add MentionHighlighter plugin

00re3se:
# MentionHighlighter

Customizes the appearance of messages where you are mentioned.

You can set a custom highlight color, enable a glow effect, and toggle the default Discord mention background in the plugin settings.


Vendicated:
this can be done with css

PR #3802, opened 2025-11-26
Migrate Off Forms

thororen1234:
Migrate to vencord components for forms text and button

PR #4078, opened 2026-03-26
refactor: remove deprecated components and update relevant files to new standard

justjxke:
Cleaned up a bunch of the old components being used in settings pages and plugins, swapped them out for the new `@components/` standard

the only plugin I didn't test was `SortFriendRequests`, I do not have any friend requests 💔

I quadruple checked this before I PR'd but i feel like somehow something slipped through the cracks

oh and also i noticed a deprecation notice with `src/plugins/_core/settings.tsx` but I decided not to touch it, i didnt see anywhere else where `customSections` was being used. if you want me to take a look and move it to the new `customEntries` let me know

also if u didnt want this to be done at all, or in this specific way. apologies in advance

### DIFF
--- a/src/api/Notifications/notificationLog.tsx
+++ b/src/api/Notifications/notificationLog.tsx
@@ -18,12 +18,15 @@
 
 import * as DataStore from "@api/DataStore";
 import { Settings } from "@api/Settings";
+import { BaseText } from "@components/BaseText";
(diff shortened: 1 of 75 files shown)


thororen1234:
https://github.com/Vendicated/Vencord/pull/3802

PR #4081, opened 2026-03-28
feat(plugin): LastFM: added variables to custom status

Kisakay:
Add support for `{artist}`, `{album}`, and `{title}` variables in the LastFM rich presence custom status field, allowing users to dynamically display track information in their status text.


src/plugins/lastfmRichPresence/index.tsx
```diff
@@ -308,7 +308,10 @@ export default definePlugin({
                 case NameFormat.AlbumName:
                     return trackData.album || settings.store.statusName;
```
Vendicated: here statusName is potentially used without variables

that also brings up the case of what should be done if a variable is empty?

PR #4084, opened 2026-03-31
feat(showHiddenChannels): show live streamers in hidden channels

qrewy:
now showing live streamers in hidden channels


sillyElvy:
wtf does bro need 1.7k lines for you can do this in well under 100 probably and if not its not rlly worth it

PR #4085, opened 2026-03-31
PinDMs: Allow directly moving channels from one category to another

AntoKek:
added the option to move channels from one category to another via DM ContextMenu, without having to waste time unpinning them first. just a small QoL feature


moxie-coder:
this I would really like honestly, I hate having to unpin first before moving them

PR #4086, opened 2026-04-02
feat(plugin): DisableOverlaySoundboard

Andrew-DLO:
Simply disables the entire soundboard while using the in-game overlay, usually accessed by ``Ctrl + ` `` or `Ctrl + /`. For some reason the soundboard in the in-game overlay can not be disabled and the default keybind can not be disabled in the default Discord client.🙃

I made this plugin because I use ``Ctrl + ` `` as my toggle mute keybind and was tired of Discord playing super loud sounds every time I try to mute while in a game. After searching online I saw tens of Reddit posts and people in the Vencord discord complaining about the same thing.
soundboard6

### DIFF
--- a/src/plugins/disableOverlaySoundboard/index.ts
+++ b/src/plugins/disableOverlaySoundboard/index.ts
@@ -1,0 +1,25 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2026 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


Vendicated:
Pretty sure you can disable keybinds by adding the same keybind in the Keybinds page and not assigning any action to it

Either way this is way too niche

Andrew-DLO:
Would be better as a user tweak. I definitely was thinking it's too niche. Thank you!

PR #4094, opened 2026-04-04
feat(vcNarrator): add screen share narration

zoinbase:
added screen share event narration
to disable just remove everything from the text box


Vendicated:
why did you rewrite the entire plugin for no reason?

PR #4095, opened 2026-04-05
feature/local-reminders

youtsuhodev:
# feat: LocalReminders plugin (local-only reminders)

## Summary

Adds the **LocalReminders** plugin: reminders are **local only** (no server). Data is persisted with Vencord’s DataStore, rescheduled when the plugin starts, and surfaced through the existing Notifications API.

## Features

- **`/remind in`** — `minutes` (1–28,800) + `message` (max 500 characters); confirmation includes a Discord timestamp.
- **`/remind list`** — lists pending reminders, sorted by due time.
- **`/remind cancel`** — cancel by full UUID or a **unique prefix** (e.g. first 8 characters).
- **Discord restart** — on `start()`, reminders that were due while the client was closed are notified once (*missed while offline*) then removed; future ones are rescheduled with `setTimeout`.
- **Setting** — `maxReminders` (default: 50).

## Technical details

| Item | Detail |
|------|--------|
| Storage | Key `LocalReminders_v1` (IndexedDB / `@api/DataStore`) |
| Notifications | `showNotification` with `noPersist: true` |
| Plugin stop | `stop()` clears timers without wiping persisted data |

**28,800 minutes** (~20 days) is the upper bound so `setTimeout` stays within safe engine limits.

## Manual testing

All manual tests below have been completed successfully.

- [x] Create a short reminder (1–2 min) → notification fires at the expected time.
- [x] Restart Discord before the due time → reminder still fires after relaunch.
- [x] `/remind list` and `/remind cancel` (full id, then unique prefix).

## Commands (reference)

```text
/remind in minutes:<number> message:<text>
/remind list
/remind cancel id:<uuid or prefix>
```

### DIFF
--- a/src/plugins/localReminders/index.ts
+++ b/src/plugins/localReminders/index.ts
@@ -1,0 +1,281 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2022 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


Vendicated:
aislop

PR #4096, opened 2026-04-06
feat(plugin): add SelectiveNR per-user noise gate

ARM9000:
## What does this plugin do?
Adds per-user noise gating to voice channels. Right-click any user in a VC → **Suppress** to apply a noise gate to their audio stream. Right-click again → **Unsuppress** to restore clean audio.

## How it works
Each Discord user's audio is a separate `MediaStream`. This plugin intercepts those streams via a `MutationObserver` and routes suppressed users through a custom `AudioWorkletProcessor` (noise gate) running on the dedicated audio thread. Non-suppressed users get a clean passthrough with no processing.

## Settings
- **Threshold** — gate open threshold in dBFS
- **Attack** — how fast the gate opens when someone speaks
- **Release** — how fast the gate closes after they stop
- **Hold** — keeps gate open briefly after signal drops (prevents word clipping)
- **Reduction** — gain applied when gate is closed

### DIFF
--- a/src/plugins/selectiveNR/README.md
+++ b/src/plugins/selectiveNR/README.md
@@ -1,0 +1,30 @@
+# SelectiveNR
+
+Per-user noise gate for Discord voice channels.
+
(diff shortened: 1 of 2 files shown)


Vendicated:
nice aislop did u prompt it urself

PR #4102, opened 2026-04-07
feat(plugin): MessageCorrector - Corrects the order of messages in the chat, based on when the message reached Discord's servers.

Calebh101:
Small QOL plugin I made :)
Excerpt from my readme (cuz I don't feel like typing all this again):

> This plugin is very simple; it reorders messages based on their timestamps. This might *seem* useless, but it's really not. (in my humble opinion)
> 
> Basically, I've found that the client's message store just kinda receives a message, whether from itself, or from the server, and just accepts it. But this creates some inconsistencies; if you and someone else send a message at around the same time, at least one of the clients will show a different order than the server received it.
> 
> This plugin basically uses the timestamps of each message, and sorts them how they should be. Each message's timestamp represents when the server received it, *not* the client.

I talked with people on the discord about it, and some suggested that it might be good as a built-in or default plugin? I just left it as a normal plugin, but it can also ofc be made into a required plugin, or hidden plugin, or whatever is preferred. (It has 1 non-hidden setting, and it can be toggled via a slash command.)

Also, if you're concerned about that there's a slash command, it can be removed... it's mainly there for debugging and verifying my plugin works, but it also works as just adding timestamps to the message imo. So I decided to keep it. Here's a preview of those timestamps:

Sorry if this is a duplicate, but there are a ton of PRs and I'm not sure what to even search for. Thanks for seeing this!
(Btw the actual source and commit history to this is in [another repo](https://github.com/Calebh101/VC-MessageCorrector))

Apparently there's a template now too


Vendicated:
Sorry but I think this is too niche of a plugin to be added. I don't think many people care enough about this to warrant it being a plugin

Anyway seems like Discord might be working on fixing this too? There's this experiment

image

PR #4108, opened 2026-04-12
muting the "You are currently muted!" tooltip

Twigee:
If you are muted in a vc for a certain amount of time, or on joining a vc it will pop a tool tip saying "You are currently muted". IKnowImMuted replaces that method checking if you are muted while talking 
pk4ujr63yzwd1

also this is my first pull request to vencord, pls alert me to anything i may have done wrong ;3


constants.ts
```diff
@@ -0,0 +1,649 @@
+/*
```
claration: why was the entire `constants.ts` file copied to the root of the project

Twigee: I'm not sure how developer names and id's are added otherwise. my bad

claration: you need to edit the original file, not copy an entirely new one over

programminglaboratorys:
for removing a tooltip, making `getSpeakingWhileMuted` always return false could result in unintended side effects. could you patch the component that displays it instead?

Vendicated:
too niche and probably doable with css

PR #4109, opened 2026-04-13
ChannelNameFormatter

Gh0sTyNZ:
Adds a ChannelNameFormatter plugin to improve readability of channel names by removing dashes/underscores and applying proper capitalization for a cleaner UI.


Gh0sTyNZ:
Hi, this is my first contribution for the ChannelNameFormatter plugin. The workflow looks like it’s awaiting maintainer approval before tests can run. Thanks for taking a look.

Nuckyz:
This plugin is too niche to be merged to the main repo. It should also be done using patches instead of overwriting the local cache entry. Regardless, thanks for the contribution!

PR #4112, opened 2026-04-14
fix Select menus not showing

thororen1234:


thororen1234:
also removes NoDefaultHangStatus because they removed hang statuses again... you can check [here](https://discord.com/channels/1015060230222131221/1105582084237570088/1493733904581005483)

PR #4120, opened 2026-04-17
MessageLogger: fix attachment deletion detection

### LINKED ISSUE #3877: [Bug] Messagelogger does not support new attachment deletion feature

When someone deletes an attachment from a message, the attachment is deleted, and it is not preserved by Vencord's messagelogger.

The expected behavior is that it would be logged by messagelogger, and still visible to the user.

1. Send message including a text file
2. Delete the file attachment from the message
3. Observe from the messagelogger user, that it is not logged

### Errors

N/A

### Request Agreement

xNasuni:
I opted for a red overlay instead of grayscale since it is hard to notice on certain images and impossible on other types of attachments
I haven't seen any bugs with this after testing for a bit, and I don't think any of my changes have any negative unintended side effects

fixes #3877

I'm open to suggestions and criticism image


xNasuni:
image

xNasuni:
@Vendicated please don't let me end up like ImSkully 💔

sadan4:
seems superseded by https://github.com/Vendicated/Vencord/pull/4342

xNasuni:
> seems superseded by https://github.com/Vendicated/Vencord/pull/4342

to be honest, I don't know if superseded is the correct word
I don't even have a computer to use anymore because of something that happened but I will leave this pr in the past I guess if you think so

I really liked my red overlay on deleted attachments because you literally can't tell for non image attachments when it's only a grayscale, and even that was hard to see on images sometimes

I don't know if the filter for the history modal that I implemented is required in your implementation so it doesn't show entries with no actual changes since it doesn't show attachments in the modal anyway

I don't know if you made sure toggle message highlight work properly for the attachments or remove message history works so it would clean up the attachments etc

I’m not really in the state to look through code on my phone I’ve had a long day
as much as I wanted to contribute to something it doesn't really matter
that's all I have to say

PR #4123, opened 2026-04-18
feat(plugin): closeDMs

zoinbase:
Adds a button to close DMs or leave group DMs.


Vendicated:
too niche and not very intuitive

PR #4126, opened 2026-04-18
feat: add FixLinkEmbeds plugin

kikkudayo:
Added a plugin that rewrites supported social media URLs to embed-fixing alternative domains before sending messages to improve Discord embed previews.


Vendicated:
TextReplace can do this

PR #4129, opened 2026-04-18
CharacterCounter: add selected text count

thororen1234:


src/plugins/characterCounter/index.tsx
```diff
@@ -56,16 +62,37 @@ export default definePlugin({
...
+        const [selectedCount, setSelectedCount] = useState(0);
+        const showSelected = selectedCount > 0;
         const premiumType = UserStore.getCurrentUser().premiumType ?? 0;
         const charMax = premiumType === 2 ? 4000 : 2000;
-
         const color = getCounterColor((text.length / charMax) * 100);
 
+        useEffect(() => {
+            function onSelectionChange() {
+                const selection = window.getSelection();
```
Vendicated: no guarantee this is inside the chat input

src/plugins/characterCounter/index.tsx
```diff
@@ -56,16 +62,37 @@ export default definePlugin({
...
+                if (!selection || selection.isCollapsed) {
+                    setSelectedCount(0);
+                    return;
+                }
+                setSelectedCount(selection.toString().length);
+            }
+
+            eventEmitter.on("selection-changed", onSelectionChange);
```
Vendicated: you could use document.addEventListener("selectionchange") too

Vendicated:
still doesn't really work correctly :sob: 


https://github.com/user-attachments/assets/0e56f9d3-1679-43e7-abcf-80be1d71bd0e

thororen1234:
> still doesn't really work correctly 😭
> 
>  Screencast.From.2026-06-14.17-24-07.mp4

probably fixed now I couldnt repro it when I switched to document which apperantly just doesnt track outside now so 🔥

src/plugins/characterCounter/style.css
```diff
@@ -12,4 +12,8 @@
 
 .vc-charCounter-count {
     transition: color 0.3s ease;
+}
+
+.vc-charCounter-selected {
+    color: #4285F4
```
Vendicated: please don't hardcode colours, use a discord variable

Vendicated: brand-500, same as reply highlight

or var(--focus-primary, #00b0f4)

thororen1234: I listened to sadan for once and I found a good one so this is done

thororen1234: I did brand-500

PR #4130, opened 2026-04-19
chore: add TypeScript output folder and Vesktop files to .gitignore

lunaynx:
Adds the `who-fucking-cares-dude` folder (which is the defined output folder in `tsconfig.json`) and files starting with `vencordDesktop` (created by Vesktop if you point it at the source tree) to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ src/userplugins
 
 ExtensionCache/
 /settings
+who-fucking-cares-dude/
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,9 @@ src/userplugins
 
 ExtensionCache/
 /settings
+
+# TypeScript output
 who-fucking-cares-dude/
+
+# Vesktop files
+vencordDesktop*


nin0-dev:
it’s called `who-fucking-cares-dude` because vencord uses esbuild to properly build to the `dist` folder so this folder is never made in normal circumstances

lunaynx:
I see, my bad. I used to be a regular / contributor but I haven't touched Vencord code in years. I promise you no AI was used for this pull request.

~~Not sure about the Vesktop files, it seems like I'm probably also doing something wrong there, since Vesktop seems to intentionally delete the package.json if I set it to use my git clone as the Vencord directory.~~ Found the answer in the docs.

nin0-dev:
oh my bad, we frequently get AI submissions

vesktop shouldn’t do that at all, would you mind rejoining the server to see what’s going on? that’ll be easier than github pr comments

lunaynx:
Yeah, I asked there but I didn't get an answer yet. It was because I was directing it to `~/src/Vencord` instead of `~/src/Vencord/dist`.

Vendicated:
The folder wouldn't be named this (and be gitignored) if it was meant to be used :P

And Vesktop shouldn't even let you choose the wrong directory, are you sure you're up to date? If yes, I'll have to revisit that

PR #4135, opened 2026-04-21
Badges: fix crashing

sadan4:


sadan4:
works on stable and canary

PR #4138, opened 2026-04-22
fix(modal): support both legacy and new Discord modal API

nghiphaam:
The modal API was only searching for `MODAL_ROOT_LEGACY` which doesn't exist in newer Discord builds. 

This adds fallback patterns to detect both legacy and new modal components using `anyCode()` helper that tries multiple filter patterns.

- Works on old Discord builds (MODAL_ROOT_LEGACY)
- Works on new Discord builds (headerIdIsManaged)
- No breaking changes to existing plugins

### DIFF
--- a/src/utils/modal.tsx
+++ b/src/utils/modal.tsx
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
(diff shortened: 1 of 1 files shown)


nghiphaam:
merge to dev branch ...

PR #4139, opened 2026-04-22
fix context menu crashes & some plugins

sadan4:


src/plugins/characterCounter/index.tsx
```diff
@@ -50,8 +50,8 @@ export default definePlugin({
         {
             find: "#{intl::PREMIUM_MESSAGE_LENGTH_UPSELL_TOOLTIP}",
             replacement: {
-                match: /return \i\?\i\(\):\i\(\)(?<=#{intl::PREMIUM_MESSAGE_LENGTH_UPSELL_TOOLTIP_WITHOUT_LINK}.{0,200}?)/,
-                replace: "return null"
+                match: /return .+?,\i\?\(0,\i\.jsxs?.+?:\(0,\i\.jsxs?.+?#{intl::PREMIUM_MESSAGE_LENGTH_UPSELL_TOOLTIP}.+?\}\)\]\}\)(?=\})/,
```
Vendicated: now the beginning of this match is useless. it will find the first return in the file, then match until comma

sillyElv:
```suggestion
                match: /#{intl::PREMIUM_MESSAGE_LENGTH_UPSELL_TOOLTIP}.{0,20}\}\)\]\}\)(?=\})/,
```

this is at least better

PR #4145, opened 2026-04-24
feat(plugin): ExpressiveKeysmash

josiauh:
well it makes it LESS expressive but it works

the plugin is simple, if you press a bunch of keys in a short amount of time (i will make that ms value changeable immediately after i make this pr) it sends a message

you can add custom messages
i'm also pretty short on messages so if you want to add some maintainers, go ahead

https://github.com/user-attachments/assets/1a6c6f69-d158-4500-98ad-4fe37154f62a

if this gets accepted which it might not then that would be crazy... YOU COULD SAY THAT AGAIN

features:
* customizable cooldown between smashes so that you dont spam the chat because you're literally dying of laughter
* extra mesages like LoadingQuotes
* functionality for capital smashes
* other than that its not that special
* nobody needed it but we're getting it

### DIFF
--- a/src/plugins/expressiveKeysmash/index.ts
+++ b/src/plugins/expressiveKeysmash/index.ts
@@ -1,0 +1,109 @@
+import { definePluginSettings } from "@api/Settings";
+import { Devs } from "@utils/constants";
+import { getCurrentChannel, sendMessage } from "@utils/discord";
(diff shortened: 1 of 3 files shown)


josiauh:
theres audio in my video 😭
thanks a lot obs desktop audio channel

Vendicated:
insane

PR #4160, opened 2026-04-30
fix location settings

RoScripter999:
fixes the vencord section not being positioned at the logout section cuz discord changed it to utility_section. 

I dont think ANYONE uses the bottom section but whatever


Vendicated:
thanks

PR #4162, opened 2026-05-01
Fix NoTrack, Experiments, ImplicitRelationships

thororen1234:
notrack experiments and implicitrelationships are stable compat however decor isnt

### DIFF
--- a/src/plugins/_core/noTrack.ts
+++ b/src/plugins/_core/noTrack.ts
@@ -44,8 +44,8 @@ export default definePlugin({
             find: "AnalyticsActionHandlers.handle",
             predicate: () => settings.store.disableAnalytics,
             replacement: {
-                match: /^.+$/,
-                replace: "()=>{}",
+                match: /\(0,\i\.analyticsTrackingStoreMaker.{0,800}scheduleWhenIdle:\i\.\i\}\);/,
+                replace: "",
             },
         },
         {
--- a/src/plugins/_core/noTrack.ts
+++ b/src/plugins/_core/noTrack.ts
@@ -44,8 +44,8 @@ export default definePlugin({
             find: "AnalyticsActionHandlers.handle",
             predicate: () => settings.store.disableAnalytics,
             replacement: {
-                match: /\(0,\i\.analyticsTrackingStoreMaker.{0,800}scheduleWhenIdle:\i\.\i\}\);/,
-                replace: "",
+                match: /\(0,\i\.analyticsTrackingStoreMaker\)/,
+                replace: "(()=>{})",
             },
         },
         {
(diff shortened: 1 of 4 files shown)


src/plugins/_core/noTrack.ts
```diff
@@ -44,8 +44,8 @@ export default definePlugin({
             find: "AnalyticsActionHandlers.handle",
             predicate: () => settings.store.disableAnalytics,
             replacement: {
-                match: /^.+$/,
-                replace: "()=>{}",
+                match: /\(0,\i\.analyticsTrackingStoreMaker.{0,800}scheduleWhenIdle:\i\.\i\}\);/,
```
sadan4: would it be easier to match the start and replace with with a noop fn here.

src/plugins/implicitRelationships/index.ts
```diff
@@ -96,7 +96,7 @@ export default definePlugin({
         {
             find: ".REQUEST_GUILD_MEMBERS,",
             replacement: {
-                match: /\.send\(\i\.\i\.REQUEST_GUILD_MEMBERS,{/,
+                match: /\.send\(\i\.(?:\i\.)?REQUEST_GUILD_MEMBERS,{/,
```
thororen1234: this is for stable compat

src/plugins/decor/index.tsx
```diff
@@ -112,10 +112,6 @@ export default definePlugin({
         {
             find: "#{intl::PREMIUM_UPSELL_PROFILE_AVATAR_DECO_INLINE_UPSELL_DESCRIPTION}",
             replacement: [
-                {
-                    match: /(#{intl::PREMIUM_UPSELL_PROFILE_AVATAR_DECO_INLINE_UPSELL_DESCRIPTION}.+?return null!=(\i)&&\()(null==\i)/,
-                    replace: (_, rest, avatarDecoration, hasPurchase) => `${rest}(${avatarDecoration}.skuId!==$self.SKU_ID&&${avatarDecoration}.skuId!==$self.RAW_SKU_ID&&${hasPurchase})`
-                },
```
thororen1234, line 118: this will break it on stable so I feel like the stable compat below might be useless

src/plugins/implicitRelationships/index.ts
```diff
@@ -96,7 +96,7 @@ export default definePlugin({
...
                 replace: "$&nonce:arguments[1].nonce,"
```
Vendicated:
```suggestion
                match: /\.REQUEST_GUILD_MEMBERS,{/,
replace: "$&nonce:arguments[1]?.nonce,"
```

src/plugins/decor/index.tsx
Vendicated: just revert this change

thororen1234:
done

PR #4163, opened 2026-05-01
Add Windows background material support

PL7963:
Add Windows background material support.  

This feature doesn't require transparency. And transparency would introduce some issues like missing window animation, unable to actually maximize discord app etc.. So I also added note for enable transparency option for Windows users.  

The transparency option cannot be removed, because backgroundMaterial doesn't have working full transparent option, at least in my knowledge.


Vendicated:
Hi, thanks for your contribution! ❤️ 

I wasn't fully satisfied with various aspects of this PR (using deprecated components, no check if it's even supported, bad naming/description, merging with mac vibrancy logic) so I ended up doing it myself via https://github.com/Vendicated/Vencord/commit/b8a515dff5163ab31d157a204aa134c09ed7f20d

PR #4165, opened 2026-05-02
remove deprecated plugin.options field

Vendicated:
neeeeeds testing


src/plugins/betterNotes/index.tsx
Vendicated, line 1: deleted because very niche

PR #4171, opened 2026-05-03
feat: add safari extension build

claration:
## Description

Allows Vencord to be used within the Safari browser, the submodule here is located at [Vencord/Vencord-Safari-Extension](https://github.com/Vencord/Vencord-Safari-Extension).

```sh
# Run this with Xcode installed on Mac
pnpm buildWebStandaloneXcode
```

Preferably this should get merged once we get a proper workflow for uploading builds to Testflight/AppStore, but it does properly build the extension on its own.

### Notes

- There will be no themes/quickcss window, similar to the userscript on firefox.

image

###### This PR was remade, smite me


ssalggnikool:
image

ssalggnikool:
image

.github/workflows/publish.yml
```diff
@@ -43,3 +44,20 @@ jobs:
                   CLIENT_ID: ${{ secrets.CHROME_CLIENT_ID }}
                   CLIENT_SECRET: ${{ secrets.CHROME_CLIENT_SECRET }}
                   REFRESH_TOKEN: ${{ secrets.CHROME_REFRESH_TOKEN }}
+
+            - name: Setup Certificates
+              uses: apple-actions/import-codesign-certs@v5
```
Vendicated: to prevent supply chain attacks, all third party actions should be pinned to a long commit hash instead of tag

Vendicated: does this even need an action? looking at https://github.com/Apple-Actions/import-codesign-certs/blob/main/src/security.ts it seems like this is just a few shell commands

claration: It doesn't but it's mainly used for convenience

PR #4174, opened 2026-05-04
add HideMediaEverywhere

t6rtar:
Right click a message to hide specific media links everywhere they appear in discord. good against people who spam naughty gifs.

image
image


claration:
There is already a plugin inside of Vencord that hides media: https://vencord.dev/plugins/HideMedia
image

It should be better to modify that plugin.

Also, this is AI generated.

t6rtar:
the contributing guidelines never said anything banning ai usage. Regarding hidemedia, hidemedia does it on an individual basis while hidemediaeverywhere stores the url and blocks it everywhere. just felt like keeping them separate made more sense.

Vendicated:
not sending aislop should be common sense

PR #4175, opened 2026-05-04
feat: add Pankaj to contributors and implement VoiceRPC plugin

Pankajkumar4433:
### Overview
This PR introduces a new plugin called **VoiceRPC** and adds **Pankaj** to the developers' list in the constants.

### Changes
1. **VoiceRPC Plugin**:
    - Adds a rich presence (RPC) status when joined to a voice channel.
    - Displays the guild name, channel name, and a list of participants in the status.
    - Automatically updates when the user switches channels or when participants change.
    - Uses a debounced update mechanism to prevent rate-limiting.
2. **Constants Update**:
    - Added `Pankaj` to the `Devs` object in `src/utils/constants.ts` with the corresponding Discord ID (`1168970294678003826n`).

### Testing
- [x] Verified that the RPC updates correctly when joining and leaving voice channels.
- [x] Confirmed that the participant count/names update dynamically.
- [x] Verified that the developer badge/attribution logic correctly recognizes the new entry.

### DIFF
--- a/src/plugins/voiceRPC/README.md
+++ b/src/plugins/voiceRPC/README.md
@@ -1,0 +1,30 @@
+# VoiceRPC
+
+A Vencord plugin that adds a custom Rich Presence (RPC) status when you join a voice channel.
+
+## Features
+
+- **Real-time Voice Tracking:** Automatically updates your Discord status as you join, move, or leave voice channels.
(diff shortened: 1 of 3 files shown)


Vendicated:
aislop

PR #4178, opened 2026-05-05
QuickCSS: drop inherited app menu so close shortcuts only close the e…

### LINKED ISSUE #2548: [Bug] Closing QuickCss with keyboard shortcut (Alt + F4 / Ctrl + Q) closes Discord instead of just QuickCss
### Discord Account

_No response_

Closing QuickCss with keyboard shortcut (Alt + F4 / Ctrl + Q) closes Discord

it should only close the QuickCss window

1. Open QuickCss
2. Use platform specific window close keybind (Alt F4 / Ctrl Q)
3. Observe that Discord closes fully

### Errors

_No response_

### Request Agreement

thomassolcia:
## What this fixes

Pressing `Alt+F4` on the Vencord QuickCSS Editor window quits the entire
Discord client instead of only the editor, closes #2548.

## Root cause

The QuickCSS `BrowserWindow` is created without a per-window menu, so
on Windows and Linux it inherits Discord's application menu. The
inherited menu's accelerators (Quit) keep firing from this window even
though `autoHideMenuBar: true` is set, that option only hides the bar,
it doesn't disable accelerators.

## Fix

Call `win.removeMenu()` immediately after creating the BrowserWindow.

- **Windows/Linux**: per-window menu is removed, so inherited
  accelerators no longer fire from the QuickCSS window. `Alt+F4` falls
  back to the OS-level "close focused window" behavior, and any other
  inherited menu accelerator on Linux (e.g. `Ctrl+Q`) becomes inert
  here.
- **macOS**: `removeMenu` is a no-op (the menu is global). `Cmd+Q`
  quitting the focused app is the OS convention there and is left
  untouched, matching the position taken in the issue thread by
  @Tiagoquix.

This addresses the platform-coverage concern raised on #2493
("the shortcut is different on linux & mac") without trying to
intercept individual key combinations.

## Validation

End-to-end on Discord Canary (Windows 11):

- Built the dev bundle from the parent commit (without this change),
  injected, opened QuickCSS, pressed `Alt+F4` → entire Discord client
  closed (bug reproduced).
- Built with this change applied, re-injected, repeated the steps →
  only the QuickCSS window closed; Discord remained open.

### DIFF
--- a/src/main/ipcMain.ts
+++ b/src/main/ipcMain.ts
@@ -163,6 +163,8 @@ ipcMain.handle(IpcEvents.OPEN_MONACO_EDITOR, async () => {
         }
     });
 
+    win.removeMenu();
+
     makeLinksOpenExternally(win);
 
     await win.loadURL(`data:text/html;base64,${monacoHtml}`);


Tiagoquix:
Hi, thanks for the PR, but your entire PR seems to be written by a LLM, which makes me not want to talk about it.

thomassolcia:
> Hi, thanks for the PR, but your entire PR seems to be written by a LLM, which makes me not want to talk about it.

Hii, the description might've been overkill, lol... first time doing this and English isn't my native language, I just wanted to make it as descriptive and formal as possible, sorry about this. Want me to close the PR then?

PR #4180, opened 2026-05-05
fix(plugins): narrow MessageStore webpack patch find strings

punkholic:
Plain '"MessageStore"' appears in many webpack chunks, so MessageLogger and NoBlockedMessages patches could run on the wrong module factory. That produced invalid JS after substitution and eval() threw SyntaxError (e.g. Discord stable 1.0.136), often showing a blank client.

Scope patches to the flux store implementation via static displayName="MessageStore" / "ReferencedMessageStore", which is unique in the main web bundle.

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -350,7 +350,8 @@ export default definePlugin({
 
     patches: [
         {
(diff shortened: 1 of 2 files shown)


Vendicated:
no it fucking doesn't, go away with your ai

PR #4183, opened 2026-05-05
fix NoMosaic

AutumnVN:


src/plugins/noMosaic/index.ts
```diff
@@ -0,0 +1,53 @@
...
+            replacement: {
+                match: /"IMAGE"===\i\|\|"VIDEO"===\i\|\|"CLIP"===\i/,
+                replace: "false"
+            }
+        },
+        {
+            find: "return{visualMediaItems:",
+            replacement: {
+                match: /return{visualMediaItems:.*?props:\i/,
+                replace: "$&,useFullWidth:false"
```
Vendicated: I don't think this is really correct

The problem is that our above patch makes Discord consider images as nonVisualMediaItems which makes them full width

This patch bandaid fixes that by making all nonVisualMediaItems render without full width, but this will also affect real non visual items (i assume stuff like files, maybe voice messages, etc)

We have to fix it properly some way but I couldn't figure out a good way that's why I just deleted the plugin for now xD

KingGamingYT:
I'm no vencorder but from my experience developing the betterdiscord-equivalent plugin, the state this is in now is still gonna crop images because of discord's width and height link hardcoding (unless something else in vencord prevents this)

Again I didn't test this but that's my interpretation from trying the same things in the bd plugin

src/plugins/noMosaic/index.ts
```diff
@@ -0,0 +1,53 @@
...
+            replacement: {
+                match: /"IMAGE"===\i\|\|"VIDEO"===\i\|\|"CLIP"===\i/,
+                replace: "false"
+            }
+        },
+        {
+            find: "return{visualMediaItems:",
+            replacement: {
+                match: /return{visualMediaItems:.*?props:(\i)/,
+                replace: '$&,useFullWidth:$1.item.type==="IMAGE"||$1.item.type==="VIDEO"||$1.item.type==="CLIP"?false:undefined'
```
Vendicated:
```suggestion
                replace: '$&,useFullWidth:["IMAGE","VIDEO","CLIP"].includes($1.item?.type)?false:undefined'
```

Vendicated:
Seems to work fine

PR #4187, opened 2026-05-05
MessageLogger: opt-in persistence + parity with BetterDiscord MLV2

MrMcEpic:
Closes #4186.

Most of what people use BetterDiscord's MessageLoggerV2 for is the persistence side: deleted and edited messages surviving a client reload, plus a viewer to browse the log, plus pin/save and attachment caching on top. The current MessageLogger keeps everything in memory; the markers live on the live MessageCache and disappear on reload. This PR brings the plugin to feature parity with MLV2, with a couple of deliberate exclusions called out below.

Around 2k lines added under `src/plugins/messageLogger/`. Built in five sequential phases, each its own commits on the branch, so it's reviewable in chunks if you want to read it that way.

### What's in the patch

1. **Persistence core.** Own IndexedDB database (`VencordMessageLogger`), not `@api/DataStore` (idb-keyval), because we need secondary indexes. Capture hooks the existing `handleDelete`/`makeEdit` paths. Restore via `LOAD_MESSAGES_SUCCESS` and `CHANNEL_SELECT` flux handlers, reusing the `MessageCache.receiveMessage(...).update(...)` pattern (same approach `messageLinkEmbeds` uses) so deleted ghosts come back with the red strikethrough on reload. Retention by days and count, plus retroactive purge when the user adds an entry to `ignoreUsers`/`ignoreChannels`/`ignoreGuilds`.

2. **Viewer modal** under `src/plugins/messageLogger/viewer/`. Tabs (Deleted/Edited/Saved/All), search across content+author, scope dropdown (this channel / this server / everywhere), author filter, sort, virtualized list via `ListScrollerThin`, click-row-to-jump, per-row remove, clear-visible footer. Three entry points: channel context, guild context, and a button in `settingsAboutComponent`.

3. **Saved/pinned entries.** Star toggle on rows. Saved entries are exempt from retention purge.

4. **Optional attachment caching, off by default.** Eagerly downloads attachments at delete time so they survive Discord CDN URL expiry (~24h). Desktop writes to `${DATA_DIR}/messageLogger/attachments/<id>` via a new `native.ts`; browser falls back to a Blob in IndexedDB. Type filter (images/videos/audio/other), per-file MB cap, total MB cap with LRU eviction, 3-worker download queue. Inline restore substitutes `blob:` URLs for the dead CDN URLs, with a `#mlcache` fragment guard so Discord's image renderer's `?format=webp&width=...&height=...` append doesn't break the blob-store lookup.

5. **JSON export/import.** Plain DOM `Blob` + `<a download>` for export, `<input type=file>` for import. Magic marker plus schemaVersion gate on import. Attachments are referenced by id, never embedded as bytes (a 250 MB cache base64-encoded would be unusable).

### Parity vs. MLV2

Present:
- Persistence between client reloads.
- Inline restoration of deleted/edited messages in their original channels.
- Browseable log with filtering and search.
- Pinned/saved entries that survive retention.
- Attachment caching with size caps and LRU eviction.
- JSON export and import.
- Type filter for the attachment cache (images by default).

Deferred on purpose, happy to add in follow-up PRs:
- "Save Message" from the in-channel message context menu (saving a live message, not just one that has already been deleted or edited).
- Scheduled auto-export.
- Thumbnail strip for cached attachments inside the viewer modal (the channel-render path already shows them; the modal is text-only for now).
- Refcount-based cache cleanup. Eviction is purely LRU; orphaned cache entries age out naturally rather than getting eagerly cleaned.

### Constraints respected

- No new dependencies. Raw IDB only.
- No new top-level plugin; everything lives under the existing `MessageLogger` directory.
- The `definePlugin({...})` literal default export is preserved (the AST parser in `scripts/generatePluginList.ts` requires it).
- Path-alias-only cross-folder imports (`@plugins/messageLogger/...`).
- Browser users don't break. Native IPC is gated behind `IS_DISCORD_DESKTOP || IS_VESKTOP`; everything else has a Blob-in-IDB fallback.
- All new behaviour is behind settings with sensible defaults. Existing users see no change unless they opt in. `persistEnabled` defaults true (the headline switch); `cacheAttachmentsEnabled` defaults false (the heavy one).
- Schema versioning lives in `meta.version`, bumped 1 -> 2 -> 3 across the phases. Forward-compat readonly fallback if the on-disk version is higher than the code expects.

### Smoke testing

Manual smoke tested on Discord stable (Windows 11) across all phases:
- Capture and restore on reload (deleted ghosts and edit history both come back).
- Viewer search/sort/scope/author-filter, per-row remove, clear-visible.
- Star/unstar; retention purge respects saved entries (verified by setting count cap below current size and confirming starred entry survives).
- Attachment caching with LRU eviction. Verified files written to `%APPDATA%\Vencord\messageLogger\attachments\` on desktop. Cached `blob:` URLs render in the channel after restore.
- Export round-trip (file downloads, magic marker present, entry count matches), import-restore (entries reappear in viewer and survive reload), idempotent re-import, magic-marker rejection on unrelated JSON, future-version rejection on a hand-edited `schemaVersion`.

### Settings added

| Key | Default | Notes |
| --- | --- | --- |
| `persistEnabled` | `true` | Headline switch. Disabling stops new captures and skips inline restore. |
| `persistRetentionDays` | `30` | 0 disables. Saved entries are exempt. |
| `persistRetentionCount` | `10000` | 0 disables. Saved entries don't count toward the cap. |
| `restoreInline` | `true` | If false, persistence still happens but deleted ghosts only show in the viewer, not inline in channels. |
| `viewerRowDensity` | `compact` | `compact` or `comfortable`. |
| `cacheAttachmentsEnabled` | `false` | Off by default. The only setting that introduces network/disk activity. |
| `cacheAttachmentImages` | `true` | Per-bucket toggle. |
| `cacheAttachmentVideos` | `false` | Per-bucket toggle. |
| `cacheAttachmentAudio` | `false` | Per-bucket toggle. |
| `cacheAttachmentOther` | `false` | Per-bucket toggle. |
| `cacheAttachmentsPerFileCapMB` | `5` | Drops attachments larger than this. |
| `cacheAttachmentsTotalCapMB` | `250` | LRU evicts when exceeded. |


Vendicated:
holy aislop

PR #3035, opened 2024-11-22
noBlockedMessages: remove users from member list and add override for DMs

sillyElv:
initially a remake of #2803

changes:
- allowing to hide replies to blocked messages
- hiding blocked users from the member list
- allowing an override of the DMs to look like the user was never blocked (because why look at DMs if half the history is invisible?)
- improved/clearer wording

PR #4190, opened 2026-05-06
feat(NoBlockedMessages): Remove replies to blocked messages

### LINKED ISSUE #4182: Feature Request - Hiding replies to blocked users
### Content

https://github.com/Vendicated/Vencord/blob/main/src/plugins/noBlockedMessages/index.ts

Currently the plugin only hides the actual messages from blocked/ignored users. I am thinking about working on having a setting where you can hide all messages to blocked/ignored users, too. 

Feedback?

### Request Agreement

n1ckoates:
Closes #4182 

This is my first time writing patch regexes for Vencord, so hopefully I'm doing this right. Apologies if not.

I initially thought this would be a pretty simple task but it required a lot more code than I had anticipated.
* Patches the Message component to check if its a reply to a blocked user and return `null` if so.
* That check uses `ReplyStore` to lookup message references, because messages being replied to often weren't in `MessageStore` in my testing, especially when they were old. So there's additional patches to get access to `ReplyStore`.
* Added flux handlers for relationship updates that just sends stub message updates for the current channel. Otherwise, when you blocked someone messages wouldn't re-render until you navigated away from the channel and back. If that's an ok tradeoff though, I'm happy to remove the flux handlers to make the code simpler.

### DIFF
--- a/src/plugins/noBlockedMessages/index.ts
+++ b/src/plugins/noBlockedMessages/index.ts
@@ -21,12 +21,44 @@ import { Devs } from "@utils/constants";
 import { runtimeHashMessageKey } from "@utils/intlHash";
 import { Logger } from "@utils/Logger";
 import definePlugin, { OptionType } from "@utils/types";
-import { Message } from "@vencord/discord-types";
-import { i18n, RelationshipStore } from "@webpack/common";
+import { Message, MessageJSON } from "@vencord/discord-types";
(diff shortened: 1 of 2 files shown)


sillyElv:
From my experience just patching readstatestore and messagestore is absolutely enough to filter out replies, it's the first two patches, just put an or into the if and see if that works if you didn't already (worked fine for me the last time I worked on #3035)

PR #4191, opened 2026-05-06
VoiceMessages: add Alt+V to record and Alt+B to send

sree-sanak:
## Summary
Adds keybinds to the VoiceMessages plugin: Alt+V to open the recorder and start recording, Alt+B to stop and send.

Motivation: the existing flow requires right-click → context menu → Start → speak → Stop → Send (4 clicks plus mouse navigation). Often voice-message senders just want to fire and forget.

## Behavior
- **Alt+V** (global): opens the modal with `autoStart`. Permission-checked the same way as the right-click context menu.
- **Alt+B** (modal-scoped): if recording, stops; once the blob and waveform are ready, sends and closes. If a blob is already ready (e.g. uploaded file), sends immediately.
- A single new boolean setting **Use keybinds** (default on) controls both shortcuts.
- The right-click "Send Voice Message" entry still opens the modal without auto-starting, so existing UX is preserved.

## Implementation notes
- Pattern follows `quickReply` (global keydown listener registered in `start()`, removed in `stop()`).
- Keybinds reject when any other modifier is held alongside Alt.
- The recorder receives a `stopSignal` counter prop; a watching effect in the modal observes blob+meta and calls `sendAudio` once both are ready and a send is pending.

### DIFF
--- a/src/plugins/voiceMessages/DesktopRecorder.tsx
+++ b/src/plugins/voiceMessages/DesktopRecorder.tsx
@@ -17,20 +17,33 @@
 */
 
 import { PluginNative } from "@utils/types";
-import { Button, MediaEngineStore, showToast, Toasts, useState } from "@webpack/common";
+import { Button, MediaEngineStore, showToast, Toasts, useEffect, useRef, useState } from "@webpack/common";
 
 import type { VoiceRecorder } from ".";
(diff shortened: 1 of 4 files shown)


Vendicated:
aislop

PR #4194, opened 2026-05-06
PermissionsViewer: sort roles by hierarchy, users alphabetically

Vexify4103:
Currently the roles/users list in the permissions modal is unsorted, making it hard to find a specific user or role at a glance.

This change sorts the list alphabetically by name within each type group (roles first, then members), re-sorting reactively when member data finishes loading.


Vendicated:
Ordering alphabetically doesnt make that much sense either imo. If anything it should be ordered by hierarchy, i.e. highest role first

Vexify4103:
What if we do roles per hierarchy and users alphabetically?

Vendicated:
sounds good

Vexify4103:
Also what do we think of the idea of adding a toggle button, which would allow switching from usernames to nicknames?
The current system only does discord usernames.

Vendicated:
Seems pretty over-engineered, like AI code usually is

Vexify4103:
Yeah I used claude to do this pr. The username sorting alphabetically is good, you already approved. The nickname stuff is pretty over engineered. I can remove it if you want.

Vexify4103:
I’ve gone ahead and removed the nickname toggle change now.
Idk if I should open a new PR for that change, but it feels more right to keep this PR focused on the original quality of life improvement instead of adding extra complexity for no real reason.

PR #4195, opened 2026-05-06
add dmsearch plugin

QDaves:
adds message search across all dms and group dms to the quick switcher. mobile discord has this through an internal api but desktop never got it natively. there was an experiment for it on desktop but it was removed from the client about a year ago.

opens with ctrl+k. searches dms and group dms. tabs for all, messages, media, pins, links and files like the mobile app.


Vendicated:
Stock Discord feature

PR #4200, opened 2026-05-07
Feat/MutualTimeZones

tojuszn:
Adds a local time badge next to usernames in profile popouts.

Your own timezone is auto-detected from your system. For other users, you can right-click their name and select "🕐 Set Timezone" to manually set their timezone, which is saved locally on your client.

This is useful for Discord servers with international members where knowing someone's local time at a glance is helpful.

Features:
- Auto-detects your own timezone
- Right-click any user → Set Timezone (26 timezones available)
- Shows time-of-day emoji (🌅☀️🌆🌙)
- Shows hour difference from your timezone on hover
- Fully local, no external APIs


Vendicated:
Duplicate of various other plugins, also vibecoded

PR #4203, opened 2026-05-07
feat(plugin) noNameplates

lostf1sh:
## Summary

Adds NoNameplates, an appearance plugin that hides Discord collectible nameplates in the member list.

## Rationale

Discord nameplates can make the member list visually noisy and harder to scan, especially in large servers where many users have animated or high-contrast nameplates. This plugin provides a simple toggle for users who want a cleaner member list without disabling other profile decorations.

Although this is implemented with CSS, keeping it as a plugin makes it discoverable and easy to enable/disable from Vencord settings without requiring users to maintain custom CSS selectors themselves.

## Implementation

The CSS is scoped to member list nameplate wrappers and only hides the direct visual layers used by Discord nameplates. It does not touch the member row itself, so hover states, avatars, badges, activities, status text, and popouts continue to work.

## Testing

- Enabled the plugin in Discord
- Verified member list nameplate gradients/images are hidden
- Verified hover states and user row content still work
- Ran `pnpm exec eslint src/plugins/noNameplates/index.ts src/utils/constants.ts`
- Ran `pnpm exec stylelint src/plugins/noNameplates/style.css`
- Ran `pnpm exec tsc --noEmit --pretty false`

### DIFF
--- a/src/plugins/noNameplates/index.ts
+++ b/src/plugins/noNameplates/index.ts
@@ -25,6 +25,6 @@ export default definePlugin({
     name: "NoNameplates",
     description: "Hides Discord nameplates in member lists.",
     tags: ["Appearance", "Customisation"],
-    authors: [Devs.KingFish],
+    authors: [Devs.Moli],
     managedStyle,
 });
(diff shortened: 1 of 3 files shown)


Vendicated:
Consider using QuickCSS instead of aislop
```css
[class*=nameplated]>[class*=container],
[class*=dm] [class*=fadeIn] {
    display: none;
}
```

PR #1765, opened 2023-09-29
CustomSounds

TheKodeToad:
Open
- in case I die
- so you can criticise my bad code

## Roadmap
what does Roadmap mean? I'm sure it's not important

## To-do / Issues
- ~~Upload local file (how should I store the data)~~ fixed
- ~~Settings are applied directly instead of being queued~~ won't fix
- ~~Setting initialisation is lazy which is weird~~ fixed
- ~~More sound customisation~~ won't fix in this PR
- ~~Additional sounds for specific conditions or events not normally covered~~ won't fix in this PR
- ~~Get it to play nicely with Discord's built in sound themes option~~ fixed

PR #4209, opened 2026-05-09
feat(plugin): CustomSounds

SyberiaK:
A plugin that allows to change Discord sounds, replacing them with the seasonal ones or your own sounds that you can upload.

Basically a duplicate of #1765 with some QoL changes:

- Allowed uploading multiple files within the same popup;
- Uploading existing files invokes a prompt to skip or replace files;
- Missing files at configuration import invokes a prompt to upload those files;
- Derived sounds: you can set different sounds to events that either reuse the same sound (connecting to a VC vs someone else connecting to your VC) or where the separation makes sense (mentioning you vs everyone/here vs role);
  - There are currently two derived sounds implemented: `connect` (derived from `user_join`) and `user_mentioned` (derived from `message1`)
- Various fixes and changes that improved on UX or made the plugin's behavior more predictable.

## Some context

For one of my projects I was looking for a Discord plugin that allows to replace sounds with the custom ones with configuration import. There are a couple of implementations out there:

- by TheKoadToad (#1765)
- by @ScattrdBlade (https://github.com/ScattrdBlade/customSounds)
- by Equicord's team (https://github.com/Equicord/Equicord/blob/main/src/equicordplugins/customSounds)

...and neither of those could actually satisfy my requirements:

- TheKoadToad's plugin is **very** dated and lacks configuration import;
- Scattrd's file storage uses randomly generated IDs, which makes imports *almost completely* useless. I did make a [hotfix PR](https://github.com/ScattrdBlade/customSounds/pull/20) by utilizing file contents hashes as IDs, but Scattrd [reverted](https://github.com/ScattrdBlade/customSounds/commit/263108006d1c4edbb68f81658244de6dbfbd44c4#diff-751dcb032a615aeaaf802914775e6c2731dad7727a767e85071ca13fa70caa11R20) the change later on by pulling code from Equicord's repo, *which also removed the migration mechanism, breaking existing configurations* (?);
- Equicord's plugin also uses randomly generated IDs, as well as an outdated sounds list, which now includes some unused ones.

But I was (and still is) very glad those do exist.

I have almost no background in Discord modifications, so the code might look weird in some parts or still do some quirky stuff in the background, which I haven't noticed while using it myself for a few months. It also contains some quite opinionated decisions *(like the whole "derived sounds" thingy)*, but I'm open for any suggestions for changes.

While it is a duplicate, I hope at the very least it sparks more activity into making this kind of plugin available for regular Vencord users without a hassle.

### DIFF
--- a/src/plugins/customSounds/audioStore.ts
+++ b/src/plugins/customSounds/audioStore.ts
@@ -1,0 +1,278 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2026 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { get, set } from "@api/DataStore";
+import { Logger } from "@utils/Logger";
+
+const STORAGE_KEY = "CustomSounds";
+const METADATA_KEY = "CustomSounds_Metadata";
+
+// Default maximum file size: 15MB per file
+const DEFAULT_MAX_FILE_SIZE_MB = 15;
+
+// Configurable max file size (set by plugin settings)
+let maxFileSizeMB = DEFAULT_MAX_FILE_SIZE_MB;
+
+const logger = new Logger("CustomSounds");
+
+export function setMaxFileSizeMB(sizeMB: number): void {
+    maxFileSizeMB = sizeMB;
+}
+
+export function getMaxFileSizeMB(): number {
+    return maxFileSizeMB;
+}
+
+export interface StoredAudioFile {
+    id: string;
+    name: string;
+    type: string;
+    dataUri: string;
+}
+
+export interface AudioFileMetadata {
+    id: string;
+    name: string;
+    type: string;
+    size: number; // Size of dataUri in bytes
+}
+
+// Lightweight metadata stored separately for fast access
+type MetadataStore = Record<string, AudioFileMetadata>;
+
+// Full audio data (only dataUri, no redundant buffer)
+type AudioStore = Record<string, StoredAudioFile>;
+
+/**
+ * Returns audio files metadata.
+ */
+export async function getAllAudioMetadata(): Promise<MetadataStore> {
+    return (await get(METADATA_KEY)) as MetadataStore ?? {};
+}
+
+async function setMetadataStore(new_: MetadataStore): Promise<void> {
+    await set(METADATA_KEY, new_);
+}
+
+/**
+ * Returns a single audio file's data URI.
+ */
+export async function getAudioDataURI(id: string): Promise<string | undefined> {
+    const all: AudioStore = await getAllAudio();
+    return all?.[id]?.dataUri;
+}
+
+/**
+ * Returns all audio files (use sparingly as it loads all the data)
+ */
+export async function getAllAudio(): Promise<AudioStore> {
+    return (await get(STORAGE_KEY)) as AudioStore ?? {};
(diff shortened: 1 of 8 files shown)


ScattrdBlade:
Funny enough, I was working on a bunch of fixes (including the ones you mentioned), improvements, and additions to the plugin right as you made the PR. 

Also note, TheKoadToad was the original dev of the plugin but abandoned it early on, so I started working on it and improved it. I've been the maintainer/dev for the plugin for the last 2 years. As for Equicord's version, I worked with the Equicord team to implement customSounds into Equicord. So in the end, its all just 1 plugin with 3 different versions available (with the https://github.com/ScattrdBlade/customSounds being the most updated)

Edit: I just released an update for the plugin with all fixes (including an improved version of the ID/export system) + some improvements, and new features. Dunno if this PR should get closed and resubmitted or what since this version of customSounds is missing many things.

SyberiaK:
> Funny enough, I was working on a bunch of fixes (including the ones you mentioned), improvements, and additions to the plugin right as you made the PR.
> [...]
> Edit: I just released an update for the plugin with all fixes (including an improved version of the ID/export system) + some improvements, and new features. Dunno if this PR should get closed and resubmitted or what since this version of customSounds is missing many things.

Hey, glad to hear you were working on an update as well!

I went through the commit and honestly, I have quite mixed feelings about it.

- Customizable playback speed and boosted max volume, at least to me, sound like very niche features that won't see much usage outside of a one-time giggle;
- Exports having actual audio data is nice, not exactly sure why I haven't implemented it in the first place, I'll look into it;
- AudioPlayerAPI is also nice, it definitely deserves a separate PR.

Everything else is pretty much unchanged and lacks things addressed here, so I don't see a reason to close the PR just yet.

imjustprism:
looks really overengineered ngl, feels like 70% of this is just bloat

SyberiaK:
> looks really overengineered ngl, feels like 70% of this is just bloat

At some point I felt exactly the same, like: "cmon, there is no way you would need 1k+ lines just to replace sounds :Clueless:"
It does a bit more than just replace sounds though. You could argue that it shouldn't do anything but just that which is fair I guess ¯\\\_(ツ)\_/¯

src/plugins/customSounds/audioStore.ts
```diff
@@ -0,0 +1,278 @@
...
+ */
+export async function clearStore(): Promise<void> {
+    await setAudioStore({});
+    await setMetadataStore({});
+}
+
+/**
+ * Migrates old storage format to new format (run once on startup).
+ */
+export async function migrateStorage(): Promise<{ [oldId: string]: string; } | null> {
```
SyberiaK, line 149: This is the main bloat for me. The only usage I could think of is people moving their existing configurations from the userplugin which would still require to move the store data. Easily -100 lines.

imjustprism:
this entire plugin should be like 500 loc

src/plugins/customSounds/audioStore.ts
imjustprism: i dont see the point of this, they'll just reconfigure it for themselves anyway, it only adds extra complexity

ScattrdBlade:
> this entire plugin should be like 500 loc

I just pushed out an update that compacts the plugin to 470 loc.
https://github.com/ScattrdBlade/customSounds

PR #4210, opened 2026-05-10
New plugin: PdfViewer (supersedes #2876)

RlxChap2:
Adds a `PdfViewer` plugin (`.desktop`) that previews PDF attachments inline so the user no longer has to download them just to see what's inside.

This supersedes the dormant #2876 by @TheGreenPig — the original concept. His last code commit was September 2024 and the PR has been inactive since. This is a fresh implementation that incorporates the review feedback you left there:

- Renamed to `PdfViewer`, folder marked `.desktop` since native is required.
- Strict allowlist using the `URL` API (host + path prefix + `.pdf` extension + `https:` only).
- `fetch` instead of the `https` module, with `.catch(() => null)` so a failed connection produces a meaningful error.
- `redirect: "manual"` so the allowlist cannot be sidestepped through a CDN redirect.
- Streaming download with both a `content-length` pre-check and a per-chunk size cap, so an oversized response is aborted mid-flight rather than buffered to completion.
- `%PDF` magic-byte check before the bytes leave the native side.
- React state for the preview toggle, no DataStore persistence.
- `ErrorBoundary.wrap(..., { noop: true })`.
- All class names prefixed with `vc-`.

### Why PDF.js instead of `<embed>`

The original PR used `<embed type="application/pdf">`, which relies on the host having Chromium's PDF plugin enabled. Recent vanilla Discord builds disable it (`navigator.pdfViewerEnabled` is `false` on current Canary), so the embed would render blank for most users. PDF.js renders to canvas in plain JS so it doesn't depend on the host plugin, and it's loaded lazily from `cdnjs.cloudflare.com` — a host already on Vencord's CSP allow-list (same pattern `shikiCodeblocks.desktop` uses for grammar JSON). No new build dependencies, no CSP changes.

PDF.js is configured with `isEvalSupported: false`, `disableAutoFetch: true`, `disableStream: true`, and pages are rasterised lazily via an `IntersectionObserver` (only the visible page plus its neighbours are rendered at any time).

Happy to add a `<embed>` path back as a runtime fallback for clients where `navigator.pdfViewerEnabled === true` if you'd prefer; left it out here to keep the PR scope tight.

### Settings

- **Largest PDF that can be previewed inline** — slider, 5/10/25/50/100 MB, default 25.
- **Number of recently opened PDFs to keep in memory** — slider, 0/1/3/5/10, default 3 (LRU cache of fetched bytes).

### Files

- `src/plugins/pdfViewer.desktop/{index.tsx,native.ts,styles.css,README.md}` — new
- `src/utils/constants.ts` — adds `Devs` entries for the plugin authors

CSP and every other core file are untouched.

### Screenshots

Plugin entry in settings

PDF inline preview

Page rendering with toolbar

image

### Tested on

- Discord Canary (Chromium PDF plugin disabled, `navigator.pdfViewerEnabled === false`) — works.

Thanks to @TheGreenPig for the original concept, and to @semon009 for the toolbar's Fit Width icon, the per-page hover badges, the resize grip indicator, and the updated Discord theme variables.

### DIFF
--- a/src/plugins/pdfViewer.desktop/README.md
+++ b/src/plugins/pdfViewer.desktop/README.md
@@ -1,0 +1,12 @@
+# PdfViewer
+
+Preview PDF attachments inline without having to download them first.
+
+## How it works
+
+PDFs are fetched through a small desktop-native helper (CORS would otherwise block the renderer when downloading from `cdn.discordapp.com`) and rendered with [PDF.js](https://mozilla.github.io/pdf.js/), loaded lazily from `cdnjs.cloudflare.com`. The viewer supports zoom, fit-to-width and lazy per-page rasterisation, so even large documents stay responsive.
+
+## Settings
+
+- **Largest PDF that can be previewed inline** — above this size the inline preview is disabled. The file remains downloadable from Discord's normal attachment UI.
+- **Number of recently opened PDFs to keep in memory** — LRU cache size for the fetched bytes. `0` disables caching.
--- a/src/plugins/pdfViewer.desktop/index.tsx
+++ b/src/plugins/pdfViewer.desktop/index.tsx
@@ -721,7 +721,7 @@ export default definePlugin({
     name: "PdfViewer",
     description: "Preview PDF attachments inline without downloading them first",
     tags: ["Media", "Utility", "Chat"],
-    authors: [Devs.vp9],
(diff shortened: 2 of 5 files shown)


Vendicated:
ai garbage

PR #4213, opened 2026-05-12
feat: add AutoReactMention plugin

wiruwiru:
## Summary
- Adds a new plugin that automatically reacts to messages where the user is mentioned
- Supports both unicode emojis (👍, 😂, etc.) and custom server emojis (name:id format)
- Configurable via plugin settings
## Context
A small, fun plugin for users who don’t like to be notified and want to add a touch of humor to mentions. Although not essential, it can add a bit of fun to mentions.
## How it works
Listens to `MESSAGE_CREATE` flux events, checks if the current user is mentioned, and automatically adds a reaction using the configured emoji.


Vendicated:
aislop selfbot pr, incredible

PR #4228, opened 2026-05-17
CustomRPC Fix

SwezyDev:
Fix `TimestampMode.TIME` not resetting after 24h in CustomRPC.
The timestamp was only calculated once, causing the timer to continue past midnight.

Now it recalculates based on the start of the current day so it correctly resets at 00:00 without needing timers or scheduling.


snp5164:
bruh i was hoping this would get merged

SwezyDev:
> bruh i was hoping this would get merged

rip bro

PR #4231, opened 2026-05-18
feat(plugin): server tag everywhere

oatsfx:
Similar to RoleColorEverywhere, this plugin adds a server tag everywhere, specificially to 4 locations: Server Tooltip, Server Header, App Title Bar, and User Name Tag.

There are toggles for each location in plugin settings.

image

image


Vendicated:
too niche

PR #4237, opened 2026-05-22
fix various Plugins

thororen1234:
plugins fixed
FullSearchContext
CallTimer
ThemeAttributes
ShowConnections
TypingTweaks


thororen1234:
added ThemeAttributes patch fix

PR #4239, opened 2026-05-24
Disable FakeNitro for Nitro subscribers

EmirX3D:
When I have the FakeNitro plugin enabled (enabled it before I got a Nitro subscription) while my Nitro is activated, and I use for example - an external sticker from another server, FakeNitro gets prioritized over the real Nitro, and it will send a FakeNitro hyperlink instead of a regular sticker. This becomes annoying for some users at some point, and will eventually lead them to disable the extension altogether. To combat this, I've:
- Disabled FakeNitro when the current user already has an active Nitro subscription
- Gated FakeNitro patches and runtime handlers behind the user’s premium status
- Prevented emoji/sticker/theme/stream bypass behaviour from running for Nitro users
When the user's Nitro subscription expires, it will read that the user does not currently have an active Nitro subscription, and will allow them to use the plugin again. All of these changes are silent. A better approach would be to disable the plugin automatically when it detects that the user has an active Nitro subscription, or to give them a notification that the plugin is silently disabled.

(ran ESLint for testing)

### DIFF
--- a/src/plugins/fakeNitro/index.tsx
+++ b/src/plugins/fakeNitro/index.tsx
@@ -73,6 +73,10 @@ const fakeNitroStickerRegex = /\/stickers\/(\d+?)\./;
 const fakeNitroGifStickerRegex = /\/attachments\/\d+?\/\d+?\/(\d+?)\.gif/;
 const hyperLinkRegex = /\[.+?\]\((https?:\/\/.+?)\)/;
 
+function hasNitroSubscription() {
+    return (UserStore?.getCurrentUser?.()?.premiumType ?? 0) > 0;
+}
+
 const settings = definePluginSettings({
     enableEmojiBypass: {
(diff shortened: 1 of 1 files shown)


EmirX3D:
Committed some of Gemini's suggestions

Vendicated:
This isn't a real issue, the plugin already handles those cases. They will only be sent as FakeNitro emojis/stickers if that expression is unavailable for some reason

EmirX3D:
@Vendicated Could it be because Ext. Stickers and Emojis are disabled in the server?
If so, much love

Vendicated:
yes if you don't have permission to use external emotes/stickers it will also send them as FakeNitro emoji (assuming you have embed perms)

People who don't want this behaviour can just turn off emoji/sticker spoofing in settings

EmirX3D:
Alright, thanks!

PR #4241, opened 2026-05-25
Update ShowHiddenChannels plugin

prixmzz:
1. updated some of the descriptions to make more sense
2. updated the api to use basetext instead of text


Vendicated:
please don't send meaningless refactor prs.

PR #4245, opened 2026-05-25
Fix PermissionsViewer modal scrolling

oasdnbv301:
Discord's modal scroller started handling the entire PermissionsViewer content, so the roles and permissions lists no longer scrolled independently.

This change sizes the modal content to the visible height of the modal scroller so the existing list scrollers can handle overflow again. The layout and panel widths stay unchanged.

Tested with:
- pnpm build


oasdnbv301:
looked at the feedback:
rerun height sync when the selected item container mounts
guard the scroller target before observing it
avoid redundant ResizeObserver writes
move scroller-node resolution into a helper

Rebuilt with pnpm build

MallyDev2 (commented):
Small edge case I noticed while reading through this: if the permissions array changes after a user has selected an item near the end, selectedItemIndex can point past the new array length. That would make selectedItem undefined even when permissions still has entries, so the modal could briefly render the empty state instead of clamping back to an available row. Might be worth resetting or clamping selectedItemIndex when permissions.length changes, especially now that the scroller height sync is tied to whether a selected item exists.

PR #4246, opened 2026-05-27
Experiments: improve settings UI

90gq29:
Discord writes shortcuts as keyboard keys one after another, like "CTRL K" with its own styling. The plugin description already uses this styling but it combines it with the plus symbol, like you do in text, e.g. CTRL+K. 

However, Discord doesn't use the plus sign when displaying keybinds, as the custom display buttons make this intuitively unnecessary, so this change just standardizes that display to be more in line with other Discord menus. See imgur link below for comparison.

https://imgur.com/a/kDybW16

Super minor thing, but I thought I'd point it out. Found in the description of the Experiments plugin.

### DIFF
--- a/src/plugins/experiments/index.tsx
+++ b/src/plugins/experiments/index.tsx
@@ -140,8 +140,8 @@ export default definePlugin({
                 <Paragraph size="md">
                     You can open Discord's DevTools via {" "}
                     <div className={KbdStyles.combo} style={{ display: "inline-flex" }}>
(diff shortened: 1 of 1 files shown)


MallyDev2 (commented):
One accessibility/readability thought: removing the visible '+' separators makes the shortcut read more like separate labels than a key combo in plain text or copied output. If the visual styling now handles spacing, it may still be worth adding an aria-label like 'Ctrl plus Alt plus O' on the combo wrapper so screen readers keep the shortcut meaning clear.

PR #4286, opened 2026-06-09
WikiLinks: New Plugin :shiggy:

patttterson:
## Describe your Changes
Plugin that transforms [[wiki]] and {{template}} links in your chat messages into actual hyperlinks, with the ability to use MediaWiki APIs to check the links existence.
>[!IMPORTANT]
> That feature may lead to an unexpected delay between hitting "Enter" and the message actually sending, so it is disabled by default.

For example, it turns [[Anubis (software)]] into [[[`Anubis (software)`]]](https://en.wikipedia.org/wiki/Anubis_(software)) and {{Citation needed}} into [{{`Citation needed`}}](https://en.wikipedia.org/wiki/Template:Citation_needed)

This is mainly inspired by https://github.com/stjohann/DiscordWikiBot

### DIFF
--- a/src/plugins/wikiLinks/index.tsx
+++ b/src/plugins/wikiLinks/index.tsx
@@ -1,0 +1,109 @@
+/*
+ * Vencord, a modification for Discord's desktop app
+ * Copyright (c) 2026 Vendicated and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


Vendicated:
Can mostly be done with TextReplace no?

patttterson:
Yeahh I thought about that, but with all the interwiki prefixes, template syntax, different languages, per-wiki capitalization rules, etc. I figured it was worth a whole plugin

Also regex is scary

Vendicated:
Yeah sorry Patty but this is too niche imo! I think for most people TextReplace would be enough, also I feel like any server where this is very relevant would already be using the bot you mentioned? Also if I were to use something like this I would want a wiki search with autocomplete instead of raw typing article names, which a bot could provide via slash command much better than a plugin ever could

Of course, you may still post it in our plugin channel though if you would like to

PR #4290, opened 2026-06-10
plugin: MoreAccounts

H7LT:
## Describe your Changes

removes the 5 account cap in the account switcher.

discord drops accounts past 5 every connect but keeps the token saved, so they vanish from the dropdown until you log in again. this patches both the trim and the "add an account" block.

also restores accounts that were already evicted: reads their saved tokens, validates against `/users/@me`, adds them back. expired ones get the red "please log in again". nothing is removed.

tested on canary, windows, 7 accounts.


Vendicated:
no

PR #4292, opened 2026-06-10
fix openPrivateChannel passes string instead of object to ChannelActionCreators

justjxke:
openPrivateChannel was passing a bare userId string to ChannelActionCreators.openPrivateChannel, but that function expects an object { recipientIds: string[], navigateToChannel?: boolean }. The string got destructured silently and recipientIds landed as undefined, the dm was never opened and it was just a noop with no error.

### DIFF
--- a/src/utils/discord.tsx
+++ b/src/utils/discord.tsx
@@ -90,7 +90,7 @@ export function getCurrentGuild(): Guild | undefined {
     return GuildStore.getGuild(getCurrentChannel()?.guild_id!);
 }
 
-export function openPrivateChannel(options: { recipientIds: string[]; navigateToChannel?: boolean; }) {
+export function openPrivateChannel(options: { recipientIds: string[]; joinCall?: boolean; joinCallVideo?: boolean; location?: string; onBeforeTransition?: () => void; navigateToChannel?: boolean; }) {
     ChannelActionCreators.openPrivateChannel(options);
 }
 
--- a/src/utils/discord.tsx
+++ b/src/utils/discord.tsx
@@ -90,8 +90,8 @@ export function getCurrentGuild(): Guild | undefined {
     return GuildStore.getGuild(getCurrentChannel()?.guild_id!);
 }
 
-export function openPrivateChannel(userId: string) {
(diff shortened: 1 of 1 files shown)


src/utils/discord.tsx
```diff
@@ -90,8 +90,8 @@ export function getCurrentGuild(): Guild | undefined {
     return GuildStore.getGuild(getCurrentChannel()?.guild_id!);
 }
 
-export function openPrivateChannel(userId: string) {
-    ChannelActionCreators.openPrivateChannel(userId);
+export function openPrivateChannel(options: { recipientIds: string[]; navigateToChannel?: boolean; }) {
+    ChannelActionCreators.openPrivateChannel(options);
```
sadan4, line 94: any reason to not type the full props?

justjxke: honestly didnt think to do it, my fault

src/utils/discord.tsx
```diff
@@ -90,8 +90,8 @@ export function getCurrentGuild(): Guild | undefined {
     return GuildStore.getGuild(getCurrentChannel()?.guild_id!);
 }
 
-export function openPrivateChannel(userId: string) {
-    ChannelActionCreators.openPrivateChannel(userId);
+export function openPrivateChannel(options: { recipientIds: string[]; joinCall?: boolean; joinCallVideo?: boolean; location?: string; onBeforeTransition?: () => void; navigateToChannel?: boolean; }) {
```
Vendicated, line 93: avoid this inline type, just update ChannelActionCreators type and import

Vendicated:
tbh this function is useless and not even used so it should just be removed instead

justjxke:
> tbh this function is useless and not even used so it should just be removed instead

the only reason i found this issue is because i needed to use it and it was just creating a groupchat with myself. i personally think you should keep it.

Vendicated:
just use ChannelActionCreators.openPrivateChannel directly xD

justjxke:
lmk if you want to remove it and i'll close the PR, or if you prefer to keep it i'll resolve your earlier comment and keep it inline

PR #4314, opened 2026-06-17
Added a vectimized plugin

jsanotheruser:
Vectimized applies optimization to Electron part of Vencord.

It's can be useful for low-end PC, that barely can run Vencord because of runtime.


Vendicated:
This is either the sloppiest aislop that has ever slopped in human history or you have great creativity

PR #4316, opened 2026-06-17
Call Indicator

commandblocks0:
* I added a call indicator plugin
* Shows indicator on DMs with active calls
* Setting to customize the indicator color


Vendicated:
too niche

PR #4317, opened 2026-06-17
YoutubeAdblock: Fix error causing memory leak

lolsuffocate:
## Describe your Changes
Scopes the adguard variables into a block statement to fix the `Uncaught SyntaxError: Identifier 'hiddenCSS' has already been declared` error that causes Sentry to relog a very long error string recursively with each video you play, until Discord eventually crashes and leaves you with hundreds of MB of logs.

image


Vendicated:
Meh this is kind of a bad solution

Instead of fixing the root problem (script is injected multiple times), it fixes the symptom

lolsuffocate:
Undone the scoping and switched it to fix the double injection, seems that every youtube embed loads an inner frame that then retriggers the parent.url condition in native.ts and injects again. I haven't had any issues after just removing the parent url check.

Vendicated:
hmm but the parent was for the Youtube Watch Together activity, does that still work with this change?

lolsuffocate:
You're right, both seem to be necessary in that case so I think just adding a global flag in the frame context is probably best. Have tested and works in message embeds and Watch Together fine

Vendicated:
Did you forget to commit your changes to the other file? 0 files changed lmao

Vendicated:
I'm curious what will happen if I merge it, will it just make an empty commit :sob:

lolsuffocate:
Embarrassment of a PR ngl

src/plugins/youtubeAdblock.desktop/adguard.js
```diff
@@ -18,199 +18,202 @@
  * You should have received a copy of the GNU General Public License
  * along with AdGuard's Block YouTube Ads.  If not, see <http://www.gnu.org/licenses/>.
  */
+if(!window.adguardInjected) {
+    window.adguardInjected = true;
```
Vendicated: Would be better to make this an iife
```suggestion
(() => {
	if (window.adguardInjected) return;
	window.adguardInjected = true;
```

src/plugins/youtubeAdblock.desktop/adguard.js
```diff
@@ -18,199 +18,202 @@
...
+    hideElements();
+    // Some changes should be re-evaluated on every page change
     hideDynamicAds();
     autoSkipAds();
-});
+    observeDomChanges(() => {
+        hideDynamicAds();
+        autoSkipAds();
+    });
+}
```
Vendicated:
```suggestion
})();
```

lolsuffocate:
Done

PR #4325, opened 2026-06-22
feat: add ToS safe splitLargeMessages plugin

AfoxDesignz:
## TOS safe splitLargeMessages plugin

Adds a frequently requested plugin that splits large messages into multiple chunks if they exceed discord's char limit. To comply with vencord's plugin guidelines this plugin does not send messages automatically.

I'm aware of the issues discussed at https://github.com/Vencord/plugin-requests/issues/77 and that similar requests and prs have been made, including those that abused the API.

Therefore, I've chosen a strictly safe approach. No messages are sent automatically, as user interaction is always required.

### Functionality

The plugin intercepts the message, splits it into chunks, and allows the first chunk to be sent through the users initial action. It then waits for a configurable delay before automatically prefilling the chat input box with the next chunk. The user then has to manually hit `Enter` to send each subsequent chunk. Technically, it's identical to the user splitting the message themselves and pasting the pieces one by one, just much more convenient. Also works with codeblocks.

#### Credits

Text splitting and codeblock repair logic based on a js implementation by [Mirco  Wittrien](https://github.com/mwittrien).

### DIFF
--- a/src/plugins/splitLargeMessages/README.md
+++ b/src/plugins/splitLargeMessages/README.md
@@ -1,0 +1,5 @@
+# Split Large Messages
+
+Splits long messages that exceed the character limit and automatically queues the next part in your chat box after sending.
+
+![](https://github.com/user-attachments/assets/80435572-302c-4780-ab81-16df3e449f97)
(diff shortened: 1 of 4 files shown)


AfoxDesignz:
@Vendicated I'd appreciate a short feedback on why this was closed.
There is demand for this; code strictly follows the guidelines; solution is tos friendly; no ai.

PR #4334, opened 2026-06-24
MessageEventsApi: fix missing props on options object

sadan4:
discord changed the object we grab so it no longer has the sickers and attachments properties. as a result we need to obtain these properties from other sources.

the current fix is poor, but working. alas, we can't use arguments[0] as this is an arrow function.

### DIFF
--- a/src/plugins/_api/messageEvents.ts
+++ b/src/plugins/_api/messageEvents.ts
@@ -38,9 +38,9 @@ export default definePlugin({
             find: ".handleSendMessage,onResize:",
             replacement: {
                 // https://regex101.com/r/7iswuk/1
-                match: /let (\i)=\i\.\i\.parse\((\i),.+?\.getSendMessageOptions\(\{.+?\}\)?;(?=.+?(\i)\.flags=)(?<=\)\(({.+?})\)\.then.+?)/,
-                replace: (m, parsedMessage, channel, replyOptions, extra) => m +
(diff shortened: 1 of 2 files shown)


src/plugins/_api/messageEvents.ts
```diff
@@ -38,9 +38,9 @@ export default definePlugin({
             find: ".handleSendMessage,onResize:",
             replacement: {
                 // https://regex101.com/r/7iswuk/1
-                match: /let (\i)=\i\.\i\.parse\((\i),.+?\.getSendMessageOptions\(\{.+?\}\)?;(?=.+?(\i)\.flags=)(?<=\)\(({.+?})\)\.then.+?)/,
-                replace: (m, parsedMessage, channel, replyOptions, extra) => m +
-                    `if(await Vencord.Api.MessageEvents._handlePreSend(${channel}.id,${parsedMessage},${extra},${replyOptions}))` +
+                match: /let (\i)=\i\.\i\.parse\((\i),.+?\.getSendMessageOptions\((\{.+?\})\),.{0,100}?\};(?=.+?(\i)\.flags=)(?<=\)\(({.+?})\)\.then.+?)/,
```
sadan4, line 41: we could pull out the individual arguments we need here; however, that would make the patch even more complex.

Vendicated:
Not really a fan of this, because options really isn't options, it's (pretty useless) props. And options2 is not a nice name xD

I did it via https://github.com/Vendicated/Vencord/commit/590522a40db414ee428e69bfa68879c6abbb0b36

PR #4335, opened 2026-06-24
Experiments: Allow linking experiments by negative index and label.

EtorixDev:
Allow linking experiments by negative index, which mainly pertains to -1 for Not Eligible, and their labels, such as "noteligible" or "variant1".

image

### DIFF
--- a/src/plugins/experiments/index.tsx
+++ b/src/plugins/experiments/index.tsx
@@ -127,8 +127,32 @@ export default definePlugin({
                 replace: "$&if($1==null)return;"
             }
         },
-
+        {
+            // Expands the experiment regex to allow negative numbers as well as text in the last segment of the URL.
+            find: '"^dev://experiment/',
+            replacement: {
+                match: /(\[0-9\]\+)/,
+                replace: "[a-zA-Z0-9-]+"
+            }
+        },
+        {
+            // Allow linking experiments by their label as well as their value.
+            find: ".EXPERIMENT_TREATMENT&&null",
+            replacement: [
+                {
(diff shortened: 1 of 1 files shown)


Vendicated:
isnt -1 the same as 0?

src/plugins/experiments/index.tsx
```diff
@@ -127,9 +127,33 @@ export default definePlugin({
...
+        },
     ],
 
+    matchExperiment(url: string, label: string): boolean {
+        const items = url.split("/");
+        const labelCleaned = label.replace(/[^a-zA-Z0-9]+/g, "").toLowerCase();
+        const urlEndCleaned = items[items.length - 1]?.replace(/[^a-zA-Z0-9]+/g, "").toLowerCase();
+        return !!labelCleaned && urlEndCleaned !== undefined && labelCleaned === urlEndCleaned;
+    },
+
```
Vendicated:
```suggestion
```

src/plugins/experiments/index.tsx
```diff
@@ -127,7 +127,14 @@ export default definePlugin({
                 replace: "$&if($1==null)return;"
             }
         },
-
+        {
+            // Expands the experiment regex to allow negative numbers as well as text in the last segment of the URL.
+            find: '"^dev://experiment/',
+            replacement: {
+                match: /(\[0-9\]\+)/,
+                replace: "[a-zA-Z0-9-]+"
```
Vendicated, line 135: this isn't very safe, the match should be more specific to make sure we don't change the wrong regex

PR #4342, opened 2026-06-26
MessageLogger: fix patches

sadan4:


sadan4:
as far as i can tell, the other patches seem vestigial

https://github.com/Vendicated/Vencord/blob/6eb22070407713b9d0dcfc5ee24861db71816183/src/plugins/messageLogger/index.tsx#L422-L440

above has been silently broken for over a year (i compared it with year old copies of that file with the patch applied).it references arguments[1], which doesn't exist. typo for arguments[0]?

https://github.com/Vendicated/Vencord/blob/6eb22070407713b9d0dcfc5ee24861db71816183/src/plugins/messageLogger/index.tsx#L441-L447

above was made redundant with the recent discord change.

sadan4:
on stable

sadan4:
support for partially deleted attachments should be re-implemented at some point (either in this PR or a follow-up)

src/plugins/messageLogger/index.tsx
```diff
@@ -452,9 +461,16 @@ export default definePlugin({
...
+                {
+                    // we can't use arguments[0] because we patch a nested **non-arrow** function
+                    match: /\.SPOILER,(?=\[\i\.\i\]:)(?<=item:(\i),.{0,200}?)/,
+                    replace: '$&"messagelogger-deleted-attachment": $1?.originalItem?.deleted,'
+                },
+                // dont allow deleting attachments from deleted messages
                 {
-                    match: /\.SPOILER,(?=\[\i\.\i\]:)/,
-                    replace: '$&"messagelogger-deleted-attachment":arguments[0]?.item?.originalItem?.deleted,'
+                    match: /(?<=function \i\(\i\)\{let\{[^}]*?item:(\i),\i:\i,)canRemoveItem:(\i)(?=,onRemoveItem:)/,
```
Nuckyz: Is it needed to match the function definition here or just the let is enough?

Nuckyz (approved):

PR #4343, opened 2026-06-26
PermissionsViewer, CharacterCounter: fix bugs

gnustella-lab:
## Describe your Changes

PermissionsViewer was showing every unset role permission bit as denied. Those should just be hidden.

Also copy arrays before sorting so we don't mutate props, and actually use the return value of sortUserRoles.

CharacterCounter was counting selected text from anywhere on the page. Only count selection inside the chat editor.


Vendicated:
> CharacterCounter was counting selected text from anywhere on the page. Only count selection inside the chat editor.

Can you show an example? I can't reproduce
image

PR #4348, opened 2026-06-28
feat(showHiddenChannels): optionally hide hidden text channels

NightmareSan7:
## Describe your Changes
This adds an option to hide hidden text channels from the channel list.

This is useful for users who mainly use the plugin to see hidden voice channels (like me :D) and who do not want the channel list to be cluttered by hidden text channels.


src/plugins/showHiddenChannels/index.tsx
```diff
@@ -48,6 +50,12 @@ export const settings = definePluginSettings({
         default: true,
         restartNeeded: true
     },
+    hideHiddenTextChannels: {
+        description: "Prevent hidden text channels from being shown",
```
Vendicated: I think this option will be very confusing to the average user

You could change it to a select option:
#### Channel Types to show
- All Channels
- Text Channels
- Voice Channels

I think that would be less confusing

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -25,14 +25,16 @@ import { classNameFactory } from "@utils/css";
...
+import { ChannelType, type ChannelTypesSets as ChannelTypesSetsType } from "@vencord/discord-types/enums/channel";
+import { findByPropsLazy, findCssClassesLazy } from "@webpack";
 import { ChannelStore, PermissionsBits, PermissionStore, Tooltip } from "@webpack/common";
 
 import HiddenChannelLockScreen, { setChannelBeginHeader } from "./components/HiddenChannelLockScreen";
 
 export const cl = classNameFactory("vc-shc-");
 
 const ChannelListClasses = findCssClassesLazy("modeSelected", "modeMuted", "unread", "icon");
+const ChannelTypesSets = findByPropsLazy("GUILD_TEXT_ONLY") as ChannelTypesSetsType;
```
Vendicated: It's in commons

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -83,7 +91,7 @@ export default definePlugin({
                 // Remove the special logic for channels we don't have access to
                 {
                     match: /if\(!\i\.\i\.can\(\i\.\i\.VIEW_CHANNEL.+?{if\(this\.id===\i\).+?threadIds:\[\]}}/,
-                    replace: ""
+                    replace: m => `if($self.shouldHideHiddenTextChannel(this.record)){${m}}`
```
Vendicated:
```suggestion
                    replace: "if($self.shouldHideHiddenTextChannel(this.record)){$&}"
```

src/plugins/showHiddenChannels/index.tsx
```diff
@@ -514,6 +522,12 @@ export default definePlugin({
         }
     },
 
+    shouldHideHiddenTextChannel(channel: Channel) {
+        return settings.store.hideHiddenTextChannels
```
Vendicated: this needs a try catch to avoid crashes if the webpack find for ChannelTypesSets ever breaks or ChannelTypesSets.GUILD_TEXT_ONLY is renamed to something else

src/plugins/showHiddenChannels/index.tsx
NightmareSan7: I can't seem to find ChannelTypeSets in common. Should i move this to utils.ts similar to MessageTypeSets and then import it from webpack/common?

Vendicated: Oh, right, I confused it with MessageTypeSets

Also yeah I guess

src/plugins/showHiddenChannels/index.tsx
NightmareSan7: That's a fair point, i implemented it. Had to extend the patch for this otherwise it would show unsubscribed voice channels when not using "show all channels" with "Text Channels" selected.

PR #4361, opened 2026-07-01
feat: TempIgnore plugin - timed user ignore with auto-expiry

senkuthegreat:
Adds a plugin to temporarily ignore users for a set duration, hiding their messages, DMs, status, voice presence, and friends list entries.


Vendicated:
more aislop OwO

PR #4368, opened 2026-07-03
SmartPaste plugin

Luckfiel:
## Describe your Changes
SmartPaste is a plugin that attempts to auto detect the most common programming languages and then wraps the pasted text in a fenced Markdown code block.

1. The plugin can easily be toggled on/off via a button in the message input box.
 
2. Auto Detect Language can be turned off and the language can be manually specified each time, or approved after a language is detected and Ask Before Wrapping is enabled. The manually selected language is remembered.

3. Always Use Plaintext can be specified if the user wishes to ignore the remembered language and always select Plain Text as the language by default. Auto Detect Language will still recommend the language the text should be wrapped in, but will not force it.

4. Minimum Lines and Minimum Characters can be specified to only wrap text if both of these conditions are met.

5. Finally, there is an option to ignore any pasted text that is a singular line long.

The two icons in the message input box were not made by me. The language detection was partially assisted by AI, as it is quite a bit of code to try and get it accurate enough.

image

image

- [ x ] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file and made sure this pull request complies with it
- [ x ] This pull request was written by me, and not an AI agent

### DIFF
--- a/src/plugins/smartPaste/README.md
+++ b/src/plugins/smartPaste/README.md
@@ -1,0 +1,1 @@
+Attempts to detect the pasted programming language and automatically wrap the pasted text in a fenced Markdown code block.
--- a/src/plugins/smartPaste/README.md
+++ b/src/plugins/smartPaste/README.md
@@ -1,1 +1,0 @@
-Attempts to detect the pasted programming language and automatically wrap the pasted text in a fenced Markdown code block.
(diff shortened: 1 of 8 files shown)


Vendicated:
Thank you for your contribution :heart: But this is too niche, sorry. Someone pasting code (especially anyone who'd enable this plugin) would know about codeblocks and have little use for this plugin considering it takes like 2 seconds to type a codeblock

PR #4385, opened 2026-07-11
feat(plugins): SoundboardUnlocker

Adversing:
## Describe your Changes
This plugin allows you to use soundboards without having a Nitro subscription. 
The workaround is simple: Discord asks 'Can this user use soundboard sounds everywhere?' and I force the answer to always be positive; then when the Discord client checks whether the soundboard sound is locked behind Nitro, I set the value to false.

Known limitations:
1) If Discord ever moves the nitro subscription check for this feature on their end, this plugin won't work anymore.
2) You must have the permission to use the soundboard feature in the voice channel you're into.
3) The plugin could also stop working if Discord simply renames or restructures the client code.

### DIFF
--- a/src/plugins/soundboardUnlocker/README.md
+++ b/src/plugins/soundboardUnlocker/README.md
@@ -1,0 +1,5 @@
+# Soundboard Unlocker
+
+A Vencord user plugin that makes soundboard sounds from your other guilds selectable without Nitro.
+
+The plugin patches Discord's client-side Soundboard Everywhere entitlement and removes the Nitro-locked marker from external guild sections. It does not grant guild permissions, therefore you must still be able to view the source guild, join the destination voice channel, and have Use External Sounds permission there.
--- a/src/plugins/soundboardUnlocker/index.ts
+++ b/src/plugins/soundboardUnlocker/index.ts
@@ -1,0 +1,31 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2026 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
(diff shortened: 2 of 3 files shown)


Vendicated:
If this actually works, it should just be added to FakeNitro plugin

meowpurrpur:
I tested this plugin, the sound does play on the client side but no body else in the VC can hear it

Adversing:
> I tested this plugin, the sound does play on the client side but no body else in the VC can hear it

I'm currently investigating on this, thanks for the report.

Vendicated:
Yeah that's what I thought, afaik it's not possible lol.

Whenever you use soundboard it actually does two things:
1. Play sound locally
2. Send API request to play for other people

First one is totally bypassable so that's probably where the confusion came from

meowpurrpur:
> > I tested this plugin, the sound does play on the client side but no body else in the VC can hear it
> 
> I'm currently investigating on this, thanks for the report.

You wont be able to find a workaround as the restriction is server-sided. When sending a request to `send-soundboard-sound` without Nitro you will just get this response:
```json
{"message": "This action requires a premium subscription.", "code": 20015}
```

I think realistically the only way you could do it is maybe by "injecting" the sound into the audio stream but im not really sure how that would work.

GodTierCutie:
> I think realistically the only way you could do it is maybe by "injecting" the sound into the audio stream but im not really sure how that would work.

or you can make a plugin that hook into discord’s voice stream and uses the web audio API to mix your mic input with local audio files, playing them directly into the voice channel. Tried it, but only works on Vesktop.

Vendicated:
It's indeed very easy on Vesktop/web but likely challenging or straight up impossible on desktop

Gonna close this because it doesn't work and this is fairly niche anyway, if there's a simple solution for this it can be added to FakeNitro (send pr). If it's more complex then I'd rather just not have it

PR #4387, opened 2026-07-12
Fix StickerPaste/FakeNitro conflict – #3691

### LINKED ISSUE #3691: Collison bewteen FakeNitro and StickerPaste
### Content

https://canary.discord.com/channels/1015060230222131221/1015063227299811479/1421692189930422373

https://github.com/user-attachments/assets/0d948bed-9c91-4c31-9cfc-61b2f67e9cb3

i did a terrible job of describing the issue in the video
you couldn't watch that and reproduce that
the thing is that the sticker is no longer accessible because the server fell below the boosting tier
but fake nitro bypasses that
and that's what cuases the bug (afaik)

### Request Agreement

meowpurrpur:
## Describe your Changes

Fixes https://github.com/Vendicated/Vencord/issues/3691

StickerPaste already handles sending stickers later, so FakeNitro shouldn't call `promptToUpload` and upload the generated GIF a second time. Pasted stickers now get attached to the current send options instead. I have no idea if there's a better way to do this but it doesn't seem like it as I tried a few different fixes.

### DIFF
--- a/src/api/MessageEvents.ts
+++ b/src/api/MessageEvents.ts
@@ -17,7 +17,7 @@
 */
 
 import { Logger } from "@utils/Logger";
-import type { Channel, CustomEmoji, Message } from "@vencord/discord-types";
+import type { Channel, CloudUpload, CustomEmoji, Message } from "@vencord/discord-types";
 import { MessageStore } from "@webpack/common";
 import type { Promisable } from "type-fest";
 
@@ -38,6 +38,7 @@ export interface SendMessageOptions {
     };
     location: string;
     stickerIds?: string[];
+    attachmentsToUpload?: CloudUpload[];
     alsoForwardToChannelId?: string;
 
     // If you end up using these, update their type
--- a/src/api/MessageEvents.ts
+++ b/src/api/MessageEvents.ts
@@ -17,7 +17,7 @@
 */
 
 import { Logger } from "@utils/Logger";
-import type { Channel, CloudUpload, CustomEmoji, Message } from "@vencord/discord-types";
+import type { Channel, CustomEmoji, Message } from "@vencord/discord-types";
 import { MessageStore } from "@webpack/common";
 import type { Promisable } from "type-fest";
 
@@ -38,7 +38,6 @@ export interface SendMessageOptions {
     };
     location: string;
     stickerIds?: string[];
-    attachmentsToUpload?: CloudUpload[];
     alsoForwardToChannelId?: string;
 
     // If you end up using these, update their type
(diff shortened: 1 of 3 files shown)


src/api/MessageEvents.ts
```diff
@@ -38,6 +38,7 @@ export interface SendMessageOptions {
     };
     location: string;
     stickerIds?: string[];
+    attachmentsToUpload?: CloudUpload[];
```
sadan4: This feels wrong to me. 

Image

sadan4 (commented):
This code is incorrect and breaks if the user uploads a sticker and a attachment at the same time. The sticker is omitted.

src/api/MessageEvents.ts
sadan4: more context 
image

meowpurrpur: image


I dont really get whats wrong about it, it is a thing

sadan4: here is how you should access and mutate uploads
image

meowpurrpur: I did try using that, but it caused the same bug that I was trying to fix lol, I'll try using it again though

meowpurrpur: alright, that did end up working and it also fixes the issue where it will get rid of the sticker when other attachments are added

sadan4 (commented):
Not sure if this behaviour is intentional or not. 

If sticker paste is enabled & the user has nitro & the sticker is locked because the server lost the nitro boost, and the user selects the sticker, it will be shown as an attached sticker even though it will be sent as an apng/gif.

sadan4:
niche edge case. other than that LGTM

sadan4:
Not sure how it interacts when trying to upload more than one sticker because that behaviour is currently broken in dev.

meowpurrpur:
> Not sure if this behaviour is intentional or not.
> 
> If sticker paste is enabled & the user has nitro & the sticker is locked because the server lost the nitro boost, and the user selects the sticker, it will be shown as an attached sticker even though it will be sent as an apng/gif.

I would try and fix this but I dont have any locked stickers to test it with, sorry :sob:

meowpurrpur:
> Not sure if this behaviour is intentional or not.
> 
> If sticker paste is enabled & the user has nitro & the sticker is locked because the server lost the nitro boost, and the user selects the sticker, it will be shown as an attached sticker even though it will be sent as an apng/gif.

i think this behavior should be intentional as "fixing" this will result in an inconsistency between sending  static & animated stickers

(just putting this here "for the record", I know i already said in the discord)

meowpurrpur:
please merge :3

sadan4 (commented):
looks good

PR #4388, opened 2026-07-12
small type improvments

sadan4:


packages/discord-types/enums/misc.ts
```diff
@@ -3,7 +3,7 @@ export const enum CloudUploadPlatform {
     WEB = 1,
 }
 
-export const enum DraftType {
+export declare enum DraftType {
```
Vendicated: no

sadan4: As far as I'm aware, the alternative is some typescript fuckery, I'll do that instead.

Vendicated: tbh the entire purpose of that file is to import enums from there to use them

I don't remember why we made DraftType a webpack common, we could just change that and use our enum instead

sadan4: We can't have it be a const enum because we iterate over it in at least crash handler 
https://github.com/Vendicated/Vencord/blob/main/src%2Fplugins%2FcrashHandler%2Findex.ts#L144-L150

sadan4: it now uses the type system instead of changing the enum to be declare

PR #4390, opened 2026-07-13
WebScreenShareFixes: fix Discord bug causing high CPU usage

DBosley:
## Describe your Changes
While streaming my full monitor today, I noticed that my CPU usage was increasing over time. I ran some tests which showed it increasing CPU usage across all cores by ~55-65 every 300 seconds. Restarting the share fixed it for a while, and shutting off the stream preview made it better, but didn't resolve it completely.

I was able to determine that the discord stream preview uploader grabs a thumbnail every 5 minutes, lining up with the CPU increase. The issue is from a buggy line of minified code:
`.finally(()=>{a.ontimeupdate=null,a.removeAttribute("srcObject"),a.load()})`
Which should be removing a javascript property with `removeAttribute` but that is not the correct way to do it. `removeAttribute` searches the element's attribute map which doesn't have the `srcObject` on it, and does nothing. So instead we need to set it to null manually, which then removes the property, which stops the element from continuing to render the video which was used to capture the screenshot. Essentially, over time, it would be rendering a new instance of the streamed video in the background and wouldn't shut them down.


wundervrc:
Tested this overnight and it fixed the problem for me.                                                                                                                     
                                                                                                                                                                                
     Setup: Vesktop 1.6.5, CachyOS (Arch), KDE Plasma on Wayland, NVIDIA RTX 4070 Ti, Ryzen 7 5800X. Kimi K3 applied the diff from this PR on top of latest main (4b9c27d), built it, and loaded it through the custom Vencord location setting. WebScreenShareFixes stayed enabled by default.                                                                    
                                                                                                                                                                                
     Before: during any long stream the renderer CPU grew in steps of roughly half a core every 5 minutes. After about 8 hours it was sitting at over 500% CPU with load average around 15 on 16 threads, and my audio and mouse were lagging until I restarted the share.                                                                                    
                                                                                                                                                                                
     After: streamed overnight, about 6 hours on one client instance. Load average is around 1.1, the renderer shows no more growth in steps, and I did not have to restart the stream once. System stayed smooth the whole time.                                                                                                                            
                                                                                                                                                                                
     Thanks for the fix, the 300 second cadence theory matched my machine exactly <3 :3

src/plugins/webScreenShareFixes.web/index.ts
```diff
@@ -27,6 +27,13 @@ export default definePlugin({
                     replace: '$&${$1?";stereo=1;sprop-stereo=1":""}'
                 },
             ]
+        },
+        {
+            find: "ApplicationStreamPreviewUploadManager",
+            replacement: {
+                match: /(\i)\.removeAttribute\("srcObject"\)/,
+                replace: "$1.pause(),$1.srcObject=null"
```
Nuckyz, line 35:
```suggestion
                replace: "($1.pause(),$1.srcObject=null)"
```

~~wrap it in a single expression so it works regardless if the bundler uses the comma operator or semicolons~~

Nuckyz (changes requested):

PR #4391, opened 2026-07-13
feat: Add SpotifyLyricsStatus plugin

Papa17217:
Adds the SpotifyLyricsStatus plugin, which automatically updates the user's custom status to show the current lyric line of the song playing on Spotify, fetched from LRCLIB.


Vendicated:
api abuse

PR #4393, opened 2026-07-14
Fix #4376: Clear stale platform indicators when users go offline

### LINKED ISSUE #4376: [Bug] Platform indicators sometimes show stale online status until Discord restart

After Discord has been running for several hours, the Platform Indicators plugin sometimes shows users as being online on specific platforms, such as desktop, mobile, or web, even though they are actually offline.

The shown platform status can remain stuck as online, idle, or dnd (whatever the user's previous status was). When checking the same user from another device, such as the Discord mobile app, the user appears offline correctly. After restarting the Discord desktop client, the platform indicators update correctly and the user is shown as offline.

This happens pretty much every day, tested on multiple devices and discord client versions (stable, canary, ptb, web), after an amount of time that is hard to pinpoint.

My guess would be, that stale platform presence data may remain cached in Discord's local presence/client status store and the plugin continues to display it.

Platform indicators should disappear or update correctly when a user goes offline.

If a user is offline, the plugin should not continue showing old platform indicators such as desktop, mobile, or web with an outdated online, idle, or dnd status.

Keep the client running for several hours.
Observe the member list.
At some point, a user may still appear as online/idle/dnd via the platform indicators.
Check the same user from another Discord client (mobile, ptb, web, canary).
The user appears offline there.
Restart the Discord desktop client.
The platform indicator disappears and the user is now shown as offline correctly.

Discord presence will also display the cached game activity status / rich presence and their timers will keep on going infinitely.
It is especially easy to catch with users who have their spotify status shared, since the song that was played when they go offline, will permanently display as playing with a finished playback bar. 

The issue does not occur with the plugin disabled.

### Errors

I have not been able to catch any errors so far. 

### Request Agreement

- [x] I've searched existing issues and this hasn't been reported before

wasteofwifi:
# Fix for issue #4376: stale platform indicators after a user goes offline

## Root cause

Discord's internal `PresenceStore` keeps two separate pieces of state:

- `statuses` / `getStatus(userId)` - the overall aggregate status (online, idle, dnd, offline)
- `clientStatuses` / `getClientStatus(userId)` - a per platform breakdown (desktop, mobile, web...)

When a user goes offline, Discord updates the overall `statuses` entry correctly, but the per platform `clientStatuses` entry is not reliably cleared. Since PlatformIndicators only reads from `getClientStatus`, it kept rendering the last known platform/status combo (and, for rich presence, the last known activity) until a full client restart forced the stores to rebuild from scratch. This matches the report: other clients (mobile, web) show the user as offline correctly because they only rely on the overall status, and restarting the desktop client fixes it because the stale in-memory store gets discarded.

## Fix

In both places the plugin reads `getClientStatus` (the profile badge and the member list / message indicator), first check `PresenceStore.getStatus(user.id)`. If the overall status is `"offline"`, treat the user as having no platform indicators at all, ignoring whatever stale data is left in `clientStatuses`. The member list / message indicator also needed this check added to its `useStateFromStores` selector so it actually re-renders and clears the indicator the moment the overall status flips to offline, instead of only being reevaluated at the next unrelated re-render.

## Files changed

- `src/plugins/platformIndicators/index.tsx`
  - `getBadges`: added an early return of `[]` when the user's overall status is offline
  - `PlatformIndicator`: the `useStateFromStores` selector now returns `null` (instead of the stale `clientStatus` object) when the user's overall status is offline

## Testing

- Verified the plugin still shows indicators normally for online/idle/dnd users across desktop, mobile, web, and embedded platforms
- Simulated a stale `clientStatuses` entry (an entry left in `PresenceStore`'s internal `clientStatuses` map for a user whose overall status was manually set to `offline`) and confirmed the indicator no longer renders in the member list, messages, or profile badges

### DIFF
--- a/src/plugins/platformIndicators/index.tsx
+++ b/src/plugins/platformIndicators/index.tsx
@@ -169,6 +169,10 @@ function getBadges({ userId }: BadgeUserArgs): ProfileBadge[] {
 
     ensureOwnStatus(user);
 
+    // Discord's clientStatuses store can retain stale per platform entries after a user
+    // goes offline, so only trust them while the user's overall status is not offline
+    if (PresenceStore.getStatus(user.id) === "offline") return [];
+
     const status = PresenceStore.getClientStatus(user.id);
     if (!status) return [];
 
@@ -191,7 +195,12 @@ function getBadges({ userId }: BadgeUserArgs): ProfileBadge[] {
 const PlatformIndicator = ({ user, small = false }: { user: User; small?: boolean; }) => {
     ensureOwnStatus(user);
 
-    const status = useStateFromStores([PresenceStore], () => PresenceStore.getClientStatus(user.id));
+    const status = useStateFromStores([PresenceStore], () => {
+        // Discord's clientStatuses store can retain stale per platform entries after a user
+        // goes offline, so only trust them while the user's overall status is not offline
+        if (PresenceStore.getStatus(user.id) === "offline") return null;
+        return PresenceStore.getClientStatus(user.id);
+    });
     if (!status) return null;
 
     const icons = Object.entries(status).map(([platform, status]) => (
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -646,6 +646,10 @@ export const Devs = /* #__PURE__*/ Object.freeze({
         name: "NightmareSan",
         id: 304239816466235392n
     },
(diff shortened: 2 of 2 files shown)


byeoon:
oh my god yes this NEEDS an update 🙏🙏🙏🙏🙏🙏🙏🙏🙏

PR #4396, opened 2026-07-15
Fix MemberCount causing high idle CPU and heap growth

### LINKED ISSUE #4394: [Bug] MemberCount causes severe idle CPU usage and continuous JS heap allocation

Enabling the MemberCount plugin causes Discord's renderer to consume substantially more CPU while completely idle.

In the same server and channel, with no user interaction:

- MemberCount enabled: approximately 3–8% sustained CPU usage, with frequent much larger spikes
- MemberCount disabled: approximately 0.25–1.5% CPU usage

The DevTools Performance Monitor also shows continuous JS heap growth while MemberCount is enabled. The heap repeatedly grows from roughly 300–500 MB toward 800 MB–1 GB before garbage collection reduces it again. With MemberCount disabled, the heap remains around 200–250 MB and is considerably more stable.

This appears to be continuous allocation churn rather than necessarily a retained memory leak: garbage collection reclaims much of the memory, but the repeated allocation and collection causes significant idle CPU usage.

I isolated MemberCount by testing my enabled plugins in groups of six, then testing the remaining plugins individually. MemberCount alone consistently reproduces the issue.

The problem occurs while idling inside a server. No scrolling, typing, voice activity, or other interaction is required.

I have attached two DevTools Performance Monitor screenshots:

1. MemberCount enabled

Image

3. MemberCount disabled

Image

Both were recorded under the same general idle conditions.

MemberCount should not cause continuous background CPU usage or sustained heap allocation while Discord is idle.

The plugin should only subscribe to and recompute the data required for the currently visible feature.

In particular:

- Server tooltip counts should only be queried while the user is hovering over a server icon and the tooltip is mounted.
- Member-list counts should only update when relevant guild, channel, member-list, or voice-state data actually changes.
- Flux selectors should ideally return stable primitive values such as the final count, rather than newly allocated objects or arrays that can trigger unnecessary rerenders.

Idle CPU and heap behaviour should remain close to the baseline observed with MemberCount disabled.

1. Install Vencord.
2. Use Discord Stable.
3. Enable the MemberCount plugin.
4. Enable the Member List option. I also had Tool Tip and Voice Activity enabled during testing.
5. Restart Discord.
6. Open any server and remain idle in a text channel.
7. Open DevTools with Ctrl+Shift+I.
8. Open More tools → Performance monitor.
9. Enable CPU usage and JS heap size.
10. Observe CPU usage and heap behaviour for approximately 1–2 minutes.
11. Disable MemberCount and restart Discord.
12. Repeat the same test in the same server and channel.

Observed result:

- With MemberCount enabled, idle CPU remains around 3–8% and the JS heap continuously grows until garbage collection occurs.
- With MemberCount disabled, idle CPU falls to approximately 0.25–1.5% and heap usage remains much lower and more stable.

### Errors

No relevant console errors are produced.

This is a performance issue rather than a crash.

### Request Agreement

- [x] I've searched existing issues and this hasn't been reported before

jaxx-devv:
## Describe your Changes

Fixes #4394

While testing MemberCount, I noticed my CPU usage was slowly increasing and the JS heap kept growing over time.

After looking into it, I found that MemberCount was causing unnecessary rerenders because `ChannelMemberStore.getProps()` and `ThreadMemberListStore.getMemberListSections()` were creating new objects/arrays every time they were called. Since `useStateFromStores` compares references, it thought the value changed even when the actual count stayed the same.

I changed the selectors to return the member count directly instead of the whole object/array, so it only updates when the count actually changes.

Also skipped the store calls for the tooltip version since it doesn't use those values anyway.

- [x] I have read the CONTRIBUTING.md file and made sure this pull request complies with it

### DIFF
--- a/src/plugins/memberCount/MemberCount.tsx
+++ b/src/plugins/memberCount/MemberCount.tsx
@@ -48,23 +48,36 @@ export function MemberCount({ isTooltip, tooltipGuildId }: { isTooltip?: true; t
         () => OnlineMemberCountStore.getCount(guildId)
     );
 
-    const { groups } = useStateFromStores(
+    const memberListOnlineCount = useStateFromStores(
         [ChannelMemberStore],
-        () => ChannelMemberStore.getProps(guildId, currentChannel?.id)
+        () => {
+            if (isTooltip) return null;
+
+            const { groups } = ChannelMemberStore.getProps(guildId, currentChannel?.id);
+            if (groups.length >= 1 || groups[0].id !== "unknown") {
+                return groups.reduce((total, curr) => total + (curr.id === "offline" ? 0 : curr.count), 0);
+            }
+
+            return null;
+        }
     );
 
-    const threadGroups = useStateFromStores(
+    const threadListOnlineCount = useStateFromStores(
         [ThreadMemberListStore],
-        () => ThreadMemberListStore.getMemberListSections(currentChannel?.id)
+        () => {
+            if (isTooltip) return null;
+
+            const threadGroups = ThreadMemberListStore.getMemberListSections(currentChannel?.id);
+            if (threadGroups && !isObjectEmpty(threadGroups)) {
+                return Object.values(threadGroups).reduce((total, curr) => total + (curr.sectionId === "offline" ? 0 : curr.userIds.length), 0);
+            }
+
+            return null;
+        }
     );
 
-    if (!isTooltip && (groups.length >= 1 || groups[0].id !== "unknown")) {
-        onlineCount = groups.reduce((total, curr) => total + (curr.id === "offline" ? 0 : curr.count), 0);
-    }
-
-    if (!isTooltip && threadGroups && !isObjectEmpty(threadGroups)) {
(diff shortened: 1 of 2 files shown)


Vendicated:
@sadan4 review

sadan4:
Why did you close this?

jaxx-devv:
issues already solved

chloricacid:
The plugin hasn't been touched for ~6 months, your pr is still welcome

jaxx-devv:
my mistake, remade the PR #4403

src/plugins/memberCount/MemberCount.tsx
sadan4, line 20:
```suggestion
    const currentChannel = useStateFromStores([SelectedChannelStore], () => isTooltip ? undefined : getCurrentChannel(), undefined, (a, b) => a?.id === b?.id);
```

sadan4:
While i can't measure performance (because it was never an major performance drag for me), this fixes the issue of MemberCount constantly re-rendering and should re-opened.

PR #4404, opened 2026-07-17
add autoDismissCalls plugin

commandblocks0:
* Adds to DM context menu a option for auto dismiss calls where you can pick in how many seconds after a call starts ringing it will be rejected
* In the plugin settings screen is a table of configured delays where you can also change the delay


Vendicated:
Really pointless, if someone is spam calling you just block them and find better friends

commandblocks0:
I think its not pointless currently there is no way to mute calls and also you notice a ringing call after few seconds if you in middle of a game it wont bother you when a call starts ringing just few seconds

PR #4412, opened 2026-07-18
MessageLogger: fix new attachments being dropped on edit

yuna0x0:
## Describe your Changes
fix new attachments added to the same message are dropped without pushing back into the `attachments`

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -216,7 +216,7 @@ export default definePlugin({
     name: "MessageLogger",
     description: "Temporarily logs deleted and edited messages.",
     tags: ["Chat", "Utility"],
-    authors: [Devs.rushii, Devs.Ven, Devs.AutumnVN, Devs.Nickyux, Devs.Kyuuhachi, Devs.sadan],
+    authors: [Devs.rushii, Devs.Ven, Devs.AutumnVN, Devs.Nickyux, Devs.Kyuuhachi, Devs.sadan, Devs.yuna0x0],
     dependencies: ["MessageUpdaterAPI"],
     settings,
     contextMenus: {
@@ -274,15 +274,12 @@ export default definePlugin({
         if (!newMessage.attachments?.length) {
             return oldMessage.attachments.map((a): MLAttachment => ({ ...a, deleted: true }));
         }
-        const attachments: MLAttachment[] = [];
-        for (const oldAttachment of oldMessage.attachments) {
-            const wasDeleted = newMessage.attachments.every(a => a.id !== oldAttachment.id);
-            if (wasDeleted) {
-                attachments.push({ ...oldAttachment, deleted: true });
-            } else {
-                attachments.push(oldAttachment);
-            }
-        }
+        const attachments = oldMessage.attachments
+            .map((oldAttachment): MLAttachment =>
+                newMessage.attachments.find(a => a.id === oldAttachment.id)
+                ?? { ...oldAttachment, deleted: true }
+            )
+            .concat(newMessage.attachments.filter(a => oldMessage.attachments.every(o => o.id !== a.id)));
         return attachments;
     },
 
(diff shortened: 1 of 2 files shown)


Vendicated:
Can you explain a bit better please? I don't really understand the issue

yuna0x0:
sure, so basically when a bot attaches a new attachment to the message, the message logger will only show the old attachment, where new attachments are dropped.

https://github.com/user-attachments/assets/f4e34ce6-4e28-4777-9b41-cbafa025b0f2

src/plugins/messageLogger/index.tsx
```diff
@@ -274,14 +274,12 @@ export default definePlugin({
...
-                attachments.push({ ...oldAttachment, deleted: true });
-            } else {
-                attachments.push(oldAttachment);
-            }
+        const attachments: MLAttachment[] = oldMessage.attachments
+            .filter(oldAttachment => newMessage.attachments.every(a => a.id !== oldAttachment.id))
+            .map((a): MLAttachment => ({ ...a, deleted: true }));
+        for (const newAttachment of newMessage.attachments) {
+            const oldAttachment = oldMessage.attachments.find(a => a.id === newAttachment.id);
+            attachments.push(oldAttachment ?? newAttachment);
```
Vendicated: why do we need to use oldAttachment? couldn't we just do 
```js
 const attachments: MLAttachment[] = oldMessage.attachments
            .filter(oldAttachment => newMessage.attachments.every(a => a.id !== oldAttachment.id))
            .map((a): MLAttachment => ({ ...a, deleted: true }))
			.concat(newMessage.attachments);
```

sadan4:
Oops, I didn't know that was possible when I wrote it.

src/plugins/messageLogger/index.tsx
sadan4: Would that preserve attachment order?

Vendicated: good point. neither solution preserves order and that should be fiixed

sadan4: Yeah that why I wrote it that way originally

Vendicated: should just be able to sort by attachment id

Vendicated: Actually I think it's possible to reorder attachments, so there's no reliable way to always preserve correct order. Probably most reasonable to use the new attachments and then intersperse deleted ones based on old order? i.e. right before where they were in the old attachments

sadan4: Is attachment ID based on the time the attachment is uploaded or the time the attachment is attached to the message. Are those even different things?

sadan4: > Actually I think it's possible to reorder attachments, so there's no reliable way to always preserve correct order. Probably most reasonable to use the new attachments and then intersperse deleted ones based on old order? i.e. right before where they were in the old attachments

I think whatever we do is fine as long as if the user clears the history it looks the same as if they just reloaded the client.

yuna0x0: > why do we need to use oldAttachment? couldn't we just do
> 
> ```js
>  const attachments: MLAttachment[] = oldMessage.attachments
>             .filter(oldAttachment => newMessage.attachments.every(a => a.id !== oldAttachment.id))
>             .map((a): MLAttachment => ({ ...a, deleted: true }))
> 			.concat(newMessage.attachments);
> ```

Yeah, this is simpler, I am going to stick with this solution. Though if we use oldAttachment, we are able to distinguish old and new attachments if needed, but since we aren't using this data, using new attachments should be sufficient.

Since V mentioned that it's possible to reorder attachments, I think we should just concat deleted messages with new attachments. Intersperse deleted ones based on old orders didn't make sense for me. And if the user clears the history, this approach looks the same as if they just reloaded the client.

sadan4: this is wrong, it re-orders deleted attachments in the common case where one is just deleted

sadan4: >I think whatever we do is fine as long as if the user clears the history it looks the same as if they just reloaded the client.

i mean that in the case where attachments are added to an existing message

Issue #4410, opened 2026-07-18
PinDMs custom pins causes some DMs not to load


Aware-ok:
### What happens when the bug or crash occurs?

When I turn on PinDMs, I have about 3 categories, sometimes when scrolling down the DMs list, some users DMs don't load, where there would be a user on your DM list instead it is just empty and unclickable.

For example imagine this, you have 70 DMs total, maybe 30 are in 3 categories, the top 2 categories load all the users, upon scrolling down, the DMs load, but it gets to a point where you're only seeing the top half of the whole list that SHOULD be in view, everything else is blank, upon scrolling further down, more DMs load of course, but it never loads ALL of them, this causes sometimes when a user is inbetween where it would load, that DM just becomes not clickable at all, you scroll down just far enough to where it loads and you can see it at near the top of your DM list, and then scrolling back up unloads the DMs that are at the bottom, the exact same DMs don't load again upon scrolling down, and those DMs only load upon scrolling far enough down to where it about near the top of the DM list. I hope that made sense. Even the unpinned DMs don't load (the ones that are not in a cetegory) Scrolling all the way to the bottom of the DM list, loads all the DMs, then upon scrolling up it just unloads the bottom half near the middle of your entire DM list.

### What is the expected behaviour?

Well all the DMs should just show up and be loaded, that is about it, or they shouldn't unload at least.

### How do you recreate this bug or crash?

I have a lot of DMs, so it may be because of that, but I have 3 or 4 categories of custom pins, with at least 7 users in each, but even the unpinned DMs don't load upon scrolling further down. 

### Errors

```
Replace this text with your crash-log.
```

### Request Agreement

- [x] I've searched existing issues and this hasn't been reported before
- [x] I am using Discord Stable or tried on Stable and this bug happens there as well
- [x] I installed Vencord via an official method (download from vencord.dev or from source), not via something like AUR or Nix packages
- [x] This is not a support request
- [ ] I am willing to submit a fix for this myself

PR #4414, opened 2026-07-18
Fix PinDMs virtualization breaking with large DM lists

jaxx-devv:
## Describe your Changes

Fix an incorrect chunk size calculation used by PinDMs when keeping pinned sections mounted in Discord's DM list.
Solves: #4410 

The chunk size was calculated by adding the number of pinned rows to the row height instead of converting the row count into pixels. With larger pinned sections, this significantly underestimated the size of the pinned area, causing Discord's virtualized list to unmount rows too aggressively. As a result, scrolling through large DM lists could produce blank, unclickable entries until they were scrolled back into view.

This changes the calculation to use the correct row height multiplication so the virtualized list keeps the expected range of items mounted.


Vendicated:
oh my god lmao

Vendicated:
Thanks for the fix! Since this is a one letter change it's unfortunately not substantial enough to be added as an author so I undid that change. But you can be added if another bigger contribution is merged

jaxx-devv:
I understand, im still happy to help 🥹

PR #4415, opened 2026-07-18
fix MutualGroupDMs

meowpurrpur:
## Describe your Changes

Fixes the broken patch that pushes the mutual group section to the user profile modal.


sadan4:
The new patch is bad. Not only does it use minified identifiers, it will also cause a syntax error if discord declares multiple things in a variable declaration

meowpurrpur:
> The new patch is bad. Not only does it use minified identifiers, it will also cause a syntax error if discord declares multiple things in a variable declaration

Ah okay will find another way to patch

meowpurrpur:
> The new patch is bad. Not only does it use minified identifiers, it will also cause a syntax error if discord declares multiple things in a variable declaration

Would this patch be more appropriate?
```
{
    match: /items:(\i),initialSection:\i,onClose:\i\}=\i.+?(?=return\(0,\i\.jsxs?\)\()/,
    replace: "$&$self.pushSection($1,arguments[0].user);"
},
```

PR #4421, opened 2026-07-19
MusicRRPC: add caching & custom listenbrainz instance support

Vendicated:
- Caches musicbrainz and coveractarchive api lookups for 15mins
- Adds support for self hosted instances
- If no valid cover is found, checks if the scrobble originates from a youtube video and if yes uses its thumbnail as cover art
 
image


Vendicated:
@grok @claude @gemini @codex any feedback?

PR #4423, opened 2026-07-19
Fix NewGuildSettings not applying when joining via browser invites

### LINKED ISSUE #4422: [Bug] (New Guild Settings): Does not work when joining servers via Disboard.

I've been joining a lot of Disboard.org servers, everytime I join one. The plugin always fails to apply the settings. 

Join a server via Disboard, have your client open and logged into an account. Let the browser be able to access your client. (I use Brave). When joining a server, a pop up should mention if Disboard/YourBrowser open Discord. And when joining, the plugin fails to put the settings.

Join a server via Disboard. (Ensure your client is running and open)
Accept the invite link, and allow your browser to open your Discord client aka communicate with it. 
Upon joining, you should see that the plugin has failed to input the settings. 

### Errors

```
Replace this text with your crash-log.
```

### Request Agreement

- [ ] I've searched existing issues and this hasn't been reported before

jaxx-devv:
## Describe your Changes

Fix `NewGuildSettings` not applying default settings when joining a guild through a browser opened invite (e.g. Disboard).
Fixes #4422

The plugin only handled specific join paths (`acceptInvite`, `joinGuild`, and membership screening). Browser hand off joins bypass those hooks, so `applyDefaultSettings` was never called.

Handle the `GUILD_JOIN` Flux event so defaults are applied regardless of how the guild was joined.

### DIFF
--- a/src/plugins/newGuildSettings/index.tsx
+++ b/src/plugins/newGuildSettings/index.tsx
@@ -144,6 +144,9 @@ export default definePlugin({
     settings,
     applyDefaultSettings,
     flux: {
+        GUILD_JOIN({ guildId }) {
+            applyDefaultSettings(guildId);
+        },
         GUILD_JOIN_REQUEST_UPDATE({ guildId, request, status }) {
             if (status === "APPROVED" && request.user_id === UserStore.getCurrentUser().id)
                 applyDefaultSettings(guildId);


Vendicated:
I don't think that's right. It doesn't handle cases like being a lurker. And flux events like this can do unexpected things

PR #4433, opened 2026-07-21
Fix newGuildSettings unable to work

MoriitoDev:
## Describe your Changes
Fix `NewGuildSettings` not applying any settings when joining a new guild through the client or browser from webs like disboard..

Use `GUILD_CREATE` event, so it handles all types of server invites. Then make a check in case the user is a lurker.


src/plugins/newGuildSettings/index.tsx
```diff
@@ -144,9 +144,16 @@ export default definePlugin({
     settings,
     applyDefaultSettings,
     flux: {
-        GUILD_JOIN_REQUEST_UPDATE({ guildId, request, status }) {
-            if (status === "APPROVED" && request.user_id === UserStore.getCurrentUser().id)
-                applyDefaultSettings(guildId);
+        GUILD_CREATE(event: any) {
```
Nuckyz, line 147: GUILD_CREATE is not only fired when a user joins a guild. This will incorrectly apply default settings

PR #4436, opened 2026-07-22
fix many bugs found by Claude

Vendicated:
Basically Claude Fable audited the codebase (thanks Zere) and found many bugs.

This PR fixes most of them which I found relevant enough to fix

### TODO:
- [x] test themes on web to ensure no regression
- [x] test React DevTools install still works
- [x] test themes & quickcss watchers still work

### Included Fixes:
- Webpack Patcher: incorrect patch error handling could lead to some functional patches not being applied 
- CommandsAPI: subcommands were impossible to unregister
- useAwaiter: on error, the value was set to null even if a fallbackValue was specified, violating the type contract
- Patcher: build check boolean logic was grouped incorrectly due to missing parantheses
- CustomCommands: commands weren't unregistered on stop, leading to error when enabling, disabling & re-renabling the plugin
- Notification Log: delete was by timestamp instead of ID and non-atomic
- FixSpotifyEmbeds: settings listener was registered inside browser-window-created event handler, leading to duplicate listeners
- CustomRPC: changing activity type or timestamp didn't update the RPC
- CopyStickerLinks: message patch had non-sense code, just removed it
- XSOverlay: UDP socket wasn't closed on stop
- VoiceMessages: Web recorder never closed the mic media stream, which continues showing the OS mic indicator
- PictureInPicture: Video clone wasn't properly cleaned up, which leaked memory
- Themes (Extension version): theme object urls were never cleaned up, which leaked memory on theme change
- Themes (Extension version): Upload Theme button hitbox was only the text, not the whole button. Also added Drag & Drop from file manager support
- Dearrow: toggling for one embed flipped the global dearrow by default setting
- Extension Installer: the zip extract function was vulnerable to path traversal, although we only ever used this for React DevTools obtained via https so it has low impact
- FileWatchers: ensure always properly cleaned up
- components/Link: mutated props which is a very bad practice

### DIFF
--- a/src/plugins/pictureInPicture/index.tsx
+++ b/src/plugins/pictureInPicture/index.tsx
@@ -57,16 +57,42 @@ export default definePlugin({
 
                             videoClone.loop = settings.store.loop;
                             videoClone.style.display = "none";
-                            videoClone.onleavepictureinpicture = () => videoClone.remove();
 
-                            function launchPiP() {
-                                videoClone.currentTime = video.currentTime;
-                                videoClone.requestPictureInPicture();
-                                video.pause();
-                                videoClone.play();
+                            let cleaned = false;
+                            function cleanup() {
+                                if (cleaned) return;
+                                cleaned = true;
+
+                                videoClone.onloadedmetadata = null;
+                                videoClone.onleavepictureinpicture = null;
+
+                                videoClone.pause();
+                                videoClone.removeAttribute("src");
+                                videoClone.load();
+                                videoClone.remove();
+
+                                // resume original if still in the document
+                                if (video.isConnected) {
+                                    video.currentTime = videoClone.currentTime;
+                                    video.play().catch(() => 0);
+                                }
                             }
 
-                            if (videoClone.readyState === 4 /* HAVE_ENOUGH_DATA */)
+                            videoClone.onleavepictureinpicture = cleanup;
+
+                            async function launchPiP() {
+                                try {
+                                    videoClone.currentTime = video.currentTime;
+                                    video.pause();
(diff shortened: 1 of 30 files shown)


src/plugins/previewMessage/index.tsx
```diff
@@ -44,6 +44,8 @@ const getAttachments = async (channelId: string) =>
         UploadAttachmentStore.getUploads(channelId, DraftType.ChannelMessage)
             .map(async (upload: CloudUpload) => {
                 const { isImage, filename, spoiler, item: { file } } = upload;
+
+                // FIXME: revoke object url to fix memory leak
```
Vendicated, line 49: TODO: this is a bit tricky because we need to revoke these once the ephemeral message is dismissed. Maybe detectable via MESSAGE_DELETE flux event, otherwise probably have to patch or find some better solution or maybe just auto revoke after like 10 minutes

src/plugins/pictureInPicture/index.tsx
```diff
@@ -57,16 +57,42 @@ export default definePlugin({
...
+                                videoClone.onleavepictureinpicture = null;
+
+                                videoClone.pause();
+                                videoClone.removeAttribute("src");
+                                videoClone.load();
+                                videoClone.remove();
+
+                                // resume original if still in the document
+                                if (video.isConnected) {
+                                    video.currentTime = videoClone.currentTime;
```
Vendicated: fixme: doesnt work, probably because it needs to be higher above

Vendicated:
Also the branch name is kinda a misnomer, these aren't claude's fixes they're my fixes to claude's feeback

PR #4439, opened 2026-07-23
feat(readAllNotificationsButton): add Read All DMs button

tekkenkkk:
## Describe your Changes

Adds a button to Read All DMs to the existing readAllNotificationsButton plugin, along with options to display both buttons, only the Guilds (Servers) button, or only the DMs button.

### DIFF
--- a/src/plugins/readAllNotificationsButton/index.tsx
+++ b/src/plugins/readAllNotificationsButton/index.tsx
@@ -90,9 +90,9 @@ const settings = definePluginSettings({
         description: "Which buttons to show",
         type: OptionType.SELECT,
         options: [
-            { label: "Both", value: 0, default: true },
-            { label: "Only Guilds (Servers)", value: 1 },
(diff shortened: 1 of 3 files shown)


Nuckyz:
What about a context menu option instead? And a setting to just make Read All include dms?

fucktekken:
> What about a context menu option instead? And a setting to just make Read All include dms?

So just leave "Read All" button as it was but instead of reading all servers notifications open context menu instead where u can choose what to read 1. Both 2. Only Servers 3. Only DMs?

Nuckyz:
> > What about a context menu option instead? And a setting to just make Read All include dms?
> 
> So just leave "Read All" button as it was but instead of reading all servers notifications open context menu instead where u can choose what to read 1. Both 2. Only Servers 3. Only DMs?

Sorry for the late reply. Yes, but you can also add a context menu to the button to separately allow to just read DMs or servers regardless of the setting

PR #2091, opened 2024-01-04
feat(plugin): FavoriteAnything

MrDiamondDog:
Cause of discord's silly code I had to change how the favorite gifs menu is rendered and re-make parts of how the favorite accessory works

PR #4444, opened 2026-07-24
New plugin: FavouriteAnything

Davr1:
This plugin allows you to add common media and file types to your favourites (including images, videos, audio files, text files, content within rich embeds, external youtube/tiktok embeds, cv2 file/media components, clips, etc), and is compatible with the vanilla client, so anything you favourite will be displayed in the gif picker as a regular image/video (if possible) or a simple thumbnail with the proper source url.

*Favorite button is displayed either on the left for basic attachments, or on the right grouped with other actions*
image
*Favourite files live in a separate tab, and are fully interactive. Plugin is made to be extensible so adding another type of favouritable item would take ~50 lines tops, however the split between "media" and "files" already covers most use cases*
image
*What custom non-media items look like with the plugin disabled. Clicking on the thumbnail will send the original attachment url. Attachments are still searchable given the fact that the filename is used in the thumbnail url itself*
image

This plugin has been in Equicord (and vencor #unofficial-plugins) for a while now so most edge cases are already handled correctly.
Supersedes #2091.


src/plugins/favouriteAnything/components.tsx
```diff
@@ -0,0 +1,363 @@
...
+import { AttachmentContextProviderProps, AttachmentItem, AttachmentsComponentProps, CustomItemFormat, FavoriteButtonProps, FavouriteItemFormat, FilePickerItemProps, FilePickerProps, ManaSearchBarProps, MessageComponentClass } from "./types";
+import { cl, defs, hasPermission, ImageUtils, sendAttachment, transformAttachment, useFavourites, useListScroller, useResizeObserver } from "./utils";
+
+export const EmbedContext = proxyLazyWebpack(() => React.createContext<null | Embed>(null));
+export const EmbedMosaicContext = proxyLazyWebpack(() => React.createContext<null | number>(null));
+const AttachmentContext = proxyLazyWebpack(() => React.createContext<null | AttachmentItem>(null));
+
+const ManaSearchBar = findComponentByCodeLazy<ManaSearchBarProps>("#{intl::SEARCH}),ref");
+const FavoriteButton = findComponentByCodeLazy<FavoriteButtonProps>("#{intl::GIF_TOOLTIP_ADD_TO_FAVORITES}");
+const SendIcon = findComponentByCodeLazy("M6.6 10.02 14 11.4a.6.6");
```
sadan4: Better to vendor icons

src/plugins/favouriteAnything/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    managedStyle,
+    patches: [
+        // EMBEDS
+        {
+            find: "this.renderInlineMediaEmbed",
+            replacement: [
+                {
+                    // Wrap the embed component's render method in a custom context to avoid having to drill props
+                    match: "render()",
+                    replace: "$&{return $self.renderEmbed.call(this)}__render()"
```
sadan4: i dont think this will work, top-level plugins keys are bound to the object at the top level, bind has more precedence than call

sadan4: oops, only [these](https://github.com/Vendicated/Vencord/blob/main/src/api/PluginManager.ts#L331-L334) keys are bound

src/plugins/favouriteAnything/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+            find: '["VIDEO","CLIP","AUDIO"]',
+            replacement: [
+                {
+                    // Wrap the attachment component in a custom context to avoid having to drill props
+                    match: /(?<=children:)(\i)=>(\i\(\1\))\}\):(\i\(\))/,
+                    replace: "$1=>$self.renderAttachment($2,arguments[0])}):$self.renderAttachment($3,arguments[0])"
+                },
+                {
+                    // Always add our custom accessory to the attachment's adjacent content
+                    match: "=[];",
```
sadan4: too generic, is there something to anchor

src/plugins/favouriteAnything/index.tsx
Davr1: Removed the call just in case the key list changes in the future

src/plugins/favouriteAnything/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+                match: '.sortBy("order").reverse()',
+                replace: "$&.filter($self.filterGifs)"
+            }
+        },
+        // FAVOURITE BUTTON
+        {
+            find: "#{intl::GIF_TOOLTIP_REMOVE_FROM_FAVORITES}",
+            replacement: {
+                // Intercept the onClick callback to replace the placeholder thumbnail with a valid CDN link
+                match: /\(0,(\i\.\i)\)\((\{[^}].{40,60}?\})\)/,
```
sadan4: this also feels too generic

src/plugins/favouriteAnything/stores.ts
```diff
@@ -0,0 +1,95 @@
...
+import { proxyLazyWebpack } from "@webpack";
+import { Constants, Flux, FluxDispatcher, RestAPI } from "@webpack/common";
+
+import { RefreshedUrlsResponse } from "./types";
+import { BatchedRequestQueue, isAllowedHost } from "./utils";
+
+/** Used for storing and automatically refreshing signed CDN/Media proxy urls ({@link https://docs.discord.food/reference#signed-attachment-urls}). */
+export const SignedUrlsStore = proxyLazyWebpack(() => {
+    class SignedUrlsStoreClass extends Flux.Store {
+        public static readonly displayName = "SignedUrlsStoreClass";
```
sadan4:
```suggestion
        public static readonly displayName = "SignedUrlsStore";
```

Davr1:
I can finally remove the polyfill
image

Davr1:
Updated the attachment thumbnails to use tiny svgs instead of an external placeholder image service (Thanks fres!)
image
This also means one less patch since the thumbnail can be generated fairly cheaply at runtime

nin0-dev:
does it work on iPhone

Davr1:
Mobile can't render inline svgs I think (haven't tested on iphone)
at least the expression picker search still works so if you know what the file is called you can find it blindly

Maybe I could add a toggle to choose between inline svg and external bitmap thumbnails (which would have to use an external service, since storing a bitmap in the protobuf would be really wasteful)

Davr1:
Added mobile compatible thumbnails generated with placeholder.nin0.dev

PR #4453, opened 2026-07-27
fix(musicRichPresence): misc metadata lookup things

weightedangelcube:
## Describe your Changes
- `replace` on L18 is because Apache Lucene interprets those as special characters, so a name like "(The) Evening" wouldn't lookup properly
- find the release corresponding to that submitted by the scrobbler when using MusicBrainz lookup for cover arts
- if MusicBrainz lookup returns an album, and the submitted scrobble doesn't contain one, use the MusicBrainz album name and cover art

### DIFF
--- a/src/plugins/musicRichPresence/listenbrainz.ts
+++ b/src/plugins/musicRichPresence/listenbrainz.ts
@@ -74,7 +74,7 @@ async function getUrls(additionalInfo: Record<string, string> | undefined, track
     let rawQuery = `artist:"${artistName}" AND recording:"${trackName}"`;
     if (releaseName)
         rawQuery += ` AND album:"${releaseName}"`;
-    const query = encodeURIComponent(rawQuery).replace(/[\\!\\(\\)\\*\-\\~]/g, c => `\\${c}`,);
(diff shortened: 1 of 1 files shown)


src/plugins/musicRichPresence/listenbrainz.ts
```diff
@@ -74,7 +74,7 @@ async function getUrls(additionalInfo: Record<string, string> | undefined, track
     let rawQuery = `artist:"${artistName}" AND recording:"${trackName}"`;
     if (releaseName)
         rawQuery += ` AND album:"${releaseName}"`;
-    const query = encodeURIComponent(rawQuery);
+    const query = encodeURIComponent(rawQuery).replace(/[\\!\\(\\)\\*\-\\~]/g, c => `\\${c}`,);
```
Vendicated: escapes shouldn't be necessary (also ur double escapes likely just added "\\" literal multiple times) and replacement can be simplified to use a string
```suggestion
    const query = encodeURIComponent(rawQuery).replace(/[!()*\-~]/g, "\\$&");
```

Vendicated:
Thanks! Added them to the other pr

PR #4455, opened 2026-07-27
spotifyDisplayer plugin

Tinrii:
spotifyDisplayer plugin, adds song info above username in the bottom left corner. This is perfect for us who can't have discord and spotify opened at the same time but would like to see the song playing / skipping via keyboard shortcut.


lolsuffocate:
This is already a plugin called SpotifyControls

Tinrii:
yea I found that one when I finished making this one xD
This one doesn't have controls and is more basic and uses discord activity.
It might be useful to someone 😄

lolsuffocate:
You'd be better off making this a userplugin then as they won't have two of pretty much the same plugin. You could also use quickCss to get rid of the controls if you don't want them

Tinrii:
I've made it a bit prettier and without controls, but yea if it's not needed I will make it a userplugin and use it like that. 
I saw the other one once I was done making mine, I wouldn't have made it otherwise :/
image

Vendicated:
that's a real bummer, great work anyway!

PR #4456, opened 2026-07-27
New plugin: AutoJumpToMessage

bk304:
## Describe your Changes

This plugin is a quality-of-life improvement for anyone who keeps Discord open on a second monitor.

I often find myself having to stop what I'm doing just to click on Discord, switch chats, and view a message I've received. This frequently happens when someone I'm on a call with wants to show me something, like: "Hey, I sent a screenshot/something to your chat."

I'm tired of that.

With this plugin, if someone sends me a message, the plugin automatically switches the active chat to the one containing the new message.

I tried to make the plugin as user-friendly as possible. By default (though this is configurable), it only performs these switches when the Discord window isn't in focus; we don't want to take control away from the user.

Also, muted channels are ignored. I see no point in jumping to every single message from every server I'm in, lol.

The trick I used to build this plugin is simple: the behavior hooks into Discord's own notification-generation event.

### DIFF
--- a/src/plugins/AutoJumpToMessage/index.ts
+++ b/src/plugins/AutoJumpToMessage/index.ts
@@ -1,19 +1,21 @@
+import { definePluginSettings } from "@api/Settings";
 import { Devs } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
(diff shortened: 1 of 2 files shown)


bk304:
What was the reason for it being closed?

PR #4473, opened 2026-08-01
New plugin: Tagging for channels, threads, DMs

SteffanDonal:
## Describe your Changes

- Allows adding colored shapes to channels, threads and DMs that appear in the channels or DMs lists
- When adding/editing a tag, click the shape next to the name to switch between circle, triangle or square
- Can remove tags by clicking the inline tag decorations
- Can add/remove multiple tags at a time via context menu
- In the plugin's settings modal, all tags can be viewed, edited or deleted. A shortcut to the plugin's settings dialog has been added to the context menu (where relevant)
- Viewing a single tag shows all channels/threads/dms where it's used, grouped either by server or "DMs"
- You can navigate directly to a channel/thread/DM where a tag is used from the "tag view"

### Of special note

I widened the `label` property of the `MenuCheckboxItem` type to allow passing a full `ReactNode`, which I needed so I could add the tag swatch/shape in the context menus

I also widened the modal code filter on `closeAllModals` as I was encountering issues where only a single level was being closed (or all-but-one.. can't quite remember)

As this is my first contribution I've tried my best to keep to the existing patterns; as far as I can tell this plugin should fit in neatly. Let me know if I missed anything obvious and I'd be happy to clean up!


src/plugins/channelTags/ChannelTags.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { Tooltip } from "@webpack/common";
+
+import { removeTagFromChannel } from "./data";
+import { settings } from "./settings";
+import { TagShapeIcon } from "./TagShape";
+
+export function ChannelTags({ channelId }: { channelId: string; }) {
+    // CUSTOM settings do not expose wildcard paths in definePluginSettings' types.
+    settings.use(["tags.*" as "tags", "channelTags.*" as "channelTags", "clickTagsToRemove"]);
```
Vendicated: I think you can fix this by using
```js
const settings = definePluginSettings({...}).withPrivateSettings<{
	tags?: TagMap;
}>();
```
but then you are responsible for handling the null case and setting a default value

SteffanDonal: Looking into it, yeah, surprised I missed this!

SteffanDonal: Easy fix, cheers for the review

SteffanDonal:
Video of functionality:

https://github.com/user-attachments/assets/fe1f77f2-2dd0-4f36-897f-d27d42b4af03

src/webpack/common/modals.ts
```diff
@@ -17,7 +17,7 @@ const ModalAPI: t.ModalAPI = mapMangledModuleLazy(".modalKey?", {
     openModalLazy: filters.byCode(".modalKey?"),
     openModal: filters.byCode(",instant:"),
     closeModal: filters.byCode(".onCloseCallback()"),
-    closeAllModals: filters.byCode(".getState();for")
+    closeAllModals: filters.byCode(".getState();for", ".key,")
```
Nuckyz, line 20: What's the reason for this addition

SteffanDonal: Without it, `closeAllModals` matches against the wrong implementation - it matches `hasModalOpenedSince` not `closeAllModals`

Nuckyz: You are right, I did not notice it

PR #4508, opened 2026-08-10
new browser plugin WebPWA: install discord website as an app

ThaUnknown:
continuation of https://github.com/Vendicated/Vencord/pull/2994/ from 2024 maybe this time ven will review it :P

## Describe your Changes

This plugin makes discord installable as a PWA when used in browser, it's effectively a *more than slightly* worse version of Vesktop, but since it can re-use a browser that's already in use, does wonders for reducing RAM usage even further on devices which have little of it.

This enables vencord on web to:
- makes vencord installable as a PWA
  - lets discord open as a standalone window just like an app
  - lets discord be launched at startup
- adds support fo discord's custom title bar
  - supports dynamic sizing, zoom, display pixel density etc
- shows a notification badge
- added global keybinds
  - updated keybinds notice to say how to configure it
- added a shortcuts field, it could be modified to add more things, but for now it's just friends
- protocol handlers was added, but it's niche, won't do much
- launch handlers was added, this will open the PWA or navigate to the given URL in the already opened instance if the url is opened externally, say via another app on the PC or via another website
- nice install screenshots taken from discord's official website, why not

Most of this code is re-used from vesktop with slight alterations

there are still a few things to consider:
- what to do on firefox? firefox doesnt support ANY of this, but vencord also doesn't ship on firefox anymore?
- discord FORCES 720p30fps screenshare on web, for no good reason, idk how to not do that, while on the JS side it's easy since it's just changing track constraints, I have 0 success with trying to patch discord's garbled mess, so this might need to be a separate plugin, I'm sure copying vesktops screenshare logic for this wouldn't be hard, just out of scope
- if the user changes the theme, the PWA needs to be "re-installed", aka click remove and add, which is like 10 seconds, but still, this is because theme and background colors need to be hardcoded, and webmanifest doesn't really support updating itself

other things that could be done:
- the "download discord button" instead of being hidden could be modified to install the PWA if its not yet installed, yes this is possible, like for example on: https://haven.pages.dev/img-viewer/public/, but its just extra effort that adds little value
- protocol handler could and maybe should have a custom hook into discords protocol handler, but it's unlikely to be ever used, so it's fine as is

about://apps settings for auto-start:

image

keybinds:
image
the link is clickable and opens:
image

title bar with window controls overlay:
image
without:
image
css updates it automagically

the install prompt:
image


src/plugins/webPWA.browser/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+            },
+            {
+                src: icon,
+                sizes: "256x256",
+                type: "image/png",
+                purpose: "maskable"
+            }
+        ]
+    };
+    const url = URL.createObjectURL(new Blob([JSON.stringify(manifest)], { type: "application/json" }));
```
Vendicated: I think we need an URL.revokeObjectURL(linkEl.href) on top so we don't leak old manifest blobs?

ThaUnknown: yes we do!

sadan4:
>what to do on firefox? firefox doesnt support ANY of this, but vencord also doesn't ship on firefox anymore?

vencord doesnt ship a firefox extension, it's still available via a userscript

sadan4:
>discord FORCES 720p30fps screenshare on web, for no good reason, idk how to not do that, while on the JS side it's easy since it's just changing track constraints, I have 0 success with trying to patch discord's garbled mess, so this might need to be a separate plugin, I'm sure copying vesktops screenshare logic for this wouldn't be hard, just out of scope

maybe take a look at WebScreenshareFixes/add to that

PR #4510, opened 2026-08-11
feat(translate): add in-place translation and reply reference transla…

Josue2004exe:
This PR adds two user-friendly features to the Translate plugin:
1. An "in-place" translation display mode (to hide the original message and show the translated text directly inline).
2. Automatic translation of referenced reply messages (reply quotes).


Vendicated:
doesn't comply with our contribution guidelines

PR #4514, opened 2026-08-12
feat(MessageLogger): Add Whitelist/Blacklist Toggle

honeyxm8:
## Describe your Changes
Adds a new field to the MessageLogger plugin called "Exclusive Guilds"
Allows people to have the message logger work in only select servers/guilds.
(Useful for moderators who feel it's unethical to be using them otherwise)

image

### DIFF
--- a/src/plugins/messageLogger/index.tsx
+++ b/src/plugins/messageLogger/index.tsx
@@ -104,7 +104,7 @@ const settings = definePluginSettings({
     },
     whitelistEnabled: {
         type: OptionType.BOOLEAN,
-        description: "Whether to enforce the whitelist, logging only messages in whitelisted channels",
+        description: "Whether to enforce the whitelist, logging only channels or users",
         default: false,
     },
     whitelistedUsers: {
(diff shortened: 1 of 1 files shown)


lolsuffocate:
Would it be easier to add a toggle between whitelist and blacklist instead? Having an input for both will probably be confusing

honeyxm8:
That is an excellent suggestion and I'm updating it to work like that immediately

honeyxm8:
I've made the change:
image

lolsuffocate:
I mean more along the lines of using the same ID inputs that already exist but just toggling the behaviour of those lists between blacklists and whitelists

honeyxm8:
I think it would be better to keep the blacklist and whitelist boxes separate, so that you don't have to get all the guild IDs again when swapping between the two.
thoughts?

honeyxm8:
hopefully this version works :)
image
image

honeyxm8:
it occurs to me these commit names are probably wrongly formatted (this is my first time contributing to an open source repo)

nin0-dev:
> I think it would be better to keep the blacklist and whitelist boxes separate, so that you don't have to get all the guild IDs again when swapping between the two. thoughts?

image

this is from IgnoreActivities, and I wouldn't expect people to change the black/whitelist setting frequently

I was thinking too you should add an option to black/whitelist all DMs, that way I can choose to log specific server, and DMs

honeyxm8:
Consider that change added! Thank you for the feedback.

image

honeyxm8:
I forgot to include a screenshot of the new system earlier (the filter system)

image

PR #4520, opened 2026-08-13
README: fix broken Star History chart

OctoBored:
The Star History chart at the top of the README is broken. Switching it to a working data source that needs no API token so the chart renders again. All query params and the linking behaviour are preserved.


OctoBored:
Why?

PR #4525, opened 2026-08-16
Plugin filters multiselect

gitb0i:
# Changes

"Filter by Type" in the plugins tab changed to multiselect to be like the Tags filter to the right of it.

# Screenshots (if applicable)
image

## Checklist 


Vendicated:
I can't really think of a use case for this, especially considering you combine them via AND instead of OR

If you check both Enabled & Disabled you will get no results

PR #4527, opened 2026-08-16
SpotifyControls "improvements"

zoinbase:
## Describe your Changes
Added a volume slider (option), and some style options (centered text and blurred cover as the background of the player)

image


src/plugins/spotifyControls/spotifyStyles.css
```diff
@@ -218,4 +223,33 @@
...
+    inset: -20px;
+    background-image: var(--vc-spotify-track-image);
+    background-size: cover;
+    background-position: center;
+    filter: blur(7px);
+    opacity: 0.4;
+}
+
+.vc-spotify-cover-background > * {
+    position: relative;
```
ssalggnikool, line 254: all of this css can be done by a user if they want it, this shouldnt be a default style/design

src/plugins/spotifyControls/PlayerComponent.tsx
ssalggnikool, line 289: there is already a volume slider this pr is useless

Vendicated:
Album cover background was deliberately not included because it should be up to themes to implement it in a way that fits in. There is not one single style that will match everyone's expectation

Volume is already an option by right clicking the album cover. I don't think we should add more clutter like that

Aligning the text should also be done with css, makes little sense to be an option

PR #4532, opened 2026-08-19
ExpressionCloner: show number of available expression slots

sadan4:
image

image

### DIFF
--- a/src/plugins/expressionCloner/index.tsx
+++ b/src/plugins/expressionCloner/index.tsx
@@ -17,19 +17,22 @@
 */
 
 import { findGroupChildrenByChildId, NavContextMenuPatchCallback } from "@api/ContextMenu";
-import { migratePluginSettings } from "@api/Settings";
+import { definePluginSettings, migratePluginSettings } from "@api/Settings";
 import { BaseText } from "@components/BaseText";
 import { CheckedTextInput } from "@components/CheckedTextInput";
 import { Flex } from "@components/Flex";
 import { PlusIcon } from "@components/Icons";
+import { Span } from "@components/Span";
 import { Devs } from "@utils/constants";
 import { getGuildAcronym, hasGuildFeature } from "@utils/discord";
 import { Logger } from "@utils/Logger";
-import definePlugin from "@utils/types";
+import { EMDASH } from "@utils/text";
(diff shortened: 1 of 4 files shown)


Vendicated:
Greyed out isnt noticeable enough and should be at the end
image

Besides that the number would look better as a badge or smth

sadan4:
image

src/plugins/expressionCloner/index.tsx
```diff
@@ -16,22 +16,49 @@
...
+    openModalLazy,
+    PermissionsBits,
+    PermissionStore,
+    React,
+    RestAPI,
+    StickersStore,
+    Toasts,
+    Tooltip,
+    UserStore,
+} from "@webpack/common";
```
Vendicated: hop off this formatting, so many wasted lines for something you should never have to manually edit/look at

src/utils/text.ts
```diff
@@ -171,7 +171,9 @@ export function stripIndent(strings: TemplateStringsArray, ...values: any[]) {
     return string.replace(new RegExp(`^[ \\t]{${minIndent}}`, "gm"), "").trim();
 }
 
+export const EMDASH = "\u2014";
```
Vendicated: just use `&mdash;` or the actual character

src/plugins/expressionCloner/index.tsx
```diff
@@ -247,76 +335,68 @@ function CloneModal({ data }: { data: Sticker | Emoji; }) {
...
-                                )}
-                            </div>
-                        )}
-                    </Tooltip>
-                ))}
+            <div className={cn("guild-select-modal")}>
+                {guilds.map(({ guild: g, totalSlots, usedSlots }) => {
+                    const hasFreeSlots = usedSlots < totalSlots;
+                    const canClone = hasFreeSlots && !isCloning;
+                    const cursor: CSSProperties["cursor"] = canClone ? "not-allowed" : "pointer";
```
Vendicated: 😭 
```suggestion
                    const cursor: CSSProperties["cursor"] = canClone ? "pointer" : "not-allowed";
```

Vendicated:
The badge is really distracting tbh. Maybe the badge could just be number of available slots like 5 or 0, then elaborate in the tooltip

PR #4537, opened 2026-08-20
feat: GPU accelerated typing dots animation

ThaUnknown:
## Describe your Changes
expands on the idea of the no typing animation plugin to instead use a GPU bound animation, which makes it a LOT less resource intensive [but not costless]

makes the plugin on by default, since its pretty much impossible to tell them apart now

performance of this can easily be compared by opening chromium devtools > rendering > show paint flashing and show frame stats

one of these is the original animation, one of them is the GPU one, I don't know which :)

https://github.com/user-attachments/assets/e70f02e3-f971-4748-a802-451c71174eff

https://github.com/user-attachments/assets/fa407443-1e9d-4292-a64c-118dbfe648d4

### DIFF
--- a/src/plugins/fastTypingAnimation/index.ts
+++ b/src/plugins/fastTypingAnimation/index.ts
@@ -1,0 +1,28 @@
+/*
+ * Vencord, a Discord client mod
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+import managedStyle from "./styles.css?managed";
+
+export default definePlugin({
+    name: "FastTypingAnimation",
+    authors: [Devs.ThaUnknown],
+    description: "Reaplces the CPU-intensive typing dots animation with a fast fully GPU-bound one.",
+    tags: ["Utility"],
+    patches: [
+        {
+            find: "dotCycle",
+            replacement: {
+                match: /width:2\*(\i)\*3\+\1\/2\*2,height:2\*\1,className:/,
+                replace: "$&'vc-fast-typing-animation '+",
+            }
+        }
+    ],
+    enabledByDefault: true,
+    managedStyle
+});
--- a/src/plugins/fastTypingAnimation/styles.css
+++ b/src/plugins/fastTypingAnimation/styles.css
@@ -9,7 +9,6 @@
 }
 
 @keyframes dot-pulse {
-
   0%,
   50%,
   100% {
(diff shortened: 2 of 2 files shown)


src/plugins/fastTypingAnimation/index.ts
```diff
@@ -0,0 +1,34 @@
...
+        {
+            find: "dotCycle",
+            replacement: [
+                {
+                    match: /focused:(\i)/g,
+                    replace: (_, focused) => `_focused:${focused}=false`
+                },
+                {
+                    match: /width:2\*(\i)\*3\+\1\/2\*2,height:2\*\1,className:/,
+                    replace: "$&'nex-is-incapable-of-reading '+",
```
sadan4: change the className

src/plugins/fastTypingAnimation/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) 2023 Vendicated and contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+import { Devs } from "@utils/constants";
+import definePlugin from "@utils/types";
+
+import managedStyle from "./styles.css?managed";
+
+export default definePlugin({
```
sadan4, line 12: use migratePluginSettings, see ExpressionCloner for usage

src/plugins/fastTypingAnimation/index.ts
nexpid: don't

src/plugins/fastTypingAnimation/index.ts
ThaUnknown: TLDR from discord: no because its on by default

src/plugins/fastTypingAnimation/index.ts
sadan4: i say this because plugins use a standard naming convention `vc-plugin-name-class-name` and this makes things confusing for anybody who wants to, for example, disable the animation completely

nexpid:
```suggestion
                    replace: "$&'vc-fasttypinganimation-nex-cannot-read '+",
```

Nuckyz: do

ThaUnknown:
this plugin will be made obsolete by `dev://experiment/2026-07-typing-users-css-dots`, but for now I've changed the styling to align the old typing dots with this experiment's styling, as this experiment has a new animation is ever so slightly different

also we can't forcefully enable the experiment by default, as the typing indicators plugin uses the old component not new

PR #4555, opened 2026-08-26
WebKeybinds: Re-add missing CTRL+, keybind

Davr1:
## Describe your Changes
Added back the ctrl+, shortcut which I accidentally removed in a previous PR.

### DIFF
--- a/src/plugins/webKeybinds.web/index.ts
+++ b/src/plugins/webKeybinds.web/index.ts
@@ -19,6 +19,7 @@
 import { definePluginSettings } from "@api/Settings";
 import { Devs, IS_MAC } from "@utils/constants";
 import definePlugin, { OptionType } from "@utils/types";
+import { SettingsRouter } from "@webpack/common";
 
 
 const settings = definePluginSettings({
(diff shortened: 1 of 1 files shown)


src/plugins/webKeybinds.web/index.ts
```diff
@@ -64,6 +65,26 @@ export default definePlugin({
...
+
+    stop() {
+        document.removeEventListener("keydown", this.onKey);
+    },
+
+    onKey(e: KeyboardEvent) {
+        const hasCtrl = e.ctrlKey || (e.metaKey && IS_MAC);
+
+        // Cmd+, is a native shortcut on macos (Vesktop/src/main/mainWindow.ts)
+        const hasNativeShortcut = IS_VESKTOP && IS_MAC;
```
Nuckyz, line 80: the original code didn't have this, does this fixes an issue?

Davr1: The original code would capture the event and stop propagation so it never actually reached the system handler. There's functionally no difference for most users, but you can now safely override it without having a duplicate cmd+, shortcut

image

Nuckyz: alright

PR #4569, opened 2026-08-30
Feat auto translate

raspkey692:
## Describe your Changes
a better translate plugin that automatically translates different languages sent in dms, you can automatically set dms to be translated and set it so that in specific dms it automatically translates your own message sent to that person. also works in server channels.
image
image

- [x ] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file and made sure this pull request complies with it
- [x ] This pull request was written by me, and not an AI agent


Vendicated:
Hi! Unfortunately this is a duplicate of other pull requests and we also don't allow multiple versions of the same plugin so I'll have to reject this, but thanks nevertheless

PR #4574, opened 2026-09-02
feat(NoBlockedMessages): option to hide replies to blocked/ignored users

ssalggnikool:
## Describe your Changes

this adds a setting to NoBlockedMessages to hide messages which are replying to blocked and ignored (if enabled) users

this has a couple of issues hence why it's a draft, i'm not even sure if it's possible to fix them, messages aren't always hidden or take a second to get hidden because the message for messageReference won't always be in MessageStore

this probably isn't the best way to do this either and i'm not very happy with the code, i'm hoping someone smarter than me can pick this up and fix this and do it properly

i wouldn't have PRed this in this state but it took me a while to figure out the patch and the types so I don't really want to let that effort go to waste xD

also, this renames the ignoreMessages setting to suppressUnread because i think it's more indicative of what it's actually doing

- [x] This pull request was written by me, and not an AI agent.
- [x] I understand that if any of the above conditions are not met, this pull request will be closed and I will be banned from future contributions without further warning


src/plugins/noBlockedMessages/index.ts
```diff
@@ -46,10 +52,21 @@ const settings = definePluginSettings({
     }
 });
 
+type NonnullReference = Exclude<Message["messageReference"], undefined>;
```
sadan4:
```suggestion
type NonNullReference = Message["messageReference"] & {};
```

packages/discord-types/src/stores/MessageStore.d.ts
```diff
@@ -39,6 +39,8 @@ export interface ChannelMessages {
     forEach(callback: (message: Message, index: number, array: Message[]) => void, thisArg?: any): void;
     receiveMessage(msg: any): this;
     get(msgId: string): Message | undefined;
+    map(callback: (message: Message) => Message): Message[];
```
sadan4:
```suggestion
    map(callback: (message: Message) => Message, thisArg?: any): Message[];
```

src/plugins/noBlockedMessages/index.ts
```diff
@@ -63,6 +80,14 @@ export default definePlugin({
                 }
             ]
         },
+        {
+            find: "_channelMessages={}",
+            predicate: () => settings.store.hideReplies,
+            replacement: {
+                match: /static commit\((\i)\)\{/g,
```
sadan4:
```suggestion
                match: /static commit\((\i)\)\{/,
```

just checked and this only appears once in it's file

sadan4 (commented):
does this handle relationship updates

ssalggnikool:
> does this handle relationship updates

~~i think it does but i'll test later~~ it doesn't, it will only update if you click on a reply to the ignored message or click it inside of search

i'm not really sure what could be done about this tbh

PR #4577, opened 2026-09-03
Experiments: support playground & remove bug report button

thororen1234:


thororen1234:
closes https://github.com/Vendicated/Vencord/pull/4335 I got permission

src/plugins/experiments/index.tsx
```diff
@@ -127,8 +126,41 @@ export default definePlugin({
...
+                }
+            ]
+        },
     ],
+    matchExperiment(url: string, label: string): boolean {
+        const items = url.split("/");
+        const labelCleaned = label.replace(/[^a-zA-Z0-9]+/g, "").toLowerCase();
+        const urlEndCleaned = items[items.length - 1]?.replace(/[^a-zA-Z0-9]+/g, "").toLowerCase();
+        return !!labelCleaned && urlEndCleaned !== undefined && labelCleaned === urlEndCleaned;
+    },
```
Vendicated: I would prefer not having linking by label. It seems pointless and brittle. and just makes links non-standard

thororen1234: it is done

src/plugins/experiments/index.tsx
```diff
@@ -127,7 +126,24 @@ export default definePlugin({
                 replace: "$&if($1==null)return;"
             }
         },
-
+        // Enable playground embed on sent playground links
+        // dev://playground/mana, dev://playground/payments, dev://playground/virtual-currency,
+        // dev://playground/nitro, dev://playground/mfa, dev://playground/cms, dev://playground/void
+        {
+            find: "{PlaygroundEmbed:()=>",
```
Nuckyz: Is this find depending on the unmangled export?

thororen1234: fixed

PR #4578, opened 2026-09-03
TextReplace: allow namin g & reordering rules

thororen1234:


Vendicated:
I like this but the DnD UX sucks, it needs to preveiw drop location. Maybe you can try using Discord's reorderable list component (used in e.g. role list)

Also the search bar styling is ugly, needs some padding

PR #4580, opened 2026-09-03
WhoReacted: Disable Avatar Click Setting

thororen1234:
some people accidentally click the avatar a lot apparently so this disabled clicking it

### DIFF
--- a/src/plugins/whoReacted/index.tsx
+++ b/src/plugins/whoReacted/index.tsx
@@ -17,13 +17,14 @@
 */
 
 import { isPluginEnabled } from "@api/PluginManager";
+import { definePluginSettings } from "@api/Settings";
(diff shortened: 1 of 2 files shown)


src/plugins/whoReacted/index.tsx
```diff
@@ -113,12 +118,21 @@ function ReactionUsers({ message, users }: { message: Message, users: User[]; })
     );
 }
 
+const settings = definePluginSettings({
+    disableAvatarClick: {
+        description: "Toggle clicking avatars in reactions",
```
Vendicated:
```suggestion
    clickableAvatars: {
        description: "While this is enabled, clicking a reacting user's avatar will open their profile instead of adding the reaction",
```

thororen1234: done

PR #4597, opened 2026-09-14
fix(VoiceMessages): record real Ogg Opus in Chromium

### LINKED ISSUE #1512: VoiceMessages: Web recorder records webm instead of ogg

noita-player:
Fixes #1512

Chrome MediaRecorder can only emit WebM, which iOS Discord will not play even if the upload is named `.ogg`. Firefox already records real Ogg Opus, so that path is used as-is.

On Chromium we encode 20ms VoIP Opus with WebCodecs and mux the Ogg container ourselves. If WebCodecs is missing we remux a WebM recording. No extra dependencies (unlike #2303).

## Test plan
- Chrome / Vesktop: record a voice message, confirm the file is Ogg Opus (`OggS` / `OpusHead`), play it on iOS Discord
- Firefox: recording still works via native Ogg MediaRecorder
- Upload a non-Ogg file: the existing iOS warning still shows


Vendicated:
Who tf uses cursor in 2026

PR #4605, opened 2026-09-16
fix `pnpm inject` url on macos

claration:
## Describe your Changes

The installer is being distributed in a `.dmg` file now, `pnpm inject` needs to be adjusted for this new file type.

- [x] This pull request was written by me, and not an AI agent. 
- [x] I understand that if any of the above conditions are not met, this pull request will be closed and I will be banned from future contributions without further warning


LoganDark:
Can we extract the dmg instead of mounting it as a volume? Then I think we should also still use spctl and xattr as before.

Also, on newer macOS releases, `hdiutil: WARNING: 'hdiutil attach -nobrowse -mountpoint ...' is deprecated. Please use 'diskutil image attach --mountOptions nobrowse --mountPoint ...' instead.`

claration:
> Can we extract the dmg instead of mounting it as a volume? Then I think we should also still use spctl and xattr as before.

Mounting the DMG is standard for getting the contents inside.

No point in using `xattr` as we're not using a browser that applies the quarantine. The installer is also now notarized which negates the need of running this command anyway, it being notarized is also why it copies the entire bundle instead of the single binary.


> Also, on newer macOS releases, `hdiutil: WARNING: 'hdiutil attach -nobrowse -mountpoint ...' is deprecated. Please use 'diskutil image attach --mountOptions nobrowse --mountPoint ...' instead.`

You mean the one introduced in an OS released 1-2 days ago? They stopped using `hdiutil` in favor of a newer tool which is fine, but `hdiutil` will continue to work fine. The installer supports 12.0+, not purely 26.0+.

LoganDark:
> Mounting the DMG is standard for getting the contents inside.

My concern was more that I wouldn't want the mounted volume sitting around in case the script is killed for whatever reason.

> it being notarized is also why it copies the entire bundle instead of the single binary.

Copying the entire bundle is good

> You mean the one introduced in an OS released 1-2 days ago?

My bad, didn't know that was so recent.

claration:
> My concern was more that I wouldn't want the mounted volume sitting around in case the script is killed for whatever reason.

The volume is detached after the contents have been copied, it leaves an empty folder afterwards but a reboot will remove this.

The script keeps the installer inside of `dist/` and doesn't download/mount a new instance unless you tell it to do that by deleting it.

LoganDark:
Oh, you dismount it before running the installer. Ok, that's a small enough window that it looks fine to me.

iBreeilyRBLX:
Not 100% sure if this works might be worth it to take another look at it.

The .dmg is the only macOS asset so VencordInstallerCli-darwin just doesnt exist.
also dropping the whole xattr/quarantine removal step is a bold move, since without stripping it Gatekeeper just kills the downloaded binary on modern macOS versions (im using 26.6 Beta (25G5043d) so just keep that in mind)

claration:
> The .dmg is the only macOS asset so VencordInstallerCli-darwin just doesnt exist.

https://github.com/Vencord/Installer/commit/e56f3630d79c7b528f3caa20a4329cdd89827e28

> since without stripping it Gatekeeper just kills the downloaded binary on modern macOS versions

Only downloading via a browser does it quarantine downloaded files. This uses nodejs.

LoganDark:
> The .dmg is the only macOS asset so VencordInstallerCli-darwin just doesnt exist.

It will exist eventually, for now just use the commit immediately prior.

PR #4608, opened 2026-09-17
ClearReactions

Kyuvie:
Allows you to instantly remove all reactions from a message by holding a modifier key (Shift / Ctrl / Alt) and clicking it.

**Target audience:** Built for server moderators to handle reaction spam

**Permissions:** Requires the Manage Messages permission. Works in guilds only.


Vendicated:
Your brain is so AI poisoned you didn't realise this is already a stock feature

